PR #17, opened 2021-04-07
Add detailed source requirements.

MarkLodato:
This PR provides a starting point for discussions of detailed requirements for source systems. It likely does not get everything right, but hopefully it shows the level of detail we're expecting.

### DIFF
--- a/source-requirements.md
+++ b/source-requirements.md
@@ -1,0 +1,123 @@
+# SLSA Source Requirements
+
+This document enumerates all of the detailed requirements for a source to meet
+[SLSA](README.md).
+
+## Definitions
+
+A source **revision** is an immutable, coherent state of the source. In Git, for
+example, a revision is a commit in the history reachable from a specific branch
+in a specific repository. Different revisions within one repo MAY have different
+levels. Example: the most recent revision on a branch meets SLSA 3 but very old
+historical revisions before the cutoff do not.
+
+A source repository has a set of "trusted persons" who own the project and a
+"platform" that runs the infrastructure. For example, for
+https://github.com/MarkLodato/dotfiles, there is just one "trusted person" (Mark
+Lodato) and the platform is GitHub, while for
+https://hg.mozilla.org/mozilla-central the set of "trusted persons" are those
+with write access to the mozilla-central repository and the platform is
+Mozilla's mercurial hosting.
+
+## SLSA 1
+
+There are no source requirements at SLSA 1.
+
+## SLSA 1.5
+
+A revision meets SLSA 1.5 if all of the following are true:
+
+-   **[Version Controlled]** There exists a record of the history of changes
+    that went into the revision. Each change must contain: the identities of the
+    uploader and reviewers (if any), timestamps of the reviews (if any) and
+    submission, the change description / justification, the content of the
+    change, and the parent revisions.
+
+    -   **[Immutable Reference]** There exists a way to indefinitely reference
+        this particular, immutable revision. In git, this is the {repo URL +
+        branch/tag/ref + commit ID}.
+
+Almost any popular version control system meets this requirement, such as git,
+Mercurial, Subversion, or Perforce.
+
+NOTE: This does NOT require that the code, uploader/reviewer identities, or
+change history be made public. Rather, some organization must attest to the fact
+that these requirements are met, and it is up to the consumer whether this
+attestation is sufficient.
+
+## SLSA 2
+
+_NOTE: The SLSA 2 requirements are subject to change._
+
+A revision meets SLSA 2 if all of the following are true:
+
+-   The revision meets [SLSA 1.5](#slsa-15).
+
+-   **[Verified History]** Every change in the revision's history has at least
+    one strongly authenticated actor identities (author, uploader, reviewer,
+    etc.) and timestamp. It must be clear which identities were verified, and
+    those identities must use
+    [two-step verification](https://www.google.com/landing/2step/) or similar.
+    (Exceptions noted below.)
+
+    -   **[First-Parent History]** In the case of a non-linear version control
+        system, where a revision can have more than one parent, only the "first
+        parent history" is in scope. In other words, when a feature branch is
+        merged back into the main branch, only the merge itself is in scope.
+    -   **[Historical Cutoff]** There is some TBD exception to allow existing
+        projects to meet SLSA 2/3 even if historical revisions were present in
+        the history. Current thinking is that this could be either last N months
+        or a platform attestation guaranteeing that future changes in the next N
+        months will meet the requirements.
+
+-   **[Retained Indefinitely]** The revision and its change history are
+    preserved indefinitely and cannot be deleted, except when subject to an
+    established and transparent policy for obliteration, such as a legal or
+    policy requirement.
+
+    -   **[Immutable History]** It must not be possible for persons to delete or
+        modify the history, even with multi-party approval, except by trusted
+        platform admins with two-party approval following the obliterate policy.
+    -   **[Limited Retention for SLSA 2]** At SLSA 2 (but not 3), it is
+        acceptable for the retention to be limited to 18 months, as attested by
+        the source control platform.
+        -   Example: If a commit is made on 2020-04-05 and then a retention
+            attestation is generated on 2021-01-01, the commit must be retained
+            until at least 2022-07-01.
+
+## SLSA 3
+
+_NOTE: The SLSA 3 requirements are subject to change._
+
+A revision meets SLSA 3 if all of the following are true:
+
+-   The revision meets [SLSA 2](#slsa-2).
+
+-   **[Two-Person Reviewed]** Every change in the revision's history was agreed
+    to by two trusted persons prior to submission, and both of these trusted
+    persons were strongly authenticated. (Exceptions from [Verified History]
+    apply here as well.)
+
+    -   The following combinations are acceptable:
+        -   Uploader and reviewer are two different trusted persons.
+        -   Two different reviewers are trusted persons.
+    -   **[Different Persons]** The platform ensures that no person can use
+        alternate identities to bypass the two-person review requirement.
+        -   Example: if a person uploads with identity X then reviews with alias
+            Y, the platform understands that this is the same person and does
+            not consider the review requirement satisfied.
+    -   **[Informed Review]** The reviewer is able and encouraged to make an
+        informed decision about what they're approving. For example, it is not
+        acceptable to only display "path X was updated from hash abcd to hash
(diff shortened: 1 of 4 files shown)


MarkLodato:
One question for @kimsterv and @TomHennen: are we ready to submit such detailed requirements now, or do we still want to hold off?

source-requirements.md
```diff
@@ -0,0 +1,126 @@
...
+This document enumerates all of the detailed requierements for a source to meet
+[SLSA](README.md).
+
+## Definitions
+
+A source **revision** is an immutable, coherent state of the source. In Git, for
+example, a revision is a commit in the history reachable from a specific branch
+in a specific repository. Different revisions within one repo MAY have different
+levels. Example: the most recent revision on a branch was reviewed and meets
+SLSA 3, but very old historical revisions before the cutoff were not reviewed
```
TomHennen, line 12: Are we imagining that when the repo adopts SLSA 3 controls that the entire state of the repo will be reviewed?

If not then it's still true that many files in the most recent revision won't have been reviewed.

source-requirements.md
```diff
@@ -0,0 +1,126 @@
...
+
+    -   **[Binary Review]** A binary blob is OK if it can be viewed easily and
+        its usage is unambiguous. For example, a JPEG image is OK so long as it
+        is used as an image and not, say, a PE executable.
+    -   **[Generated Content]** Unreviewable machine-generated content MUST be
+        considered a separate resource with its own provenance and SLSA level.
+        For example, an executable checked into an otherwise human-readable git
+        repository must be visible to analysis and have its own SLSA level.
+        (Technical details TBD.)
+        -   TBD: Is it acceptable to have a policy on check-in that ensures SLSA
```
TomHennen, line 37: It seems like this will be the only way to say that a repo meets the requirements?  Or do we know of a method that could be used to analyze the content of a repo looking for generated or binary content?

source-requirements.md
```diff
@@ -0,0 +1,126 @@
...
+            3 deployments?
+    -   **[Vendoring]** Bulk import of third-party code, such that the change is
```
TomHennen, line 39: I'm wondering how we'd implement this technically.  Would we have to start including paths in the provenance?  foo/bar -> SLSA 3, foo/baz -> SLSA 2 (bulk import of third party code)

source-requirements.md
```diff
@@ -0,0 +1,126 @@
...
+
+    -   **[Trusted Metadata]** The identity of at least the uploader or one
+        reviewer, as well as the timestamps, must be non-falsifiable. It must be
+        clear which identity was verified, and that identity should use strong
+        authentication, such as
+        [two-step verification](https://www.google.com/landing/2step/).
+        -   Example: A GitHub user may push a commit with a fake
+            author/committer/reviewer/timestamp, but GitHub must record the
+            authenticated uploader and upload timestamp.
+    -   **[Multiple Commits]** A "change" may be more than one commit. If
```
TomHennen, line 67: So this suggests that the first commit after adopting SLSA 3 requires reviewing all content in the repo?

That's a pretty high bar...

Maybe we need to have a bootstrapping story too or else people may wind up getting concerned about how they could possibly meet these requirements.

source-requirements.md
```diff
@@ -0,0 +1,126 @@
...
+        distinguish the reviewed/verified commits from the
+        unreviewed/non-verified commits. (See also: "Informed review" below.)
+        -   As an extreme example, a revision's entire contents may be reviewed
+            using no parents, in which case only the revision itself meets SLSA
+            3 but not any of its ancestors.
+    -   **[Historical Data]** It is acceptable for the platform to implement
+        some cutoff revision whose ancestors do not meet these requirements.
+        Exact details TBD.
+
+-   **[Two-Person Review]** All changes in the revision's history were agreed to
```
TomHennen, line 79: I think talking about trusted robots might be very helpful here.  E.g. if a trusted robot does a copy of previously reviewed code then maybe it _doesn't_ need review?

source-requirements.md
```diff
@@ -0,0 +1,126 @@
...
+        -   Git example: If a fully reviewed commit in one repo is merged into a
+            different repo, or a commit in one branch is merged into a different
+            branch, then the merge still requires review.
+
+**Important:** The requirements above do NOT imply that the code,
+uploader/reviewer identities, or change history need to be public. Rather, some
+organization must attest to the fact that these requirements are met, and it is
+up to the consumer whether this attestation is sufficient.
+
+## SLSA 2
```
TomHennen, line 113: FWIW I've found this formulation of our internal requirements to be pretty difficult to read.  Could we go with the [L2,L3] tags or something like that?

TomHennen (changes requested):

source-requirements.md
MarkLodato: Removed "was reviewed."

source-requirements.md
MarkLodato: Removed this entirely. It's certainly an open question, but for now let's omit it.

source-requirements.md
MarkLodato: Reworded to be the traditional L1, L2 = L1 + ..., L3 = L2 + ....

source-requirements.md
MarkLodato: If it's OK with you, I'd like to omit it for now since we're running out of time and this is already taking a long time. We will likely need to add it in a future revision.

source-requirements.md
MarkLodato: Agreed. PTAL at the current exception to see if it seems OK.

MarkLodato:
OK, I think this is now ready for review. It's not perfect, but I think it is hopefully good enough to commit and iterate.

source-requirements.md
```diff
@@ -0,0 +1,123 @@
...
+    that went into the revision. Each change must contain: the identities of the
+    uploader and reviewers (if any), timestamps of the reviews (if any) and
+    submission, the change description / justification, the content of the
+    change, and the parent revisions.
+
+    -   **[Immutable Reference]** There exists a way to indefinitely reference
+        this particular, immutable revision. In git, this is the {repo URL +
+        branch/tag/ref + commit ID}.
+
+Almost any popular version control system meets this requirement, such as git,
```
inferno-chromium, line 40: nit: s/Almost any popular version control system meets/Most popular version control systems meet

source-requirements.md
```diff
@@ -0,0 +1,123 @@
...
+
+A revision meets SLSA 1.5 if all of the following are true:
+
+-   **[Version Controlled]** There exists a record of the history of changes
+    that went into the revision. Each change must contain: the identities of the
+    uploader and reviewers (if any), timestamps of the reviews (if any) and
+    submission, the change description / justification, the content of the
+    change, and the parent revisions.
+
+    -   **[Immutable Reference]** There exists a way to indefinitely reference
```
inferno-chromium, line 36: nit: is indent wrong, this one appears slightly right.

source-requirements.md
```diff
@@ -0,0 +1,123 @@
...
+        -   Uploader and reviewer are two different trusted persons.
+        -   Two different reviewers are trusted persons.
+    -   **[Different Persons]** The platform ensures that no person can use
+        alternate identities to bypass the two-person review requirement.
+        -   Example: if a person uploads with identity X then reviews with alias
+            Y, the platform understands that this is the same person and does
+            not consider the review requirement satisfied.
+    -   **[Informed Review]** The reviewer is able and encouraged to make an
+        informed decision about what they're approving. For example, it is not
+        acceptable to only display "path X was updated from hash abcd to hash
```
inferno-chromium, line 111: nit: instead of path, is dependency a slightly more realistic word here.

source-requirements.md
```diff
@@ -0,0 +1,123 @@
...
+        1234" because the reviewer would not be able to differentiate between a
+        benign change and a malicious one. Instead, the reviewer should be
+        presented with a diff between the proposed revision and the previous
```
inferno-chromium, line 114: nit: expand on diff, maybe "full code diff"

inferno-chromium (approved):
Nice!

source-requirements.md
MarkLodato: I removed this section entirely for now. It's certainly an issue, but I'd like to gloss over it since it's not really relevant until SLSA 3 and SLSA 3 isn't fully defined yet.

source-requirements.md
MarkLodato: This was intentional because it was intended to be a sub-bullet for "Version Controlled". The idea (perhaps misguided) was that the only top-level entries correspond to the ones in the main table, and sub-bullets are clarifications of the main one. Basically, there are a bunch of smaller items that are necessary but (I thought) would be too cluttered to add to the main table.

Is that too confusing? Would it be better to remove this two-level system? Or perhaps only add exceptions as sub-bullets rather than additional requirements?

source-requirements.md
MarkLodato: What about just "file" instead of "path X"? The point of this is that it's not sufficient for a "code review" to just indicate that a file changes, without showing the reviewer the contents of the change. Is there a way to make that more clear?

Updating a dependency, say in a package-lock.json file, actually would be fine IMO. There, it's perfectly normal for the content _of the package-lock.json_ to in turn contain a hash. Rather, it wouldn't be OK to just say "package-lock.json was updated" without explaining how.

WDYT?

source-requirements.md
MarkLodato: Done.

source-requirements.md
MarkLodato: Reworded this entirely. PTAL.

source-requirements.md
MarkLodato: Done.

source-requirements.md
inferno-chromium: slightly confused on the transition from main to sub-point. don't need to add to table, but can connect them by saying
**[Version Controlled]** Each revision has:
      -- **[Recorded History]
      -- **[Immutable Reference]

Not a big deal either way, either as-is or use this suggestion.

source-requirements.md
inferno-chromium: LGTM

source-requirements.md
MarkLodato: Good idea. Done.

source-requirements.md
MarkLodato: Resolving but we can always have a follow-up to fix.

MarkLodato:
Thanks Abhishek. I'll submit now. We can always add further refinements down the road (which we will almost definitely need.) This is a good start.

Issue #5, opened 2021-03-31
Add verified reproducible builds, maybe as SLSA 4


david-a-wheeler:
The SLSA proposal omits reproducible builds, and I think that's a mistake.

Currently the proposal assumes that there are magical build systems that cannot be subverted. No such systems exist. You could use build systems that are protected/hardened, and that obviously reduces risk, but subverting a build environment is a *known* attack that *just* happened to SolarWinds' Orion & is one of the main reasons people are calling for improvements in supply chain integrity. Ignoring a well-known attack vector, especially one that is *specifically* the kind of attack people are trying to prevent, is a mistake.

Implementing reproducible builds is usually not *that* hard. Typical challenges are enforcing order where things were in arbitrary order (e.g., by sorting) & by forcing timestamps to fixed values. It typically requires changes because historically no one tried to make builds reproducible, but it's doable.

I recommend that verified reproducible builds. It's fine if they're added as a new SLSA level (say, "level 4").

david-a-wheeler:
It's true that many organizations won't want to make their closed source public, but that's easily solved. It would be relatively easy to sign a contract to have some external party verify a reproducible build *and* provide many mechanisms to keep it from going public.

Even Oracle, which is *very* reticent about releasing source code for some of its products, has released their closed source code to an external lab to get a higher-level Common Criteria evaluation. A reproducible build is in many ways easier, because very few people in the rebuilder need access to the source code.

Hixie:
In https://github.com/slsa-framework/slsa/issues/34 I argue that verified reproducible builds isn't really a solution to the Solarwinds attack either. It arguably increases the cost of such an attack — you have to hit two systems instead of one — but is that enough?

There's an argument to be made that if you have reproducible builds it's actually _easier_ to attack your build system. With one CI/CD system, there's one target to attack, and all the efforts can go into securing it. With two CI/CD systems plus some logic to verify they are doing the same thing, there's three systems to attack, and you need to break two of them (either both build systems, or one of the build systems plus the verifier), and the effort for securing these systems is now split in three, so isn't it possible they're net easier to attack than the one system?

david-a-wheeler:
> so isn't it possible they're net easier to attack than the one system?

***NO.***

First, in many setups the "separate verifier" is likely (in part) to be the recipient, who by definition must receive something to be a recipient. This isn't "voting"; if there's *any* difference then by definition there's a *problem to be investigated*.

[...]

PR #18, opened 2021-04-09
Add Reproducible as a recommendation at SLSA 3

MarkLodato:
At SLSA 3, we now recommend reproducible builds. This is not a strict requirement because not all builds can become reproducible, as explained in the text. Once we write the detailed requirements, we will likely want to somehow explain that reproducible should be the default, while still allowing individual projects to opt-out.

The reason for adding this recommendation is to move the industry towards reproducibility, which is a generally useful property. By having it as the "default" path, most software will just go with the past of least resistance rather than opting out.

Note that this does not require *verifying* the reproduction for security (#5). Instead, the builder just claims that it was reproducible, presumably by building it twice and making sure that the output is identical.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,25 +179,34 @@ more, see [Threats, Risks, and Mitigations in the Open Source Ecosystem].
 
 ### What about reproducible builds?
 
-[Reproducible](https://reproducible-builds.org) and hermetic builds
-[provide](https://reproducible-builds.org/docs/buy-in/)
+When talking about [reproducible builds](https://reproducible-builds.org)
+builds, there are two related but distinct concepts: "reproducible" and
+"verified reproducible."
+
+"Reproducible" means that repeating the build with the same inputs results in
+bit-for-bit identical output. This property
+[provides](https://reproducible-builds.org/docs/buy-in/)
 [many](https://wiki.debian.org/ReproducibleBuilds/About)
 [benefits](https://static.googleusercontent.com/media/sre.google/en//static/pdf/building_secure_and_reliable_systems.pdf#page=357),
 including easier debugging, more confident cherry-pick releases, better build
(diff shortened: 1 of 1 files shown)


MarkLodato:
@david-a-wheeler Could you please review?

(Ignore the first two commits - GitHub doesn't have a way of marking one PR a dependency of another. ☹

david-a-wheeler (approved):
This makes sense to me, and it's what we discussed.

david-a-wheeler (approved):
This makes sense to me, it's what we agreed on.

README.md
```diff
@@ -184,10 +184,8 @@ more, see [Threats, Risks, and Mitigations in the Open Source Ecosystem].
 [many](https://wiki.debian.org/ReproducibleBuilds/About)
 [benefits](https://static.googleusercontent.com/media/sre.google/en//static/pdf/building_secure_and_reliable_systems.pdf#page=357),
 including easier debugging, more confident cherry-pick releases, better build
-caching and storage efficiency, and accurate dependency tracking. Most builds
-ought to be reproducible, but they often are not simply because it is not the
-default mode of operation by build tools. We strongly support any effort to make
-reproducibility the default mode of operation and the norm in the industry.
+caching and storage efficiency, and accurate dependency tracking. For these
+reasons, we require reproducibility and hermeticity at SLSA 3 by default.
```
TomHennen, line 188: "we require reproducibility" seems to be in conflict with "we only _recommend_ reproducibility" below.

TomHennen (changes requested):

README.md
MarkLodato: I rewrote most of this section to hopefully better explain. (With the old wording it was hard to make a small tweak.) PTAL

TomHennen (approved):

PR #30, opened 2021-05-18
Merge slsa-controls into slsa

MarkLodato:
The slsa-controls repo is very small, so it doesn't make sense to keep it separate. Having it all in one will hopefully aid discoverability.


joshuagl (approved):
Agree with the motivation for the merge, LGTM.

kimsterv (approved):

PR #40, opened 2021-06-02
Add "definition version" date

MarkLodato:
Addresses the most critical piece of #40. Down the road, it might be good to further document how we use this date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,8 @@ _Reminder: the proposed requirements here are based on what Google currently
 uses and has found useful. We expect these requirements—and perhaps even the
 concept of levels—to evolve based on community feedback._
 
+**Definition Version: 2021-04-21**
+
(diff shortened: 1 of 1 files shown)


dlorenc:
If this applies to the entire readme and needs to be updated each time the readme changes, do we even need it?

MarkLodato:
The intention is that this only applies to the normative requirements and should only be updated when there are significant requirements changes. (Think SemVer2 major version.) Clarifications changes and changes unrelated to requirements (e.g. motivation) would not necessitate a change.

Ideally, with more time, we'd (a) split up the README into multiple files to make this more clear and (b) add more guidance on how to use the version. But I thought it's useful to at least give people a common way to refer to this version now.

kimsterv:
lgtm!

kimsterv (approved):

PR #44, opened 2021-06-03
Rework SLSA requirements.

MarkLodato:
SLSA 1 is now fully defined and has lower requirements.

Major changes:

- Replace "Readability" with "Version Control", removed at SLSA 1.
- Add "Build Service", required at SLSA 2.
- Split "Provenance" into three rows, and do not require tamper
  resistance (e.g. signing) until SLSA 2.
- Remove "Enforcement". This is now orthogonal to the level.
- Do not require "Logging", "Policy", or "Common" until SLSA 2.
- Simplify "Policy" to not talk about build entry point.

Minor changes:

- Rename "Change History" to "Verified History".

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,9 +309,12 @@ effort, so intermediate milestones are important.
    <td>SLSA 2
    <td>"Auditable." Moderate confidence that one can trace back to the original source code and change history. However, trusted persons still have the ability to make unilateral changes, and the list of dependencies is likely incomplete.
   </tr>
+  <!--
+    Old wording for SLSA 1, can be used for 1.5: Stepping stone to higher levels. Moderate confidence that one can determine either who authorized the artifact or what systems produced the artifact. Protects against tampering after the build.
+  -->
   <tr>
    <td>SLSA 1
-   <td>Stepping stone to higher levels. Moderate confidence that one can determine either who authorized the artifact or what systems produced the artifact. Protects against tampering after the build.
+   <td>Stepping stone to higher levels. Unauthenticated assertion of how the artifact was produced.
   </tr>
  </tbody>
 </table>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -330,23 +329,26 @@ Each SLSA level has a set of requirements.
...
-  <tr>                      <td>Provenance       <td>↓ <td>✓ *   <td>✓     </tr>
-  <tr><td rowspan="4">Deploy<td>Provenance Chain <td>↓ <td>✓     <td>✓     </tr>
-  <tr>                      <td>Policy           <td>↓ <td>✓     <td>✓     </tr>
-  <tr>                      <td>Enforcement      <td>† <td>†     <td>✓     </tr>
-  <tr>                      <td>Logging          <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td rowspan="3">Common<td>Security         <td>↓↓<td>↓     <td>✓     </tr>
-  <tr>                      <td>Access           <td>↓↓<td>↓     <td>✓     </tr>
-  <tr>                      <td>Superusers       <td>↓↓<td>↓     <td>✓     </tr>
+  <tr><td rowspan="3">Provenance
+                            <td>Available        <td>✓ <td>✓     <td>✓     </tr>
```
MarkLodato, line 343: Any ideas on a better word? I don't like "Available" but I couldn't think of anything better.

msuozzo: Maybe "Published" or "Provided"? If you're more concerned with its existence over the ability of a user to get it, maybe "Generated"?

We're still hand-waving the propagation scheme(s) so just conveying that the "next hop" consumer can access it should be good enough.

README.md
```diff
@@ -356,29 +358,30 @@ Legend:
...
-    other things, authenticated uploader and reviewer identities as well as
-    trusted review and submission timestamps.
 *   **[Two-Person Review]** At least two trusted persons agreed to every change
     in the history.
 
 **[Build]** A build process qualifies for SLSA 3 if:
 
-*   **[Automation]** All build steps were fully automated, i.e. no manual
-    commands.
+*   **[Automation]** All build steps were fully defined in some sort of "build
```
msuozzo, line 381: "Automation" and "Build Service" still feel like they have a lot of overlap. What about "Build Definition" instead of "Automation"? I can definitely see people getting tripped up over whether `make` on a dev machine counts as automated.

I'm not opposed to Build Service but maybe adding "Remote" somehow could convey that we want builds to happen off of a dev machine.

README.md
dlorenc: Available is the best I can come up with too.

README.md
```diff
@@ -388,20 +391,27 @@ nuanced. We only provide a brief summary here for clarity.
...
+
+**[Provenance]** An artifact's provenance qualifies for SLSA 3 if:
+
+*   **[Available]** Provenance is available to the consumer of the artifact, or
+    to whomever is verifying the policy, and it contains at least the
+    cryptographic hash of the aritfact, the identity of the system that
+    performed the build, and the top-level source repository (i.e. the one
+    containing the build script).
+*   **[Tamper Resistant]** The build platform generates the provenance and the
+    platform's users cannot falsify it.
```
msuozzo, line 403: I was under the impression that we landed more on "immutable" instead of "non-falsifiable". I think ACL lockdown and human user restriction would be required to prevent falsifiability while a transparency log gets us immutability.

MarkLodato: The requirement here is that regular users of the build platform (e.g. pyyaml) cannot influence the provenance, e.g. claim that the source was X when it was really Y. Instead, only the platform admins can do that (e.g. GitHub Actions team). I don't know whether "tamper resistant" is the right term for that. My inclination is "authenticated" but it sounded like others didn't like this term.

The notion of transparency around deployments was what is currently called "logging".

README.md
MarkLodato: What about "Scripted" instead of "Automation"?

msuozzo: Sgtm!

README.md
msuozzo: I missed that this point is fully conditional on using a build platform/service. I think the "immutable" discussion centered around the case when provenance may be generated locally.

I think this is good as is, in that case.

README.md
inferno-chromium: Available, Accessible, Usable, Generated, lets pick any one?

README.md
inferno-chromium: "Scripted" sgtm as well

inferno-chromium (approved):
Thanks! Confirmed that @dlorenc reviewed this. And seems like @msuozzo reviewed as well.

README.md
MarkLodato: I'll leave it as Available for now, since none of the others seem to be significantly better.

README.md
MarkLodato: Done.

README.md
MarkLodato: Got it, thanks.

dlorenc (approved):

PR #47, opened 2021-06-04
Add back the link to the SRS book PDF

joshuagl:
The PDF has been made available again at the same location, so revert 85861abf2e7f68e685dd611c43833e472a1d2896 which removed the link.


MarkLodato (approved):
Thanks!

PR #49, opened 2021-06-07
Add SLSA 1.5

MarkLodato:
As discussed on Friday, this PR adds a new level, 1.5, with guarantees between 1 and 2. It also cleans up some wording.


dlorenc (approved):
Nice!

MarkLodato:
Thanks for the review! I'll just submit this now to allow future changes. If there is any more feedback, I'm happy to make follow-up changes.

PR #50, opened 2021-06-07
Differentiate between Source and Built requirements

### LINKED ISSUE #21: Split Source and Built Artifact ("Package"?) requirements
We have already defined SLSA to be non-transitive, such that the SLSA rating of built artifact does not depend on the ratings of its dependencies. But currently it does depend on its sources' ratings, which is odd because we don't have a strong definition of source vs dependency.

I propose that we fully split, such that each resource always has its own rating. Then we could split the requirements, one set for sources and one set for built artifacts. This would make things like two-person control more clear. For example, a GitHub repo that is owned by a single person may only be SLSA 2 (no two-party control), yet its built packages are all SLSA 3 because it uses an approved builder with provenance and reproducibility.

By the way, perhaps we should call the output of a build a "Package." That term fits the most common usage, and it is arguably correct for all examples. (Thanks @david-a-wheeler for the suggestion.)

MarkLodato:
This makes the table easier to read. Also rename "Retained" to "Retained Indefinitely."

Fixes #21.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,10 @@ the ideal end state. SLSA 1–2 offer lower security guarantees but are easier
 to meet. In our experience, achieving SLSA 3 can take many years and significant
 effort, so intermediate milestones are important.
 
+Furthermore, we differentiate between
+[source artifacts](#source-artifact-requirements) and
+[built artifacts](#built-artifact-requirements).
+
 <table>
  <thead>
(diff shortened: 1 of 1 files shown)


MarkLodato:
Discussed offline. Sounds like this is a change for the worse. Dropping.

PR #52, opened 2021-06-07
Clarify SLSA terms and fill in 2/3.

dlorenc:
This PR makes a few (hopefully) clarifying tweaks to the level and
term definitions:

* Split isolation into two rows, one called "Ephemeral Environments".
  I think we could roll all of this into isolation, but this makes it
  easier to understand and incrementally adopt IMO.
* Define Source Integrity, and leave it as a requirement at L2.
* Define Dependencies, and add it as a requirement at L2.
* Add one new requirement to Common for separate system identities. This
  is important for ephemeral/hermetic builds and system-generated
  provenance to be meaningful at lower levels.
* Move the "Security" requirement to be purely L3. I like having this
  reference a thorough existing compliance standard. We should decide
  which one(s), but that's fine to do later.
* Make Access and SuperUsers a requirement at L2.
* Define Verified History, including examples.
* Clarify the definition of Hermetic, authenticated and Dependencies.

Signed-off-by: Dan Lorenc <dlorenc@google.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,31 +331,30 @@ Each SLSA level has a set of requirements.
  </thead>
  <tbody>
   <tr><td rowspan="4">Source
-      <td>Version Controlled        <td>  <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Verified History          <td>  <td>  <td>✓     <td>✓     </tr>
-  <tr><td>Retained                  <td>  <td>  <td>18 mo.<td>indef </tr>
-  <tr><td>Two-Person Reviewed       <td>  <td>  <td>      <td>✓     </tr>
-  <tr><td rowspan="6">Build
-      <td>Scripted                  <td>✓ <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Build Service             <td>  <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Isolated                  <td>  <td>  <td>✓     <td>✓     </tr>
-  <tr><td>Hermetic                  <td>  <td>  <td>      <td>✓     </tr>
-  <tr><td>Reproducible              <td>  <td>  <td>      <td>○     </tr>
-  <tr><td>Source Integrity          <td>  <td>  <td>✓ *   <td>✓     </tr>
+      <td>Version Controlled        <td> <td>✓<td>✓     <td>✓</tr>
+  <tr><td>Verified History          <td> <td> <td>✓     <td>✓</tr>
+  <tr><td>Retained Indefinitely     <td> <td> <td>18 mo.<td>✓</tr>
+  <tr><td>Two-Person Reviewed       <td> <td> <td>      <td>✓</tr>
+  <tr><td rowspan="7">Build
+      <td>Scripted                  <td>✓<td>✓<td>✓     <td>✓</tr>
+  <tr><td>Build Service             <td> <td>✓<td>✓     <td>✓</tr>
+  <tr><td>Ephemeral Environment     <td> <td>✓<td>✓     <td>✓</tr>
+  <tr><td>Isolated                  <td> <td> <td>✓     <td>✓</tr>
+  <tr><td>Hermetic                  <td> <td> <td>      <td>✓</tr>
+  <tr><td>Reproducible              <td> <td> <td>      <td>○</tr>
+  <tr><td>Source Integrity          <td> <td> <td>✓     <td>✓</tr>
   <tr><td rowspan="5">Provenance
-      <td>Available                 <td>✓ <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Authenticated             <td>  <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Service Generated         <td>  <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Non-Falsifiable           <td>  <td>  <td>✓     <td>✓     </tr>
-  <tr><td>Dependencies              <td>  <td>  <td>      <td>✓     </tr>
-  <tr><td rowspan="3">Deploy
-      <td>Provenance Chain          <td>✓ <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Policy                    <td>  <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Logging                   <td>  <td>  <td>✓     <td>✓     </tr>
+      <td>Available                 <td>✓<td>✓<td>✓     <td>✓</tr>
+  <tr><td>Authenticated             <td> <td>✓<td>✓     <td>✓</tr>
+  <tr><td>Service Generated         <td> <td>✓<td>✓     <td>✓</tr>
+  <tr><td>Non-Falsifiable           <td> <td> <td>✓     <td>✓</tr>
+  <tr><td>Dependencies              <td> <td>✓<td>✓     <td>✓</tr>
+  <tr><td rowspan="1">Deploy
+      <td>Provenance Chain          <td>✓<td>✓<td>✓     <td>✓</tr>
   <tr><td rowspan="3">Common
-      <td>Security                  <td>  <td>  <td>↓     <td>✓     </tr>
-  <tr><td>Access                    <td>  <td>  <td>↓     <td>✓     </tr>
-  <tr><td>Superusers                <td>  <td>  <td>↓     <td>✓     </tr>
+      <td>Security                  <td> <td> <td>      <td>✓</tr>
+  <tr><td>Access                    <td> <td> <td>✓     <td>✓</tr>
+  <tr><td>Superusers                <td> <td> <td>✓     <td>✓</tr>
  </tbody>
 </table>
 
@@ -377,7 +376,10 @@ nuanced. We only provide a brief summary here for clarity.
     when that change occurred.
 *   **[Verified History]** The version control history indicates which actor
     identities (author, uploader, reviewer, etc.) and timestamps were strongly
-    authenticated.
+    authenticated. Note: in GitHub, direct commits **are not** strongly
+    authenticated unless they are cryptographically signed and the key is managed
+    by some form of external PKI, however Pull Requests merged by the GitHub
+    system can be if the account is using multi-factor authentication.
 *   **[Retained]** The artifact and its change history are retained indefinitely
     and cannot be deleted.
 *   **[Two-Person Review]** At least two trusted persons agreed to every change
@@ -389,15 +391,25 @@ nuanced. We only provide a brief summary here for clarity.
     script". The only manual command, if any, was to invoke the build script.
 *   **[Build Service]** All build steps ran using some build service, such as a
     Continuous Integration (CI) platform, not on a developer's workstation.
+*   **[Ephemeral Environment]** The build steps ran in an ephemeral environment,
+    such as a container or VM, provisioned solely for this build, and not reused
+    by other builds.
 *   **[Isolated]** The build steps ran in an isolated environment free of
-    influence from other build instances, whether prior or concurrent.
+    influence from other build instances, prior or conccurent. Build caches
+    should not be used for production builds unless they are purely
+    content-addressable.
 *   **[Hermetic]** All build steps, sources, and dependencies were fully
-    declared up front and the build steps ran with no network access.
+    declared up front and the build steps ran with no network access. The
+    control plane may fetch external dependencies and make them available to
+    the build step, as long as the build step declares them via immutable
+    references.
 *   **[Reproducible]** Re-running the build steps with identical input artifacts
     results in bit-for-bit identical output. (Builds that cannot meet this must
     provide a justification.)
 *   **[Source Integrity]** All input artifacts were fetched in a manner that
-    prevents tampering, such as TLS.
+    prevents tampering (such as TLS), and these artifacts were referenced using
+    immutable identifiers (specific commit instead of branch/tag, or a tarball
+    along with a cryptographic hash).
 
 **[Provenance]** An artifact's provenance qualifies for SLSA 3 if:
 
@@ -406,25 +418,22 @@ nuanced. We only provide a brief summary here for clarity.
     cryptographic hash of the artifact, the identity of the system that
     performed the build, and the top-level source repository (i.e. the one
     containing the build script).
-*   **[Authenticated]** Provenance's authenticity and integrity can be verified,
-    such as through a digital signature.
+*   **[Authenticated]** Provenance's authenticity and integrity can be verified
+    against the identity of the build system, such as through a digital
+    signature.
 *   **[Service Generated]** Provenance is generated by the build service itself,
     as opposed to user-provided tooling running on top of the service.
 *   **[Non-Falsifiable]** Provenance cannot be falsified by the build service's
     users.
-*   **[Dependencies]** Provenance records all build dependencies, meaning every
-    artifact that was available to the build script. This includes the initial
-    state of the machine, VM, or container of the build worker.
+*   **[Dependencies]** Provenance records build dependencies via immutable
+    references, meaning every artifact that was available to the build script.
+    This includes the initial state of the machine, VM, or container of the build
+    worker. Note: without `Hermeticity`, this list may still be incomplete. 
 
 **[Deploy]** An artifact deployed to a resource meets SLSA 3 if:
 
 *   **[Provenance Chain]** There is an unbroken chain of provenance linking the
     artifact back to its original sources and dependencies.
-*   **[Policy]** The resource's security policy only accepts artifacts with
-    provenance by specific builders and specific top-level source repositories.
-    The policy itself is a SLSA 3 source.
-*   **[Logging]** The history of deployed artifacts and their provenances is
-    retained for 18 months and is resistant to tampering.
 
 **[Common]** In addition to the requirements above, every trusted system
 involved in the supply chain (source, build, deploy, etc.) must meet the


README.md
```diff
@@ -338,24 +338,26 @@ Each SLSA level has a set of requirements.
...
   <tr><td>Reproducible              <td>  <td>  <td>      <td>○     </tr>
-  <tr><td>Source Integrity          <td>  <td>  <td>✓ *   <td>✓     </tr>
+  <tr><td>Source Integrity          <td>  <td>  <td>✓     <td>✓     </tr>
   <tr><td rowspan="5">Provenance
       <td>Available                 <td>✓ <td>✓ <td>✓     <td>✓     </tr>
   <tr><td>Authenticated             <td>  <td>✓ <td>✓     <td>✓     </tr>
   <tr><td>Service Generated         <td>  <td>✓ <td>✓     <td>✓     </tr>
   <tr><td>Non-Falsifiable           <td>  <td>  <td>✓     <td>✓     </tr>
-  <tr><td>Dependencies              <td>  <td>  <td>      <td>✓     </tr>
+  <tr><td>Dependencies              <td>  <td>✓ <td>✓     <td>✓     </tr>
```
MarkLodato, line 351: This change (Requiring Deps at 1.5 instead of 3) seems impratical. Doesn't this require hermeticity? For example, how does the GitHub Actions framework know that your Dockerfile did a `apt-get install`?

README.md
```diff
@@ -377,7 +379,8 @@ nuanced. We only provide a brief summary here for clarity.
     when that change occurred.
 *   **[Verified History]** The version control history indicates which actor
     identities (author, uploader, reviewer, etc.) and timestamps were strongly
-    authenticated.
+    authenticated (via MFA or similar). Note: in GitHub, commits **are not**
+    strongly authenticated unless they are signed, however Pull Requests are.
```
MarkLodato, line 383: Do you mean automatic merge commits generated by GitHub pull requests? (Pull Requests themselves are not part of the version history.)

Also, I assume MFA = multi-factor auth? If so, do those automatic merge commits satisfy that? In other words, how can one know that the signature was from an MFA-enabled account?

README.md
```diff
@@ -338,24 +338,26 @@ Each SLSA level has a set of requirements.
   <tr><td rowspan="6">Build
```
MarkLodato, line 338: this is now 7

README.md
```diff
@@ -390,14 +393,24 @@ nuanced. We only provide a brief summary here for clarity.
 *   **[Build Service]** All build steps ran using some build service, such as a
     Continuous Integration (CI) platform, not on a developer's workstation.
 *   **[Isolated]** The build steps ran in an isolated environment free of
-    influence from other build instances, whether prior or concurrent.
+    influence from other concurrent build instances. Build caches should not
+    be used for production builds unless they are purely content-addressable
+    and use cryptographically-secure hash algorithms.
+    **[Ephemeral]** The build steps ran in an ephemeral environment, such as a
```
MarkLodato, line 399: nit: [Ephemeral Environment] and move it above Isolated to match table.

README.md
inferno-chromium, line 383: Pull requests is common workflow in Github, nice that subsitutes for this.

README.md
```diff
@@ -338,24 +338,26 @@ Each SLSA level has a set of requirements.
   <tr><td rowspan="6">Build
       <td>Scripted                  <td>✓ <td>✓ <td>✓     <td>✓     </tr>
   <tr><td>Build Service             <td>  <td>✓ <td>✓     <td>✓     </tr>
+  <tr><td>Ephemeral Environments    <td>  <td>✓ <td>✓     <td>✓     </tr>
```
MarkLodato, line 341: nit: Environment (singular)

README.md
```diff
@@ -390,14 +393,24 @@ nuanced. We only provide a brief summary here for clarity.
...
+    control plane may fetch external dependencies and make them available to
+    the build step, as long as the build step declares them via immutable
+    references.
 *   **[Reproducible]** Re-running the build steps with identical input artifacts
     results in bit-for-bit identical output. (Builds that cannot meet this must
     provide a justification.)
 *   **[Source Integrity]** All input artifacts were fetched in a manner that
-    prevents tampering, such as TLS.
+    prevents tampering (such as TLS), and these artifacts were referenced using
+    immutable identifiers (specific commit vs. branch/tag, or a tarball along
```
inferno-chromium, line 412: nit: "instead of" reads slightly better than "vs. "

README.md
```diff
@@ -406,15 +419,17 @@ nuanced. We only provide a brief summary here for clarity.
...
+    against the identity of the build system, such as through a digital
+    signature.
 *   **[Service Generated]** Provenance is generated by the build service itself,
     as opposed to user-provided tooling running on top of the service.
 *   **[Non-Falsifiable]** Provenance cannot be falsified by the build service's
     users.
-*   **[Dependencies]** Provenance records all build dependencies, meaning every
-    artifact that was available to the build script. This includes the initial
-    state of the machine, VM, or container of the build worker.
+*   **[Dependencies]** Provenance records all build dependencies via immutable
```
inferno-chromium, line 429: Should we call this [Build Dependencies], just dependencies is slightly confusing.

README.md
```diff
@@ -390,14 +393,24 @@ nuanced. We only provide a brief summary here for clarity.
 *   **[Build Service]** All build steps ran using some build service, such as a
     Continuous Integration (CI) platform, not on a developer's workstation.
 *   **[Isolated]** The build steps ran in an isolated environment free of
-    influence from other build instances, whether prior or concurrent.
+    influence from other concurrent build instances. Build caches should not
```
MarkLodato, line 396: nit: The following wording would be more to the point:
"Build caches, if used, are purely content-addressable."
I think we can leave the hash algorithm bit to the details.

README.md
```diff
@@ -390,14 +393,24 @@ nuanced. We only provide a brief summary here for clarity.
 *   **[Build Service]** All build steps ran using some build service, such as a
     Continuous Integration (CI) platform, not on a developer's workstation.
 *   **[Isolated]** The build steps ran in an isolated environment free of
-    influence from other build instances, whether prior or concurrent.
```
MarkLodato, line 393: I'm a bit confused by the removal of "prior." It sounds to me like build caches could be the result of a prior build, not a concurrent one?

inferno-chromium (approved):
LGTM. @MarkLodato @msuozzo - thoughts?

README.md
MarkLodato, line 412: How is this clause different from "hermetic"? They sound the same to me.

I interpreted this item as "in the absence of hermeticity, you at least fetch the mutable references over TLS."

README.md
MarkLodato: I suggest [Dependencies Complete]. The main point of this bullet is that the list of dependencies is complete. This new change makes it seem like the immutability is the important bit, which is misleading. I suggest moving this to an additional sentence at the end.

README.md
```diff
@@ -435,6 +451,8 @@ following requirements:
     transport security, secure boot, machine identity, etc. Perhaps
     [NIST 800-53](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-53r5.pdf)
     or a subset thereof.)
+*   **[System Identity]** Any platform components run as a separate identity
```
MarkLodato, line 454: Isn't this (a) build-specific and (b) part of non-falsifiable?

README.md
```diff
@@ -338,24 +338,26 @@ Each SLSA level has a set of requirements.
...
   <tr><td>Policy                    <td>  <td>✓ <td>✓     <td>✓     </tr>
   <tr><td>Logging                   <td>  <td>  <td>✓     <td>✓     </tr>
-  <tr><td rowspan="3">Common
-      <td>Security                  <td>  <td>  <td>↓     <td>✓     </tr>
-  <tr><td>Access                    <td>  <td>  <td>↓     <td>✓     </tr>
-  <tr><td>Superusers                <td>  <td>  <td>↓     <td>✓     </tr>
+  <tr><td rowspan="4">Common
+      <td>Security                  <td>  <td>  <td>      <td>✓     </tr>
+  <tr><td>System Identity           <td>  <td>✓ <td>✓     <td>✓     </tr>
+  <tr><td>Access                    <td>  <td>  <td>✓     <td>✓     </tr>
```
MarkLodato, line 359: I'm not sure this is practical at SLSA 2. Let's discuss.

MarkLodato (changes requested):

README.md
msuozzo: Yeah +1. In general, I'd be fine with some lesser, more informational accounting of dependencies at these lower levels but I agree that this requirement would end up either being a) too difficult to meet or b) not meaningful without additional constraints.

Also, given that we've punted on even discussing dependencies in any sort of depth, I don't think we should include this in the initial set of requirements. I still have a lot of questions and concerns.

README.md
msuozzo: @MarkLodato This is in reference to just Access or to System Identity, as well?

For Access, the MPA requirement in particular seems tricky. Would that refer to the Build Service? The build workers? (and not to over-complicate this but...) The VCS server?

If we walk back the MPA into a suggestion, I'm fine requiring logged access and locked-down ACLs at L2.

README.md
msuozzo: +1 to point b). I think one implies the other and they're now required at different levels.

README.md
msuozzo: I had assumed this would only apply to the top-level source. "All input artifacts" would be too broad if that's the intent here.

msuozzo (changes requested):

README.md
dlorenc: Fixed.

README.md
dlorenc: Tried to clarify here. Is that more clear?

README.md
dlorenc: Fixed.

README.md
dlorenc: I thought Common applied to everything - source, build, artifact storage, etc.

README.md
dlorenc: Completeness and accuracy are orthogonal IMO. The immutability is the important part here. I'll try to clarify in the text: the difference is: what is specified is accurate, vs everything is specified. Both are useful, and accuracy is both important and feasible and lower levels.

README.md
dlorenc: Replied above, this is intentional and will aid in the generation of SBOMs, which are likely to be required WAY before hermeticity:

> Completeness and accuracy are orthogonal IMO. The immutability is the important part here. I'll try to clarify in the text: the difference is: what is specified is accurate, vs everything is specified. Both are useful, and accuracy is both important and feasible and lower levels.

README.md
dlorenc: Fixed!

dlorenc: Yes, anything declared as input artifacts. Basically, no curl | bash.

README.md
dlorenc: Happy to drop - the intention was to cover other parts of the system beyond simply provenance capture, but maybe I'm reading too much into what "Common" is for.

README.md
dlorenc: Exactly! Requiring pull requests and 2FA is actually **stronger** than signing with random GPG keys and not using 2FA.

README.md
msuozzo: Gotcha. Can we make the ladder a bit more nuanced then? I like the idea of describing completeness on a scale. Maybe "best-effort" or "partial" -> "source and data inputs" -> "all inputs and build tools".

README.md
msuozzo: Maybe this could be moved to the Build header and expressed as something like "Isolated Identity" and have a similar intent: The builder identity doesn't have sensitive access in the system.

I'm not exactly sure where GCB would sit with these constraints, though, at least with the current architecture.

inferno-chromium: Lets drop this then.

README.md
inferno-chromium: Can you please note this one down and discuss with @MarkLodato in next meeting. For now, fine to commit.

README.md
inferno-chromium: Sure, can add a "*" that this will be on best-effort basis for 1.5 ? And complete for 2 and 3?

inferno-chromium (approved):
LGTM with some minor changes below and:

1) I think we need to drop the policy,logging part from table (i thought we discussed dropping it from now, but still see in table)
2) Common table needs concensus, so fine to commit now, but please discuss in meeting/chatroom.

README.md
dlorenc: Clarified that it is not "complete" until hermeticity.

README.md
dlorenc: I think we're good now, I dropped this one.

README.md
```diff
@@ -331,31 +331,32 @@ Each SLSA level has a set of requirements.
...
-      <td>Available                 <td>✓ <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Authenticated             <td>  <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Service Generated         <td>  <td>✓ <td>✓     <td>✓     </tr>
-  <tr><td>Non-Falsifiable           <td>  <td>  <td>✓     <td>✓     </tr>
-  <tr><td>Dependencies              <td>  <td>  <td>      <td>✓     </tr>
+      <td>Available                 <td>✓<td>✓<td>✓     <td>✓</tr>
+  <tr><td>Authenticated             <td> <td>✓<td>✓     <td>✓</tr>
+  <tr><td>Service Generated         <td> <td>✓<td>✓     <td>✓</tr>
+  <tr><td>Non-Falsifiable           <td> <td> <td>✓     <td>✓</tr>
+  <tr><td>Dependencies              <td> <td>✓<td>✓     <td>✓</tr>
```
MarkLodato, line 351: This row should be called "Dependencies Complete" and should only be required only at SLSA 3. It can't be required earlier because it depends on hermeticity.

If you want another "best effort" requirement earlier, (a) that should be another row, (b) that should not be required at 1.5, and (c) how would that work?Let's use GitHub Actions as an example. Since GHA is the one generating provenance, how would it know what the dependencies were if there is no hermeticity? Until that is worked out, I don't think it should be added.

README.md
```diff
@@ -389,15 +393,25 @@ nuanced. We only provide a brief summary here for clarity.
     script". The only manual command, if any, was to invoke the build script.
 *   **[Build Service]** All build steps ran using some build service, such as a
     Continuous Integration (CI) platform, not on a developer's workstation.
+*   **[Ephemeral Environment]** The build steps ran in an ephemeral environment,
```
MarkLodato, line 396: We lost the other thread (man, GitHub is a terrible review system). In another thread, @msuozzo brought up that this seems intertwined with Isolated. I'm also not sure we can require it at 1.5. What is the motivation for breaking it up?

README.md
```diff
@@ -389,15 +393,25 @@ nuanced. We only provide a brief summary here for clarity.
...
+    declared up front and the build steps ran with no network access. The
+    control plane may fetch external dependencies and make them available to
+    the build step, as long as the build step declares them via immutable
+    references.
 *   **[Reproducible]** Re-running the build steps with identical input artifacts
     results in bit-for-bit identical output. (Builds that cannot meet this must
     provide a justification.)
 *   **[Source Integrity]** All input artifacts were fetched in a manner that
-    prevents tampering, such as TLS.
+    prevents tampering (such as TLS), and these artifacts were referenced using
```
MarkLodato, line 412: Lost the original thread. I don't agree with this change. It doesn't make any sense to me. What is "referenced"? Why is it related to TLS?

Overall, I think we should drop this requirement. I don't see how it can be implemented except with hermeticity, and in that case, we can just make that a line-item under hermeticity.

@msuozzo who also commented on this.

MarkLodato (changes requested):
I think all of the new 1.5 requirements should be reverted, but if we ultimately agree on adding them, the commit title needs to be updated to reflect that. (Right now it just sounds like it's clarifying, not adding any new requirements.)

README.md
```diff
@@ -406,15 +420,17 @@ nuanced. We only provide a brief summary here for clarity.
...
+    against the identity of the build system, such as through a digital
+    signature.
 *   **[Service Generated]** Provenance is generated by the build service itself,
     as opposed to user-provided tooling running on top of the service.
 *   **[Non-Falsifiable]** Provenance cannot be falsified by the build service's
     users.
-*   **[Dependencies]** Provenance records all build dependencies, meaning every
-    artifact that was available to the build script. This includes the initial
-    state of the machine, VM, or container of the build worker.
+*   **[Dependencies]** Provenance records build dependencies via immutable
```
MarkLodato, line 430: Whoops, for this this one. Please revert this change. It completely changes the meaning of the requirement.

README.md
dlorenc: > I think all of the new 1.5 requirements should be reverted, but if we ultimately agree on adding them, the commit title needs to be updated to reflect that. (Right now it just sounds like it's clarifying, not adding any new requirements.)

I'm happy to delete requirements completely, but we need to move this forward. Reverting to an undefined state is the wrong direction. If you can suggest clarifications that improve, that works too.

MarkLodato: Sorry, I thought all of my suggestions were very concrete. Is there something that is confusing?

dlorenc:
> * I think we need to drop the policy,logging part from table (i thought we discussed dropping it from now, but still see in table)

Also dropped this one for now too.

dlorenc:
Closing this one for now.

PR #53, opened 2021-06-08
Remove Resource and Deploy to simplify the model.

MarkLodato:
Previously, we differentiated between Resources and Artifacts, and SLSA
was a property of a Resource's security policy. However, many readers
found this concept very confusing.

Now, SLSA is purely a property of the artifact. If provenance exists
showing that it met the requirements, the artifact meets the level. No
policy or notion of "resource" is required. This simplifies the model at
some cost of security, which we have collectively decided is worth the
trade-off.


MarkLodato:
NOTE: This is still a draft because I have not yet updated the Motivating Example section. I wanted to get early review to see if it looks good to others before spending the time on that section.

inferno-chromium (approved):
Thanks!

dlorenc (approved):

MarkLodato:
I'll update the Vision section in a future revision.

PR #54, opened 2021-06-08
Update Vision section with latest changes.

### LINKED ISSUE #31: Relationship between resources and artifacts
(Upfront disclosure: I'm an engineer at Microsoft who is starting to look into supply chain security and I'm reaching out with some of my personal thoughts and opinions; I am not representing Microsoft as a whole).

Resources as friendly, non-content-based identifiers are assumed mutable. This tells me that what artifact a resource links to could change over time. Furthermore, I assume there may be multiple resources that point to the same artifact. For instance, the alpine image on Docker Hub currently has tags 3.13.5, 3.13, 3 and latest that all point to the same underlying image sha.

In the curlimages/curl example provided it looks like the build information (and thus attestation) is tied to the artifact via the resource, but if the resource changes which artifact it points to, the attestation would no longer be valid.

Would it make sense for builds to link directly to artifacts and pull resources out to the side as independent, friendly identifiers of artifacts (that can be updated over time)? Or was this the intent and I'm reading too much into what the implied model is from the diagram?

I appreciate the great work done here so far!

MarkLodato:
- Make the vision diagrams consistent with the terminology section:
  - Output is on the right, input is on the left.
  - Use colors consistently.
  - Rename "resource" to "artifact locator".
  - Simplify the diagram to reduce confusion (fixes #31).
- Update the level explanations based on recent changes:
  - SLSA 1 is unsigned.
  - Add SLSA 1.5 (merged with the SLSA 2 section).
  - Minor wording updates.
- Remove Deployment Policies section. We will eventually need to explain
  policies, but for now let's omit it until we agree on what that should
  look like.


msuozzo (approved):

README.md
```diff
@@ -559,54 +556,15 @@ too early to develop such a measure without real-world data. Once SLSA becomes
 more widely adopted, we expect patterns to emerge and the task to get a bit
 easier.
 
-### Deployment policies
```
inferno-chromium, line 562: Thank you for removing policy as discussed.

inferno-chromium (commented):
LGTM

PR #55, opened 2021-06-09
Clarify SLSA requirements.

MarkLodato:
Changes to requirements:
- Remove "Source Integrity", add immutable references to "Hermetic".
- Drop "Common" from SLSA 2 because it is likely expensive.

Clarifications:
- Split out "Ephemeral Environment" from "Isolation" (from #52).
- Explain that GH-generated merge commits meet Verified History (from #52).
- Clarify that all artifact references are immutable (from #52).
- Rename "Dependencies" to "Dependencies Complete" to avoid confusion.
- Define "SLSA level", "provenance", and "top-level source."
- Other minor cleanups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,13 +284,22 @@ The following diagram shows the relationship between concepts.
 
 Special cases:
 
-*   A ZIP file is containing source code is package, not a source, because it is
-    built from some other source, such as a git commit.
+*   A ZIP file is containing source code is a package, not a source, because it
+    is built from some other source, such as a git commit.
 
 ## Proposed SLSA definitions
 
 _Reminder: The definitions below are not yet finalized and subject to change. In
 particular, the levels will be renumbered to be all integers._
+
+An artifact's **SLSA level** describes the integrity strength of its direct
+supply chain, meaning its direct sources and build steps. To verify that the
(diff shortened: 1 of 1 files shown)


dlorenc (approved):

dlorenc:
LGTM. I'm not sure I fully understand what it means for 2/3 to be provisional though.

inferno-chromium (approved):
Thanks a lot!

provisional can be changed to * and can explain this * in more detail below as provisional and that will be iterating on it with feedback from community.

inferno-chromium:
> Thanks a lot!
> 
> provisional can be changed to * and can explain this * in more detail below as provisional and that will be iterating on it with feedback from community.

it can be even done in a generic way at bottom, as lets say 1.5 can be improved as we implement demo. so provisional should allow changes to all levels if needed in future.

inferno-chromium:
since you are here, can you rephrase "Our draft proposal is called SLSA", we shouldnt call it draft now ?

MarkLodato:
Let me revert the "provisional" and removal of "proposed" and instead do that in a separate PR, because that is kind of a loose thread that we can keep uplling.

MarkLodato:
Sent #55 for address the "provisional" and "proposed" aspect.

PR #56, opened 2021-06-09
Remove "proposed" wording.

MarkLodato:
Minor changes to remove the notion that this is a "proposal" and instead
just describe SLSA as it is.

Also explain that levels 2-3 are likely to change in the future, rather
than using some sort of symbol (*) or term (provisional), since
technically all requirements are subject to change. It's just that 2-3
are more likely to change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,23 +1,20 @@
-# SLSA: Supply-chain Levels for Software Artifacts, Proposal
+# SLSA: Supply-chain Levels for Software Artifacts
 
-## Objective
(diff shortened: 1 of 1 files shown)


MarkLodato:
@olivekl If you have a time, a quick review would be appreciated. I know that more substantive changes are needed, but is this at least a step in the right direction?

inferno-chromium (approved):

MarkLodato:
Thanks Abhishek and Kara!

PR #57, opened 2021-06-09
Split README.md into multiple files.

MarkLodato:
This splits up the doc as we discussed, @olivekl. I left the Related Work on the main page for now since I didn't know what to do with it.

If you think it's just as readable as before, we can merge. If it's a step backwards, let's just keep this in a separate branch until we improve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ wide-ranging feedback via GitHub issues, [email][mailing list], or
 
 SLSA consists of:
 
-1.  **Standards:** (this doc) Industry consensus on the definition of a "secure"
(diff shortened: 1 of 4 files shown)


olivekl:
Thanks! Let's keep it separate for now and merge once we've filled in some of the remaining missing pieces and connective tissue.

MarkLodato:
OK, sounds good. I just pushed it here:
https://github.com/slsa-framework/slsa/tree/multi-page

You can edit that branch and make changes, then once it's ready, we can merge back into the "main" branch.

PR #59, opened 2021-06-14
Define build requirements; add "Parameterless".

MarkLodato:
Fully define all build requirements. The document still needs an introduction, including diagram, threat model, and high-level description.

Add a "Parameterless" requirement to SLSA 4, which we forgot previously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,11 +342,12 @@ effort, so intermediate milestones are important.
   <tr><td>Verified History          <td> <td> <td>✓     <td>✓</tr>
   <tr><td>Retained Indefinitely     <td> <td> <td>18 mo.<td>✓</tr>
   <tr><td>Two-Person Reviewed       <td> <td> <td>      <td>✓</tr>
-  <tr><td rowspan="6">Build
+  <tr><td rowspan="7">Build
       <td>Scripted                  <td>✓<td>✓<td>✓     <td>✓</tr>
   <tr><td>Build Service             <td> <td>✓<td>✓     <td>✓</tr>
   <tr><td>Ephemeral Environment     <td> <td> <td>✓     <td>✓</tr>
   <tr><td>Isolated                  <td> <td> <td>✓     <td>✓</tr>
+  <tr><td>Parameterless             <td> <td> <td>      <td>✓</tr>
   <tr><td>Hermetic                  <td> <td> <td>      <td>✓</tr>
   <tr><td>Reproducible              <td> <td> <td>      <td>○</tr>
   <tr><td rowspan="5">Provenance
@@ -390,10 +391,12 @@ meaning the one containing the build script:
     Continuous Integration (CI) platform, not on a developer's workstation.
 *   **[Ephemeral Environment]** The build steps ran in an ephemeral environment,
     such as a container or VM, provisioned solely for this build, and not reused
-    by other builds.
+    from a prior build.
 *   **[Isolated]** The build steps ran in an isolated environment free of
     influence from other build instances, whether prior or concurrent. Build
     caches, if used, are purely content-addressable to prevent tampering.
+*   **[Parameterless]** The build output cannot be affected by user parameters
+    other than the build entry point and the top-level source location.
 *   **[Hermetic]** All build steps, sources, and dependencies were fully
     declared up front with immutable references, and the build steps ran with no
     network access. All dependencies were fetched by the build service control
--- a/build-requirements.md
+++ b/build-requirements.md
@@ -1,3 +1,318 @@
 # SLSA Build and Provenance Requirements
 
-TODO
+TODO: Add an introduction, diagram, threat model, and high-level description of
+each level.
+
+_Reminder: The definitions below are not yet finalized and subject to change,
+particularly SLSA 3-4._
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
+
+## Definitions
+
+**Provenance**: Metadata about how an artifact was produced.
+
+<a id="immutable-reference"></a>**Immutable reference:** An identifier that is
+guaranteed to always point to the same, immutable artifact. This MUST allow the
+consumer to locate the artifact and SHOULD include a cryptographic hash of the
+artifact's contents to ensure integrity. Examples: git URL + branch/tag/ref \+
+commit ID; cloud storage bucket ID + SHA-256 hash; Subversion URL (no hash).
+
+[immutable reference]: #immutable-reference
+[immutable references]: #immutable-reference
+
+## Requirements
+
+### Build Requirements
+
+Requirements on build process:
+
+<table>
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -394,6 +395,9 @@ meaning the one containing the build script:
 *   **[Isolated]** The build steps ran in an isolated environment free of
     influence from other build instances, whether prior or concurrent. Build
     caches, if used, are purely content-addressable to prevent tampering.
+*   **[Parameterless]** The build output cannot be affected by user parameters
+    other than the build entry point and the top-level source location. In other
+    words, the build is
```
inferno-chromium, line 400: finish sentence.

kimsterv, line 400: is there more to this sentence, or is it meant to be be joined with hermetic below?

build-requirements.md
```diff
@@ -1,3 +1,317 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
+
+## Definitions
+
+**Provenance**: Metadata about how an artifact was produced.
+
+<a href="immutable-reference"></a> **Immutable reference:** An identifier that
```
inferno-chromium, line 17: nit: move **Immutable reference:** inside the <a> tags

build-requirements.md
```diff
@@ -1,3 +1,317 @@
 # SLSA Build and Provenance Requirements
 
-TODO
+TODO: Add an introduction, diagram, threat model, and high-level description of
+each level.
+
+_Reminder: SLSA 3 and 4 are still subject to change._
```
kimsterv, line 6: I think we're still open to changes to 1 and 2 too, or we should be, but those are a bit more firmed up than the higher levels.

build-requirements.md
```diff
@@ -1,3 +1,317 @@
...
+interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
+
+## Definitions
+
+**Provenance**: Metadata about how an artifact was produced.
+
+<a href="immutable-reference"></a> **Immutable reference:** An identifier that
+is guaranteed to always point to the same, immutable artifact. This MUST allow
+the consume to locate the artifact and SHOULD include a cryptographic hash of
```
inferno-chromium, line 19: s/consume/consumer

build-requirements.md
```diff
@@ -1,3 +1,317 @@
...
+  <td> <td>✓<td>✓<td>✓
+ </tr>
+ <tr id="ephemeral-environment">
+  <td>Ephemeral Environment
+  <td>
+
+  The build service ensured that the build steps ran in an ephemeral
+  environment, such as a container or VM, provisioned solely for this build, and
+  not reused by other builds.
+
```
kimsterv, line 66: Is this a feature on build systems today, or how would one test for this, or should we add a note that this is something that a would potentially need to be audited by an accreditation process?

build-requirements.md
```diff
@@ -1,3 +1,317 @@
...
+  <td> <td>✓<td>✓<td>✓
+ </tr>
+ <tr id="ephemeral-environment">
+  <td>Ephemeral Environment
+  <td>
+
+  The build service ensured that the build steps ran in an ephemeral
+  environment, such as a container or VM, provisioned solely for this build, and
+  not reused by other builds.
```
inferno-chromium, line 65: nit: other builds or any other workloads ?

build-requirements.md
```diff
@@ -1,3 +1,317 @@
...
+  * MUST verify the integrity of each artifact.
+    * If the [immutable reference] includes a cryptographic hash, the service
+      MUST verify the hash and reject the fetch if the verification fails.
+    * Otherwise, the service MUST fetch the artifact over a channel that ensures
+      transport integrity, such as TLS or code signing.
+  * MUST prevent network access while running the build steps.
+    * This requirement is "best effort." It SHOULD deter a reasonable team from
+      having a non-hermetic build, but it need not stop a determined adversary.
+      For example, using a container to prevent network access is sufficient.
+
```
inferno-chromium, line 133: nit: Any example of this we should point to ? e.g. bazel ? can be done later too.

inferno-chromium (approved):
Thanks!

build-requirements.md
dlorenc: +1, I think the whole document is subject to change

inferno-chromium: i think we bubble it up at high level here - https://github.com/slsa-framework/slsa#slsa-definitions , so this one can be removed i think ?

README.md
MarkLodato: Oops, no. I meant to delete it.

build-requirements.md
TomHennen: I think this would be part of the accreditation process.

My understanding is that Tekton does this.  Not sure about others.

MarkLodato: Can we leave that to a future commit? We need to do an entire pass on how the consumer verifies these properties. In this PR, I tried to use the language "build service ensured" to indicate things that we have to trust/accredit, but I didn't want to get into details.

README.md
MarkLodato: See Kim's comment.

build-requirements.md
MarkLodato: Oops, thanks. Fixed.

build-requirements.md
MarkLodato: I'm not sure I quite follow, but the real point is about the past history of the environment, not its future. Changed to "from a prior build". Does that help?

build-requirements.md
MarkLodato: I don't have an example handy. I'm not super familiar with Bazel, but I would have guessed that it does _not_ satisfy this property since a genrule or user-defined rule can perform network requests. Let's defer.

build-requirements.md
MarkLodato: Oops, meant to be id, not href. (It seems like the standard in markdown is to just have an empty a tag for this case, so I left it as-is but fixed the attribute.)

Thanks for catching.

build-requirements.md
MarkLodato: Fixed. I copied the wording verbatim from the other doc. I think it's valuable have the reminder here in case the reader didn't see it in the other doc or forgets.

build-requirements.md
inferno-chromium: Fine to defer.

build-requirements.md
inferno-chromium: yes that is great!

build-requirements.md
inferno-chromium: Great!

build-requirements.md
inferno-chromium: SGTM

inferno-chromium:
Fine to merge, thanks for addressing these all comments!

MarkLodato:
Thanks for the review. @dlorenc and @TomHennen if you have any more comments, let me know and I can do a follow-up PR.

build-requirements.md
kimsterv: SGTM

PR #73, opened 2021-06-22
Linkify reqs table; remove duplicate descriptions.

MarkLodato:
In the "Level requirements" section, add links in the table from each
requirement the corresponding entry in the detailed source/build/common
doc, and remove the duplicate "summary" descriptions below it. This
avoids confusion and the need to keep the two in sync, at the cost of
making it slightly harder to skim. This seems like a worthwhile
tradeoff.

Also convert the table to pure Markdown, for two reasons:
- To make the table easier maintain.
- To allow us links to work in both GitHub and GitHub Pages. This is
  achieved by the jekyll-relative-links plugin, which only works with
  markdown links, not HTML links.

The downside to converting to Markdown is that we can't use rowspan and
colspan. This isn't too bad of a hit to readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,109 +250,55 @@ mean the lack of a SLSA level.
 
 ### Level requirements
 
-<table>
- <thead>
-  <tr><th colspan="2">           <th colspan="4">Required at</tr>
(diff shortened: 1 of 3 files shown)


inferno-chromium (approved):
Could be slightly annoying to click back and forth for seeing the definitions, but lets give it a try. But i do like reducing the duplication.

PR #76, opened 2021-06-23
Convert Source and Common requirements to tables

MarkLodato:
* Use the same formatting as the Build requirements, which is much easier to read.
* Reformat the Build requirements table as explained in 7fafe1e8.


inferno-chromium (approved):
Someone should double check the table ticks after this is checked in. It is hard to review in this CL itself. @kimsterv @TomHennen

PR #79, opened 2021-06-25
Allow a service other than the 'build service' to generate provenance

TomHennen:
This is just an initial thought.  We might only want this adjustment at L2 but still make the build service generate the provenance at L3+.

On the other hand, maybe as long as the builder _reports_ the data and it's confident in that data, it would be fine for some other service to generate the provenance?

That would allow for a 'trusted service' to translate one provenance format to another (in addition to gathering the data from API calls).


requirements.md
```diff
@@ -294,14 +294,14 @@ all the other requirements.
...
+the service generating the provenance.
 
 <td> <td>✓<td>✓<td>✓
 <tr id="service-generated">
-<td>Service Generated
+<td>Build Service Reported
 <td>
 
-The provenance was populated by the build service, not by user-provided tooling
+The provenance was populated by data reported by the build service, not by user-provided tooling
```
msuozzo, line 304: nit: populated _with_ data reported by...

or maybe "_using_"

requirements.md
```diff
@@ -294,14 +294,14 @@ all the other requirements.
...
 running on top of the service.
```
msuozzo, line 305: nit: I think I'd also reword the "on top of the service" to indicate that we actually me _in_ the service and not _over the top_ which we want to explicitly allow here.

msuozzo (approved):

msuozzo:
And yeah I think we could potentially tighten this language for higher levels but it may not be necessary.

I'd be curious as to what we'd call the "builder ID" in this case and whether this additional system would need to be represented somehow in the provenance but overall, I think it could work as-is.

requirements.md
TomHennen: went with 'with'

requirements.md
TomHennen: 'within'

requirements.md
```diff
@@ -294,15 +294,15 @@ all the other requirements.
 
 The provenance's authenticity and integrity can be verified by the consumer.
 This SHOULD be through a digital signature from a private key accessible only to
-the build service.
+the service generating the provenance.
 
 <td> <td>✓<td>✓<td>✓
 <tr id="service-generated">
-<td>Service Generated
+<td>Build Service Reported
```
MarkLodato, line 301: "Reported" seems very weak. Doesn't the current GHA demo meet this requirement because it pulls all the data from the service?

Instead, I think what you're trying to say is that _some_ service generates the provenance, but it doesn't have to be the "builder" per se. For example, if there is a trusted service that scrapes build logs and generates provenance, such that the user of the system cannot trivially fake it, that would comply, right?

If so, then maybe rename to "Trusted Generation" and update the text below to better explain the idea?

MarkLodato (changes requested):
If you rename the requirement, you also need to update README.md.

requirements.md
TomHennen: @msuozzo can you comment on Mark's first question "Doesn't the current GHA demo meet this requirement because it pulls all the data from the service?"

msuozzo: It's an interesting point. I'd disagree with the premise, though, that a build step within a build should be considered part of the build service. This gets down to a difference in trust boundaries (though, which we haven't rigorously specified yet): The data access is made inside the worker is done from the build's execution environment which we have much less confidence in considering the user-controlled nature of it. (And in some cases, this context data is also necessarily incomplete because the full build hasn't yet completed.)

I'm still for the current phrasing in the PR, personally.

TomHennen: Updated.  PTAL

MarkLodato: LGTM

MarkLodato (approved):

Issue #75, opened 2021-06-23
Style: Wrapping at sentences instead of lines


dlorenc:
This is definitely a preference thing, but I find markdown easier to work with and diff when we wrap at sentences instead of characters. See the OCI style guide for an example: https://github.com/opencontainers/image-spec#markdown-style

Any opposition to switching to this from the current 80 char convention?

inferno-chromium:
Sure, seems fine, just need to do at a time when there are no other active edits.

MarkLodato:
I've been using `mdformat`, which is Google internal but avoids formatting arguments altogether. I'd prefer to keep using that as long as possible.

dlorenc:
> I've been using `mdformat`, which is Google internal but avoids formatting arguments altogether. I'd prefer to keep using that as long as possible.

Is there an OSS equivalent we could point contributors to?

inferno-chromium:
> > I've been using `mdformat`, which is Google internal but avoids formatting arguments altogether. I'd prefer to keep using that as long as possible.
> 
> Is there an OSS equivalent we could point contributors to?

there is https://mdformat.readthedocs.io/en/stable/ , might be better or worse, not sure. we can try running on current spec and see if major diff.

dlorenc:
> there is https://mdformat.readthedocs.io/en/stable/ , might be better or worse, not sure. we can try running on current spec and see if major diff.

That one doesn't handle line-wrapping explicitly, for similar reasons to the one I outlined:

https://mdformat.readthedocs.io/en/stable/
https://sembr.org/

joshuagl:
> > I've been using `mdformat`, which is Google internal but avoids formatting arguments altogether. I'd prefer to keep using that as long as possible.
> 
> Is there an OSS equivalent we could point contributors to?

AIUI the desire is to present industry consensus, so a big +1 to using open tools

MarkLodato:
Sounds good. I'll take a look at one of the open-source mdformat to see if it works, but if not, we can just go with the OCI style guide.

MarkLodato:
I found the open-source [mdformat](https://mdformat.readthedocs.io/en/stable/) needlessly opinionated, so I set up [markdownlint](https://github.com/igorshubovych/markdownlint-cli) in #83. The only main change is consistent use of bullet character.

It doesn't address line length, but I have on objections to wrapping at sentences. Would we still wrap at 80 columns AND sentences, or just have really long lines?

06kellyjac:
[...]

PR #83, opened 2021-06-29
Set up and use markdownlint.

MarkLodato:
This does not directly address line length (#75) but it does make the Markdown more consistent. [markdownlint](https://github.com/igorshubovych/markdownlint-cli) can automatically fix most issues _except line wrapping_.

Do you think this is worth using?


dlorenc:
I don't have any direct experience with this one but it looks fine to me! There's a conflict right now - were you intending to actually merge this or just start a discussion?

dlorenc (approved):

.github/workflows/lint.yml
```diff
@@ -0,0 +1,11 @@
+name: Lint
+on: [pull_request]
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Markdown Lint
+      uses: nosborn/github-action-markdown-cli@v1.1.1
```
joshuagl, line 8: I may be a little paranoid, but I like to pin actions by SHA, as [recommended by the GHA docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses):
```suggestion
      uses: nosborn/github-action-markdown-cli@9fc95163471d6460c35cccad13645912aaa25d5f
```

Note also that:
i) 1.1.1 is shown as "Unverified" on GitHub:
> This commit is not signed, but one or more authors requires that any commit attributed to them is signed.

ii) 2.0.0 of the action has been released, perhaps this commit should add a dependabot config for the GHA?
That's as easy as adding a `.github/dependabot.yml` with contents like:
```
version: 2
updates:
- package-ecosystem: github-actions
  directory: "/"
  schedule:
    interval: daily
```

joshuagl (approved):
I added some minor comments on the lint workflow, but otherwise LGTM. Thanks for working on this.

.github/workflows/lint.yml
dlorenc: cc @jeffmendoza and @laurentsimon who have been working on some automation to help enforce/automate the GitHub actions pinning stuff.

inferno-chromium: @joshuagl suggestion is the one we recommend in scorecard too, e.g. https://github.com/ossf/scorecard/blob/main/.github/workflows/codeql-analysis.yml#L52 just at end, we also do "  # version x.y.z" for reference as otherwise hashes are not human readable.

MarkLodato: Thanks! Done. (Note that the Dockerfile still fetches things without a specific version tag (jq and curl) but it's better than nothing.)

jeffmendoza: More important than pinning, is restricting permissions with something like
```
permissions:
  contents: read
```
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/disabling-or-limiting-github-actions-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository

joshuagl: Thanks for the links!

PR #84, opened 2021-06-30
Test to see if mdlint GH Action works.

MarkLodato:
Not a real PR, just testing.


MarkLodato:
Well, it does run, but it doesn't do anything! It just prints the usage message.
https://github.com/slsa-framework/slsa/pull/84/checks?check_run_id=2954458079

PR #86, opened 2021-06-30
Set up `npm run lint` to call markdownlint.

MarkLodato:
This makes it easier for other developers to run markdownlint, and
checking in package-lock.json ensures that we're all using the same
version.

Futhermore, this commit calls `npm run lint` directly instead of using
the GitHub Action because the action doesn't seem to be working.

Example in action: https://github.com/slsa-framework/slsa/pull/84/checks?check_run_id=2955334338

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,12 +1,11 @@
 name: Lint
-on: [pull_request]
+on: [workflow_dispatch, pull_request]
 permissions: read-all
 jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-    - name: Markdown Lint
-      uses: nosborn/github-action-markdown-cli@6e34c67b8bd3406dd8ee6eeba353ca4c66793501 # v2.0.0
-      with:
-        files: .
-        config_file: ".markdownlint.yaml"
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+    - run: npm ci --ignore-scripts
+    - run: npm run lint --silent
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,7 +5,9 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-node@v2
+    - name: Checkout
+      uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # v2.3.4
+    - name: Setup Node
+      uses: actions/setup-node@38d90ce44d5275ad62cc48384b3d8a58c500bb5f # v2.2.0
     - run: npm ci --ignore-scripts
     - run: npm run lint --silent
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "scripts": {
     "lint": "markdownlint *.md controls/*.md",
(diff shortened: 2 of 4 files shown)


.github/workflows/lint.yml
```diff
@@ -1,12 +1,11 @@
...
   lint:
     runs-on: ubuntu-latest
     steps:
-    - name: Markdown Lint
-      uses: nosborn/github-action-markdown-cli@6e34c67b8bd3406dd8ee6eeba353ca4c66793501 # v2.0.0
-      with:
-        files: .
-        config_file: ".markdownlint.yaml"
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
```
joshuagl, line 9: Pin these both by sha digest, as was done with the 3rd party action?

package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "scripts": {
+    "lint": "markdownlint *.md controls/*.md",
+    "fix": "markdownlint *.md controls/*.md --fix"
```
joshuagl, line 4: Do we want fix? We don't actually push any changes back to git (and I don't think we should, either).

joshuagl: Oh, I see. This is so contributors can do `npm run fix`?

joshuagl (approved):
LGTM, thanks.

.github/workflows/lint.yml
inferno-chromium: +1

.github/workflows/lint.yml
```diff
@@ -1,12 +1,11 @@
...
+    - run: npm ci --ignore-scripts
+    - run: npm run lint --silent
```
inferno-chromium, line 11: Can remove --silent, good to have full output in logs.

package.json
inferno-chromium: instead of fix, can say format, we usually keep format and lint command in several of our projects.

inferno-chromium (approved):
lgtm with minor changes.

package.json
06kellyjac: Markdownlint-cli can do it's own globbing if you quote it.

`markdownlint '**/*.md'`

MarkLodato: Yes, it is for contributors to run. Renamed for `format`. The specific globs are to avoid recursing into node_modules or a pre-rendered Jekyll output.

.github/workflows/lint.yml
MarkLodato: Done.

.github/workflows/lint.yml
MarkLodato: Without --silent the error is significantly harder to read. Compare:
* Silent: https://github.com/slsa-framework/slsa/pull/84/checks?check_run_id=2955334338
* Non-silent: https://github.com/slsa-framework/slsa/runs/2955296872?check_suite_focus=true

06kellyjac: yeah, the flag is sent to `npm`, all it does is stop it from saying "oops it looks like the program you ran failed" and some other unrelated mess

package.json
06kellyjac: you can use ignore flags `--ignore node_modules` or an `.markdownlintignore` file

Or you can feed it the `.gitignore` file `--ignore-path .gitignore` as it has the same rules for the `.markdownlintignore` file

https://github.com/igorshubovych/markdownlint-cli#globbing-examples

Issue #42, opened 2021-06-03
Add threat model, in/out-of-scope attacks, and real-world examples


MarkLodato:
Right now it is not clear what attacks SLSA prevents. Example issues: #34, #35.

Tasks:

* [x] Add a threat model diagram and list of attacks.
* [x] Explain what attacks are mitigated (and to what extent) at each level.
* [x] Explain what other threats remain that SLSA does not address.
* [x] Add real-world examples like SolarWinds, event-stream, Great Suspender, and CodeCov, along with how those attacks could have been mitigated with SLSA.

**Status update, 2021-11-15:** https://slsa.dev/threats exists but still has several TODOs. I'll leave this issue open as a reminder to finish it.

MarkLodato:
Note: I gave a draft explanation in https://github.com/slsa-framework/slsa/issues/34#issuecomment-853363638 about how this would mitigate attacks like SUNBURST.

joshuagl:
Looks like the threat model diagram and example of mitigated attacks was done for the blog post here https://security.googleblog.com/2021/06/introducing-slsa-end-to-end-framework.html 

It would be good to get these added to the repo here.

MarkLodato:
The main page now has the list of example attacks from the blog post.

Next step is to add an exhaustive list of attacks. I started a draft in #92, but I'm not sure if it's helpful. Any feedback would be appreciated.

joshuagl:
> The main page now has the list of example attacks from the blog post.

💯 
 
> Next step is to add an exhaustive list of attacks. I started a draft in #92, but I'm not sure if it's helpful. Any feedback would be appreciated.

I think that for many the real-world examples are most useful for conveying the value of SLSA.
If we do add the detailed threat model, it should be a separate document, like the detailed requirements. 

However, I too am not sure how helpful the detailed threat model is. We already have mitigations proposed in form of the requirements. The levels indicate priority for implementing said mitigations.

MarkLodato:
**Status update, 2021-11-15:** https://slsa.dev/threats exists but still has several TODOs. I'll leave this issue open as a reminder to finish it.

Issue #35, opened 2021-05-27
Is the event-stream attack style in scope of SLSA?


Hixie:
It's tempting to think the attacks like the event-stream incident could be avoided by requiring all one's dependencies to be SLSA3-compliant, but I wonder if it is the intent of the SLSA document to make that claim.

The event-stream incident happened because the previous maintainers handed over control to a new team. As it happens, there was only one maintainer in the first group and only one in the second group, but it seems like it would be reasonable to imagine a situation where event-stream had two maintainers (as would be required to be SLSA3-compliant), both of which would be happy to hand over the reigns to a new team, which could easily consist of two collaborators. In this situation, wouldn't the same attack as affected event-stream users still work?

The downstream users of the package, when rolling their dependency, would do their due diligence, see that there was a new team in place, verify that they are still operating as a SLSA3-compliant team, and miss that the team's values have changed.

joshuagl:
My understanding is that event-stream like attacks are outside of the scope of the SLSA, but definitely interested in hearing the author's thoughts.

MarkLodato:
It depends on what you mean by the [event-stream attack](https://schneider.dev/blog/event-stream-vulnerability-explained/).

If you mean malicious changes to source code, which includes adding a malicious dependency, then the only protection from SLSA is the two-person review of SLSA 3. As you stated, this is far from perfect, and SLSA probably wouldn't help.

However, it sounds like event-stream NPM include malicious code that was not present in the upstream repo. That is **precisely** the attack that SLSA is intending to protect against.

Hixie:
I mostly meant the way event-stream's erstwhile maintainer handed off the project to a hostile maintainer, but you're right that there were other, more SLSA-relevant, aspects of that attack.

It would be useful for the SLSA docs to be more explicit about the kinds of attacks that are considered out-of-scope, so that potential adopters (who aren't security experts themselves and are using SLSA as a proxy for expertise) could have a better understanding of what other problems to consider. For example, if a dependency turning hostile is explicitly not in scope, then a potential SLSA-adopter may consider regularly auditing dependencies or using dependencies with more safeguards against that problem. If it's out of scope but not explicitly listed then we may not think of such things. :-)

MarkLodato:
Agreed. I just filed #42 to document this more generally. I'll mark this one as a duplicate.

Issue #34, opened 2021-05-27
Unclear how SLSA would help with the SolarWinds attack


Hixie:
The SLSA document says:

> * Attack any of the systems involved in the supply chain, as in the SolarWinds attack.
>
> SLSA intends to cover all of these threats.

It's not clear to me how the proposed model would help in cases like the SolarWinds attack. For example, suppose an open source project uses Travis CI to build their binaries, and Travis CI's infrastructure is infiltrated in a manner similar to the SUNBURST hack: how would the open source project detect it? Short of having actual verified reproducible builds using a different CI, we could never trust that the binaries built by this open source project are correct, could we? (The SLSA document says "SLSA does not require verified reproducible builds directly".)

But even if we did have two CI systems building the code of this open source project, say both Travis CI and Cirrus CI, what would stop an attacker from compromising both? Surely then even verified reproducible builds wouldn't help?

MarkLodato:
Totally agree that the doc does not make it clear, and we intend to better explain this. What follows is a hasty blurb.

The thinking is that we trust a small number of hardened **platforms** to avoid having to trust thousands of individual projects/users. It is very hard and expensive to lock down a build system sufficiently to protect against nation-state actors, and few organizations have the resources, skill, and priorities to do so. I won't comment on the security of Travis CI or Cirrus CI, but my biased view is large organizations like Google (my employer) are sufficiently resourced to protect against these threats. Thus, it's much harder to attack Google Cloud Build than it is to attack a random open source project.

Verified reproducible builds are another possibility. You can theoretically assemble a more secure "platform" out of several less secure pieces. What is stopping an attacker from compromising both is cost and risk of detection. If the N individual builders are independent, then the attacker has to perform N attacks and then coordinate among them. If they screw up and they do not properly compromise all, then they risk detection.

I know that's not a perfect answer, but hopefully it's a start.

Hixie:
That makes sense (and is similar to what @david-a-wheeler suggested in #5). Thanks.

MarkLodato:
Agreed. I just filed #42 to document this more generally. I'll mark this one as a duplicate.

PR #92, opened 2021-07-07
Add Threat and Mitigations page.

MarkLodato:
The intention is to address #42 by describing specific attacks that must be addressed at each level to guide implementers.

The doc is not yet complete, but hopefully it is still of some use. It seems better to submit a mostly complete page rather than wait forever for me to finish it.

The page is linked from the "levels" page but not from the bar at the top. This is because it is (a) not really ready for general consumption and (b) not intended for most readers.


TomHennen:
I do find this useful.

It's not quite clear to me how these various threats are protected against at these levels (maybe that's fine?).  Would it be worth numbering the requirements and then referencing them with these threats?  Maybe it would be too long a list.

MarkLodato:
I'm working on an alternative format that will hopefully better address it. @joshuagl also mentioned that moving this to a separate doc is probably a good idea.

One thing that I think the detailed threat model can convey is how the notion of policy comes in - in other words, how to *use* the provenance to prevent specific attacks.

I'll send out an update once the next version is ready for review.

MarkLodato:
Hey folks, I rewrote this to be much more exhaustive and explanatory. It's not yet complete, but I'm running out of time before my vacation (Aug 13 - Sep 13) so I wanted to get this draft out now. Please take a look to see if you find it useful or if you have any other feedback.

TomHennen:
I'm not sure "(H) Use a bad package" needs to be outside the scope of SLSA.  Users _could_ define a policy for the packages they use.  That policy could indicate what acceptable builder & sources are used for this package.  That information _may_ provide enough context for the user to determine this isn't what they want.

MarkLodato:
Hi @joshuagl would you mind reviewing this? Do you think it's valuable to submit in its current state rather than waiting until I have time to fully flesh it out?  I added an intro so hopefully it isn't quite so rough.

joshuagl (approved):
This looks great, thanks Mark!

Note that DCO bot is failing because 5f3edaca19ca550141ba4ea38f6e6c88c73a81f2 fails has a co-authored-by but no signed-off-by

MarkLodato:
Thanks @joshuagl! DCO issue is fixed.

Any other feedback on this? @hixie you had wrote #34 and #35. Does this doc (or the existing page) address your original question?

Hixie:
This is fantastic. I think for #34 and #35 it would be great to name those attacks specifically in this doc to make it clear to what extent and how those specific attacks are mitigated.

PR #98, opened 2021-07-12
Explicitly allow multiple provenance hashes

jchestershopify:
This change updates the "Identifies Artifact" requirement to allow multiple hashes to be computed for artifacts. This is already a capability of the In-Toto attestation format.


TomHennen (approved):
Looks great, thanks for the contribution!

PR #100, opened 2021-07-19
Remove source & entrypoint reqs at SLSA 1/2

TomHennen:
Many builders don't necessarily have this information available to populate the provenance.

In the interest of making the lower levels easier to adopt I propose we remove this requirement at
the lower levels so that builders can comply without having to make major architectural changes
or requiring that users use them in any particular manner.

### DIFF
--- a/requirements.md
+++ b/requirements.md
@@ -392,7 +392,7 @@ provenance. This represents the entity that the consumer must trust. Examples:
 The provenance identifies the source containing the top-level build script, via
 an [immutable reference]. Example: git URL + branch/tag/ref + commit ID.
 
-<td>✓<td>✓<td>✓<td>✓
+<td><td><td>✓<td>✓
 <tr id="identifies-entry-point">
 <td>Identifies Entry Point
 <td>
@@ -400,7 +400,7 @@ an [immutable reference]. Example: git URL + branch/tag/ref + commit ID.
 The provenance identifies the "entry point" or command that was used to invoke
 the build script. Example: `make all`.
 
-<td>✓<td>✓<td>✓<td>✓
+<td><td><td>✓<td>✓
 <tr id="includes-all-params">
 <td>Includes All Build Parameters
 <td>


MarkLodato (commented):
I'm OK with the change, but I'm a teeny bit concerned that SLSA 1 then becomes so weak that it's not particularly useful.

Do you think it would be valuable to convert these two to RECOMMENDED, meaning that they should be included unless there is a strong reason not to? Of course all of the stuff from higher levels falls into that boat, but these in particular are often inexpensive to add, so maybe it makes sense here?

msuozzo:
What builders are we talking about and why are we prioritizing onboarding for them?

I was under the impression that these requirements were the main contributions of this level on the SLSA ladder. Without them, we cause a sizeable discontinuity in effort to achieve higher levels.

TomHennen:
It's my understanding that Tekton, for example, doesn't always (if ever) know the source repo that the build config is stored in.  My limited understanding is that the build config is always (?) stored as metadata someplace else.  Config-as-code has been proposed, but hasn't been implemented yet (for example).  Making these required would require new features to be added to Tekon.

CloudBuild is another one, you _can_ configure it so that you can drive everything from some yaml checked into source, but I guess you don't have to.  Making this required would mean that users need to reconfigure how they run their Cloud Builds.  (Again I have limited understanding here).

This is discussed a bit in the first bullet here: https://docs.google.com/document/d/1RMzZP9dnz3JKFJglKco-cv0PBq_KkCWuNdi5pTVGeVQ/edit?ts=60eec1ad#heading=h.k99gx52wov3k

RECOMMENDED is fine with me.

msuozzo:
You can configure most build systems to be non-compliant but I don't think this warrants changing the design of the ladder. Per the discussion today, it seems Tekton will absolutely support tying a build definition to source and GCB already does for L1.

I think this is a viable constraint to enforce, even at L1, and I think we should have a strong rationale for weakening it, if we decide to do so.

TomHennen:
I'd argue that at the lower levels we should make it as easy as possible to adopt to include allowing builders to produce this information _without requiring their users to change their workflow_.

This can provide some benefits:

* At SLSA L2 the provenance provides chain of custody protection after the package leaves the builder.
* At SLSA L2 it provides an audit trail if something were to go wrong.
* At SLSA L1 & L2 this provides downstream users with a better understanding of the supply chain security of their dependencies.  If we forbid builds that cannot provide this information from entering the levels then we lose valuable insight like what builders packages are using and (if provided) what dependencies _those_ packages have (or at least a partial list).  This can help people prioritize work on their dependencies, and what builders to invest in.

TomHennen:
@MarkLodato marked both as recommended at L1/L2.

joshuagl:
> This is discussed a bit in the first bullet here: https://docs.google.com/document/d/1RMzZP9dnz3JKFJglKco-cv0PBq_KkCWuNdi5pTVGeVQ/edit?ts=60eec1ad#heading=h.k99gx52wov3k
> 

I am unable to access this document and would appreciate the context.



I don't think Tekton is particularly prevalent as a build platform for open source projects, so I'm not particularly inclined to weaken SLSA requirements for it's current features -- especially if work is already underway to support this case.

I agree that we should make SLSA 1 & 2 easy too adopt. Should we try and establish personas we are targeting? Or are we really targeting making it easy for all projects to comply with the early levels?

MarkLodato:
> I agree that we should make SLSA 1 & 2 easy too adopt. Should we try and establish personas we are targeting? Or are we really targeting making it easy for all projects to comply with the early levels?

Yes, we definitely need that. Filed #111.

Currently it seems like there is not enough consensus to remove these requirements? @TomHennen should we revisit once we've clarified goals (#111) and perhaps made more headway on implementation?

inferno-chromium:
> > I agree that we should make SLSA 1 & 2 easy too adopt. Should we try and establish personas we are targeting? Or are we really targeting making it easy for all projects to comply with the early levels?
> 
> Yes, we definitely need that. Filed #111.
> 
> Currently it seems like there is not enough consensus to remove these requirements? @TomHennen should we revisit once we've clarified goals (#111) and perhaps made more headway on implementation?

SGTM, should we move this PR to draft OR close, any of these is fine.

TomHennen:
Happy to revisit later.

PR #99, opened 2021-07-19
Document some SLSA use cases

TomHennen:
Still very much draft.  Lots of work left to do.

PR #103, opened 2021-07-20
new folder for community members to submit case studies

kimsterv:
trying to reach SLSA levels


TomHennen:
Love this idea.  FWIW I just sent PR #99 which is somewhat related.  Real case studies would be great to have.

TomHennen (approved):

PR #105, opened 2021-07-21
Use .markdownlintignore.

MarkLodato:
This allows us to automatically ignore all gitignore'd files, rather
than having to maintain the patterns in package.json.


MarkLodato:
An alternative to using a symlink would be to add `-p .gitignore` to the package.json file, or to copy the contents to the ignore file. Let me know if either of those sound preferable.

inferno-chromium (approved):
Thanks!

PR #116, opened 2021-07-29
adding community info

kimsterv:


dlorenc (approved):

inferno-chromium (approved):

inferno-chromium:
This would need to be connected to slsa.dev docs, so will let @devmoran do this as part of his cl.

PR #119, opened 2021-08-03
Import the provenance predicate from in-toto/attestation

MarkLodato:
This PR moves the in-toto Provenance predicate to the SLSA repo, as per in-toto/attestation/issues/54, including the full git history. Commits 0e86a0a thru 690b50c are from the old repo, while commits 9a70f02 onward are for this PR.

Summary of changes from the in-toto version:

- Rename https://in-toto.io/Predicate/v0.1 to https://slsa.dev/predicate/v0.1. (I used lowercase because there was previously some discussion about doing so, and using lowercase seems like a good idea.)
- Redirect https://slsa.dev/predicate to the latest version (currently v0.1).
- Remove obsolete references to "SLSA policy", since policy is not currently part of the SLSA spec.
- Add introductory text so that the page can stand alone. (The previous version assumed that the reader was familiar with in-toto attestations.)
- Fix links and markdown formatting.

Note: I manually set the DCO check to pass because the old commits lack the signed-off-by tag.


ci_survey.md
```diff
@@ -0,0 +1,93 @@
+# Appendix: Review of CI/CD systems
+
+The intention is to do a comprehensive review of all widely used CI/CD systems
+and make sure they cleanly fit into the schema.
+
+Lists:
```
TomHennen, line 6: What is 'Lists' supposed to be?  Unclear to me.

ci_survey.md
```diff
@@ -0,0 +1,93 @@
+# Appendix: Review of CI/CD systems
```
TomHennen, line 1: Should this page be included in this PR?  I don't see it mentioned in the PR description.

docs/provenance/v0.1.md
```diff
@@ -0,0 +1,481 @@
...
+## Schema
+
+```jsonc
+{
+  // Standard attestation fields:
+  "_type": "https://in-toto.io/Statement/v0.1",
+  "subject": [{ ... }],
+
+  // Predicate:
+  "predicateType": "https://in-toto.io/Provenance/v0.1",
```
TomHennen, line 46: Should this get updated to something SLSA specific?

TomHennen (changes requested):

TomHennen:
Looks like you beat me to it.  Thanks!  Re-assigned issue to you.

ci_survey.md
MarkLodato: It was referenced from the predicate page, but I'll remove it. We can add it back as a separate PR if it's useful.

docs/provenance/v0.1.md
MarkLodato: Oops! Fixed.

TomHennen (approved):

PR #125, opened 2021-08-06
Set page title to "Overview"; avoid custom header.

MarkLodato:
For index.md, manually set the page title to "Overview" rather than relying on the h1 ("Improving artifact integrity..."). This has two
effects:
1. The title in the titlebar is now "Overview", which reads better.
2. We no longer need a custom header.html template, and instead can  include the page in _config.yml.

Otherwise, there is no impact on how the page looks.

@devmoran Could you please review this change? In particular, is it OK to change the page title, or is there some SEO reason why we wanted it to be what it was before? Thanks!


docs/_layouts/home.html
```diff
@@ -0,0 +1,12 @@
+---
+layout: default
+---
+
+{% comment %}
+We cut out all feature of the minima "home":
```
TomHennen, line 6: nit: 'minimal' ?

docs/index.md
```diff
@@ -1,3 +1,6 @@
+---
+title: Overview
```
TomHennen, line 2: Does this just wind up rendering as 'Overview' in a tab?

Would 'SLSA Overview' be preferred?

TomHennen (changes requested):

docs/index.md
MarkLodato: It renders as "Overview | SLSA".

docs/_layouts/home.html
MarkLodato: The theme name is "minima".

MarkLodato:
Setting to draft so we don't accidentally submit without @devmoran's approval.

devmoran (approved):
LGTM!

TomHennen (approved):

MarkLodato:
Thanks for the reviews!

Issue #130, opened 2021-08-10
Clarify that SLSA is intended to be automatically verifiable


MarkLodato:
As mentioned in #127, we need to clarify that all of the SLSA requirements are intended to be automatically verifiable to some degree. Right now that is completely unstated, which results in confusion and disagreement.

Futhermore, we should explain *how* we picture verifying these requirements. Examples:

* Version Control: The source listed in the build provenance is either a known version control system (e.g. `git+https` URI scheme), has an attestation showing it came directly from version control (by matching on `digest`), or has some TBD chained verification.
* Retention: The source has an attestation from the source control platform claiming that it meets the retention requirement, and the verifier has some global configuration saying which platforms the verifier trust to make that claim. For example, a verifier might trust GitHub's claims about retention.

TomHennen:
Regarding the source control system issuing an attestation to these requirements please see https://github.com/in-toto/attestation/issues/47

joshuagl:
Note we added mention of the intent for all requirements to be automatically verifiable in #133 https://github.com/slsa-framework/slsa/pull/133#discussion_r690470200

TomHennen:
I talked about this a bit with @joshuagl 

The idea is that _artifacts_ can be automatically verified to see if they meet a given SLSA level.  _Systems_, however, may need some human analysis to see if they meet their SLSA requirements (e.g. Source & Build requirements).

For example, a security engineer may analyze the architecture and implementation of a build system to ensure that it meets (or _can_ meet) the [build requirements](https://github.com/slsa-framework/slsa/blob/main/docs/requirements.md#build-requirements).  Once the analysis is complete the keys used by the build system can then be 'trusted' up to a given SLSA level.

During verification of artifacts, the verifier can then automatically determine that the build system used meets the build requirements by checking if the keys used to sign the provenance are 'trusted' at the target SLSA level.

joshuagl:
Issue https://github.com/slsa-framework/slsa/issues/46 "Policy & Verification" is related to this. As is #478 on achieving automatic verification.

MarkLodato:
I'm repurposing this issue to cover the general issue of how systems and artifacts are verified to meet the SLSA requirements, de-duplicating other issues here. See the updated top comment for the latest.

PR #133, opened 2021-08-11
Improves comms around project ethos

devmoran:
- New copy on `index` to explain the high-level ethos of the project

image

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -50,6 +50,7 @@ Provenance checks to help evaluate risks and security
 ![Level 2](images/levelBadge2.svg)
 
 </div>
+<<<<<<< Updated upstream
 
 ### Medium protection
 
@@ -59,6 +60,17 @@ Further checks against the origin of the software
 
 <div class="level">
 
+=======
+
+### Medium protection
+
+Further checks against the origin of the software
+
+</div>
+
+<div class="level">
+
+>>>>>>> Stashed changes
 <div class="level-badge">
 
 ![Level 3](images/levelBadge3.svg)
@@ -116,18 +128,16 @@ Strict auditability and reliability checks
 <section class="col-2">
 <span>
 
-## How do you mitigate risks and threats to your supply chain?
+## How do you mitigate risks to your supply chain?
 
-Recent high profile supply chain attacks prove how costly an attack can be. It’s difficult to check the integrity of software artifacts today, but SLSA wants to fix that with a set of integrity requirements developers can follow to improve the security of the software they produce.
-
-See SLSA compared to [known supply chain attacks](levels.md#threats).
+Any software can introduce vulnerabilities into a supply chain, with recent high profile supply chain attacks proving how costly an attack can be. The requirements and checks that make up the SLSA framework aim to empower developers and software consumers to easily check the integrity of software artifacts, developed in direct response to [known supply chain attacks](levels.md#threats).
 
 </span>
 <span>
 
 ## Standard security guidelines that scale for your future
 
-Software consumers can choose software that provides the needed level of security. SLSA levels are a way to better understand your current security posture, and plan for the future. You can check that the security information for any software in your supply chain is accurate manually, and help develop and share tools that automate the process.
+SLSA levels are a way to better understand your current security posture, and plan for the future. If you’re a software consumer, you can check that the security information for any software in your supply chain is accurate, whether it provides the exact level of security you need, and help develop, share and promote tools that automate the process.
 
 <div class="pseudo-button">
 
@@ -144,9 +154,11 @@ Software consumers can choose software that provides the needed level of securit
 <div class="wrapper">
 <span class="subtitle flushed">In collaboration</span>
 
-## Building towards an industry consensus
+## Building towards the future
 
-We’re developing SLSA collectively to tackle common threats across the supply chain.
+<span class="subtitle">
+Today’s projects, products and services are increasingly complex and open to attack. As that trend continues, we need to scale up our effort to provide more secure, accessible ways to protect the development, distribution and consumption of the software we use, and all the impacted communities behind it.
+</span>
 
 <div class="pseudo-button m-t-l">
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -118,7 +118,7 @@ Strict auditability and reliability checks
 
 ## How do you mitigate risks to your supply chain?
 
-Any software can introduce vulnerabilities into a supply chain, with recent high profile supply chain attacks proving how costly an attack can be. The requirements and checks that make up the SLSA framework aim to empower developers and software consumers to easily check the integrity of software artifacts, developed in direct response to [known supply chain attacks](levels.md#threats).
+Any software can introduce vulnerabilities into a supply chain, with recent high profile supply chain attacks proving how costly an attack can be. The requirements and checks that make up the SLSA framework were developed in direct response to [known supply chain attacks](levels.md#threats) with the aim to empower developers and software consumers to easily check the integrity of software artifacts in an automatically verifiable way.
 
 </span>
 <span>
--- a/docs/index.md
+++ b/docs/index.md
@@ -50,7 +50,6 @@ Provenance checks to help evaluate risks and security
 ![Level 2](images/levelBadge2.svg)
 
 </div>
-<<<<<<< Updated upstream
 
 ### Medium protection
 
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -64,9 +64,9 @@ SLSA guides you through gradually improving the security of your software, build
 
 <section class="breakout">
 
-## Building towards an industry consensus
+## Building towards the future
 
-We’re developing SLSA collectively to tackle common threats across the supply chain.
+Almost all software projects today contain open source software. We expect that trend to continue. As it does, we need to scale up our effort to provide more secure, more accessible means to protect the development, distribution and consumption of that software and all the impacted communities behind it. That’s our ethos, our vision and our commitment.
```
TomHennen, line 69: Perhaps it's because I'm reading this out of context, but is this trying to restrict this to open source?  SLSA should be able to apply to open and closed source.

stevenif: Thanks both! I've taken on that feedback and tweaked this (incoming) to add in a mention of closed source, though @06kellyjac your feedback's got me drafting up a new direction for this. I'll keep that bubbling and present back once it gets somewhere. The main aim for this version, just for context, was to try keeping a focus on the positive ideal future (if security was already taken care of), so making the reader think less acutely about threats and risks.

docs/index.md
```diff
@@ -64,9 +64,9 @@ SLSA guides you through gradually improving the security of your software, build
 
 <section class="breakout">
 
-## Building towards an industry consensus
+## Building towards the future
 
-We’re developing SLSA collectively to tackle common threats across the supply chain.
+Almost all of today’s projects, products and services rely on open source software. We expect that trend to continue. As it does, we need to scale up our effort to provide more secure, accessible means to protect the development, distribution and consumption of open and closed source software, and all the impacted communities behind it.
```
06kellyjac, line 69:
```suggestion
Almost all of today’s projects, products and services rely on many dependencies with those having dependencies in turn. We expect that trend to continue. As it does, we need to scale up our effort to provide more secure, accessible means to protect the development, distribution and consumption of software, and all the impacted communities behind it.
```

I'd rather not cover the licenses of dependencies as it's not relevant to whether it's a secure dependency or not.
I guess it might be important for non-technical readers who are only really concerned with the split between libre/open source and proprietary dependencies so mentioning both sides of the divide might make things clearer? :shrug: 

---

Also I haven't got a proper suggestion but I'm not a huge fan of "Almost all". I know that this needs to be readable by technical and non-technical people alike and that's why I'm not sure to suggest here but these are my thoughts:

For all intents and purposes all projects, products, and services have dependencies. When we think software projects with dependencies we might think a React project with 10 direct dependencies and 2k transitive dependencies or a go project with a `go.mod` file that has a handful of libraries. But even if you're writing hello world in C you're going to depend on a compiler and you should arguably know it's providence and all that other good stuff being documented in SLSA and other supply chain security efforts

TomHennen: I like this suggestion.

stevenif: Thanks both for the comments! 
Ace getting this conversation going - goals-y and vision-y statement copy is _always_ the toughest to get right and usually takes forever, so it's great seeing that this might be close-ish to the mark. 

**tl;dr**
The last couple of changes made might be throwing this copy a little too far off course from what it's design needs are, and (hopefully I'm not overstepping here!) I'd suggest rolling back to the original PR copy for now - I can keep working off these comments to see where a) points are already being covered in contextual content, b) if I can create another viable alternative for this block for a future PR/discussion?

_rationale_ 
It's tricky to get the tone needed and balance it with being just specific/referential as possible. It doesn't need to get too far into the weeds though, just enough to nod to the context and still be accessible to your widest possible audience. To get there though, I'd plumbed for the original version so you don't have to reiterate the negative problem context in words, especially as it would be reiterating content that has a home in the supply chain/threats diagram content used on the landing page. Even if putting the focus on 'using open source' vs 'dependencies' isn't all that SLSA's about, it _does_ mean you can use this goals content to power home a really strong rallying call (i.e. the last two sentences), and a tone that feels peer-to-trusted-peer with elements like "We expect", "As it does", which are really doing the bulk of the work here. 

I've taken a stab at revising based on your comments, and end up with this one below. But I don't think it's fit for purpose yet - because it really still needs the "which presents... we use" clause for the first sentence to make sense, it throws the "We expect" line way off in its tone and impression, but cutting that particular line out then means cutting "As it does", which in turn loses the idea that SLSA's development is in concert with the wide audience/industry... and by then I think you just lose too much of the necessary elements a good goals/vision statement needs to have.

e.g.
_"Today's projects, products and services rely on dependencies more than ever, which presents serious problems for securing the software supply chains we use. We need to scale up our effort to provide more secure, accessible means of protecting software development, distribution and consumption, and all the impacted communities behind it."_

Definitely happy to keep the conversation going, and I'll take any of these ideas back into the drafting stage <3

devmoran:
Updated to address PR comments. 
@TomHennen I'd like to get this merged in so that we can lock this up with #134. If you're happy, pls review and merge.
If we feel there's further improvements to make we can always raise another PR.

TomHennen:
This looks good to me, but I'd like someone outside Google to review since it is adjusting the mission statement.

@inferno-chromium added @joshuagl to provide that review.

docs/index.md
TomHennen: LGTM

docs/index.md
TomHennen: LGTM

inferno-chromium:
Adding all the goverance board members
@joshuagl
@mlieberman85
@trishankatdatadog
@david-a-wheeler
@MarkLodato (ooo)
@zakgreant
@brunodom

inferno-chromium:
@devmoran - you would need to rebase after last commit.

docs/index.md
```diff
@@ -84,18 +84,16 @@ We’re developing SLSA collectively to tackle common threats across the supply
 <section class="col-2">
 <span>
 
-## How do you mitigate risks and threats to your supply chain?
+## How do you mitigate risks to your supply chain?
 
-Recent high profile supply chain attacks prove how costly an attack can be. It’s difficult to check the integrity of software artifacts today, but SLSA wants to fix that with a set of integrity requirements developers can follow to improve the security of the software they produce.
-
-See SLSA compared to [known supply chain attacks](levels.md#threats).
+Any software can introduce vulnerabilities into a supply chain, with recent high profile supply chain attacks proving how costly an attack can be. The requirements and checks that make up the SLSA framework aim to empower developers and software consumers to easily check the integrity of software artifacts, developed in direct response to [known supply chain attacks](levels.md#threats).
```
joshuagl, line 89:
```suggestion
Any software can introduce vulnerabilities into a supply chain, with recent high profile supply chain attacks proving how costly an attack can be. The requirements and checks that make up the SLSA framework were developed in direct response to [known supply chain attacks](levels.md#threats) with the aim to empower developers and software consumers to easily check the integrity of software artifacts in an automatically verifiable way.
```
Re-ordered a bit and added an emphasis on the desire for SLSA checks to be automatically verifiable (#130). Is that addition out of scope for this PR?

joshuagl (commented):
Thanks for the ping, I'm a bit behind after a long weekend AFK. This LGTM and in-line with the mission we discussed in the recent community meeting.

I added a suggestion to mention that we want SLSA to be automatically verifiable.

docs/index.md
msuozzo: I like the emphasis on automation. I think a framework for automated SSCI checking is a big selling point.

docs/index.md
```diff
@@ -50,6 +50,7 @@ Provenance checks to help evaluate risks and security
 ![Level 2](images/levelBadge2.svg)
 
 </div>
+<<<<<<< Updated upstream
```
inferno-chromium: Probably needs to be removed?

docs/index.md
```diff
@@ -59,6 +60,17 @@ Further checks against the origin of the software
...
+
+### Medium protection
+
+Further checks against the origin of the software
+
+</div>
+
+<div class="level">
+
+>>>>>>> Stashed changes
```
inferno-chromium: Probably needs to be removed?

docs/index.md
inferno-chromium: @devmoran - can you make this change, then the CL should be good to go.

inferno-chromium (approved):

PR #140, opened 2021-08-23
New supply chain threat diagram

devmoran:
- Updated supply chain threat diagram
- Updated supporting copy

image


inferno-chromium:
I like it. Adding @TomHennen @trishankatdatadog for formal review.

trishankatdatadog:
I really like this diagram. Very nice overview of the general types of things that can go wrong. Reminds me of the diagram we made at Datadog to try to illustrate all the things that could go wrong, too:

Secure_Delivery_DD_TUF+in-tutu_FIG2

One thing I might add is a _signing key compromise_, which could happen with developers and/or machines.

inferno-chromium:
> I really like this diagram. Very nice overview of the general types of things that can go wrong. Reminds me of the diagram we made at Datadog to try to illustrate all the things that could go wrong, too:
> 
> Secure_Delivery_DD_TUF+in-tutu_FIG2
> 
> One thing I might add is a _signing key compromise_, which could happen with developers and/or machines.

Thanks Trishank. Signing key compromise is F which is you compromise the signing key (/bypass build pipeline) and upload something malicious directly to package manager.

trishankatdatadog:
> Thanks Trishank. Signing key compromise is F which is you compromise the signing key (/bypass build pipeline) and upload something malicious directly to package manager.

Hmm, yes, covers machine signing keys, but what about developer signing keys? Say someone has stolen a developer's GPG keys, and is pretending to be the developer. Right now it's not covered by A, B, or C.

inferno-chromium:
> > Thanks Trishank. Signing key compromise is F which is you compromise the signing key (/bypass build pipeline) and upload something malicious directly to package manager.
> 
> Hmm, yes, covers machine signing keys, but what about developer signing keys? Say someone has stolen a developer's GPG keys, and is pretending to be the developer. Right now it's not covered by A, B, or C.

ah interesting, right now we don't seem to account for that. anyone else has thoughts, i think it should be around A, or atleast slightly mitigated by A with code reviews ?

docs/_sass/minima/_custom.scss
```diff
@@ -168,16 +199,16 @@ h4 {
...
         padding: 12px 18px;
 
         &:hover {
             text-decoration: none;
-            background: darken($brand-color, 5%);
+            background: $brand-color;
         }
     }
-}
+}spo
```
marcofranssen, line 211: woops

devmoran: 🤦 👍

06kellyjac:
> > Hmm, yes, covers machine signing keys, but what about developer signing keys? Say someone has stolen a developer's GPG keys, and is pretending to be the developer. Right now it's not covered by A, B, or C.
> 
> ah interesting, right now we don't seem to account for that. anyone else has thoughts, i think it should be around A, or atleast slightly mitigated by A with code reviews ?

Just to add to this

Unless you have a strong policy around securely managing keys e.g. mandating and reviewing keys only exist on a yubikey *and* also distribute the public key in other locations in addition to GitHub (or whatever other VCS server) you don't even need to steal the key for impersonation.
It just shows you have the password to the account because you can just change the key used for signing and the new one will show as valid/verified.


A much better summary: https://dlorenc.medium.com/should-you-sign-git-commits-f068b07e1b1f
cc @dlorenc

joshuagl:
Nice work making a more concise and less busy diagram, a few thoughts:
* Was the removal of the "source integrity" and "build integrity" indicators an explicit choice?
* E seems to have moved from the arrow between the dependency and build boxes to the dependency box itself.
* I'm not a huge fan of "consumer" over "user", but I don't have a good alternative to suggest.

TomHennen:
> > > Thanks Trishank. Signing key compromise is F which is you compromise the signing key (/bypass build pipeline) and upload something malicious directly to package manager.
> > 
> > 
> > Hmm, yes, covers machine signing keys, but what about developer signing keys? Say someone has stolen a developer's GPG keys, and is pretending to be the developer. Right now it's not covered by A, B, or C.
> 
> ah interesting, right now we don't seem to account for that. anyone else has thoughts, i think it should be around A, or atleast slightly mitigated by A with code reviews ?

I do think A (Submit Bad Code) covers compromise of developer keys (or passwords, thanks @06kellyjac).  If someone manages to steal these credentials they would presumably use them to Submit Bad Code.

I do also think that the mitigation for this is code review by another trusted party.  I don't think it's just a slight mitigation either, but perhaps I'm missing something?

trishankatdatadog:
> It just shows you have the password to the account because you can just change the key used for signing and the new one will show as valid/verified.

Correcto. Just compromising GitHub account can add/remove/swap GPG/SSH keys. The key distribution itself needs to be secured, preferably using [X.509](https://github.com/github/smimesign).

inferno-chromium:
> Nice work making a more concise and less busy diagram, a few thoughts:
> 
> * Was the removal of the "source integrity" and "build integrity" indicators an explicit choice?

Unsure if it makes sense to complicate, this seems ok.

> * E seems to have moved from the arrow between the dependency and build boxes to the dependency box itself.

I think it was a mistake in previous diagram, dependency itself can be bad.

> * I'm not a huge fan of "consumer" over "user", but I don't have a good alternative to suggest.

sure, can update in future commit if we have better ideas.

inferno-chromium (approved):

inferno-chromium:
I think all review comments are addressed. Would need an additional reviewer to +1, @trishankatdatadog / @joshuagl / others

devmoran:
@joshuagl thanks!
> * Was the removal of the "source integrity" and "build integrity" indicators an explicit choice?

It was, for (visual) simplicity's sake and that we don't reuse those 2 concepts of integrity elsewhere in the content. Happy to challenge that though if you think it helps make the diagram more understandable.

> * E seems to have moved from the arrow between the dependency and build boxes to the dependency box itself.

Good spot, will tweak

> * I'm not a huge fan of "consumer" over "user", but I don't have a good alternative to suggest.

Similar to the first point, we're using consumer elsewhere to differentiate between _types_ of user

trishankatdatadog (approved):
I think we could worse than to start with this. Great work!

Issue #111, opened 2021-07-26
Clarify the goals / use-cases for each level


MarkLodato:
Issues like #100 show us that we need to clarify what _specific_ things we expect users to get out of each level. Without this, we don't have a clear reason to add or remove a requirement at any particular level.

Example use cases for SLSA 1:

* Inventory all of the source and build systems used across a large organization.
* Prevent project maintainers from making mistakes in the release process, such as building from unsubmitted code.
* Adopt all the expected workflow changes now, as required by higher SLSA levels, while waiting for implementation (and security) to follow. For example, "enforcing" code review on all code changes, even though it can be bypassed by faking the provenance.

TomHennen:
Hey @brunodom @david-a-wheeler @joshuagl @marklodato @mlieberman85 @trishankatdatadog @zakgreant

If you have time can you take a look at @msuozzo's doc above and see if generally does what you're looking for?  If so a PR could be put together to formalize this take, and if not we can take another stab at it.

(Might be worth discussing at Wednesday's meeting)

trishankatdatadog:
@msuozzo @TomHennen are scenarios in the doc supposed to be the use cases @MarkLodato was describing above?

brunodom:
@msuozzo @TomHennen just thinking out loud: In enterprise, there's a lot of code produced by the System Integrators that may develop software components of a major application/service. So we should somehow include the levels 3rd party organizations being part of the supply chain and how to assess the downstream.

MarkLodato:
Sorry for the delay in reviewing. Overall LGTM! I added some comments, but once those get resovled, we should merge it into the site. (We'll have to figure out how to organize as we add more pages.)

> are scenarios in the doc supposed to be the use cases @MarkLodato was describing above?

Yes

zakgreant:
Sorry for the late review. I find that the document clarifies the goals and use cases for each level. I've given a detailed review and made minor suggestions for clarity.

joshuagl:
Apologies, I provided feedback in a couple of meetings where this came up but did not do so in a durable way. I like the document and agree we should merge it into the site.

Nice work @msuozzo !

MarkLodato:
I tried to address for v1.0 in #502. Please take a look. Unless I hear otherwise, I will consider that PR to fix this issue.

PR #141, opened 2021-08-27
Require config as code at levels L3+

### LINKED ISSUE #115: SLSA 1 & 2 should not require config-as-code
As currently worded, SLSA 1 unintentionally requires that the build steps were defined in some source material, i.e. "config as code". This is a problem because it excludes many build systems where such a mapping is not present or infeasible. Examples: [Tekton](https://tekton.dev) ([more info](https://github.com/in-toto/attestation/issues/43)), [LUCI](https://github.com/luci/), and [Google Cloud Build](https://cloud.google.com/build) (unless cloudbuild.yaml-based build triggers are used).

SLSA 1 and 2 were intended to be easily applicable to *any* build system without significant modification to how that system works. We should therefore alter the requirements to make it more clear that this mapping SHOULD be specified if present, rather than blocking something from reaching SLSA 1 or 2 if the mapping is not present or available.

In particular, the problematic requirements are [Identifies Source](https://slsa.dev/requirements#identifies-source) and [Identifies Entry Point](https://slsa.dev/requirements#identifies-entry-point). 

#100 attempted to fix this by removing the requirements altogether from SLSA 1 & 2. This was met with resistance since it weakened the spec too much.

My suggestion is to rework the requirements:
* At SLSA 1+, require that the provenance identify the **independent inputs** to the build. If it is config-as-code, this SHOULD be the source repo and entry point if available. Otherwise, it it is acceptable to list the build steps, even if they were ultimately derived from some other source.
* At SLSA 3+, require **config-as-code**, where the independent inputs are boiled down to a source location and entry point.

We may also need to move the [Version Controlled](https://slsa.dev/requirements#version-controlled) requirement to SLSA 3.

TomHennen:
* Require build instructions to be tracked at **all** levels.
  * Lets people know how a build was produced.
  * Should support all build methods without requiring architectural changes.
  * If we adopt [this document](https://docs.google.com/document/d/1fLtR5-AfauKlgTvQ1yRrupIZVnX3YqBhAjO_2X3d4qE/edit#heading=h.q2vn2vj1ydg5) as level motivation (#111) then we may only need to require build instructions at level 2+ or 3+.  We can easily relax this requirement later, while making the requirement more strict would be more difficult.
* Require config-as-code at L3+.
  * Allows people to review the details of how builds are performed.
  * Can require changes to how builds are performed to be reviewed by additional parties.
  * Tracks changes to how builds are produced over time.
  * Provides an easier method of writing policy to make sure an artifact followed the build steps you expect (just point at the build config in source control)
  * May require architectural changes in some builders or changes to how users interface with their build service (e.g. Tekton, LUCI, Cloud Build)
* Combine "identifies source" with "identifies entry point".
  * The 'identifies source' requirement was _already_ specific to the top-level build script which was the 'entry point'.
    So there doesn't seem to be a reason to have these as separate requirements.
  * 'Identifies source' might have had an overly broad name since it was only talking about where the top-level build script was, but could have been interpreted to mean "identifies **all** source" (which it did not!).
* Only require "identifies entry point" at L3+.
  * As discussed with 'config-as-code' above, not all build architectures support  singular entry-points that are defined in a script.
  * Requiring 'identifies build instructions' at all levels should cover cases where folks aren't using config-as-code.
* Small updates to provenance examples to make linking easier.

This PR _does not_ take a position on how/if requiring all sources or dependencies be recorded in the provenance. That should
be handled in a different PR addressing  a different issue.

Note #115 suggested the language "require that the provenance identify the independent inputs", but I found "independent inputs" to be fairly broad.  The rest of the text of #115 indicated that this was really about tracking what the build did, so "build instructions" seems to cover the spirit of #115.

fixes #115

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -443,8 +443,8 @@ used to drive the build including what source repo the the configuration was
 read from.
 
 Example:
-source repo: git URL + branch/tag/ref + commit ID
-entrypoint: path to config file(s) (e.g. ./.zuul.yaml) + job name within config
+- source repo: git URL + branch/tag/ref + commit ID
+- entrypoint: path to config file(s) (e.g. ./.zuul.yaml) + job name within config
     (e.g. envoy-build-arm64)
 
 <td><td><td>✓<td>✓
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -199,6 +199,14 @@ workstation.
 Examples: GitHub Actions, Google Cloud Build, Travis CI.
 
 <td> <td>✓<td>✓<td>✓
+<tr id="config-as-code">
+<td>Config as code
+<td>
+
+The build service configuration is driven from files stored in source control.
+
+Examples: zuul.yaml, cloudbuild.yaml, .github/workflows
+<td> <td> <td>✓<td>✓
 <tr id="ephemeral-environment">
 <td>Ephemeral environment
 <td>
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -203,7 +203,7 @@ Examples: GitHub Actions, Google Cloud Build, Travis CI.
 <td>Config as code
 <td>
 
-The build service configuration is driven from files stored in source control.
+The build service configuration is defined in files stored in source control.
 
 Examples: zuul.yaml, cloudbuild.yaml, .github/workflows
 <td> <td> <td>✓<td>✓
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -205,7 +205,15 @@ Examples: GitHub Actions, Google Cloud Build, Travis CI.
 
 The build service configuration is defined in files stored in source control.
 
-Examples: zuul.yaml, cloudbuild.yaml, .github/workflows
+When using config-as-code the <a href="identifies-entry-point">entry point</a>
+MUST identify the build service configuration.
+
+Examples:
+
+- cloudbuild.yaml:gcr.io/cloud-builders/docker
+- .github/workflows/build.yaml:build-npm
+- .zuul.yaml:envoy-build-arm64
+
 <td> <td> <td>✓<td>✓
 <tr id="ephemeral-environment">
 <td>Ephemeral environment
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -424,13 +424,13 @@ build does not use config-as-code it MAY list details of the build steps.
 
 Config-as-code example:
 
-```json
+```json5
 "recipe": {
   "type": "https://github.com/Attestations/GitHubActionsWorkflow@v1",
   // Identifies the source repo of the build config.
   "definedInMaterial": 0,
-  # Identifies the entrypoint within the source repo at the path
-  # .github/workflows/build.yaml, using the job "build".
+  // Identifies the entrypoint within the source repo at the path
+  // .github/workflows/build.yaml, using the job "build".
   "entryPoint": ".github/workflows/build.yaml:build",
 },
 …
@@ -442,10 +442,10 @@ Config-as-code example:
 
 Build steps example:
 
-```json
+```json5
 "recipe": {
-  # Build steps were provided as an argument. No `definedInMaterial` or
-  # `entryPoint`.
+  // Build steps were provided as an argument. No `definedInMaterial` or
+  // `entryPoint`.
   "type": "https://tekton.dev/chains/recipe/buildSteps@v1",
   "arguments": {
       "steps": [
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -425,11 +425,12 @@ If the build uses <a href="#config-as-code">config-as-code<a>, this SHOULD be th
 source repo and entry point of the build config (as in
 [the GitHub Actions example](https://slsa.dev/provenance/v0.1#github-actions)).
 
-If the build doesn't use config-as-code it MAY list the details of what it did
-as in
+If the build doesn't use config-as-code it MAY list the details of what it was
+asked to do (as in
 [the Google Cloud Build RPC example](https://slsa.dev/provenance/v0.1#cloud-build-rpc)
 or
-[the Explicitly Run Commands example](https://slsa.dev/provenance/v0.1#explicitly-run-commands).
+[the Explicitly Run Commands example](https://slsa.dev/provenance/v0.1#explicitly-run-commands)
+).
 
 <td>✓<td>✓<td>✓<td>✓
 <tr id="identifies-entry-point">
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -433,7 +433,7 @@ Build steps example:
 ```
 TODO
 ```
-
+<td>✓<td>✓<td>✓<td>✓
 <tr id="identifies-entry-point">
 <td>Identifies entry point
 <td>
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -415,10 +415,10 @@ provenance. This represents the entity that the consumer must trust. Examples:
 <td>Identifies build instructions
 <td>
 
-Identifies the top-level instructions used to execute the build.
+The provenance identifies the top-level instructions used to execute the build.
 
 The identified instructions SHOULD be at the highest level available to the build
-(e.g. if the build it told to run build.sh it should list build.sh and NOT the
+(e.g. if the build is told to run build.sh it should list build.sh and NOT the
 individual instructions in build.sh).
 
 If the build uses <a href="#config-as-code">config-as-code<a>, this SHOULD be the
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -424,14 +424,51 @@ build does not use config-as-code it MAY list details of the build steps.
 
 Config-as-code example:
 
-```
-TODO
+```json
+"recipe": {
+  "type": "https://github.com/Attestations/GitHubActionsWorkflow@v1",
+  // Identifies the source repo of the build config.
+  "definedInMaterial": 0,
+  // Identifies the entrypoint within the source repo at the path
+  // .github/workflows/build.yaml, using the job "build".
+  "entryPoint": ".github/workflows/build.yaml:build",
+},
+…
+"materials": [{
+   "uri": "git+https://the.repo/foo",
+   "digest": {"sha1": "abc123..."}
+}]
 ```
 
 Build steps example:
 
-```
-TODO
+```json
+"recipe": {
+  // Build steps were provided as an argument. No `definedInMaterial` or
+  // `entryPoint`.
+  "type": "https://tekton.dev/chains/recipe/buildSteps@v1",
+  "arguments": {
+      steps: [
+          {
+            "entryPoint": "",
+            "arguments": null,
+            "environment": {
+              "container": "git-source-repo-jwqcl",
+              "image": "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init@sha256:b963f6e7a69617db57b685893256f978436277094c21d43b153994acd8a01247"
+            },
+            "annotations": null
+          },
+          {
+            "entryPoint": "#!/usr/bin/env bash\nset -x\n\ncd $(inputs.resources.repo.path)\n\nbazel build --host_force_python=PY2 //package_manager:dpkg_parser.par\ncp bazel-bin/package_manager/dpkg_parser.par .\n\nbazel build //base:static_root_amd64_debian10.tar\n\ncp bazel-bin/base/static_root_amd64_debian10.tar .\n\nfind /workspace/repo\npwd\n\necho \"gcr.io/foo/bar\" > $(results.IMAGE_URL.path)\n",
+            "arguments": null,
+            "environment": {
+              "container": "build",
+              "image": "gcr.io/cloud-marketplace-containers/google/bazel@sha256:010a1ecd1a8c3610f12039a25b823e3a17bd3e8ae455a53e340dcfdd37a49964"
+            },
+            "annotations": null
+          }
+    ]
+}
 ```
 <td>✓<td>✓<td>✓<td>✓
 <tr id="identifies-entry-point">
@@ -439,8 +476,8 @@ TODO
 <td>
 
 The provenance identifies the "entry point" of the build service configuration
-used to drive the build including what source repo the the configuration was
-read from.
+used to drive the build including what source repo the configuration was read
+from.
 
 Example:
 - source repo: git URL + branch/tag/ref + commit ID
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -429,8 +429,7 @@ If the build doesn't use config-as-code it MAY list the details of what it was
 asked to do (as in
 [the Google Cloud Build RPC example](https://slsa.dev/provenance/v0.1#cloud-build-rpc)
 or
-[the Explicitly Run Commands example](https://slsa.dev/provenance/v0.1#explicitly-run-commands)
-).
+[the Explicitly Run Commands example](https://slsa.dev/provenance/v0.1#explicitly-run-commands)).
 
 <td>✓<td>✓<td>✓<td>✓
 <tr id="identifies-entry-point">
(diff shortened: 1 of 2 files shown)


docs/requirements.md
```diff
@@ -199,6 +199,14 @@ workstation.
 Examples: GitHub Actions, Google Cloud Build, Travis CI.
 
 <td> <td>✓<td>✓<td>✓
+<tr id="config-as-code">
+<td>Config as code
+<td>
+
+The build service configuration is defined in source control and drives the build automatically.
```
msuozzo, line 206: > "drives the build automatically"

I'm a little hesitant about requiring there to be automation triggering the build. I think it'd be fine to just limit build configuration to "use this workflow" regardless of how/when it's triggered.

docs/requirements.md
```diff
@@ -403,22 +411,81 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
 
-The provenance identifies the source containing the top-level build script, via
-an [immutable reference]. Example: git URL + branch/tag/ref + commit ID.
-
+Identifies the top-level instructions used to execute the build.
+
+The identified instructions SHOULD be at the highest level available to the build. 
+If the build uses <a href="#config-as-code">config-as-code<a>, this SHOULD be the
+source repo and entry point of the build config and NOT the build steps within the
+build config.  If the build does not use config-as-code it MAY list details of the
```
msuozzo, line 423: I feel like this MAY indicates we may not want to require it at lower levels, especially L1. One option would be to make a SHOULD at L1 as it may be more onerous to meet in some situations (e.g. inside build systems that don't expose the workflow to all steps). I think this might even mean our github actions demo wouldn't qualify since it's not always possible to access to correct workflow defintion in actions.

docs/requirements.md
```diff
@@ -403,22 +411,81 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
+"recipe": {
+  // Build steps were provided as an argument. No `definedInMaterial` or
+  // `entryPoint`.
+  "type": "https://tekton.dev/chains/recipe/buildSteps@v1",
+  "arguments": {
+      "steps": [
+          {
+            "entryPoint": "",
+            "arguments": null,
+            "environment": {
```
msuozzo, line 456: not thrilled with advocating for this syntax inside the spec. This was mainly just a way to get build steps encoded in the existing schema but I don't think the intent was for `environment` to permanently host this metadata.

I don't have great alternatives, though.

docs/requirements.md
```diff
@@ -403,22 +411,81 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
+              "container": "git-source-repo-jwqcl",
+              "image": "gcr.io/tekton-releases/..."
```
msuozzo, line 458: not really sure what the `...` is doing here. do you want to give a literal image name?

docs/requirements.md
```diff
@@ -403,22 +411,81 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
+            "environment": {
+              "container": "build",
+              "image": "gcr.io/cloud-marketplace-containers/..."
+            },
+            "annotations": null
+          }
+    ]
+}
+```
 <td>✓<td>✓<td>✓<td>✓
```
msuozzo, line 475: as noted above, this might be loosened to suggested at lower levels (maybe just L1)

msuozzo (changes requested):

docs/requirements.md
TomHennen: What information _do_ we have available about what the builder actually did?

docs/requirements.md
TomHennen: This was borrowed from https://github.com/tektoncd/chains/blob/main/PROVENANCE_SPEC.md and the actual images are _quite_ long.  I could include the whole thing?

docs/requirements.md
TomHennen: Can you expand on what you're worried about?  Maybe I could address that?

docs/requirements.md
msuozzo: Is there an alternative phrasing that avoids the implication of automation?

docs/requirements.md
msuozzo: afaiu `environment` is really meant to be "env variables" (as can be provided by gha or gcb) but this field was overloaded to provide the container/image by grafeas. advocating for this use in the spec doesn't feel ideal...

docs/requirements.md
msuozzo: I wasn't clear whether `...` was intentional or meant as an abbreviation. I'd probably include the full names and then `...` the hash values.

docs/requirements.md
msuozzo: In GCB, effectively nothing. In GHA, a pointer to the *file* that contains the workflow but IIUC no metadata on which workflow from that file.

docs/requirements.md
msuozzo: I also didn't understand the `container` field (and honestly I'm still not sure I have the right idea). Maybe rename to `name`? (and since this is shadowing the tekton doc, I guess suggest the same upstream?)

docs/requirements.md
TomHennen: I thought GCB had some API that could give you all sorts of data?  Or is that only available if you're using cloudbuild.yaml?  Does GCB provide _any_ of the things used in the build?  Source repos fetched?  Deps?

In GHA what if we were to just list the file?

In #115 Mark suggested that we should require this information at L1.

What if that were weakened to 'all the details the builder has available'?  That's fine with me, but I get the impression other folks want a lot more detail.

docs/requirements.md
TomHennen: So, I didn't mean to say "use this format" I meant this to be an example of how _a_ builder could include these details.  Builders could use whatever format they want, they'd just specify their own type (like how here we're showing 'https://tekton.dev/chains/recipe/buildSteps@v1').

How could I change this to make it clear the level of detail that's being requested without being prescriptive about format (since I _don't_ want to be prescriptive about format).

msuozzo: maybe move the `image` up to the top level and remove `container`? Feels like the most concise way to put it.

docs/requirements.md
msuozzo: > I thought GCB had some API that could give you all sorts of data?

It has some env variables which provide some information but no step metadata.

> Does GCB provide any of the things used in the build? Source repos fetched? Deps?

https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values#using_default_substitutions

All metadata is made available after the build is complete via their API.

> In #115 Mark suggested that we should require this information at L1.

That's not my reading of #115 . His wording feels a lot more like a SHOULD, not a MUST.

AFAIK, this sort of requirement was never discussed and imo shouldn't be given how it would prevent the in-build mechanism for L1 we've intended to support.

> What if that were weakened to 'all the details the builder has available'?

imo that's the same as SHOULD and I don't see a compelling reason to portray it as a requirement if it's going to be purposefully incomplete. For example, we handle all other cases of "best effort" here by leaving them blank.

> That's fine with me, but I get the impression other folks want a lot more detail.

At L1? I don't know of anyone who's advocated for this as a requirement at L1.

TomHennen: It does say 

"At SLSA 1+, **require** that the provenance identify the independent inputs"

In this PR "independent inputs" -> "build instructions" (as explained in the PR description that seems to make the most sense).

Now, we don't have to stick with what the issue suggests, but we should note a departure and why.

If we didn't change "independent inputs" to "build instructions" would GHA and GCB be able to specify their 'independent inputs'?  What would those inputs be?

docs/requirements.md
TomHennen: "The build service configuration is defined in source control and drives the build automatically."  -->

"The build service configuration is defined in source control and is executed by the build service."

?

msuozzo: maybe "and is used as the build entrypoint"?

docs/requirements.md
msuozzo: Okay so I'm going to walk back my previous position. I think going from the previous config-as-code requirement to this one does make some sense and ensures we have an accounting of the build process at all levels.

I still think a recommendation could work instead of this requirement but, versioning-wise, I think it'd be safer to leave the more restrictive requirement for now and potentially relax it later than having to do the reverse.

TomHennen: Gotcha.  Updated PR comment to reflect that.

docs/requirements.md
TomHennen: I've simplified the example a bit and made it refer to a _fake_ builder.

I definitely _don't_ want this to dictate the structure that various builders use to record this information.  So if you have some suggestions as to how I can make it clear this will vary from builder to builder that might be best.

docs/requirements.md
TomHennen: Followed you suggestion regarding ...

As mentioned above I've simplified this.

docs/requirements.md
TomHennen: I'm not that hot on using 'entrypoint' here because later we distinguish between 'entrypoint' (a singular field to reference) and 'steps' (details about how the build progressed) as being two different things.  Yet here _both_ of them count as the build instructions.

I think I'll go with "The build service configuration is defined in source control and is executed by the build service." and we can see what others think.

docs/requirements.md
```diff
@@ -403,22 +425,82 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
-<td>Identifies source
+<tr id="identifies-build-instructions">
+<td>Identifies build instructions
 <td>
 
-The provenance identifies the source containing the top-level build script, via
-an [immutable reference]. Example: git URL + branch/tag/ref + commit ID.
+Identifies the top-level instructions used to execute the build.
+
+The identified instructions SHOULD be at the highest level available to the build.
```
msuozzo, line 420: "highest level" is giving me problems. what do you mean by it?

docs/requirements.md
```diff
@@ -403,22 +425,82 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
+    "steps": [
+      {
+        "entryPoint": "",
+        "arguments": null,
+        "image":
+            "gcr.io/foo-releases/github.com/foobuilder/pipeline/cmd/git-init@sha256:b963...",
+      },
+      {
+        "entryPoint":
+            "#!/usr/bin/env bash\nset -x\n\ncd src\n\nbuild.sh",
```
msuozzo, line 461: i think entry point has to be just as single executable. maybe just bash since the args look like a bash invocation?

docs/requirements.md
```diff
@@ -403,22 +425,82 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
+  // Build steps were provided as an argument. No `definedInMaterial` or
+  // `entryPoint`.
+  "type": "https://foobuilder.dev/recipe/buildSteps@v1",
+  "arguments": {
+    "steps": [
+      {
+        "entryPoint": "",
+        "arguments": null,
+        "image":
+            "gcr.io/foo-releases/github.com/foobuilder/pipeline/cmd/git-init@sha256:b963...",
```
msuozzo, line 457: nit: maybe "git-checkout" instead of "git-init"

docs/requirements.md
```diff
@@ -403,22 +411,82 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
+            "gcr.io/cloud-marketplace-containers/google/bazel@sha256:010a1e..."",
+      }
+    ]
+  }
+}
+```
 
 <td>✓<td>✓<td>✓<td>✓
 <tr id="identifies-entry-point">
 <td>Identifies entry point
```
msuozzo, line 476: I think this section name needs to change. "Entry point" is used to describe both config as code and in-Providence build instructions so using it to exclusively refer to the former is confusing.

docs/requirements.md
TomHennen: Clarified.

docs/requirements.md
TomHennen: Obsolete.  Removed inline examples.

docs/requirements.md
TomHennen: Obsolete.  Removed inline examples which conflicted with this 'entry point'.

msuozzo (approved):
Looks good!

inferno-chromium:
@trishankatdatadog @joshuagl @brunodom @david-a-wheeler @mlieberman85 @zakgreant - thoughts on this. we would definitely benefit from 1-2 additional eyes on this requirement review.

joshuagl (approved):
I made a couple of minor suggestions, but am happy to approve as is. This clarifies a common source of confusion for folks reading SLSA.

docs/requirements.md
```diff
@@ -403,22 +411,42 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
 <td>✓<td>✓<td>✓<td>✓
-<tr id="identifies-source">
-<td>Identifies source
+<tr id="identifies-build-instructions">
+<td>Identifies build instructions
 <td>
 
-The provenance identifies the source containing the top-level build script, via
-an [immutable reference]. Example: git URL + branch/tag/ref + commit ID.
+Identifies the top-level instructions used to execute the build.
```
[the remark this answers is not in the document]
TomHennen: Done.

docs/requirements.md
```diff
@@ -403,22 +411,42 @@ provenance. This represents the entity that the consumer must trust. Examples:
...
+
+The identified instructions SHOULD be at the highest level available to the build
+(e.g. if the build it told to run build.sh it should list build.sh and NOT the
```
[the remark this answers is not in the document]
TomHennen: Good catch, thanks!

docs/requirements.md
```diff
@@ -199,6 +199,14 @@ workstation.
 Examples: GitHub Actions, Google Cloud Build, Travis CI.
 
 <td> <td>✓<td>✓<td>✓
+<tr id="config-as-code">
+<td>Config as code
+<td>
+
+The build service configuration is defined in source control and is executed by the build service.
```
squee1945, line 206: I'm trying to understand what this means for something like GCF where the build definition is entirely specified by the service itself (GCF), but is not part of the user's source control at all.

(I realize there are source provenance questions for GCF, but set those aside for the moment.)

TomHennen: GCF just builds/runs whatever is in a specific repo right?

My suggestion:

1. Have a recipe.type specific to GCF 'http://google.com/cloud/gcf_recipe/v1' (or whatever)
2. Set recipe.entrypoint to 'gcf_implicit' (or whatever)
3. Put the repo that GCF actually built in materials (should be doing this anyways)
4. Set recipe.definedInMaterial to the index of the repo in #3.

Thoughts?

squee1945: GCF can build a repo, or an arbitrary zip file.

I'm not familiar with the `recipe.*` things - can you point me to a reference?

TomHennen: https://slsa.dev/provenance/v0.1#recipe.type

If GCF is building from a zip file it wouldn't qualify at SLSA Level 3 as SLSA Level 2 requires the use of source control https://github.com/slsa-framework/slsa/blob/main/docs/requirements.md#version-controlled

_but_

The way I'd represent building a zip file in the provenance for GCF would probably be very similar:

1. Have a recipe.type specific to GCF 'http://google.com/cloud/gcf_recipe/v1' (or whatever)
2. Set recipe.entrypoint to 'gcf_implicit' (or whatever)
3. Put the path of the zip file and it's hash that GCF actually built in materials (should be doing this anyways)
4. Set recipe.definedInMaterial to the index of the zip file in 3.


So the provenance would basically look the same, but the artifacts built from zip files shouldn't qualify for anything higher than SLSA L1.

squee1945: Okay, so the `recipe.*` parts are within the provenance.

My question is more specifically about the phrase "build service configuration is defined in source control" - in GCF, the build configuration is *not* defined in source control - even in the case of the source being in version control.

TomHennen: Right.  I think since GCF is basically a fixed function in that regard (you can think of the build configuration as being as simple as 'gcf <repo>') that it would be completely fine to take the approach I outlined in https://github.com/slsa-framework/slsa/pull/141#discussion_r716932691

TomHennen: I think the way to think about it is "With GCF the configuration is defined in the source repo being built."  So all you have to do is point to that repo and say "GCF built this".

PR #148, opened 2021-09-01
Add shortcut to @ committee members

TomHennen:
This should make it easier to get feedback where needed.


inferno-chromium (approved):
feel free to order if needed, alphabetically is fine too or just like in list above.

Issue #129, opened 2021-08-10
Better define "source"


MarkLodato:
Right now the term "source" is ambiguous and it is leading to confusion and disagreements, such as #127. Different people have different ideas in their head about what "source" means and how it controls on source can be implemented. In particular, I believe many readers consider it the primary "code" that is compiled, whereas @TomHennen and I had considered it the "build configuration" since there is no technical means to verify what was the actual "code." Once we clarify, I think it will be much easier to make decisions such as #127.

My proposal is to model a build as having only the following inputs (rough draft):

- **Build Interpreter:** The software that interprets the build configuration.
  - Examples: GitHub Actions (GHA), Google Cloud Build (GCB)
- **Build Configuration** (Steps? Recipe?): The way the build is configured for this particular project, usually as a list of steps.
  - Examples: GHA workflow definition at .github/workflows/build.yml within a particular repo, GCB steps defined in a trigger (not in version control).
- **Build Parameters:** Additional user-provided inputs that affect the output of the build but that are not part of the configuration.
  - Examples: GHA `inputs` for workflow_dispatch events, GCB `substitutions`.
  - Note: If the parameter cannot affect the output, it is not listed. For example, an option that sets scheduling priority would not be considered an input to the build since it doesn't affect the output.
- **Top-Level Source:** The primary input artifact to the build. This is usually a single source revision that "triggered" a CI build.
  - Note: There is no technical means to ensure that this *really* was the source. In some cases, the top-level source only contains build configuration and the actual true "code" comes from a dependency and looks like a library. Examples: Chrome [LUCI](https://luci.github.io/), [curl Alpine package](https://slsa.dev/example).
- **Build Environment:** Software artifacts that are provided by the build system that the user has no control over. Ideally this is a single VM base image or container image, which may have its own provenance.
- **Dependencies:** Software artifacts that were fetched during the build as requested by the build configuration, the top-level source, or another dependency.
  - A build MAY further differentiate between **libraries** and **build tools**. This may help prioritize between different dependencies. However, there is no technical means to differentiate between these, so consumers SHOULD consider this "best effort."

Alternatives considered:
- [SPDX Relationships](https://spdx.github.io/spdx-spec/7-relationships-between-SPDX-elements/): I'm not sure they fit cleanly into our model.
- @dlorenc had previously suggested a multi-layer model of builds in some GitHub issue, though I can't find it at the moment.

Once we decide this, we'll need to update the Provenance schema to match.

[...]

PR #149, opened 2021-09-03
Add "Identifies Source Code" requirement

TomHennen:
As discussed in #129 this requires source repositories to be identified at level 2+.

Doesn't address dependencies yet...

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -461,7 +461,8 @@ authenticated by the builder.
 At level 3+ this information MUST be authenticated by the builder (i.e. the
 builder either needs to have fetched the source itself or _observed_ the fetch).
 
-NOTE: At levels 3 and below the listed source repositories may not be complete.
+At level 4 this information MUST be complete (i.e. all source repositories used
+in the build are listed).
 
 <td><td>✓<td>✓<td>✓
 <tr id="identifies-entry-point">
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -465,7 +465,7 @@ builder either needs to have fetched the source itself or _observed_ the fetch).
 At level 4 this information MUST be complete (i.e. all source repositories used
 in the build are listed).
 
-<td><td>✓<td>✓<td>✓
+<td><td>✓<td>✓ (Authenticated)<td>✓ (Complete)
 <tr id="identifies-entry-point">
 <td>Identifies entry point
 <td>
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -446,6 +446,24 @@ or
 [the Explicitly Run Commands example](https://slsa.dev/provenance/v0.1#explicitly-run-commands)).
 
 <td>✓<td>✓<td>✓<td>✓
+<tr id="identifies-source-code">
+<td>Identifies source code
+<td>
+
+The provenance identifies repositories the source code used came from.
+
+The identified repositories SHOULD only include source used directly in the build.
+The source of dependencies SHOULD NOT be included.
+
+At level 2 this information MAY come from users and DOES NOT need to be
+authenticated by the builder.
+
+At level 3+ this information MUST be authenticated by the builder (i.e. the
+builder either needs to have fetched the source itself or _observed_ the fetch).
+
+NOTE: At levels 3 and below the listed source repositories may not be complete.
+
+<td><td>✓<td>✓<td>✓
 <tr id="identifies-entry-point">
 <td>Identifies entry point
 <td>
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -450,7 +450,8 @@ or
 <td>Identifies source code
 <td>
 
-The provenance identifies repositories the source code used came from.
+The provenance identifies the repository origin(s) for the source code used in
+the build.
 
 The identified repositories SHOULD only include source used directly in the build.
 The source of dependencies SHOULD NOT be included.


docs/requirements.md
```diff
@@ -446,6 +446,24 @@ or
...
+The identified repositories SHOULD only include source used directly in the build.
+The source of dependencies SHOULD NOT be included.
+
+At level 2 this information MAY come from users and DOES NOT need to be
+authenticated by the builder.
+
+At level 3+ this information MUST be authenticated by the builder (i.e. the
+builder either needs to have fetched the source itself or _observed_ the fetch).
+
+NOTE: At levels 3 and below the listed source repositories may not be complete.
```
msuozzo, line 464: Maybe "At level 4+ this information MUST be complete." to keep with the above pattern

TomHennen: Done.

docs/requirements.md
```diff
@@ -446,6 +446,24 @@ or
 [the Explicitly Run Commands example](https://slsa.dev/provenance/v0.1#explicitly-run-commands)).
 
 <td>✓<td>✓<td>✓<td>✓
+<tr id="identifies-source-code">
+<td>Identifies source code
+<td>
+
+The provenance identifies repositories the source code used came from.
```
msuozzo, line 453: "The provenance identifies the repository origin(s) for the source code used in the build."

TomHennen: Done

docs/requirements.md
```diff
@@ -446,6 +446,26 @@ or
...
 
 <td>✓<td>✓<td>✓<td>✓
+<tr id="identifies-source-code">
+<td>Identifies source code
+<td>
+
+The provenance identifies the repository origin(s) for the source code used in
+the build.
+
+The identified repositories SHOULD only include source used directly in the build.
```
inferno-chromium, line 456: This sentence is getting vague, next line says source of dependencies, whereas this line does not say dependencies. Do we want link to just first-party source, or all (firstparty code and all its dependency urls ?)

TomHennen: For this specific requirement I think we only want source and not dependencies.

"First party source" isn't quite right because the dependencies used in the build could also be 'first party' and we don't want _that_ source included in this provenance (it should be included in the provenance for the dependency).

msuozzo: I agree with abhishek that the concept isn't quite clear as written. To my mind, "source used directly in the build" doesn't disambiguate the concept from a source dependency or build tool. I like the level progression you've written up by I don't think I'd have enough clarity on how to apply it with the current text.

I'd suggest adding some sort of rule-of-thumb to allow the user to make the distinction: "If the source or compiled derivatives thereof appear in the final artifact, it's 'source code' as discussed here. Otherwise, it's a dependency."

trishankatdatadog (approved):

inferno-chromium (commented):
Need slight changes to clarify terminology as per comment below.. This will be nice to have before v0.1

docs/requirements.md
```diff
@@ -446,6 +446,26 @@ or
...
+The source of dependencies SHOULD NOT be included.
+
+At level 2 this information MAY come from users and DOES NOT need to be
```
msuozzo, line 459: (This is just a cosmetic note but one that we may need to address across the ladder. I'm fine with this turning into an issue and this PR going in without a resolution.)

This single line item has interior requirements that apply across different levels. Personally, I think this style has its advantages but I think the existing pattern makes more sense for the table structure currently in use.

docs/requirements.md
TomHennen: "If the source or compiled derivatives thereof appear in the final artifact, it's 'source code' as discussed here. Otherwise, it's a dependency."

I'm not sure this solves the problem, is a Debian source 'package' considered source or a dependency?  What about a compiled C library that winds up getting included in the final artifact?  Both of them wind up with "compiled derivatives" in the final artifact.  Yet they should very likely be considered dependencies and not 'source'.

My suggestion: "if the build fetched artifacts from source control systems then those artifacts are considered 'source'".

This has an important property that it allows this new "Identifies source code" requirement to be automatically verified.  Process: check the 'materials' section of the provenance for materials fetched from source control systems. E.g. check for materials that start with any of the VCS prefixes identified in [SPDX Package Download location 3.7.4](https://spdx.github.io/spdx-spec/3-package-information/#37-package-download-location).  If a provenance doesn't have any materials with VCS prefixes then it fails this requirement.

docs/requirements.md
TomHennen: If you'd like it to be split up, that's fine, but do you have any suggestions for what the names of the different requirements should be?

msuozzo: No and that's why I'm not crazy about breaking them up :)

Adding line items for each sub-requirement becomes tricky quickly and introduces more opportunities for reader confusion.

One option would be to add the complexity to the check marks as it done with retention history:

L1 | L2 | L3 | L4
----|----|----|---
|  | ✓ | ✓ (Authenticated) | ✓ (Complete)

docs/requirements.md
MarkLodato: I agree that we need to update the terminology and have several examples with common build services. My recommendation:

* Source: primary "top-level" input artifact. Most build services have a notion of a "source" and this should reflect that.
* Build Config: top-level build definition / configuration. Sometimes this is the same as Source.
* Dependency: anything else.

Examples:

Service | Source | Build Config | Dependencies
--- | --- | --- | ---
GitHub Actions | git commit containing the workflow definition | (same as source) | dependent actions, runner vm, anything fetched during the build
GCB (build-as-code) | git commit containing the cloudbuild.yaml | (same as source) | ...
GCB (config in UI) | git commit that was triggered | serialization of the config | ...

TomHennen: I'm not sure this works.  A number of build systems (e.g. Yocto) have the build-as-code config fetch from dozens of source repos and then start the build on that code.  With this recommendation it seems like all those source repos would count as dependencies, and that might lead people to believe the source requirements don't apply to them (I think we want them to).

What if, instead of trying to update terminology here I just update the text to something like:

> The provenance identifies the repository origin(s) for the source code processed by the build.

> The identified repositories SHOULD only include source processed (e.g. compiled) directly in the build. The source of dependencies processed in prior builds SHOULD NOT be included.
> For example: if the build fetches git.repo/foo, git.repo/bar, and npm package baz (built from git.repo/baz), then the provenance SHOULD list git.repo/foo and git.repo/bar and SHOULD NOT list git.repo/baz.

docs/requirements.md
TomHennen: I like that.  Done.

docs/requirements.md
inferno-chromium: @MarkLodato @TomHennen , will let you guys sync and then update terminology definition in another PR, still good to do this before v0.1 cut tmrw.

inferno-chromium (approved):
Terminology update can come in another PR, since this is reviewed by Matt, Trishank, merging.

Issue #177, opened 2021-10-06
Consider adding `steps` to `slsa-provenance`


priyawadhwa:
We have a `steps` field separate from `invocation` in `tekton-provenance`, and I think it clearly separates what started the build from what happened in the build. Especially if we do #175 , I think this would be nice to have as well. Opening this for discussion.

TomHennen:
Can you document why it would be nice to have this as a top level(?) field independent from recipe/invocation?  I.e. why not just define a custom recipe/invocation type and specify the steps in the arguments field?

priyawadhwa:
Yup, so this would go alongside the change described [here](https://github.com/slsa-framework/slsa/issues/175#issuecomment-937008279) where we would rename `recipe` to `invocation`, which corresponds to "what kicked off the build". 

We could then have a `recipe.steps` field (as we do in tekton) or just a `steps` field, which would correspond to "what happened during the build".

joshuagl:
As Tom suggested, `recipe.type` is a [`TypeURI`](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#TypeURI) and thus this _could_ be handled with a custom arguments object.

Did Tekton decide that was undesirable when designing the Tekton predicate?

dlorenc:
> As Tom suggested, `recipe.type` is a [`TypeURI`](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#TypeURI) and thus this _could_ be handled with a custom arguments object for the Tekton `recipe.type`.

I think this was changed after the Tekton format got created, but I can't remember the exact timing.

Issue #175, opened 2021-10-06
Provenance: Split up current `recipe` into `recipe.steps` and `invocation`


priyawadhwa:
EDIT: I changed the title of the issue to more accurately reflect the conversation below

Wanted to open an issue for discussion. We have this in the [tekton-provenance](https://github.com/tektoncd/chains/blob/main/PROVENANCE_SPEC.md) format and have received feedback that it's a bit more intuitive.  

This would happen alongside https://github.com/slsa-framework/slsa/issues/177, and we would be splitting up the current `recipe` into `invocation`, which would describe "what kicked off the build", and `recipe.steps`, which would describe "what happened during the build". While it is _posible_ to do this with the current `recipe` format, I believe splitting it up is better organized and so more optimal.

I'm happy to open a PR to make this change.

trishankatdatadog:
No strong feelings either way 🙂 Should we have a vote?

dlorenc:
I like separation. The invocation was meant to contain information about why the run was started. An event ID, timestamp, caller ID, anything in the overall "request" that started the build (including a pointer to the recipe via URI).

The recipe section should contain details about what happened, while invocation contains the "why".

inferno-chromium:
Looks like we have concensus on this.

dlorenc:
+1 @priyawadhwa , a lot of these only make sense when combined.

trishankatdatadog:
Yes, pls, one issue/PR, thanks

priyawadhwa:
Opened up https://github.com/slsa-framework/slsa/pull/179 for this issue

inferno-chromium:
> To clarify my position, I agree with the sentiment (separate "what" from "why") but believe that this is what we already have. Thus, the change does not makes sense to me.
> 
> In particular:
> 
> * The "why" is already captured in `metadata`. I don't understand the purpose of the additional fields.
> * The `entryPoint`, `arguments`, and `reproducibility` are proposed to move under the new `invocation` field (i.e. "why"), but they really are part of the recipe. They describe what was performed.

It is answered by @priyawadhwa, @dlorenc and @zakgreant in comments above, it is easier for understanding and for review. Nobody is saying here that it is not possible in slsa provenance, it is more about how to make slsa provenance format more intituitive and easy to use. 

I think we should use @david-a-wheeler 's wisdom here and feedback here. David, wdyt of the two formats and comments in this bug.

PR #179, opened 2021-10-08
Proposal for v0.2-alpha

### LINKED ISSUE #176: Merging tekton-provenance and slsa-provenance predicate formats
I'm opening a tracking issue where we can keep track of progress on this and consolidate opinions. Some things to consider:
- [x] https://github.com/slsa-framework/slsa/issues/175
- [x] https://github.com/slsa-framework/slsa/issues/177
- [x] https://github.com/slsa-framework/slsa/issues/178
- [ ] https://github.com/tektoncd/chains/issues/249 in Chains

Feel free to add anything I forgot! 

cc @loosebazooka @MarkLodato @inferno-chromium @bobcatwilson @TomHennen @msuozzo

### LINKED ISSUE #178: Provenance: Consider simplifying `definedInMaterial`
_[Forked from Issue #176]_

There has been some confusion about `definedInMaterial`, so it's worth considering whether there is a better design.

While we're at it, we may also want to address the issue that some builds may have recipe defined in more than one material.

### Current design: `definedInMaterial` index

Currently, all of the input artifacts are listed in `materials`, and `recipe.definedInMaterial` identifies which of these contained the recipe definition.

Benefits:

- Verifiers can handle all input artifacts identically. No need to special-casing the recipe (though they still can if they want to.)
- Reduce duplication in case a single artifact is both a recipe and a build input, such as in the case of GitHub Actions.

Drawbacks:

- More complex to read and write. Multiple implementers have already been confused by this, which highlights the need to improve.
- Only a single material (index) can be used.
- Index 0 means something different than null/undefined, which is undesirable since it's common for null and 0 to mean the same thing.
- Index can be out of range.

### Alternative 1: `definedInMaterial` URI

Make `definedInMaterial` a URI instead of an index. The semantics would be the same, except that you match by `materials[*].uri == recipe.definedInMaterial` instead of `materials[recipe.definedInMaterial]`.

Compared to current design:

- Pro: URI is more intuitive than index.
- Pro: Empty string can now mean the same as null/undefined.
- Con: If two materials have the same URI, the semantics are undefined. So we'd need to add a restriction that the recipe has to point to a unique URI. I love this particular restriction because there could be some edge case where you do want to have two URIs.
- Con: (minor) Validating and matching is slightly more difficult, since it's not just a simple array bounds check.

### Alternative 2: `recipe.definedIn` artifact reference

Instead of referring to a material, embed the artifact reference within `recipe`. That is, `"definedIn": {"url": "...", "digest": {..}}` (property name TBD). The value would be the same format as `materials`. If the same material is used as both an input to the build and defining the recipe, it gets listed in both locations.

Compared to current design:

- Pro: Makes `recipe` easier to understand and less prone to mistakes. (I think?)
- Pro: Removes possibility of using an out-of-bound index.
- Con: Duplication between `recipe` and `materials` in the common case.
- Con: Not always clear what is a `material` and what is a `recipe`. (Though we kind of have that situation now.)

### Other options?

Any other ideas? Opinions?

priyawadhwa:
This PR adds in the following changes:

1. Rename `recipe` to `invocation` as a higher level concept. Everything in invocation should be checked by policy verification. It contains the `configSource`, `environment`, and `parameters` (formerly `arguments`)
2. Rework `configSource` (formerly `definedInMaterial`) to inline the `entrypoint`, `uri`, and `digest`
3. Rename `recipe` to `buildConfig` which can be a top-level field, where the schema is dependent on `buildType`
4. Add in a high level `buildType` URI which determines the format of `invocation`, `buildConfig`, and `materials`

Closes https://github.com/slsa-framework/slsa/issues/175
Closes https://github.com/slsa-framework/slsa/issues/177
Closes #178 
Closes #176 

Signed-off-by: Priya Wadhwa <priyawadhwa@google.com>


inferno-chromium:
> Two high-level issues:
> 
> 1. I disagree with the rename of `invocation` (see #175). Let's discuss there further.
> 2. ~We need to create a new v0.2-alpha (or similar) for breaking changes. Updates to v0.1 must only have backwards-compatible changes.~ _Edit:_ Oops, I see that you mentioned this in the PR description. Could you please mark this as a draft PR to indicate that it's not ready to merge? Thanks!

There is no more work from author until furthur review. DRAFT is the wrong state for github PR here, it is ready for review (And don't worry, we don't merge things like these without steering member discussion). Also, PR discussion can also happen directly here as well, no need of "Recommended changes" forcing directly here. We should allow other steering members to chime their opinion as well here - @brunodom @david-a-wheeler @joshuagl @marklodato @mlieberman85 @trishankatdatadog @zakgreant

trishankatdatadog:
Thanks for presenting this today, Priya! A few questions, please:
* Why are there dup parameters+environments in your proposed v0.0.2-alpha SLSA predicate?
* Inside “predicate” and “metadata”
* There’s “materials” but where is “products"?

Issue #180, opened 2021-10-11
Define "strongly authenticated" in the docs


joshuagl:
The requirements refer twice to strongly authenticated users, but do not indicate what we mean by strong authentication.

trishankatdatadog:
For git repos, commit signing should indeed be used.

For other revision control systems (RCS), perhaps the RCS themselves serve as a root of trust, if they support it. Otherwise, it's pretty much blind faith (with respect to exactly who committed what and when) except for the immutable history.

joshuagl:
I wonder whether for "strong authentication" specifically we could add a definition that includes the need to map back to a specific person in a way that is resistant to account/credential compromise, i.e. signed commits or MFA with a hardware token or authenticator application (related #154).

More generally (and outside of the scope of this issue) more discussion of the threats per your WIP PR #92 @MarkLodato I think would be great.
I also have been thinking that a practical, living, document describing how to achieve SLSA compliance with GitHub would be useful for a lot of folks (I'd be happy to work on this item in a separate issue).

MarkLodato:
> I also have been thinking that a practical, living, document describing how to achieve SLSA compliance with GitHub would be useful for a lot of folks (I'd be happy to work on this item in a separate issue).

That would be great!

MarkLodato:
I believe this was fixed by #181. Please reopen if not.

PR #181, opened 2021-10-14
Define strong authentication

joshuagl:
Per the discussion in #180, an attempt at defining strong authentication.


dlorenc:
Thanks @joshuagl this looks like a great start!

MarkLodato:
I suggest we submit this now. It is only a clarification, not a requirements change, and it already has two steering members' approval (@joshuagl and mine). We can always further refine it in subsequent PRs.

PR #197, opened 2021-10-28
Provenance spec: rename to 0.2-draft, add warning, link between versions

MarkLodato:
* Rename 0.2-alpha to 0.2-draft to make it more clear that it is unstable. (Suggested by @joshuagl.)
* Add an explicit warning that 0.2 draft spec is subject to change.
* Add links to the other versions of the spec to allow readers to easily jump between versions.
* Remove boilerplate from the introductory material so that it is easier to read.


priyawadhwa (approved):
LGTM

joshuagl (approved):
LGTM, thanks!

PR #207, opened 2021-11-04
Render the threat lists as a collapsible table

MarkLodato:
Improve the display of threats so that it is easier to read by displaying as table rows, where each row is clickable to get details. Also highlight the SLSA level that addresses each threat.

Preview: https://marklodato.github.io/slsa/threats (The "view file" on GitHub doesn't use CSS, so it doesn't look as pretty)

image

### DIFF
--- a/docs/_sass/minima/_custom.scss
+++ b/docs/_sass/minima/_custom.scss
@@ -263,3 +263,25 @@ h4 {
         }
     }
 }
+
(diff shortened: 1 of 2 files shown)


joshuagl (approved):
Nice! The readability improvement here is significant. Makes the CSS and HTML tags worthwhile.

Issue #209, opened 2021-11-05
Help required on reviewing this PR regarding provenance materials


marcofranssen:
Would be great to get some community feedback from the experts.

https://github.com/philips-labs/slsa-provenance-action/pull/72

@MarkLodato would this be something we should support?

MarkLodato:
Left comments on the PR.

PR #210, opened 2021-11-05
Provenance v0.2: use enums instad of booleans.

MarkLodato:
Convert the existing `completeness.*` and `reproducibility` booleans into enums. This conveys more information, allows expressing "known incomplete" vs "unknown", allows future extensions, and aligns with SPDX 3.0.

Thanks @iamwillbar for the suggestion!

Design note: There is no standard term for "not reproducible". I went with IRREPRODUCIBLE because it mirrors INCOMPLETE (the other enum), it is the most common on [Google Books], and it doesn't risk typos between NON_REPRODUCIBLE and NOT_REPRODUCIBLE.

Any thoughts - does this increase usability? @marcofranssen FYI since you're looking as part of #209. @trishankatdatadog you brought this up in #179  - does this help?

[Google Books]: https://books.google.com/ngrams/graph?content=irreproducible%2Cnon-reproducible%2Cnonreproducible%2Cunreproducible&year_start=1800&year_end=2019&corpus=26&smoothing=3&direct_url=t1%3B%2Cirreproducible%3B%2Cc0%3B.t1%3B%2Cnon%20-%20reproducible%3B%2Cc0%3B.t1%3B%2Cnonreproducible%3B%2Cc0%3B.t1%3B%2Cunreproducible%3B%2Cc0

### DIFF
--- a/docs/provenance/v0.2-draft.md
+++ b/docs/provenance/v0.2-draft.md
@@ -97,11 +97,11 @@ See [Example](#example) for a concrete example.
       "buildStartedOn": "<TIMESTAMP>",
       "buildFinishedOn": "<TIMESTAMP>",
       "completeness": {
-        "parameters": true/false,
-        "environment": true/false,
-        "materials": true/false
+        "parameters": "<ENUM>",
+        "environment": "<ENUM>",
+        "materials": "<ENUM>"
       },
-      "reproducible": true/false
+      "reproducible": "<ENUM>"
     },
     "materials": [
       {
@@ -222,8 +222,8 @@ of the other top-level fields, such as `subject`, see [Statement]._
 >
 > This is an arbitrary JSON object with a schema defined by `buildType`.
 >
-> This is considered to be incomplete unless `metadata.completeness.parameters`
-> is true. Unset or null is equivalent to empty.
+> Unset or null is equivalent to empty. Whether or not this is considered to be
+> complete is indicated by `metadata.completeness.parameters`.
 
 <a id="invocation.environment"></a>
 `invocation.environment` _object, optional_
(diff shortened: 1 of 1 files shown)


joshuagl (approved):
I like it, nice suggestion.

marcofranssen:
To me this looks to be adding more complexity as well more combinations of possible outcomes. What would be the value? What are the practical usecases that need this feature? 

Adding this to the spec makes it more complicated because there will be more options to consider. 

E.g. when attesting the provenance there are more combo's to think about? Is that something which will add value and how does it add this value?

Trying to find the balance in what is needed vs simplicity and useability.

MarkLodato:
> To me this looks to be adding more complexity as well more combinations of possible outcomes. What would be the value? What are the practical usecases that need this feature?

Good point. Upon further thought, I'd rather defer this until v0.3 since it's not obviously a win. I'd be interested to hear other opinions as well.

The idea was to make future extensibility easier. For example, we may want to add a future value to indicate why it's incomplete. But you're right - I can't think of a concrete case where that would be valuable, and I also agree that an enum is more likely to get messed up. For example, consumers _should_ always check for `KNOWN_COMPLETE` rather than checking "not `KNOWN_INCOMPLETE`" and _should_ always treat unrecognized values as `UNKNOWN`. This doesn't seem worth the extra risk, at least not without more discussion.

On a related note, this made me reconsider the purpose of these fields in the fist place. I opened #211 to discuss and generate ideas.

So I'll close this PR to indicate that I don't intend to submit, but we can continue discussion here or on #211.

dlorenc:
The Kubernetes API design document is a great reference on this: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

They cover booleans vs. enums in detail:

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

Enums are one alternative, another one would be some kind of list of "claims", similar to the "Status" field in Kubernetes objects. List elements could consist of a mix of custom and well known values (should probably be namespaced), but here's an idea:

```shell
buildType:
- FullyReproducible
- DependenciesComplete
```

This reduces the need for Reproducible and Complete to be special-cased, top-level fields.

MarkLodato:
Thanks, Dan! That is a good idea and super helpful. @iamwillbar also mentioned something similar. I suggested it on #211.

PR #212, opened 2021-11-08
Provenance 0.2: add pseudocode to migrate from 0.1

MarkLodato:
This makes it easier to understand the mapping between versions.

@marcofranssen Does this answer your question?

### DIFF
--- a/docs/provenance/v0.2-draft.md
+++ b/docs/provenance/v0.2-draft.md
@@ -543,6 +543,39 @@ Execution of arbitrary commands:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


marcofranssen:
@MarkLodato really nice approach on showcasing the diff. I will start migrating the provenance action and come back once I finish that PR.

joshuagl (approved):
LGTM. I really like this use pseudocode to indicate the changes

MarkLodato:
Thanks again @marcofranssen for the idea, and thank you both for the review.

PR #218, opened 2021-11-10
Make provenance specification more accessible

Brend-Smits:
The provenance specification v0.2 is now accessible via the navigation bar on the website.

Co-authored-by: Marco Franssen <marco.franssen@philips.com>
Signed-off-by: Brend Smits <brend.smits@philips.com>

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -7,6 +7,7 @@ header_pages:
   - levels.md
   - requirements.md
   - use-cases/index.md
+  - provenance/v0.2.md
   - roadmap.md
   - getinvolved.md
 exclude:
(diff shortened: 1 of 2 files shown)


MarkLodato:
Thanks! I think "Specification" is ambiguous since the level definitions are also a specification. What about "Provenance"?

Brend-Smits:
> Thanks! I think "Specification" is ambiguous since the level definitions are also a specification. What about "Provenance"?

Hey, that sounds good. Thanks for the feedback. I will make the changes in my next commit.

MarkLodato (approved):

PR #226, opened 2021-11-22
Clean up steering committee list; clarify approval expectations for this repo.

MarkLodato:
Sort the steering committee list alphabetically.

Reference the new @slsa-framework/slsa-steering-committee alias.

In the contributor guide, clarify that only significant changes to the spec require two steering committee members to review, and that others can be approved by a single reviewer. In particular, document that it's OK for author and reviewer to be from the same org.

(This documents what was discussed in an email thread. It would be great if other steering committee members can weigh in, in case they disagree.)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,22 +23,20 @@ this is not currently enforced).
 
 All changes require peer review through GitHub's pull request feature.
 
-Significant changes to the SLSA requirements or provenance schema require
-approval from **two steering committee members**. (Excludes changes that only
-affect syntax or clarity without altering meaning.)
+-   Changes to the specification, such as [Requirements](docs/requirements.md)
+    or the [Provenance schema](docs/provenance/), require approval from **two
+    steering committee members**, unless the change does not affect the meaning
+    (e.g. typo fix or minor reformatting.)
 
--   Do an initial round of review with a single reviewer.
--   Add @slsa-framework/slsa-steering-committee as a reviewer and make it
-    clear you are looking for at least two approvers.
--   Wait at least a few days before submitting to give the community time to
-    comment.
+    -   Do an initial round of review with a single reviewer.
+    -   Add @slsa-framework/slsa-steering-committee as a reviewer and make it
+        clear that you are looking for at least two approvers.
+    -   Wait at least one week before submitting to give the community time to
+        comment.
 
-Otherwise, only a single approval is needed.
-
--   No requirement to include a steering committee member. In practice, several
-    members review all changes at least passively anyway.
--   No requirement for the author and reviewer to come from different
-    organizations.
+-   Other changes only require **one approver**. Any member of @slsa-framework
+    can approve. There is no requirement for the author and reviewer to
+    represent different companies/organizations.
 
 ### Signing your work
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,13 +23,22 @@ this is not currently enforced).
 
 All changes require peer review through GitHub's pull request feature.
 
--   Changes generally require two approvals before being merged.
--   Minor changes that only affect clarity or syntax without changing the
-    meaning MAY be merged with a single approval.
--   Significant changes to the SLSA requirements or provenance schema SHOULD be
-    reviewed and approved from multiple steering committee members. Furthermore,
-    such changes SHOULD wait at least a few days after proposal before
-    submitting, to give the community time to review.
+Significant changes to the SLSA requirements or provenance schema require
+approval from **two steering committee members**. (Excludes changes that only
+affect syntax or clarity without altering meaning.)
+
+-   Do an initial round of review with a single reviewer.
+-   Add @slsa-framework/slsa-steering-committee as a reviewer and make it
+    clear you are looking for at least two approvers.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -23,13 +23,22 @@ this is not currently enforced).
...
 All changes require peer review through GitHub's pull request feature.
 
--   Changes generally require two approvals before being merged.
--   Minor changes that only affect clarity or syntax without changing the
-    meaning MAY be merged with a single approval.
--   Significant changes to the SLSA requirements or provenance schema SHOULD be
-    reviewed and approved from multiple steering committee members. Furthermore,
-    such changes SHOULD wait at least a few days after proposal before
-    submitting, to give the community time to review.
+Significant changes to the SLSA requirements or provenance schema require
```
marcofranssen, line 26: When is something considered to be `significant`?

marcofranssen (approved):
Looks good, one nit picky question though 😄

CONTRIBUTING.md
MarkLodato: Good point. If I remove the word "significant" but leave the exception, does that help?

joshuagl: It's implicit below that a non-significant change is a syntax or clarify fix which doesn't alter meaning.

Is it clearer if we invert it and say:
"Trivial changes, such as clarity improvements and typo fixes which don't affect meaning, require only a single approval"
<snip bullet list>
"Non-trivial changes require approval from two steering committee members"

joshuagl (commented):
I've left a couple of comments in-line, but a meta-comment: the changes here imply the existence of a project role (those that can review/approve changes) outside of the steering committee, but that role is not formally documented.

We might consider defining that role, how to attain the role, and who currently has that role within the project. This helps communicate how the project operates and what the contributor ladder looks like for those interested in becoming SLSA contributors.

I'm not suggesting we address this as part of this PR, but I do think it's worth addressing.

CONTRIBUTING.md
marcofranssen: @joshuagl his suggestion is a good improvement to reduce the chance for interpreting significant differently.

CONTRIBUTING.md
```diff
@@ -23,13 +23,22 @@ this is not currently enforced).
...
+approval from **two steering committee members**. (Excludes changes that only
+affect syntax or clarity without altering meaning.)
+
+-   Do an initial round of review with a single reviewer.
+-   Add @slsa-framework/slsa-steering-committee as a reviewer and make it
+    clear you are looking for at least two approvers.
+-   Wait at least a few days before submitting to give the community time to
+    comment.
```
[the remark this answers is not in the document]
MarkLodato: Done.

CONTRIBUTING.md
MarkLodato: Thanks. Please take another look. My intention was that only changes to the spec would require review, while changes to other files (even if "significant") would not. Assuming you agree with that, does the updated wording better clarify that?

I'm also happy to invert, but I worried that it would put too much focus on "Trivial changes" vs "Changes to the spec".

marcofranssen: Yup much better wording with less room for interpretation.

MarkLodato:
> the changes here imply the existence of a project role [...] outside of the steering committee, but that role is not formally documented

Good point. The text now says "Any member of @slsa-framework can approve," which is hopefully an improvement but still doesn't say how to become a member. Think that's OK for now?

marcofranssen:
> > the changes here imply the existence of a project role [...] outside of the steering committee, but that role is not formally documented
> 
> Good point. The text now says "Any member of @slsa-framework can approve," which is hopefully an improvement but still doesn't say how to become a member. Think that's OK for now?

Maybe add a section below the contributors section that defines how to become a member, which privileges you will get etc. Assuming there is/should be a difference between contributors and members.

e.g.

```
### Members

Members are part of the @slsa-framework team on Github. Members help the community with reviewing PRs, …, … and …

- [Member 1](…)
- [Member 2](…)
```

CONTRIBUTING.md
joshuagl: Agreed, the updated wording is clearer.

joshuagl (approved):
> > > the changes here imply the existence of a project role [...] outside of the steering committee, but that role is not formally documented
> > 
> > 
> > Good point. The text now says "Any member of @slsa-framework can approve," which is hopefully an improvement but still doesn't say how to become a member. Think that's OK for now?

Yes, I think the change here looks good. Can we capture the need to document roles & the folks who have them, perhaps based on @marcofranssen's suggestion, in a separate issue?

MarkLodato:
I created issue #229 since that is a bigger task involving collective decision making.

@trishankatdatadog @mlieberman85 Could you approve if you agree with the updated wording, since this changes the approval requirements? (You made the mistake of :+1:'ing a comment so I know you're looking! :grin:)

trishankatdatadog:
Let me take a look at this today when I get some downtime, thanks...

trishankatdatadog (approved):
LGTM. I don't see major concerns, and we can iterate as seen fit. Thanks very much!

MarkLodato:
Thanks for the reviews!

PR #237, opened 2021-12-02
Opt-in/out of GA cookies

devmoran:
Adds cookie storage functionality that complies with GDPR (Consent, right to be informed, right to object) and to a lesser extent CCPA (Right to know, right to delete).
We're not collecting personally identifiable info, but it makes sense to be on the right side of legality.

- A banner which persists until it's interacted with
   - Adds ability to accept or deny cookie storage
   - Honors in-browser doNotTrack settings

| Format| Screenshot |
|--|--|
|Desktop|image|
|Mobile|image|

### DIFF
--- a/docs/_includes/cookie-consent.html
+++ b/docs/_includes/cookie-consent.html
@@ -1,0 +1,142 @@
+<style>
+  #cookie-notice {
+    padding: 1rem;
+    display: none;
+    z-index: 100;
+    text-align: left;
+    font-size: 0.9rem;
+    color: rgba(255, 255, 255, 1);
+    background: #000;
+
+    position: fixed;
+    bottom: 0;
+    width: 100%;
+    justify-content: space-between;
+    gap: 3rem;
+  }
+
+  #cookie-notice a {
+    display: inline-block;
+    cursor: pointer;
+    border-radius: 8px;
+    padding: 8px 10px;
+    background: white;
+    text-decoration: none;
+    text-align: center;
+    color: #000;
+    min-width: 80px;
+    margin-left: 0.2rem;
+  }
+  #cookie-notice a:hover {
+    background: rgba(255, 255, 255, 0.7);
+  }
+
+  #cookie-notice-button-wrapper {
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+    gap: 0.5rem;
+  }
+
+  @media (max-width: 767px) {
+    #cookie-notice {
+      flex-direction: column;
+    }
+
+    #cookie-notice-button-wrapper {
+      justify-content: flex-start;
+    }
+
+    #cookie-notice a {
+      position: relative;
+      bottom: 4px;
+      margin-left: 0;
+      margin-right: 0.2rem;
+    }
+  }
+</style>
+<div id="cookie-notice">
+  <span tabindex="2">
+    We'd like to use Google Analytics cookies to help us improve our website by
+    collecting and reporting information on how you use it. <br />
+    The cookies collect information in a way that does not directly identify
+    anyone. These will be set only if you accept.
+  </span>
+  <span id="cookie-notice-button-wrapper">
+    <a href="#" tabindex="2" id="cookie-notice-accept">Accept</a>
+    <a href="#" tabindex="2" id="cookie-notice-deny">Deny</a>
+  </span>
+</div>
+<script>
+  function createCookie(name, value, days) {
+    var expires = "";
+    if (days) {
+      var date = new Date();
+      date.setTime(date.getTime() + days * 24 * 60 * 60 * 1000);
+      expires = "; expires=" + date.toUTCString();
+    }
+    document.cookie = name + "=" + value + expires + "; path=/";
+  }
+  function readCookie(name) {
+    var nameEQ = name + "=";
+    var ca = document.cookie.split(";");
+    for (var i = 0; i < ca.length; i++) {
+      var c = ca[i];
+      while (c.charAt(0) == " ") c = c.substring(1, c.length);
+      if (c.indexOf(nameEQ) == 0) return c.substring(nameEQ.length, c.length);
+    }
+    return null;
+  }
+  function eraseCookie(name) {
+    createCookie(name, "", -1);
+  }
+
+  if (readCookie("cookie-notice-option") === "true") {
+    function loadScriptAsync(scriptSrc, callback) {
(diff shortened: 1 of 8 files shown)


docs/index.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: Overview
 ---
+
```
joshuagl, line 4: Nit: this seems like a stray, unnecessary, change?

docs/_sass/minima/_custom.scss
```diff
@@ -270,18 +270,16 @@ h4 {
...
     padding: ($spacer-xl / 3) ($spacer-xl / 2);
     &[open] > summary {
-      margin-bottom: $spacer-l;
+        margin-bottom: $spacer-l;
     }
     > summary > span {
-      display: block;
-      float: right;
+        display: block;
+        float: right;
```
joshuagl, line 283: Nit: Looks like no functional change (all whitespace changes?) – is this an auto-formatter?

docs/_includes/cookie-consent.html
```diff
@@ -0,0 +1,142 @@
...
+    anyone. These will be set only if you accept.
+  </span>
+  <span id="cookie-notice-button-wrapper">
+    <a href="#" tabindex="2" id="cookie-notice-accept">Accept</a>
+    <a href="#" tabindex="2" id="cookie-notice-deny">Deny</a>
+  </span>
+</div>
+<script>
+  function createCookie(name, value, days) {
+    var expires = "";
```
joshuagl, line 72: What's the effect of an empty `expires` value? A cookie which is stored forever? Can we set a sane default here _just in case_ a future caller of `createCookie()` doesn't pass a `day` value?

joshuagl (commented):
Excellent change, thank you @devmoran !

I don't do front-end development, so forgive the simplistic questions. None of my comments here should be considered as blockers for this change.

Should we wrap the cookie consent, either in the include itself or at the place it's included, in a `if Jekyll.environment == 'production'` like the removed change to custom-head.html?

david-a-wheeler:
WAIT! DO NOT ADD THIS WITHOUT FURTHER CHECKS!

Cookie consent pop-ups are a loathsome plague upon the Earth. They train people to ignore warnings & are being user-hostile to boot.

If we don't *need* this we should *NOT* be adding this. We have GDPR experts in-house at LF if you want to investigate this, but please *DO NOT* make the site user-hostile if it's not actually required by law.

devmoran:
> Cookie consent pop-ups are a loathsome plague upon the Earth.

@david-a-wheeler I completely agree with you on both points. They stink, and I don't make this proposal lightly.

// TL;DR Happy to keep talking about this and not merge until we're agreed //

However my understanding is it's a legal requirement (certainly within Europe) and, as far as IF's remit, consent and transparency are fundamentals. I'm very much hoping for one of the proposed alternatives to tracking tech makes this all a thing of the past asap.

From what I understand there is a point of ambiguity (in some territories) about whether 'anonymous' analytics tracking is exempt, so if it's possible to bend the ear of someone at LF I'd appreciate it.

In terms of UX I've made effort that this implementation a. doesn't obstruct the use of the site (it doesn't block scrolling etc.), and has a simple binary choice (Accept/Deny) which is quick to dismiss (and yes, to ignore). I.e. the least hostile solution to this.


@joshuagl thanks for these comments! Will tweak.

david-a-wheeler:
@devmoran : I *definitely* want to follow all relevant laws, and I'm a big believer in transparency & consent. However, if we have to do a cookie consent, we're probably doing it wrong. I would prefer to disable anything that would lead to requiring cookie consent.

So:

1. If reasonably possible, let's remove anything that suggests we need cookie consent.
2. If we think we might need cookie consent, let me know & I will raise it to LF legal to see if we really need it (and if there are things we do to eliminate its need.

david-a-wheeler:
TL;DR: What the heck are we doing that would suggest the need for this? Can we get rid of it? I don't see why we need such invasive analytics, let's get rid of any analytics tools that might suggest it.

MarkLodato:
We are using Google Analytics to measure improvements to the website. GA requires GDPR consent by default. @david-a-wheeler could you contact LF legal to see what the recommended way to do this is, whether that is consent popup, consentless mode of GA, something other than GA, or just "don't do that"? Thanks.

david-a-wheeler:
Okay, I've sent a request to Linux Foundation (LF) legal. They use JIRA to track legal issues, so this is Legal Requests LR-877, "Need help with legal issues on Google Analytics, GDPR, and cookie consent". I *just* submitted it. Hopefully we'll get more clarification soon; I don't know how overwhelmed they are right now (I think it comes & goes).

david-a-wheeler:
I've talked with legal folks, thanks for giving me time to do that. Errors in my summary are my own; I'm not a lawyer.

If we continue to use Google Analytics (GA) then it appears we *do* need a cookie consent form. [Cookies, the GDPR, and the ePrivacy Directive (gdpr.eu)](https://gdpr.eu/cookies/) has a simple summary. It says "Receive users’ consent before you use any cookies except strictly necessary cookies." It all revolves around defining "strictly necessary" & it seems unlikely these cookies would be considered "strictly necessary" given that summary. So it looks like cookies are needed if we keep using GA.

> We are using Google Analytics to measure improvements to the website.

Do we really need it, though? I think we could live without it. If it's just performance measurements, we can do that directly by loading various pages without a cookie-based approach. I realize GA is popular, but that doesn't mean we should use it in this case.

If we *do* keep using Google Analytics, then the cookie consent stuff seems like a step. We'll have to ensure the other requirements are met, though:

* Provide accurate and specific information about the data each cookie tracks and its purpose in plain language before consent is received.
* Document and store consent received from users.
* Allow users to access your service even if they refuse to allow the use of certain cookies
* Make it as easy for users to withdraw their consent as it was for them to give their consent in the first place

It appears we have to "document and store consent received from users". If we can just store consent via the cookie itself, great; if we have to record it centrally that'll be a challenge.

MarkLodato:
Thanks, @david-a-wheeler! I agree that we shouldn't bother users with a consent banner. As an alternative to this PR, I sent out #247, which AIUI will allow us to comply with GDPR while still running Google Analytics.

devmoran:
Great, thanks so much @david-a-wheeler et al. Valuable to rake this over.
Happy to supersede this with @MarkLodato's #247 for reasons discussed in that PR.  Will close this.

PR #238, opened 2021-12-02
provenance: fix name of configSource in descriptions

joshuagl:
There are several mentions of invocation.sourceConfig in the descriptions
of fields in the provenance format. However, the field name is
configSource.

Signed-off-by: Joshua Lock <jlock@vmware.com>


priyawadhwa (approved):
Thanks for fixing this!

PR #240, opened 2021-12-03
Fix a typo in provenance definition.

AdamZWu:
`invocation.configSource.uri` should be of type ResourceURI.


MarkLodato (approved):
Thanks for the fix! Looks like you need the "Signed-off-by" tag to pass the DCO check.

AdamZWu:
Thanks for the hint! Followed instructions and hope I've done it correctly. :D

PR #245, opened 2021-12-11
Add slsa v0.2 provenance spec CUEvalidator

mlieberman85:
This file can be used with cue cli tools to validate attestations
against provenance spec like:

`cue vet provenance-2.0.cue attestation.json`

It will return an error if it doesn't match the spec. This also
opens up the ability to use CUE for codegen for various tools down
the line allowing us to create a canonical validator and tools
written in other languages can just use the CUE spec for both
validation as well as generation of attestations that meet the spec.

---

This is a draft as I want to get folks input on doing it this way.

Using this validator, I did detect a few bugs in how Tekton Chains generates v0.2 provenance and was able to open up a PR to fix it.

The sorts of things I want to talk through, and some other thoughts are:
* Where should this code live? In this repo? In another repo?
* Certain things like URI validation are not readily available in CUE so I put it here for now. Also, I'm going to need someone to validate my regex as I'm sure I messed something up when reading through RFC3986. Also if there's a way to not do this in regex we probably shouldn't...
* This CUE spec can generate OpenAPI today, and eventually generate protobuf as well as other formats according to the CUE community. This could be helpful in SLSA community in creating some reference tooling.
* Am I missing something from the spec? Probably should include the definitions from the markdown as comments in the code.
* When building this out, I noticed a bit of ambiguity in some of the types. For example, it's not clear what the actual difference between TypeURI and ResourceURI is.
* I think in using CUE to be a canonical definition of the spec it makes it really easy to play around with changing the spec and having codegen and automated tooling handle generation and validation.

### DIFF
--- a/schema/provenance-2.0.cue
+++ b/schema/provenance-2.0.cue
@@ -1,0 +1,109 @@
+package schema
+
+import (
+    "time"
+)
+
+#Subject: {
+    name: string
+    digest: #DigestSet
+}
+
+#DigestSet: {
+            sha256?: string
+            sha224?: string 
+            sha384?: string
+            sha512?: string
+            sha512_224?: string
+            sha512_256?: string
+            sha3_224?: string
+            sha3_256?: string
+            sha3_384?: string
+            sha3_512?: string
+            shake128?: string
+            shake256?: string
+            blake2b?: string
+            blake2s?: string
+            ripemd160?: string
+            sm3?: string 
+            gost?: string
+            sha1?: string
+            md5?: string
+}
+
+//TODO: This should probably live in a shared library.
+#Alpha: "[a-zA-Z]"
+#Digit: "[0-9]"
+#Hexdig: "[0-9a-fA-F]"
+#Scheme: "[a-z]*?(\\.?|\\+?|\\-?|.?)*?"
+#Unreserved: "(\(#Alpha)|\(#Digit)|\\-|\\.|_|~)"
+#Pct_Encoded: "%\(#Hexdig){2}"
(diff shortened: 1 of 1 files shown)


inferno-chromium:
Having in this repo seems ok. But is there a way to run unit testing in github actions workflow so if someone changes provenance format, they remember to fixup this cue spec and the unittest verifies that happens.

dlorenc:
This is awesome. Nice work!

mlieberman85:
> Having in this repo seems ok. But is there a way to run unit testing in github actions workflow so if someone changes provenance format, they remember to fixup this cue spec and the unittest verifies that happens.

Yep, I can add tests for this. Generate a couple of example attestations that could run through here.

mlieberman85:
Well, one thing I'm not sure of how to do in github is to enforce that a change in the markdown must accompany a change or something to the "as code" spec.

AdamZWu:
@mlieberman85 just a small touch up -- the `invocation.configSource.uri` should be of "ResourceURI" type (the specs was corrected a couple of days ago).

mlieberman85:
> @mlieberman85 just a small touch up -- the `invocation.configSource.uri` should be of "ResourceURI" type (the specs was corrected a couple of days ago).

Thanks, will update and push out.

mlieberman85:
@inferno-chromium @MarkLodato @TomHennen Would like to talk through what sorts of tests we might want to have for something like this, and how we can ensure that a change to the spec as defined in the documentation must have an accompanying change to the spec as defined in the code.

MarkLodato:
Thanks, @mlieberman85! Putting it in this repo sounds good to me. Examples would be helpful not only for unit testing but also as documentation.

MarkLodato:
@mlieberman85 shall we close this PR? I think it is stale.

PR #247, opened 2021-12-13
Run Google Analytics in "denied consent" mode.

MarkLodato:
This stops cookies from loading so that we don't need a GDPR cookie banner.

This is an alternative to #237. I don't know how much this hampers our analytics, but I agree with @david-a-wheeler that should not do anything that requires consent.

### DIFF
--- a/docs/_includes/google-analytics.html
+++ b/docs/_includes/google-analytics.html
@@ -3,7 +3,10 @@
   window['ga-disable-{{ site.google_analytics }}'] = window.doNotTrack === "1" || navigator.doNotTrack === "1" || navigator.doNotTrack === "yes" || navigator.msDoNotTrack === "1";
   window.dataLayer = window.dataLayer || [];
   function gtag(){dataLayer.push(arguments);}
+  gtag('consent', 'default', {
+    'ad_storage': 'denied',
+    'analytics_storage': 'denied'
+  });
   gtag('js', new Date());
-
   gtag('config', '{{ site.google_analytics }}');
 </script>


david-a-wheeler:
That's **cool**, I didn't know that capability existed!!

This looks promising. Does it prevent *all* GA cookies from being set on access to the site? I saw [this answer](https://support.google.com/analytics/answer/9976101?hl=en) but it wasn't clear to me. If you could test & verify that, that'd be great. Bonus points if you can turn it into a test case :-).

If ithis prevents all GA cookie setting, that's a *great* solution as far as I'm concerned. **However**, I'm not the poor person who has to deal with analytics (and/or their lack). Who was planning to analyze the results? Can they tell us if it's "good enough"? We don't have advertizers, so we don't  need a lot of depth in our analysis results to report to advertizers. I'm assuming our needs are pretty modest... but I could be mistaken.

inferno-chromium (approved):

MarkLodato:
Oops, forgot to cc @devmoran. Dev, what do you think? Have you used this mode before? Either way, do you think this will be good enough? Basically it prevents tracking from page to page, so each page looks like a unique view. It skews stats but might be sufficient for our modest needs.

devmoran:
I think this is a good compromise.
IMO this project has modest analytics needs, the most helpful trackers are: 
- `referral` (where are people coming to the site from? I.e. Has a particular piece of press performed well?) 
- `Views by page title` (i.e. where are people spending most of their time, and what can we infer from that? Is that dwell time a good or bad thing?)

The latter will be affected by this PR because it will no longer allow us to group by new vs. returning users, or to see 'funnels' of flow through the site. **But I don't think this is at all a deal breaker.** There will be value in the information anyway, as long as the SC are comfortable with the limitations.
Would be happy for this to supercede  #237

trishankatdatadog:
Great example of a community working to reach a compromise. Thanks, everyone!

trishankatdatadog (approved):

MarkLodato:
Sounds good. I'll merge now. Thanks all!

david-a-wheeler:
@MarkLodato : Congrats & big hat-tip for finding this solution. User-friendly, meets legal requirements, still provides us with some basic analytic tools. Spectacular!

PR #250, opened 2021-12-14
README: Add steering committee list, fix GH alias.

MarkLodato:
Add the newly creating SLSA steering committee mailing list to the README file to allow people to easily contact the steering committee. This is not on the slsa.dev site since we'd prefer users to go through public channels.

Also fix the GitHub team name (no "teams/") in the README.

**Question:** Should it also go on slsa.dev?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,9 @@ SLSA is currently led by an initial cross-organization, vendor-neutral steering 
 
 Shortcut to notify the steering committee on any issues/PRs:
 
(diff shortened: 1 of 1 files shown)


joshuagl (approved):
> Question: Should it also go on slsa.dev?

I agree that we want as much discussion as possible to be through public channels, so I'd lean towards not listing it on slsa.dev

MarkLodato:
Thanks!

PR #252, opened 2021-12-16
Explain how SLSA can help with typosquatting.

MarkLodato:
Previously the doc said flatly that typosquatting was out of scope. But SLSA can be a partial mitigation, so we should clarify that.

Thanks @TomHennen for the idea. Tom, does this look good to you? If so, I'll then send it to the steering committee.


TomHennen:
Yup, LGTM.

joshuagl (approved):
LGTM, thanks for taking the time @TomHennen and @MarkLodato

PR #255, opened 2021-12-17
Fixes custom font path in production env

devmoran:
- Fixes broken path to Prodigy Sans font on production environment
| Before | After |
|--|--|
|image |image|


devmoran:
@inferno-chromium Noticed a font bug on production! Fixed here

inferno-chromium:
Minor fix, merging.

PR #256, opened 2021-12-17
Adds terminology page to specification

devmoran:
Signed-off-by: Dev Morgan <disfordisford@gmail.com>

### DIFF
--- a/docs/_includes/specifications-nav.html
+++ b/docs/_includes/specifications-nav.html
@@ -13,6 +13,11 @@
     {% endif%}
 {% endfor %}
     <li>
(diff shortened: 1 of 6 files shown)


MarkLodato (commented):
Thanks! I agree with the overall idea, and I like how this enables #257. However, I feel that it's a big loss to drop the model supply chain diagram since that's a critical concept. I also find the laundry list of terms hard to read.

I'd want to think about an alternative format for the terminology page that gets the main concepts up right away with diagrams and perhaps categorizes terms.

Since the Projects by IF contract has ended, let me revisit this in the new year and clone the PR to do further tweaking.

trishankatdatadog (approved):

joshuagl:
One thing we might consider is a

PR #262, opened 2022-01-06
Switch from Google Analytics to GoatCounter.

MarkLodato:
It ends up that the "denied consent" mode doesn't work at all. It sends the data to Google but you can't actually view it on Google Analytics. Right now it's showing our site as having no traffic.

The only free, cloud-hosted, GDPR-compliant substitute I've found is GoatCounter. The site is super basic (see [demo](https://stats.arp242.net/)) but it's good enough. If we want something better, we'll need to either pay or self-host.

Once this is live, the analytics will be world-readable at https://slsa.goatcounter.com. After this PR is merged I'll invite steering committee members to be admins.


trishankatdatadog (approved):
I don't see major objections

trishankatdatadog:
But @david-a-wheeler's voice is more critical here

MarkLodato:
@david-a-wheeler Any objections? Or @trishankatdatadog is it OK to submit?

trishankatdatadog:
> @david-a-wheeler Any objections? Or @trishankatdatadog is it OK to submit?

I have no objections but will refer to David

MarkLodato:
> Can you give me a few more days to see if I can track down more info? I think this is a winner, but I'd like to track some things before we wrap this up. I expect this to be a good solution though.

Sounds good. No problem. Thanks!

david-a-wheeler:
Let's do this! I talked with others who didn't see an issue.

Thanks for tracking down a better alternative for our situation.

david-a-wheeler (approved):

MarkLodato:
Thanks for the thorough check!

PR #271, opened 2022-01-19
Simplify the specifications list template.

MarkLodato:
Similar to e56c388 (#270), but for the list of subpages in /spec/v0.1/index. This change replace the "automatic" code with a manual list, which is much easier to understand and maintain.

No content change. The only HTML change is to switch from absolute to relative links and to eliminate empty `<li>` elements.

NOTE: Neither the original nor new code correctly handled displaying more than one specification version. This commit adds a TODO making it clear that it doesn't work.

### Testing

The old and new version result in identical HTML except as noted above, ignoring one whitespace change and build timestamps on posts.

```bash
$ git checkout main
$ bundle exec jekyll build
$ mv _site /tmp/old_site
$ git checkout spec-list
$ bundle exec jekyll build
$ diff -u {/tmp/old_site,_site}/spec/v0.1/index.html
```

Compare:

* https://deploy-preview-271--slsa.netlify.app/spec/v0.1/#specifications
* https://slsa.dev/spec/v0.1/#specifications

### DIFF
--- a/docs/_includes/specifications-list.html
+++ b/docs/_includes/specifications-list.html
@@ -1,12 +1,23 @@
+{%- comment -%}
+Displays the list of specification subpages.
+
+USAGE: {% include specifications-list subpages=page.subpages %}
+
+In the YAML frontmatter, include a snippet like this:
+    ---
+    subpages:
+      - title: Some page
+        description: Text to display below the link
+        relative_url: foo
+      - title: Another page
+        description: More text
+        relative_url: ../bar
+    ---
+{%- endcomment -%}
 <ul class="custom-list">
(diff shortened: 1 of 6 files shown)


trishankatdatadog (approved):

MarkLodato:
Thanks for the quick reviews!

NANOKINGJOKER (commented):
My accounts was lock in awebo68.net.  
It important for me

trishankatdatadog:
How do we block @NANOKINGJOKER?

MarkLodato:
> How do we block @NANOKINGJOKER?

Already done.

PR #275, opened 2022-01-24
Add terminology page and reshuffle content

MarkLodato:
Reshuffle content so that each page serves a specific purpose and jumps right into its content, without long background. Also start each page with a better into so that the reader knows how the various pages relate. This is an alternative to #256 without rewriting any significant content.

In particular:

* Add a new Terminology page to contain common definitions used across multiple pages, as noted below.
  * Note: The Requirements page retains its definitions list for terms used only on that page. 
* Move the requirements summary table to the Requirements page (from Levels).
* Move the threats summary table to the Threats page (from Levels).
* Move the definition of integrity to the Terminology page (from Threats).
* Move the supply chain model to the Terminology page (from Requirements).
* Move the "What is SLSA?" section to the Levels page (from Requirements).
* Update the intro text for each page so that they better flow and link to each other.


trishankatdatadog (approved):

joshuagl (approved):
This LGTM, thanks Mark! 

> Note: The Requirements page retains its definitions list for terms used only on that page.

Should we add some text in both of these locations pointing to further definitions being in the other place? Perhaps as a follow-on PR

MarkLodato:
> > Note: The Requirements page retains its definitions list for terms used only on that page.
> 
> Should we add some text in both of these locations pointing to further definitions being in the other place? Perhaps as a follow-on PR

Thanks! I added a link from Requirements to Terminology, which is a clear win. I couldn't think how to naturally do it the other direction, so I omitted it for now. We can always follow up with another PR.

Submitting now.

Issue #61, opened 2021-06-17
You are only as secure as your weakest dependency; SLSA 4 artifacts can have SLSA 0 dependencies?


vielmetti:
My experience of several real supply-chain vulnerabilities is that integrity of the finished product depends on the whole supply chain. In particular, incidents like [left-pad](https://www.theregister.com/2016/03/23/npm_left_pad_chaos/),
[shared-mime-info](https://www.theregister.com/2021/03/25/ruby_rails_code/), and a host of others have seen
substantial disruptions as a result of problems deep in the dependency tree.

With that in mind, I have a hard time seeing how SLSA levels can reasonably be independent of the quality assessment of the dependencies they rely on. If I read things right, you could have an SLSA 4 rating on a final product, but that rating could include components that were unrated or SLSA 0 or even I guess known to have problems.

Did I miss something crucial here? I just don't see the point of having hypercorrect build patterns for the finished product if the components it's being built from are not adequately testable.

TomHennen:
Hi Edward, thanks for your comment.

This issue is discussed in [the scope section](https://github.com/slsa-framework/slsa#scope-of-slsa).

> The reason for non-transitivity is to make the problem tractable. If SLSA 4 required dependencies to be SLSA 4, then reaching SLSA 4 would require starting at the very beginning of the supply chain and working forward. This is backwards, forcing us to work on the least risky component first and blocking any progress further downstream. By making each artifact's SLSA rating independent from one another, it allows parallel progress and prioritization based on risk. (This is a lesson we learned when deploying other security controls at scale throughout Google.)

Does that help?  Is further clarification needed?

vielmetti:
@TomHennen 

Upon reflection this seems to make sense for prioritizing the workload, if nothing else. As long as people understand the SLSA is a metric of build system quality, not code quality, I think that will be helpful overall.

However - I look at the results from Google's `deps.dev` project where big projects routinely include 100s of individual dependencies, and wonder how this process will adequately identify weak spots in big systems like `kubernetes`.

mlieberman85:
Closing as inactive — no activity since this was filed. If it's still worth looking at, feel free to reopen or file a new ticket with updated context.

Issue #46, opened 2021-06-04
v1.0: Define how artifacts are verified (automatically)


TomHennen:
*Edit 2022-10-17: This issue is being slightly repurposed to decide requirements and/or guidelines on how artifacts are automatically verified to meet the SLSA requirements.*

Tasks:

* [ ] Describe how the overall system works: builds generate provenance and downstream systems compare this to an expectation at various points (not sure where this should go)
* [x] levels.md: explain that users set expectations and downstream systems verify them
* [ ] requirements.md: add detailed requirements and recommendations for how this system must/should work
* [ ] #168

Context: #130

## Original post

There was a lot of discussion in #37 that seems to indicate there's still an apparent disconnect on policy (who produces/owns it) and evaluation (when/where to evaluate that policy).

I think, for the most part, _we actually agree with one another_.  There are just some slight misunderstandings that are confusing us.

I think the following are the open questions:

1. Who creates and owns the policy?
2. Should we have a separate set or recommendations on what a good policy looks like?
3. Should those requirements and recommendations be part of the SLSA levels (e.g. SLSA 3P) or separate from it?
4. What use cases are we targeting?
    A. complete end-to-end: producer creates and artifact & provenance and those make their way all the way to the consumer who then checks the artifact & provenance against their own policy
    B. early evaluation: producer creates and artifact & provenance and at some point of trust (e.g. code signing, ACL'd folder, repo upload) that artifact & provenance against some other policy
5. What do people mean when they talk about 'resource'?

Does that sound right?

My opinion on what the answers to these questions are:

1. Who creates and owns the policy?

Some of the discussion in #37 has changed my mind.

So, now my opinion is "**anyone that cares** can create a policy".

The question itself is a bit of a misnomer.  There isn't necessarily one single policy that should be evaluated.  There could be **many policies** created and owned by different people.  Anyone that cares about the artifact can create _their own_ policy.  The artifact producer themselves can create a policy if they want, but it's not required.  Downstream users can use the producer's policy if they wish (and there's a way for them to get it securely), but they can also create their own policy (perhaps using create on first use as has been discussed in #37).

One thing we do have to watch out for when there are lots of policies is what happens if/when the software creators refactor how their code is built.  E.g. changing the name of the Makefile target, switching repos, etc...

2. Should we have a separate set of recommendations on what a good policy looks like?

Yes, I think so.  I think having a separate clear set of requirements for policy would help.  We have this internally and I think it's helpful.

[...]

PR #284, opened 2022-01-31
Verification summary

### LINKED ISSUE #107: Suggest a method of communicating an artifact's SLSA level
Right now SLSA provides all the requirements an artifact needs to meet at various levels.

SLSA also provides guidance on how to communicate how an artifact was built (use [in-toto Provenance](https://github.com/in-toto/attestation/blob/main/spec/predicates/provenance.md)) and has [an outstanding issue to communicate what source requirements were met](https://github.com/in-toto/attestation/issues/47).

It does not, however, provide any guidance on how to communicate the SLSA level that some entity evaluated an artifact to have.

Should we create some other [in-toto predicate type](https://github.com/in-toto/attestation/tree/main/spec#predicate) to convey this information?

This would be useful for developers that do not want, or cannot, release the in-toto Provenance or source code attestations (potentially because they contain proprietary information) but still want to communicate to their downstream users the SLSA level of a given artifact.

We would probably also need to document the algorithm people should follow to determine an artifact's level (so that things are consistent at least).  That is probably part of #46.

TomHennen:
Document a SLSA verification summary attestation.
Fixes #107 by creating an attestation that indicates a `verifier`
has determined that the specified `subject` artifact(s) meets the
indicated SLSA level.

In addition this attestation also indicates the minimum aggregate
SLSA level met by the dependencies used to build the artifact
which can help to address #61.

This leaves a number of things up to the user:
* What it means to evaluate an artifact against policy. (See #46)
* How to communicate the attestations required to create this
    attestation.
* How to debug failures

I borrowed heavily from slsa.dev/provenance in terms of
formatting/documentation.

If desired I can provide detailed algorithms for how the minimum_*
fields could be computed.

This attestation is based on work done by @AdamZWu.


TomHennen:
LoL, this got messed up thanks to my being rusty with Git.  let me try again...

PR #285, opened 2022-01-31
Document a SLSA verification summary attestation.

### LINKED ISSUE #107: Suggest a method of communicating an artifact's SLSA level

TomHennen:
Fixes #107 by creating an attestation that indicates a `verifier`
has determined that the specified `subject` artifact(s) meets the
indicated SLSA level.

In addition this attestation also indicates the minimum aggregate
SLSA level met by the dependencies used to build the artifact
which can help to address #61.

This leaves a number of things up to the user:
* What it means to evaluate an artifact against policy. (See #46)
* How to communicate the attestations required to create this
    attestation.
* How to debug failures

I borrowed heavily from slsa.dev/provenance in terms of
formatting/documentation.

If desired I can provide detailed algorithms for how the minimum_*
fields could be computed.

This attestation is based on work done by @AdamZWu.

### DIFF
--- a/docs/verification_summary/v0.1.md
+++ b/docs/verification_summary/v0.1.md
@@ -28,6 +28,18 @@ Describe what SLSA level an artifact or set of artifacts was verified at
 and other details about the verification process including what SLSA level
 the dependencies were verified at.
 
+This allows software consumers to make a decision about the validity of an
+artifact without needing to have access to all of the attestations about the
+artifact or all of its transitive dependencies.  They can use it to delegate
+complex policy decisions to some trusted party and then simply trust that
+party's decision regarding the artifact.
+
+It also allows software publishers to keep the details of their build pipeline
+confidential while still communicating that some verification has taken place.
+This might be necessary for legal reasons (keeping a software supplier
+confidential) or for security reasons (not revealing that an embargoed patch has
+been included).
+
 <div class="mt-16 mb-4">
 
 ## Prerequisite
--- a/docs/verification_summary/v0.1.md
+++ b/docs/verification_summary/v0.1.md
@@ -180,9 +180,9 @@ of the other top-level fields, such as `subject`, see [Statement]._
 
 > A count of the dependencies at each SLSA level.
 >
-> Map from [SlsaResult] the number of the artifact's _transitive_ dependencies that
-> were verified at the indicated level. Absence of a given level of [SlsaResult]
-> MUST be interpreted as reporting _0_ dependencies at that level.
+> Map from [SlsaResult] to the number of the artifact's _transitive_ dependencies
+> that were verified at the indicated level. Absence of a given level of
+> [SlsaResult] MUST be interpreted as reporting _0_ dependencies at that level.
 
 <div class="mt-16 mb-4">
 
--- a/docs/verification_summary/v0.1.md
+++ b/docs/verification_summary/v0.1.md
@@ -102,8 +102,8 @@ indicate these exceptions using `vsa.minimum_downgrade_dependency_level`.
   // Required
   "verifier": {
     "id": "<URI>"
-    }
-  "time_verified": <TIMESTAMP>
+  },
+  "time_verified": <TIMESTAMP>,
   "policy": "<URI>",
   "verification_result": "<PASSED|FAILED>",
   "policy_level": "<SlsaResult>",
@@ -212,8 +212,8 @@ WARNING: This is just for demonstration purposes.
 "predicate": {
   "verifier": {
     "id": "https://example.com/publication_verifier"
-    }
-  "time_verified": "1985-04-12T23:20:50.52Z"
+  },
+  "time_verified": "1985-04-12T23:20:50.52Z",
   "policy": "https://example.com/example_tarball.policy",
   "verification_result": "PASSED",
   "policy_level": "SLSA_LEVEL_3",
--- a/docs/verification_summary/v0.1.md
+++ b/docs/verification_summary/v0.1.md
@@ -219,7 +219,10 @@ WARNING: This is just for demonstration purposes.
     "id": "https://example.com/publication_verifier"
   },
   "time_verified": "1985-04-12T23:20:50.52Z",
-  "policy": "https://example.com/example_tarball.policy",
+  "policy": {
+    "uri": "https://example.com/example_tarball.policy",
+    "digest": {"sha256": "1234..."}
+  },
   "verification_result": "PASSED",
   "policy_level": "SLSA_LEVEL_3",
   "dependency_levels": {
--- a/docs/verification_summary/v0.1.md
+++ b/docs/verification_summary/v0.1.md
@@ -1,0 +1,242 @@
+---
+title: SLSA Verification Summary Attestation (VSA)
+layout: standard
+hero_text: Verification summary attestations communicate that an artifact has been verified at a specific SLSA level and details about that verification.
+---
+<details class="mt-12">
+<summary>Note about 0.x versions</summary>
+
+We expect regular iteration on 0.x versions until 1.0. During 0.x, we lean
+towards smaller, faster releases in order to get earlier feedback on the design.
+After 1.0, we will limit the frequency of breaking changes.
+
+To make this manageable, we recommend that:
+
+-   Generators choose the latest 0.x version at the time of implementation and
+    then stick with that until 1.0 unless there is reason to upgrade before then.
+-   Consumers accept all known versions and convert internally between them.
+
+</details>
+
+<div class="mt-16 mb-4">
+
+## Purpose
+
+</div>
+
+Describe what SLSA level an artifact or set of artifacts was verified at
+and other details about the verification process including what SLSA level
+the dependencies were verified at.
+
+<div class="mt-16 mb-4">
+
+## Prerequisite
+
+</div>
+
+Understanding of SLSA [Software
+Attestations](https://github.com/slsa-framework/slsa/blob/main/controls/attestations.md),
+[SLSA Provenance](https://slsa.dev/provenance), and the larger
+[in-toto attestation] framework.
+
+<div class="mt-16 mb-4">
+
+## Model
+
+</div>
+
+A Verification Summary Attestation (VSA) is an attestation that some entity
+(`verifier`) verified one or more software artifacts (the `subject` of an
+in-toto attestation [Statement]) by evaluating the artifact and a `bundle`
+of attestations against some `policy`.  Users who trust the `verifier` may
+assume that the artifacts met the indicated SLSA level without themselves
+needing to evaluate the artifact or to have access to the attestations the
+`verifier` used to make its determination.
+
+The VSA also allows consumers to determine the minimum verified level of
+all of an artifact’s _transitive_ dependencies.  The verifier does this by
+either a) verifying the provenance of each non-source dependency listed in
+the [materials](https://slsa.dev/provenance/v0.2#materials) of the artifact
+being verified (recursively) or b) matching the non-source dependency
+listed in `materials` (by subject.digest == materials.digest and, ideally,
+subject.name == materials.uri) to a VSA _for that dependency_ and taking
+the `min(vsa.policy_level, vsa.minimum_dependency_level`).  Policy
+verifiers wishing to establish minimum requirements on dependencies SLSA
+levels may use vsa.minimum_dependency_level to do so.
+
+Since dependency graphs can be quite complex a policy verifier may wish to
+require a default SLSA level X for all dependencies but allow for explicit
+exceptions that have level < X.  This allows progress to be made without
+being held back by a handful of ‘stubborn’ dependencies.  Verifiers
+indicate these exceptions using `vsa.minimum_downgrade_dependency_level`.
+
+<div class="mt-16 mb-4">
+
+## Schema
+
+</div>
+
+```jsonc
+// Standard attestation fields:
+"_type": "https://in-toto.io/Statement/v0.1",
+"subject": [{
+  "name": <artifact-URI-in-request>,
+  "digest": { <digest-in-request> }
+}],
+
+// Predicate
+"predicateType": "https://slsa.dev/verification_summary/v0.1",
+"predicate": {
+  // Required
+  "verifier": {
+    "id": "<URI>"
+    }
+  "time_verified": <TIMESTAMP>
+  "policy": "<URI>",
+  "verification_result": "<PASSED|FAILED>",
+  "policy_level": "<SlsaResult>",
+  "minimum_dependency_level": "<SlsaResult>",
+  "minimum_downgrade_dependency_level": "<SlsaResult>"
+}
+```
+
+<div class="mt-8 mb-4">
+
+### Parsing rules
+
+</div>
+
+This predicate follows the in-toto attestation [parsing rules]. Summary:
+
+-   Consumers MUST ignore unrecognized fields.
+-   The predicateType URI includes the major version number and will always
+    change whenever there is a backwards incompatible change.
+-   Minor version changes are always backwards compatible and "monotonic." Such
+    changes do not update the predicateType.
+-   Producers MAY add extension fields using field names that are URIs.
+
+<div class="mt-8 mb-4">
+
+### Fields
+
+</div>
+
+_NOTE: This section describes the fields within predicate. For a description
+of the other top-level fields, such as subject, see [Statement]._
+
+<a id="verifier"></a>
+verifier _object, required_
+
+> Identifies the entity that performed the verification.
+>
+> The identity MUST reflect the trust base that consumers care about. How
+> detailed to be is a judgment call.
+>
+> Consumers MUST accept only specific (signer, verifier) pairs. For example,
+> "GitHub" can sign provenance for the "GitHub Actions" verifier, and "Google"
+> can sign provenance for the "Google Cloud Deploy" verifier, but "GitHub" cannot
+> sign for the "Google Cloud Deploy" verifier.
+>
+> The field is required, even if it is implicit from the signer, to aid readability and
+> debugging. It is an object to allow additional fields in the future, in case one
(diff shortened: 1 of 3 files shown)


TomHennen:
Fancy preview is available at https://deploy-preview-285--slsa.netlify.app/verification_summary/v0.1

Note that this PR does not include any links on slsa.dev to this page.  Not sure how/if people would like that done.

TomHennen:
I should have mentioned (and perhaps I should update the PR), that one of the other things this attestation allows is for software publishers to keep the details of their build pipeline confidential while still communicating that some verification has taken place.

This might be necessary for legal reasons (keeping a software supplier confidential) or for security reasons (not revealing that a embargoed patch has been included).

wietse-gmail:
Minor problem: the "View source on GitHub" link (in the preview) leads to a 404 page. Is this expected for work-in-progress, or could this be a tooling problem? I'm looking at https://deploy-preview-285--slsa.netlify.app/verification_summary/v0.1

TomHennen:
@joshuagl updated the 'Purpose' section PTAL?

@MarkLodato do you know if the 'View Source' link will just work once this is merged?  I don't see that as a property of what I've changed here.

TomHennen:
FYI for whatever reason the latest update isn't showing up in the netlify preview but is in the code...

mlieberman85:
Transitive SLSA introduces an interesting challenge that I don't see described yet in the proposal. Does transitive SLSA include build environment? Compiler? Build tools? Build OS?

docs/verification_summary/v0.1.md
```diff
@@ -0,0 +1,242 @@
...
+<a id="minimum_dependency_level"></a>
+minimum_dependency_level _string ([SlsaResult]), optional_
+
+> The minimum verified SLSA level of all the artifacts _transitive_ dependencies excluding any
+> dependencies that have been explicitly downgraded to a lower acceptable level in the policy.
+> Absence indicates no evaluation of dependencies has taken place.
+
+<a id="minimum_downgrade_dependency_level"></a>
+minimum_downgrade_dependency_level _string ([SlsaResult]), optional_
+
```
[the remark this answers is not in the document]
TomHennen: @AdamZWu and I had considered but it lead to a couple of problems.

1. Since this is the transitive level it can get _very_ large.
2. What would we do with that data if we had it?  I think we'd argue that in most cases downstream users don't need or want to know about all the dependencies used by a dependency 4 steps removed.  If they do care then they could request/use the original provenance for the entire graph?

We imagine users wanting to enforce a policy like "This thing is SLSA L4, with transitive dependencies at L3, exceptions allowed".

This might be feasible if the array of mappings only covered direct dependencies?

@AdamZWu might have more thoughts here.

TomHennen:
> Transitive SLSA introduces an interesting challenge that I don't see described yet in the proposal. Does transitive SLSA include build environment? Compiler? Build tools? Build OS?

For the purposes of this proposal it covers everything the builder includes in the [materials section](https://slsa.dev/provenance/v0.2#materials) of the provenance.

Personally I'd argue that builders should only list tooling the user is allowed to bring themselves, and that things the builder provides should be covered by the builders accreditation at a given SLSA level.  This helps to keep the build graph manageable (so we don't need to go all the way back to the hand-written assembler).  However I don't think we need to answer this question to evaluate this proposal (perhaps I'm wrong!).

docs/verification_summary/v0.1.md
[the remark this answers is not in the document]
AdamZWu: @TomHennen beat me to it. :P

I'd like to add that when we propose the verification summary attestation, we specifically cater to the needs of *verification* -- that is, efficiently make a accept/reject decision with all transitive dependencies in scope. For that, we care a lot about the scalability and verification rule complexity. As a result, our design discarded a lot of dependency details, which may be considered useful for other use cases.

For example, I can totally see project managers and product security leads would love to have a list of details of every downgraded transitive dependencies of a complex artifact (e.g. OS image). For that, we are working on a different solution which provides this information.

docs/verification_summary/v0.1.md
```diff
@@ -0,0 +1,254 @@
...
+The VSA also allows consumers to determine the minimum verified level of
+all of an artifact’s _transitive_ dependencies.  The verifier does this by
+either a) verifying the provenance of each non-source dependency listed in
+the [materials](https://slsa.dev/provenance/v0.2#materials) of the artifact
+being verified (recursively) or b) matching the non-source dependency
+listed in `materials` (by subject.digest == materials.digest and, ideally,
+subject.name == materials.uri) to a VSA _for that dependency_ and taking
+the `min(vsa.policy_level, vsa.minimum_dependency_level`).  Policy
+verifiers wishing to establish minimum requirements on dependencies SLSA
+levels may use vsa.minimum_dependency_level to do so.
```
joshuagl, line 77:
```suggestion
levels may use `vsa.minimum_dependency_level` to do so.
```

docs/verification_summary/v0.1.md
```diff
@@ -0,0 +1,254 @@
...
+
+</div>
+
+This predicate follows the in-toto attestation [parsing rules]. Summary:
+
+-   Consumers MUST ignore unrecognized fields.
+-   The predicateType URI includes the major version number and will always
+    change whenever there is a backwards incompatible change.
+-   Minor version changes are always backwards compatible and "monotonic." Such
+    changes do not update the predicateType.
```
joshuagl, line 127:
```suggestion
    changes do not update the `predicateType`.
```

docs/verification_summary/v0.1.md
```diff
@@ -0,0 +1,254 @@
...
+-   Producers MAY add extension fields using field names that are URIs.
+
+<div class="mt-8 mb-4">
+
+### Fields
+
+</div>
+
+_NOTE: This section describes the fields within predicate. For a description
+of the other top-level fields, such as subject, see [Statement]._
```
joshuagl, line 137:
```suggestion
_NOTE: This section describes the fields within `predicate`. For a description
of the other top-level fields, such as `subject`, see [Statement]._
```

docs/verification_summary/v0.1.md
```diff
@@ -0,0 +1,254 @@
...
+> debugging. It is an object to allow additional fields in the future, in case one
+> URI is not sufficient.
+
+<a id="verifier.id"></a>
+verifier.id _string ([TypeURI]), required_
+
+> URI indicating the verifier’s identity.
+
+<a id="time_verified"></a>
+time_verified _string ([Timestamp]), required_
```
joshuagl, line 162:
```suggestion
`time_verified` _string ([Timestamp]), required_
```

docs/verification_summary/v0.1.md
```diff
@@ -0,0 +1,254 @@
...
+
+> The URI of the policy that was used to verify this artifact.
+
+<a id="verification_result"></a>
+verification_result _string, required_
+
+> Either “PASSED” or “FAILED” to indicate if the artifact passed or failed the policy verification.
+
+<a id="policy_level"></a>
+policy_level _string ([SlsaResult]), required_
```
joshuagl, line 177:
```suggestion
`policy_level` _string ([SlsaResult]), required_
```

docs/verification_summary/v0.1.md
```diff
@@ -0,0 +1,254 @@
...
+
+<a id="minimum_dependency_level"></a>
+minimum_dependency_level _string ([SlsaResult]), optional_
+
+> The minimum verified SLSA level of all the artifacts _transitive_ dependencies excluding any
+> dependencies that have been explicitly downgraded to a lower acceptable level in the policy.
+> Absence indicates no evaluation of dependencies has taken place.
+
+<a id="minimum_downgrade_dependency_level"></a>
+minimum_downgrade_dependency_level _string ([SlsaResult]), optional_
```
joshuagl, line 189:
```suggestion
`minimum_downgrade_dependency_level` _string ([SlsaResult]), optional_
```

docs/verification_summary/v0.1.md
```diff
@@ -0,0 +1,254 @@
...
+
+> The minimum verified SLSA level of all the artifacts _transitive_ dependencies that have
+> been explicitly downgraded by a policy.
```
joshuagl, line 192:
```suggestion
> been explicitly downgraded by the `policy`.
```

docs/verification_summary/v0.1.md
```diff
@@ -0,0 +1,254 @@
...
+
+<a id="verifier.id"></a>
+verifier.id _string ([TypeURI]), required_
+
+> URI indicating the verifier’s identity.
+
+<a id="time_verified"></a>
+time_verified _string ([Timestamp]), required_
+
+> Timestamp indicating what time the verification occurred.
```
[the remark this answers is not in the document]
TomHennen: I think the type this links to already makes that suggestion?  https://github.com/in-toto/attestation/blob/main/spec/field_types.md#Timestamp  (though perhaps it's not explicit enough?).

Do you think that definition needs any updating?  Should we file a separate issue to clarify?

docs/verification_summary/v0.1.md
TomHennen: Updated this and other like it.  Apparently GitHub won't let me resolve?

TomHennen:
> As an alternative to giving the minimum SLSA level of dependencies, I suggest we could instead give a count of each level. Something like:
> 
> ```js
> {
> // ...
> "predicateType": "https://slsa.dev/verification_summary/v0.1",
> // ...
>   "policy_level": "<SlsaResult>",
>   "dependencies_level_4": "<Int>",
>   "dependencies_level_3": "<Int>",
>   "dependencies_level_2": "<Int>",
>   "dependencies_level_1": "<Int>",
>   "dependencies_level_0": "<Int>",
>   "dependencies_level_unknown": "<Int>"
> }
> ```

This is an interesting idea!  I believe @AdamZWu was going to look into it and see how it would affect policy evaluations of dependencies (if it could replace the downgrade level field entirely).

docs/verification_summary/v0.1.md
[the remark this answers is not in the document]
joshuagl: My fault, I hadn't clicked through to the definition here. I think the existing type definition is sufficient.

mlieberman85:
> > Transitive SLSA introduces an interesting challenge that I don't see described yet in the proposal. Does transitive SLSA include build environment? Compiler? Build tools? Build OS?
> 
> For the purposes of this proposal it covers everything the builder includes in the [materials section](https://slsa.dev/provenance/v0.2#materials) of the provenance.
> 
> Personally I'd argue that builders should only list tooling the user is allowed to bring themselves, and that things the builder provides should be covered by the builders accreditation at a given SLSA level. This helps to keep the build graph manageable (so we don't need to go all the way back to the hand-written assembler). However I don't think we need to answer this question to evaluate this proposal (perhaps I'm wrong!).

I agree we don't want to boil the ocean, but I'm just wondering how you could potentially have a let's say recursive SLSA 3 artifact if let's say the builder's own attestations are summary SLSA 1, even if the builder is certified to build SLSA 3 artifacts I would have concerns what information is actually being communicated. I am just trying to see how as a user I would consume this information and make decisions on it.

mlieberman85:
> As an alternative to giving the minimum SLSA level of dependencies, I suggest we could instead give a count of each level. Something like:
> 
> ```js
> {
> // ...
> "predicateType": "https://slsa.dev/verification_summary/v0.1",
> // ...
>   "policy_level": "<SlsaResult>",
>   "dependencies_level_4": "<Int>",
>   "dependencies_level_3": "<Int>",
>   "dependencies_level_2": "<Int>",
>   "dependencies_level_1": "<Int>",
>   "dependencies_level_0": "<Int>",
>   "dependencies_level_unknown": "<Int>"
> }
> ```

I worry that this can give folks the wrong idea. We're losing context on the upstream dependencies. For example I don't necessarily care if the vast majority of my upstream dependencies are SLSA 4 if they aren't used in critical or security focused code paths. I do care if a security focused library is SLSA level 0 or unknown.

I know a lot of enterprises will take the wrong idea from this, and people will game it to just make the numbers go up as opposed to get meaningful data intended to help make policy decisions.

jchestershopify:
>I worry that this can give folks the wrong idea. We're losing context on the upstream dependencies. ... I know a lot of people will take the wrong idea from this, and people will game it to just make the numbers go up as opposed to get meaningful data intended to help make policy decisions.

Is this an argument for leaving out information about transitive dependencies entirely? You make a good point that there will be some Goodhart's Law issues, but I am not sure if ditching the information entirely will help. Secondary tools can still crunch up these numbers.

mlieberman85:
> > I worry that this can give folks the wrong idea. We're losing context on the upstream dependencies. ... I know a lot of people will take the wrong idea from this, and people will game it to just make the numbers go up as opposed to get meaningful data intended to help make policy decisions.
> 
> Is this an argument for leaving out information about transitive dependencies entirely? You make a good point that there will be some Goodhart's Law issues, but I am not sure if ditching the information entirely will help. Secondary tools can still crunch up these numbers.

I think I might be inclined to just think of different ways to structure the data such that folks are getting the right info. I'm just wondering what does someone do with the summary information? What sorts of analysis can they do? What questions can they answer?

TomHennen:
> > > I worry that this can give folks the wrong idea. We're losing context on the upstream dependencies. ... I know a lot of people will take the wrong idea from this, and people will game it to just make the numbers go up as opposed to get meaningful data intended to help make policy decisions.
> > 
> > 
> > Is this an argument for leaving out information about transitive dependencies entirely? You make a good point that there will be some Goodhart's Law issues, but I am not sure if ditching the information entirely will help. Secondary tools can still crunch up these numbers.
> 
> I think I might be inclined to just think of different ways to structure the data such that folks are getting the right info. I'm just wondering what does someone do with the summary information? What sorts of analysis can they do? What questions can they answer?

One of the things they could do is get a glimpse of the transitive supply chain situation for a given artifact.  In the past people have been concerned that the SLSA requirements don't address transitive dependencies.  This attestation can provide a peak into the supply chain for a given artifact without needing all of the corresponding information (which they might not have access to or might be too much data).

In the original proposal you wind up knowing the worst case SLSA level of dependencies of the artifact, but don't know how widespread the problem is.  @jchestershopify's suggestion to include the counts seems an elegant way for consumers to get a feel for the extent of the problem, even if they don't know the specifics.  With the counts people can likely look at a single VSA and get an idea for how much work there would be to 'clean up' an artifacts supply chain.  99 @ SLSA_L4 + 1 @ SLSA_L0, seems pretty tractable.  2 @ SLSA_L4 + 54 @ SLSA_L3 + 44 @ SLSA_L0, that's going to take a lot of people and money to fix.

Selfishly, we want to be able to construct an automated policy like:

```proto
rules {
  ...
  dependency_verification {
    // Lowest acceptable policy level for any transitive dependency (excluding downgrades)
    minimum_level: SLSA_L3
    // Lowest acceptable downgraded level for any transitive dependency
    minimum_downgrade_level: SLSA_L1
    // Each direct dependency below `minimum_level` must be listed;
    //   has no effect on indirect dependencies. 
    downgrade_dependency {
      uri: "https://example.com/example-1.2.3.tgz"
      minimum_level: SLSA_L2
    }
    downgrade_dependency {
      uri: "https://example.com/example-tools-1.1.2.tgz"
      minimum_level: SLSA_L1
    }
  }
}
```

This type of policy expects there to be a VSA provided _for each direct dependency_ when an artifact is being evaluated for SLSA compliance.

For this to work we can use either the original proposal or @jchestershopify's proposal (which we like a little bit more).

joshuagl:
> In the original proposal you wind up knowing the worst case SLSA level of dependencies of the artifact, but don't know how widespread the problem is.

Nicely articulated. This summarises my concern very well, I too like @jchestershopify's proposal to address this concern.

TomHennen:
So how do people generally feel about @jchestershopify's proposal?  Shall I make that change?

TomHennen:
@joshuagl WDYT, should I go ahead and update to include the dependency count at each level?  Is there anything else you'd like to see?

TomHennen:
Updated with a new `dependency_levels` field replaces the minimum_dependency_level (and similar) fields.

@joshuagl PTAL and let me know if you have any concerns.

TomHennen:
Per another conversation I had it would be good to add the digest of the policy evaluated to the VSA.

joshuagl:
I like this a lot, thanks Tom!

> Per another conversation I had it would be good to add the digest of the policy evaluated to the VSA.

That's a smart idea. Are you thinking of changing policy to an object with `TypeURI` and `DigestSet` fields?

TomHennen:
> > Per another conversation I had it would be good to add the digest of the policy evaluated to the VSA.
> 
> That's a smart idea. Are you thinking of changing policy to an object with `TypeURI` and `DigestSet` fields?

Yes, exactly.

joshuagl:
Sounds good, let's do it.

TomHennen:
> Sounds good, let's do it.

Done

joshuagl (approved):
Thanks for the iterations on this, LGTM.

TomHennen:
Thanks for reviewing!

PR #290, opened 2022-02-02
terminology: define build [platform|service|step]

joshuagl:
Although Build platform is implicitly defined as part of the Build, there is still confusion around the term. 
Additionally what makes a build step is not explicitly defined. 

Expand the new terminology page to include:
* Build platform: runs builds/build processes
* Build service: typically a managed/hosted build platform
* Build step: a task in the build process

### DIFF
--- a/docs/_spec/v0.1/terminology.md
+++ b/docs/_spec/v0.1/terminology.md
@@ -27,6 +27,9 @@ supply chains plus its own sources and builds.
 | Artifact | An immutable blob of data; primarily refers to software, but SLSA can be used for any artifact. | A file, a git commit, a directory of files (serialized in some way), a container image, a firmware image. |
 | Source | Artifact that was directly authored or reviewed by persons, without modification. It is the beginning of the supply chain; we do not trace the provenance back any further. | Git commit (source) hosted on GitHub (platform). |
 | Build | Process that transforms a set of input artifacts into a set of output artifacts. The inputs may be sources, dependencies, or ephemeral build outputs. | .travis.yml (process) run by Travis CI (platform). |
+| Build platform | An automation and scheduling platform for running build processes. Better defined by what it is not: the most desirable attribute of a build platform is that builds do not happen on a developer's workstation. | GitHub Actions, Tekton, Jenkins, Buildbot. |
+| Build service | Often used interchangeably with build platform, though may additionally imply a managed/hosted build platform (rather than i.e., a developer operated server). | GitHub Actions, Google Cloud Build, Travis CI. |
+| Build step | A single task in the build process | a `step` in a GitHub Action Workflow. |
 | Package | Artifact that is "published" for use by others. In the model, it is always the output of a build process, though that build process can be a no-op. | Docker image (package) distributed on DockerHub (platform). A ZIP file containing source code is a package, not a source, because it is built from some other source, such as a git commit. |
 | Dependency | Artifact that is an input to a build process but that is not a source. In the model, it is always a package. | Alpine package (package) distributed on Alpine Linux (platform). |
 


trishankatdatadog:
Thanks! Are these terms used anywhere else?

joshuagl:
Thanks for the quick feedback @trishankatdatadog. I don't think these terms are used in this way elsewhere, which is why they seem to be a frequent cause of confusion in my discussions of SLSA.

I think there's an open question around whether we want to define these, as SLSA aims to be versatile to fit multiple artefact production pipelines.

I infer through reading the SLSA docs that build platform and build service are used interchangeable and generally assumed to be some kind of CI/CD service.

* Should we be consistent in use of build platform OR build service (but not both)?
* Should we be explicit that the build platform is typically some kind of CI/CD service (particularly for small projects), or does my definition work:
   > An automation and scheduling platform for running build processes.
  (Admittedly, it all feels a bit circular)

docs/_spec/v0.1/terminology.md
```diff
@@ -27,6 +27,9 @@ supply chains plus its own sources and builds.
 | Artifact | An immutable blob of data; primarily refers to software, but SLSA can be used for any artifact. | A file, a git commit, a directory of files (serialized in some way), a container image, a firmware image. |
 | Source | Artifact that was directly authored or reviewed by persons, without modification. It is the beginning of the supply chain; we do not trace the provenance back any further. | Git commit (source) hosted on GitHub (platform). |
 | Build | Process that transforms a set of input artifacts into a set of output artifacts. The inputs may be sources, dependencies, or ephemeral build outputs. | .travis.yml (process) run by Travis CI (platform). |
+| Build platform | An automation and scheduling platform for running build processes. Better defined by what it is not: the most desirable attribute of a build platform is that builds do not happen on a developer's workstation. | GitHub Actions, Tekton, Jenkins, Buildbot. |
```
TomHennen, line 30: Should 'build process' be defined?  Is it just the same as 'build' above?

docs/_spec/v0.1/terminology.md
```diff
@@ -27,6 +27,9 @@ supply chains plus its own sources and builds.
...
+| Build service | Often used interchangeably with build platform, though may additionally imply a managed/hosted build platform (rather than i.e., a developer operated server). | GitHub Actions, Google Cloud Build, Travis CI. |
+| Build step | A single task in the build process | a `step` in a GitHub Action Workflow. |
```
TomHennen, line 32: Do you have any plans to use 'build step' anywhere?  Is it primarily here to distinguish itself from the entire build process?

TomHennen (changes requested):
Thanks!  I like this direction.  I think it's definitely been confusing.

Would it be fair to say that the Build Service is responsible for meeting SLSA's requirements but that the Build Platform can help it to do so?  E.g. it's not enough to use Tekton by itself (even if it produces provenance and all that) you still need to manage the platform securely?

docs/_spec/v0.1/terminology.md
TomHennen, line 30: @AdamZWu pointed out that since the Build Platform is just code it _could_ run on a developer's workstation.  But it makes it possible to easily run it elsewhere, like in a build service.

MarkLodato:
Thanks for starting this discussion! We certainly need more clarity here and this is a good start. I'm not sure that defining these two terms is sufficient, and I also think we should probably stick to either platform or service rather than using both.

I put together a skeleton of how we could expand this a bit. See #294. I threw that together, but maybe we can form that into a proper model?

@joshuagl Are you available to meet to discuss? I suspect we need a higher bandwidth channel than PR comments. I chatted with @TomHennen for a half hour and we didn't even come to a conclusion. :smiley:

docs/_spec/v0.1/terminology.md
joshuagl: I defined build step here because the term is used multiple times in the [Build requirements](https://slsa.dev/spec/v0.1/requirements#build-requirements) in [scripted build](https://slsa.dev/spec/v0.1/requirements#scripted-build), [build service](https://slsa.dev/spec/v0.1/requirements#build-service), [ephemeral environment](https://slsa.dev/spec/v0.1/requirements#ephemeral-environment), [isolated](https://slsa.dev/spec/v0.1/requirements#isolated), [hermetic](https://slsa.dev/spec/v0.1/requirements#hermetic), and [reproducible](https://slsa.dev/spec/v0.1/requirements#reproducible) and to at least one reader it was not obvious what the term refers to.

docs/_spec/v0.1/terminology.md
joshuagl: That's a fair point.

joshuagl:
> Thanks for starting this discussion! We certainly need more clarity here and this is a good start. I'm not sure that defining these two terms is sufficient, and I also think we should probably stick to either platform or service rather than using both.

Agreed, the two definitions are not sufficient. I had intended to _start_ a discussion, so I probably should have started with an issue or draft PR.

> I put together a skeleton of how we could expand this a bit. See #294. I threw that together, but maybe we can form that into a proper model?

This looks like a great start, thanks for building on the vague ideas here and fleshing it out. We should close this PR in favour of #294.

> @joshuagl Are you available to meet to discuss? I suspect we need a higher bandwidth channel than PR comments. I chatted with @TomHennen for a half hour and we didn't even come to a conclusion. 😃

💯

docs/_spec/v0.1/terminology.md
joshuagl: Or alternatively, this definition is "an automation and scheduling platform for running builds".

joshuagl:
Closing in favour of the WIP model in #294

PR #294, opened 2022-02-03
Define model for a build process

MarkLodato:
To avoid confusion and misinterpretation, define the build terms used within the requirements, such as "platform", "service", and "steps". (This is a spin-off of #290.)

This PR is limited to documenting existing terminology only. It only introduces one new term ("trigger") because that concept is not mentioned in the requirements, but it is an important part of the model.

A future PR could simplify the terminology to avoid confusion (#306). For example, we could might want to use either "platform" and "service" consistently rather than having two separate terms.

Preview: https://deploy-preview-294--slsa.netlify.app/spec/v0.1/terminology#build-model

screenshot


docs/_spec/v0.1/terminology.md
```diff
@@ -26,10 +26,73 @@ supply chains plus its own sources and builds.
...
+| Workload     | whatever they ran
+| Steps        | whatever they ran
+| Environment  | developer's workstation
+| Trigger      | commands that the developer ran
+| Admin        | developer
+
+</details>
+
+[runs-on]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on
+[actions trigger]: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow
```
joshuagl, line 94:
```suggestion
```
This one appears to be unused.

docs/_spec/v0.1/terminology.md
```diff
@@ -26,10 +26,73 @@ supply chains plus its own sources and builds.
...
+| Platform     | developer's workstation
+| Workload     | whatever they ran
+| Steps        | whatever they ran
+| Environment  | developer's workstation
+| Trigger      | commands that the developer ran
+| Admin        | developer
+
+</details>
+
+[runs-on]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on
```
joshuagl, line 93:
```suggestion
```
Should move this up with the other GitHub link definitions.

docs/_spec/v0.1/terminology.md
```diff
@@ -26,10 +26,73 @@ supply chains plus its own sources and builds.
...
+| Workload     | Workflow or job (either would be OK)
+| Steps        | `steps`
+| Environment  | `runs-on`
+| Trigger      | [events that trigger workflows]
+| Source       | git commit defining the workflow
+| Dependencies | any other artifacts fetched during execution
+| Admin        | GitHub personnel
+
+[GitHub Actions]: https://docs.github.com/en/actions
+[events that trigger workflows]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
```
joshuagl, line 73:
```suggestion
[`runs-on`]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on
[`steps`]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idsteps
[events that trigger workflows]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
```

docs/_spec/v0.1/terminology.md
```diff
@@ -26,10 +26,73 @@ supply chains plus its own sources and builds.
...
+| Dependencies | Additional input artifacts required by the workload.
+| Admin | Person with administrative access to the platform, potentially allowing them to tamper with the build process or access secret material.
+
+<details><summary>Example: GitHub Actions</summary>
+
+| Term         | Example
+| ------------ | -------
+| Platform     | [GitHub Actions] + runner + runner's dependent services
+| Workload     | Workflow or job (either would be OK)
+| Steps        | `steps`
```
joshuagl, line 65:
```suggestion
| Steps        | [`steps`]
```

docs/_spec/v0.1/terminology.md
```diff
@@ -26,10 +26,73 @@ supply chains plus its own sources and builds.
...
+| Environment  | `runs-on`
```
joshuagl, line 66:
```suggestion
| Environment  | [`runs-on`]
```

joshuagl (commented):
This looks great, thanks for working on this @MarkLodato. 

One change we might consider for the diagram is a few `step` boxes inside the workload, I think that would help make the diagram even more useful without necessarily having to read the text.

docs/_spec/v0.1/terminology.md
MarkLodato: Oops. Actually I think that link is more useful, so I use that link instead (moved up).

MarkLodato:
> One change we might consider for the diagram is a few `step` boxes inside the workload, I think that would help make the diagram even more useful without necessarily having to read the text.

Good idea. Done. I used lighter color with a dotted border to make it less distracting.)

I also updated the text to clarify that a workload runs in a single environment and added an explanation and example for a distributed build. Please take a look.

MarkLodato:
> Should we capture the ideas to simplify terminology in a future spec version as an issue?

Done. #306

MarkLodato:
@joshuagl In the spirit of limiting this PR to just documenting existing terms and not creating new ones, I just pushed a commit uses "Build" and "Instruction" instead of "Workload" and "Trigger", respectively. What do you think?

(Thanks @martensryan and team for the suggestions. Any other feedback would be appreciated.)

MarkLodato:
Also made "output" plural for consistency with "dependencies". (A build may have more than one output.) Thanks @sethkoehler for the idea.

martensryan:
Overall, this LGTM, but I would like to potentially resolve "Instruction" in a subsequent PR. "Trigger" definitely felt more in-line with the intent, especially as it pertains to GitHub Actions in terms of the external event causing a build to happen, rather than the entry point itself (which is what the spec defines as instruction); e.g. a build platform executes an instruction in response to a trigger.

That said, I'm happy to submit as-is and we can resolve separately since this is an improvement overall.

MarkLodato:
Thanks both!

Good point about "trigger" vs "instruction". I reverted to "trigger" since you both prefer it and because I agree with Ryan that it is not the same as "instruction." I kept "build" for now because there did not seem to be consensus on "workload." (We can always address in a future PR.)

I'll submit since you both LGTM'd.

PR #318, opened 2022-03-08
threats: replace link to event-stream incident

### LINKED ISSUE #314: Bug: event-stream link
Link [event-stream](https://schneider.dev/blog/event-stream-vulnerability-explained/) in text is wrong.
[event-stream](https://schneider.dev/blog/event-stream-vulnerability-explained/): Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior. The update did not match the code submitted to GitHub (i.e. attack F).

The Host is not resolvable.

joshuagl:
schneider.dev is down and, sadly, unlikely to come back. This patch
replaces the link to Zach Schneider's explanation with a link to the
official npm analysis of the incident.

Fixes #314

### DIFF
--- a/docs/_spec/v0.1/threats.md
+++ b/docs/_spec/v0.1/threats.md
@@ -36,7 +36,7 @@ Many recent high-profile attacks were consequences of supply-chain integrity vul
 [php]: https://news-web.php.net/php.internals/113838
 [webmin]: https://www.webmin.com/exploit.html
 [solarwinds]: https://www.crowdstrike.com/blog/sunspot-malware-technical-analysis/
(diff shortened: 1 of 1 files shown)


MarkLodato (changes requested):
I personally find the npmjs writeup to be very sparse, not really explaining how the attack took place. Should we use the wayback version  of schneider instead?

joshuagl:
I'm more than happy to use the way back version. I'll update the PR tomorrow.

MarkLodato (approved):
Thanks for the fix!

joshuagl:
Merging with a single approval as the change doesn't affect the meaning of the spec.

PR #324, opened 2022-03-16
Clarify null vs. unset for optional fields

brandoncart:
The provenance spec explicitly dictates handling of null or unset values for certain object and array fields, but not all. With this change null or unset values of these types will be uniformly treated as empty.


MarkLodato (commented):
Brandon, do you think it would be more readable to move this up to "Parsing rules" section instead of having it peppered throughout? That is, saying that null, empty, and (for maps and arrays) unset are all equivalent? Or is it better to have it on each item?

brandoncart:
I do think it's more readable to have a uniform statement up in the "Parsing rules" section. Happy to do that instead if you agree.

MarkLodato (approved):
Thanks!

PR #336, opened 2022-03-28
Move "Provenance" under specifications menu, and add link to "VSA"

### LINKED ISSUE #334: Link to verification_summary/v0.1
https://slsa.dev/verification_summary/v0.1 exists now but you can't navigate to it from the rest of the site.

Perhaps we should turn the 'Provenance' menu at the top into a drop down to link to various things?

E.g.

* Attestations
  * Provenance
  * Verification Summary

If we really like having 'Provenance' directly in the top menu then we could do:

* Specifications
  * ...
  * Attestations (new page that links to all the defined attestations including provenance?)

WDYT @MarkLodato you seem to be the most involved in the website look & feel?

MarkLodato:
It was a bit confusing why "Provenance" got its own top-level menu, especially since it too is a specification. Now all specs are under a single menu. This makes the top-level nav a bit cleaner.

Insert a horizontal rule to differentiate between the SLSA ladder and the attestations formats. Without that, the organization of the menu is a bit unclear.

Add a link to VSA. Fixes #334.

desktop

### DIFF
--- a/docs/_data/nav.yml
+++ b/docs/_data/nav.yml
@@ -23,8 +23,13 @@
     - title: FAQ
       url: /spec/faq
 
-- title: Provenance
-  url: /provenance/v0.2
+    - spacer: true
+
+    - title: Provenance
+      url: /provenance/v0.2
+
+    - title: VSA
+      url: /verification_summary/v0.1
 
 - title: Use cases
   url: /use-cases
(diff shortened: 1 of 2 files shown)


TomHennen:
Oh nice, thanks!

MarkLodato:
Please take a look. This avoids #340 by having a single specifications menu, which @olivekl suggested for clarity.

TomHennen (approved):
Looks good, thanks!

PR #343, opened 2022-04-01
automatically pull date on build

### LINKED ISSUE #342: Now is 2022
image

06kellyjac:
Resolves #342

It will still require rebuilding when the year changes but I think that's more appropriate than some JS that pulls Date.now()
The copyright notice will only need to cover year+1 if there have been changes in year+1

image


MarkLodato (approved):
Thanks!

Issue #338, opened 2022-03-29
Missing Threat: Revoked Dependency


ThomasOwens:
I've been reviewing the SLSA documentation and it seems like there is a missing dependency threat: an instance where the authorized maintainer of a dependency destroys the dependency. This could be pulling the dependency from public repositories or updating the version to remove the functionality (especially when versioned as a patch).

This isn't quite like some of the other identified threats. For example, it's not "submit bad code to the source repository" since that threat describes outside contributors submitting changes into the review process with the intention of evading detection or "use bad dependency" since the dependency was (theoretically) well-vetted and accepted.

An example of this would be left-pad.

ThomasOwens:
> I believe the documented threats are more oriented towards cases where the attacker can gain code execution and not, as in this case, where they can DoS a build. It is certainly a threat in the general sense but perhaps less concerning from an information security perspective.

This is a valid point. It's not clear to me - even with reading the homepage and Introduction and some of the other pages - what the scope of SLSA is. There's talk of "artifact integrity" and "secure packages and infrastructure" but not the scope.

I will say, though, that the inability to build a system can indirectly lead to security issues. When you rely on public repositories, you rely on them for all your builds. Consider a security issue, like log4j, coupled with a dependency author who yanked a key dependency. You may struggle to build your patched system. But even so, if the focus is on direct attacks to systems rather than impeding or slowing down the development or deployment of systems, this would be out-of-scope.

If this threat is out-of-scope, then I'd like to see a better definition of the scope of SLSA so people know what the limitations are and that there may be classes of supply chain vulnerabilities that would not be addressed even at the highest levels.

MarkLodato:
Thanks, @ThomasOwens!

I agree that we should add this to the [threats page](https://slsa.dev/threats), presumably under the "dependencies" category. I would like that page to model all supply chain threats, whether in-scope or out-of-scope. Care to send a PR?

For the current version (v0.1), I'd call this "out of scope" since none of the requirements address this this threat. Do you agree?

For future versions, it is a good question whether SLSA should address this by adding a new requirement. Note that there is also #276 of how SLSA fits into the broader supply chain security picture. But even if we just consider the current scope, this seems to fall within "availability" which we already somewhat address with the "sources retained" piece. Any thoughts on what kind of requirement would address this, and where it should go in the ladder?

ThomasOwens:
@MarkLodato

[...]

PR #347, opened 2022-04-04
Added revoked dependency threa.

ThomasOwens:
Addresses the discussion from slsa-framework/slsaMissing Threat: Revoked Dependency #338 to add out-of-scope threat of unavailable dependencies.

This does not update the ../../images/supply-chain-threats.svg image. I did want to propose the update there, but editing SVGs is not my strong-suit, so I was hoping to get feedback on the easier part to see if that image even needed to be updated.


ThomasOwens:
I updated this PR to add the DCO sign-off. If there's any guidance or advice on this and if/how to update supply-chain-threats.svg, that would be appreciated.

MarkLodato:
Thanks! Regarding the diagram, it seems to me that we have two dimensions here:

* Location of threat
* Type of threat: integrity (tampering) vs availability

In this light, I feel that (H) and (I) share same location but for integrity vs availability, respectively. Similarly, "Delete the code" would fall into availability at location (A). Does that align with your thinking? Any thoughts on how to update the diagram to incorporate that concept?

As for actual editing, we can do so via Figma: https://github.com/slsa-framework/slsa/tree/main/resources/editable-diagrams

ThomasOwens:
@MarkLodato

I don't think so. I don't think "A" is involved at all, at least with how I'm visualizing this. I'm picturing looking at this diagram from the perspective of someone in a development organization.

"A" would be my source, my repositories. Perhaps there are threats around a malicious internal user who destroys code in the product repositories or some other internal repositories.

My newly added "I" is something that happens at an external development organization. Perhaps the deletion isn't fully malicious with respect to their reason for doing it, but it has a domino effect on my build and package.

I see "Source Threats" and "Build Threats" as internal threats around things I can control - my people, my processes, my tools. "Dependency Threats" are around external threats and I may need to add controls or mitigations, because I can't stop the author of an NPM package from making 0.1.0 point to an empty set of files or removing 0.1.0 entirely from the external repositories.

MarkLodato:
Looks great! Here's the preview: https://deploy-preview-347--slsa.netlify.app/spec/v0.1/threats

Updating the diagram would be most appreciated! If you run into any trouble, let me know.

ThomasOwens:
@MarkLodato I've updated the diagram, but I can't figure out how to export to SVG in Figma. I made the changes to the slsa-editable-diagrams, but I can only see options to export to PDF. Is this part of a build process that I'm not seeing and making the changes to the editable file is sufficient?

MarkLodato:
If you click on the "frame" (white box), on the right-hand panel there should be an export option. But if that doesn't work, don't worry about it - I'm OK leaving the diagram as-is.


image

docs/_spec/v0.1/threats.md
```diff
@@ -27,7 +27,8 @@ Many recent high-profile attacks were consequences of supply-chain integrity vul
...
+| E   | Use risky dependency (i.e. A-H, recursively)                          | [event-stream]: Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior. The update did not match the code submitted to GitHub (i.e. attack F). | Applying SLSA recursively to all dependencies would have prevented this particular vector, because the provenance would have indicated that it either wasn't built from a proper builder or that the source did not come from GitHub. |
+|     |                                                                       | [Mimemagic]: Maintainer intentionally removes package or version of package from repository with no warning. Network errors or service outages may also make packages unavailable temporarily. | SLSA does not directly address this threat.
```
MarkLodato, line 31: Could you add a summary here, e.g. "Dependency becomes unavailable"?

ThomasOwens:
@MarkLodato Should be good now, at least if you squash commits. If you don't squash-on-merge, let me know and I can clean up the commit history. I've always preferred squashing on merge and pointing a single commit to the issue, but let me know if that's not what you do here.

MarkLodato:
@ThomasOwens The diagram is a bit messed up (wrong size, transparent background) so just revert that portion of it (svg and fig) and I'll send a separate PR to fix. Sorry for the back-and-forth.

We do use the "Merge" button - I also hate the messy history but no on else seems to mind :stuck_out_tongue:

ThomasOwens:
@MarkLodato All set.

If you do want to squash merge, that's an option on the merge button. It's in the down arrow. Makes all of my commits on this branch into a single commit before merging. I find it really helpful so reviewers can see all the history but the mainline only gets the final, approved changes in its history.

MarkLodato (approved):

PR #357, opened 2022-04-12
Improve CSS on blog post pages.

MarkLodato:
This adds back spaces between paragraphs (#355), but it isn't a perfect solution because block quotes are all screwed up (use flex model, wrong color, etc.)

I didn't use this on the specification pages because there we use blockquotes. For the blog, so far it's a pure improvement.

## Before

before

## After

after


kimsterv:
Thanks for the quick improvement!

Issue #360, opened 2022-04-12
slsa.dev: make the diagrams easier to edit


MarkLodato:
Currently our drawings are created in either Figma or Google Drawings. Neither of these are good for editing.

* The Figma drawings are difficult to edit. They can't be edited on Chromebooks or Linux because they use custom fonts, which only work on Windows or Mac. Plus Figma has a non-trivial learning curve.
* The Google drawings are easy to edit, but only by the owner. We can't check it in to version control.

Ideally we'd have files that can be checked into the repository and edited with free, standard, easy-to-use tools. Bonus points if the format is text-based and can be compiled in CI/CD.

mlieberman85:
I use structurizr but it's not completely open source though free if you run it yourself. It also is really built for c4 diagrams . Here is an example of what it looks like: https://github.com/buildsec/ssf/tree/main/docs/diagrams -- the .dsl is the "code" and the .json is machine readable formatting info.

This lets you write it "as code" but tweak layout graphically to make changes. Sort of like Graphviz where you can write it as code but move around the boxes and fix the inevitable readability issues in a GUI.

It would be nice if we could find a tool that is more generic than structurizr but generally has that same flow of diagram as code, with the ability to tweak it via a GUI

david-a-wheeler:
There are a billion number of tools for editing diagrams. Options include:

* Old school, but works, is OSS, & has tons of features: LibreOffice (e.g., SVGs).
* Markdown-happy: Mermaid https://mermaid-js.github.io/mermaid

I'm sure we can fill this PR with alternatives :-).

MarkLodato:
After doing a bunch of research and experimentation, I recommend sticking with Figma. PR #363 fixes the font issue so that we can edit on any browser, and I have another PR coming to further streamline the Figma file.

Among all the [other](https://irosyadi.gitbook.io/irosyadi/app/diagram-app) [options](https://irosyadi.gitbook.io/irosyadi/app/diagram-app), none seem nearly as good. My criteria were:

* General purpose, meaning that we can recreate our existing diagrams more-or-less exactly. This rules out most tools, such as structurizr, graphviz, mermaid, which are rigid in what the output looks like.
* Web or CLI based so that we can edit on any machine, particularly Chrome OS. This rules out desktop GUI apps, such as LibreOffice.
* File-based that we can check into version control, preferably with human-readable diffs. This rules out pure-web-based, such as Google Drawings.
* Free, preferably OSS.
[...]

PR #363, opened 2022-04-13
Make the diagrams easier to edit

MarkLodato:
As part of #360, clean up the Figma diagrams to make them easier to maintain and to use a consistent visual style.

Major changes:

* Re-implement the "supply chain model" diagram in the same visual style as the other diagrams. (Previously it was made in Google Drawings and looked very different.)
* Rename "bad dependency/package" to "risky dependency/package" as per #347. (@ThomasOwens FYI)

Minor visual changes:

* Switch font to Inter, which is a Google Web Font, from Prodigy Sans. This allows editing without installing the font locally, in particular enabling editing on Linux and Chromebook.
* Make the Package → Dependencies arrow dashed (liked the Dependencies box) to hint that a Dependency is itself a package.
* Use the hexagonal Build shape consistently. (In some versions it was an oval.)
* Adjust various alignments and spacings for consistency.
* Use white background color consistency. (In some versions it was transparent, which renders poorly on GitHub with dark mode.)

Invisible cleanups:

* Refactor the Figma file so that it's easier to maintain, and update the instructions to better explain how to use it.
* Remove the fonts from our repo, since they are no longer needed.
* Remove unused PNG file.


joshuagl:
Looks like we completely lost the labels on docs/images/supply-chain-threats.svg, was that intentional?
There's a typo in docs/images/supply-chain-model.svg – depedencies --> dependencies

MarkLodato:
> Looks like we completely lost the labels on docs/images/supply-chain-threats.svg, was that intentional?

Reverted. It was intentional, but I shouldn't mix that into this refactoring PR. The diagrams are back to how they were before, and I'll send out a separate to consider removing them.

> There's a typo in docs/images/supply-chain-model.svg – depedencies --> dependencies

Thanks, fixed.

joshuagl (approved):
LGTM, thanks for tackling this!

MarkLodato:
Thanks for the review!

PR #375, opened 2022-04-28
Update build-model diagram to match visual style

MarkLodato:
Reimplement the diagram in Figma using the same visual style as the other supply chain diagrams.

Also remove the "B" labels and arrows to reduce visual clutter. I think the concept of multi-tenancy still comes through.


joshuagl (approved):
Looks great, thanks!

PR #376, opened 2022-04-28
Add SLSA + SBOM blogpost

lumjjb:
SLSA + SBOM blogpost

Signed-off-by: Brandon Lum <lumjjb@gmail.com>

### DIFF
--- a/docs/_posts/2022-05-02-slsa-sbom.md
+++ b/docs/_posts/2022-05-02-slsa-sbom.md
@@ -1,0 +1,177 @@
+---
+title: "SBOM + SLSA: Accelerating SBOM success with the help of SLSA"
+author: "Brandon Lum, Isaac Hepworth, Meder Kydyraliev"
+layout: post
+---
+
+<!-- markdownlint-disable-next-line MD033 -->
+<center>
+<img src="https://user-images.githubusercontent.com/3060102/165816019-184fdd3d-1fa6-4d33-933f-49c1642006c0.png"
+width="300" /></center>
+
+If you’re trying to build and run secure software today, you’ve probably heard
+of an [SBOM, or a Software Bill of Materials](https://ntia.gov/SBOM). Acting as
+an “ingredient” label for software, SBOMs are documents that provide a nested
+list of packages and components included in a piece of software, and are
+supported by the White House in the 2021 [Executive Order on Improving
+Cybersecurity](https://www.whitehouse.gov/briefing-room/presidential-actions/2021/05/12/executive-order-on-improving-the-nations-cybersecurity/)
+as one aspect of a secure software supply chain.
+
+SBOMs are well positioned to help consumers better manage the risks of the
+software they consume and to respond more easily to vulnerabilities. There’s
+currently a lot of work being put into SBOMs to pave the way for widespread
+adoption, and as with any industry-changing effort of this size there are
+difficult problems to solve along the way. For example:
+
+-   SBOMs currently don’t include or require enough information to help users
+    respond to build tampering and attacks like
+    [Solarwinds](https://www.zdnet.com/article/microsoft-fireeye-confirm-solarwinds-supply-chain-attack/)
+    and [Codecov](https://about.codecov.io/security-update/);
+-   There’s no well-established ecosystem to easily distribute and verify SBOM
+    documents;
+-   The most common method of generating SBOMs using only audit tools after the
+    software’s creation can result in less accurate SBOMs.
+
+We believe that [SLSA (Supply-chain Levels for Software
+Artifacts)](http://slsa.dev/), a framework for creating a secure software
+supply chain, can address each of these potential areas for improvement when
+used in conjunction with SBOMs. This blog post explains the strengths of SBOMs
+and SLSA and how they fundamentally differ, and shows how SLSA principles can
+both support the generation of high-quality SBOMs and help consumers respond to
+supply chain attacks.
+
+## SLSA vs. SBOM
+
+If an SBOM is like an ingredient list on a food product in the grocery store,
+then SLSA can be thought of as all the food safety handling guidelines that
+make that ingredient list credible. From standards for clean factory
+environments so contaminants aren’t introduced in packaging plants, to the
+requirement for tamper-proof seals on lids that ensure nobody changes the
+contents of items sitting on grocery store shelves, the entire food safety
+framework ensures that consumers can trust that the ingredient list matches
+what’s actually in the package they buy.
+
+Likewise, the SLSA framework provides this trust by providing guidelines and
+evidence for securing each step of the software production process so that the
+final SBOM attached to a package can be considered credible. As a framework,
+SLSA lays out practices and guidelines to help you securely build and verify
+the integrity of software. These guidelines help mitigate against the risk of:
+
+-   Code modification (by adding a tamper-evident “seal” to code after source
+    control);
+-   Uploaded artifacts that were not built by a CI/CD system (by marking
+    artifacts with a factory “stamp” that verifies which build service created
+    it);
+-   Threats against the build system (by providing “manufacturing facility” best
+    practices for build system services).
+
+SBOM and SLSA are helpful in different situations. In the same way that an
+ingredient list is useful if there’s a product recall so that affected items
+can be pulled from the store shelves, an SBOM helps answer the question: “Am I
+affected by
+[log4j](https://www.cisecurity.org/log4j-zero-day-vulnerability-response) (or
+the next big vulnerability)?” SLSA, on the other hand, answers the broader
+question of: “Can I trust that this software was created using safe handling
+practices at each step of the production process?” Crucially, SLSA also helps
+answer the question: “Is my software ready to meet the requirements of the
+[Secure Software Development Framework](https://csrc.nist.gov/Projects/ssdf)?”
+
+## SLSA and SBOM
+
+SBOM and SLSA are complementary, and following certain SLSA principles can make
+it easier to generate SBOMs. In particular, a major SLSA principle is the need
+to generate tamper-proof provenance data—that is, metadata that attests to who
+performed the release process for an artifact, the materials used in
+production, and whether the artifact was protected from tampering. Since SBOMs
+hinge on accuracy, completeness, and trust, having SLSA provenance for an
+artifact improves the quality and integrity of its SBOM.
+
+### Accuracy and Completeness
+
+Many tools' current approach to creating an SBOM rely on guessing or using
+heuristics and information from an end product to generate an ingredient list
+after the fact. This approach has drawbacks, though, since if you try to
+generate an SBOM by scanning compiled artifacts, you may miss things like
+statically linked binaries that don’t include their dependency metadata.
+Similarly, generating an SBOM by making guesses based on file hashes will
+likely yield false positives for vulnerabilities that you’re not actually
+affected by, reducing trust in the overall vulnerability response process.
+
+On the other hand, **generating an SBOM with the additional help of SLSA
+attestations leads to a much more accurate SBOM**. By using SLSA provenance and
+metadata generated during the build, you have high-fidelity information about
+what went into the artifact, including better dependency information so that
+you don’t have to make guesses after the fact.
+
+![SBOM and SLSA in supply
+chain](https://user-images.githubusercontent.com/3060102/165816534-653cc8dd-3ae9-4c75-ab14-47f376f9a32f.png)
+
+*Traditional sources of SBOM information are shown in green; using SLSA allows
+inclusion of build information in SBOMs as well. (Red triangles mark threats to
+the supply chain that are addressed by SLSA.)*
+
+SLSA’s accurate build metadata also helps provide tangible knowledge of how the
+software was built: build metadata specifies both the input materials and the
+factory that produced the artifact. If faulty goods are caused by poor quality
+control or machinery (such as compromised build systems or bugs in tools like
+compilers), **an SBOM with this additional data will be more complete and help
+users respond to a supply chain compromise**.
+
+### Trust
+
+An SBOM generated from SLSA provenance data boosts users’ confidence in the
+accuracy of the SBOM, since they know it was created from high-fidelity build
+metadata: the ingredient list was generated by observing what was added while
+the product was made. The SLSA provenance also tells the user that the metadata
+comes from a verifiable build, and that the provenance data itself was captured
+and signed in a secure way. **Referencing these provenance documents in SBOMs
+provides an additional level of assurance about the secure process used to
+produce the software**.
+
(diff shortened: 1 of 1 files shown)


dlorenc:
Great post!

docs/_posts/2022-04-28-slsa-sbom.md
```diff
@@ -0,0 +1,175 @@
...
+Cybersecurity](https://www.whitehouse.gov/briefing-room/presidential-actions/2021/05/12/executive-order-on-improving-the-nations-cybersecurity/)
+as one aspect of a secure software supply chain.
+
+SBOMs are well positioned to help consumers better manage the risks of the
+software they consume and to respond more easily to vulnerabilities. There’s
+currently a lot of work being put into SBOMs to pave the way for widespread
+adoption, and as with any industry-changing effort of this size there are
+difficult problems to solve along the way. For example:
+
+-   SBOMs currently don’t include enough information to help users respond to
```
mlieberman85, line 26: I would cite it as the minimum elements for an SBOM don't require this information thus meaning you can't rely on it to include build information https://www.ntia.doc.gov/files/ntia/publications/sbom_minimum_elements_report.pdf

I state this because different SBOM tools and specifications CAN include build information but don't need to leading to you not being able to rely on it to contain that information

docs/_posts/2022-04-28-slsa-sbom.md
```diff
@@ -0,0 +1,175 @@
...
+If an SBOM is like an ingredient list on a food product in the grocery store,
+then SLSA can be thought of as all the food safety handling guidelines that
+make that ingredient list credible. From standards for clean factory
+environments so contaminants aren’t introduced in packaging plants, to the
+requirement for tamper-proof seals on lids that ensure nobody changes the
+contents of items sitting on grocery store shelves, the entire food safety
+framework ensures that consumers can trust that the ingredient list matches
+what’s actually in the package they buy.
+
+SLSA offers the same trust by securing each step of the software production
```
mlieberman85, line 54: I would word it a bit less concretely. It provides evidence allowing an end user to trust. I worry that it gives the wrong impression. I wouldn't say SLSA itself secures the step.

I like the wording below that the guidelines help providence confidence that you are protected against certain kinds of attacks.

docs/_posts/2022-04-28-slsa-sbom.md
```diff
@@ -0,0 +1,175 @@
...
+process so that the final SBOM attached to a package can be considered
+credible. As a framework, SLSA lays out practices and guidelines to help you
+securely build and verify the integrity of software. These guidelines protect
```
mlieberman85, line 57: I would word it slightly differently to make sure that SLSA provides evidence of protections against certain attacks.

mlieberman85:
One thing I've done that I think is complementary and might prove to be valuable is I use SLSA attestations to provide provenance metadata for how my SBOM was built in the first place. So I would have SLSA attestation for my actual software and another one for how I generated my SBOM. I think it's valuable because it helps provide additional evidence as to what tool generated the SBOM as well as what the SBOM used as its "materials" for generation, e.g. did it build the SBOM from a container image, from a go mod file, from a binary analysis of the compiled software, etc.

docs/_posts/2022-04-28-slsa-sbom.md
lumjjb: Yup - I see where you're coming from. Yea I think from the current functional standpoint as well, most tooling doesn't do this, or at least most people are not thinking about it. Maybe, we can phrase it as:

"Today, SBOMs, don't include ..."

I am a bit hesitant to quote the minimum elements. I am afraid of going down the road of discussion of sbom minimum standards is going to be what's implemented, because today, for the most part we are doing much better than it. My view on the minimum elements is more for vendors of proprietary software that are very reluctant to put out information, that this is the absolute minimum to meet compliance.

docs/_posts/2022-04-28-slsa-sbom.md
lumjjb: Hmm, what do you think about taking the phrasing from https://slsa.dev/spec/v0.1/threats

to say "These guidelines mitigate against"

docs/_posts/2022-04-28-slsa-sbom.md
lumjjb: >I would word it a bit less concretely. It provides evidence allowing an end user to trust. I worry that it gives the wrong impression. I wouldn't say SLSA itself secures the step.

Hmm, I'm trying to tease this apart between SLSA as the provenance document and as the framework. I think we were trying to encompass some of the broader goals of the SLSA framework around securing the build system, etc. - slsa4 hardening, etc.  (i.e. the SSDF stuff).

Wondering how this can be worded to be clearer.

mlieberman85: Good point. I think we might just want to highlight that it is both. That SLSA compared to alternatives provides attestations that you are performing those practices.

docs/_posts/2022-04-28-slsa-sbom.md
mlieberman85: Good point. I think we might just want to highlight that it is both. That SLSA compared to alternatives provides attestations that you are performing those practices.

docs/_posts/2022-04-28-slsa-sbom.md
lumjjb: How about "The SLSA framework provides this trust by providing guidelines and evidence of securing each step of the software production process..."

mlieberman85: SGTM

docs/_posts/2022-04-29-slsa-sbom.md
```diff
@@ -0,0 +1,175 @@
...
+difficult problems to solve along the way. For example:
+
+-   SBOMs currently don’t include or require enough information to help users
+    respond to build tampering and attacks like
+    [Solarwinds](https://www.zdnet.com/article/microsoft-fireeye-confirm-solarwinds-supply-chain-attack/)
+    and [Codecov](https://about.codecov.io/security-update/);
+-   There’s no well-established ecosystem to easily distribute and verify SBOM
+    documents;
+-   The most common method of generating SBOMs using audit tools after the
+    software’s creation can result in less accurate SBOMs.
```
stevespringett, line 33: This generalization is inaccurate. In many cases, binary analysis of artifacts can reveal the existence of inventory that build manifests are unaware of. In the case of forensic use cases, audit tools may be the only available means to generate the data. I wouldn't recommend using audit tools as the sole SBOM generation process, but IME, using a combination of build-time, runtime, and audit tools produce the most accurate results.

docs/_posts/2022-04-29-slsa-sbom.md
```diff
@@ -0,0 +1,175 @@
...
+
+Many current approaches to creating an SBOM rely on guessing or using
+heuristics and information from an end product to generate an ingredient list
+after the fact. This approach has drawbacks, though, since if you try to
+generate an SBOM by scanning compiled artifacts, you may miss things like
+statically linked binaries that don’t include their dependency metadata.
+Similarly, generating an SBOM by making guesses based on file hashes will
+likely yield false positives for vulnerabilities that you’re not actually
+affected by, reducing trust in the overall vulnerability response process.
+
```
stevespringett, line 99: Seems to generalize two different approaches to generating SBOMs and ignoring other approaches to generating SBOMs. Not sure I understand the purpose of this paragraph.

docs/_posts/2022-04-29-slsa-sbom.md
```diff
@@ -0,0 +1,175 @@
...
+On the other hand, **generating an SBOM with the additional help of SLSA
+attestations leads to a much more accurate SBOM**. By using SLSA provenance and
+metadata generated during the build, you have high-fidelity information about
+what went into the artifact, including better dependency information so that
+you don’t have to make guesses after the fact.
+
```
stevespringett, line 105: Component pedigree seems to be missing here and is much harder to get right during a build process. There are typically processes outside of CI/CD that add this data to a build-time SBOM so that the accuracy and completeness is increased.

docs/_posts/2022-04-29-slsa-sbom.md
```diff
@@ -0,0 +1,175 @@
...
+inclusion of build information in SBOMs as well. (Red triangles mark threats to
+the supply chain that are addressed by SLSA.)*
+
+SLSA’s accurate build metadata also helps provide tangible knowledge of how the
+software was built: build metadata specifies both the input materials and the
+factory that produced the artifact. If faulty goods are caused by poor quality
+control or machinery (such as compromised build systems or bugs in tools like
+compilers), **an SBOM with this additional data will be more complete and help
+users respond to a supply chain compromise**.
+
```
stevespringett, line 119: OWASP refers to this as "formulation", which is generally a distinct subcategory of "pedigree".  MITRE and OWASP are aligned on the definition of pedigree and OWASP takes one step further by defining "formulation" as a separate but complimentary concept.

stevespringett: In general, I think if SLSA is targeting NIST SSDF, then the SLSA definition of "provenance" needs to get in line with the MITRE, OWASP, and English definition of the word.

docs/_posts/2022-04-29-slsa-sbom.md
```diff
@@ -0,0 +1,175 @@
...
+comes from a verifiable build, and that the provenance data itself was captured
+and signed in a secure way. **Referencing these provenance documents in SBOMs
+provides an additional level of assurance about the secure process used to
+produce the software**.
+
+The lessons learned by the SLSA community in creating this integrity could be
+useful to the [Software Package Data Exchange (SPDX)](https://spdx.dev/), an
+open standard for representation of SBOMs. The SPDX SBOM community is currently
+coming together to determine processes for integrity and signing of SBOM data.
+On top of that, it has kicked off its Canonicalization Committee to write a
```
stevespringett, line 135: Why is SPDX being called out specifically here? The OWASP Software Component Verification Standard (SCVS) which is referenced in its entirety by NIST SSDF, includes both CycloneDX and SPDX, but does not go into any specifics. I would refrain from mentioning specific capabilities of one format without ack the other.s

stevespringett: https://owasp.org/scvs/

docs/_posts/2022-04-29-slsa-sbom.md
lumjjb: Agree - will reword this to

"The most common method of generating SBOMs using **only** audit tools after the software’s creation can result in less accurate SBOMs."

docs/_posts/2022-04-29-slsa-sbom.md
lumjjb: This paragraph is targeted to be an explanation of why audit tools alone are insufficient. I think I can see why you interpret this as generalization of approaches. Perhaps we can reword it as

"Many current **tools'** approach ..."

docs/_posts/2022-04-29-slsa-sbom.md
lumjjb: Sounds like a discussion for a broader scope! We can create an issue to discuss this further!

docs/_posts/2022-04-29-slsa-sbom.md
```diff
@@ -0,0 +1,175 @@
...
+signing and transparency logs) and in-toto (for metadata format). These tools
+help to solve the problem of storing and distributing the metadata documents
+that need to accompany artifacts, and would help build the trust that users
+need in SBOMs.
+
+### Simplicity
+
+Adopting SLSA concepts could also simplify the tooling updates needed to
+generate SBOMs.  Enabling SBOM for all software being produced is a difficult
+problem since it requires integration with all tooling that produces software,
```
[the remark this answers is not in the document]
lumjjb: Hmm, can you elaborate a bit more on this - it does sound like a much broader discussion. Think this is a good idea to bring up during the next SLSA call if there's room in the agenda.

I think from a content delivery perspective, the discussion around reproducible builds + SBOM deserves its own story and perhaps a different audience.

docs/_posts/2022-04-29-slsa-sbom.md
lumjjb: yea let me add this in! Thanks.

docs/_posts/2022-04-29-slsa-sbom.md
lumjjb: is this something you see SLSA being able to help with? Want to understand the framing of this here.

(we've gone through cutting a lot of content to try and stay on message with SLSA, so want to make sure that SLSA can help with this. For this post, we are focusing on just problems that SLSA can help with)

docs/_posts/2022-04-29-slsa-sbom.md
lumjjb: the link you provided seems to be lead to a 404. I will add a generic link to CycloneDX instead.

docs/_posts/2022-04-29-slsa-sbom.md
lumjjb: Tracking issue: https://github.com/slsa-framework/slsa/issues/378

docs/_posts/2022-04-29-slsa-sbom.md
lumjjb: Tracking issue: https://github.com/slsa-framework/slsa/issues/378

docs/_posts/2022-04-29-slsa-sbom.md
stevespringett: Ah. I think the trailing slash made a difference. Try owasp.org/scvs

docs/_posts/2022-04-29-slsa-sbom.md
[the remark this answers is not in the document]
stevespringett: From the NTIA minimum elements:
> Timestamp records when the data is assembled -- the point of the SBOM creation. 

For every build, even if most artifacts are reproducible, the SBOM created during a build must have a different timestamp. There are many audit and assurance reasons for this requirement.

>  The data about software components can be collected at different stages in the software lifecycle, including from the software source, at build time, or after build through a binary analysis tool. Due to unique features of each of these stages, the SBOM may have some <ins>differences depending on when and where the data was created</ins >.

Timestamps are obviously tied to "when", however, a component's point-of-origin may vary depending on build tool configuration, DNS, or repository proxying and mirroring.

-------

The consensus among NTIA participants is that SBOM must a non-reproducible artifact. It can be repeatable. It can be predictable, but it will not conform to [reproducable builds](https://reproducible-builds.org/).

mlieberman85 (approved):
LGTM

Issue #309, opened 2022-03-04
SLSA Blog


MarkLodato:
I suggest that we create a SLSA Blog. The posts would be presented as opinions of the author, not necessarily of the SLSA project. Benefits:

* To give a steady update on activity to interested parties. Often work goes on behind the scenes or in detailed discussions, so it is difficult to see that progress is being made. A blog post can help communicate that progress at a higher level.
* To allow contributors to describe ideas and current thinking without requiring full community agreement or fully fleshing everything out. We can relate things that are currently in the minds of contributors but that we haven't had time to fully implement.
  * Example: Current thinking on how policy would work. This is a topic I have discussed with many people, but I haven't had time to really write it down formally or get full agreement from the community. But I could describe how *I* think it might work, which at least gives readers some notion of where SLSA might be headed.
* To allow us to post "informational" documentation that we don't necessarily want to maintain or make official. Other project use "informational" proposals for this process (PEP, TAP) - see https://github.com/slsa-framework/slsa/issues/296#issuecomment-1040339334. But in my opinion, that is an abuse of the notion of a "proposal." A blog post is a better medium for such content.

Thoughts?

joshuagl:
Great idea, demonstrating the liveliness of the project to folks not attending the community meetings and providing a space to share ideas both seem like good uses of a blog.

TomHennen:
What should the process be for approving posts?  Just have one other committee member verify that it's not spam?

I'm happy to set it up on Medium.

TomHennen:
Ok, I'm going to try to set up a Medium 'Publication' for SLSA.  Then we can have slsa.dev/blog redirect there somehow.

Someone let me know if they have a better idea.

kimsterv:
> What should the process be for approving posts? Just have one other committee member verify that it's not spam?

I think keeping it simple and easy is best. One other committee member and lazy consensus (after say 2 working days?) sgtm. If this doesn't work, we can always revisit of course.

jorydotcom:
👋🏻 qq, is Medium a strong want here? I ask because another project I work with has historically had several challenges managing its blog on Medium. It's great for the social/share component, and the UI for writers is great, but we find for a variety of reasons its hard to keep it updated (access control/permissions management, collaborating on drafts, people who don't have medium accounts & want to contribute, etc etc). If it's all the same to ya'll, hosting a blog via GH tends to be more successful long term.

joshuagl:
[...]

PR #379, opened 2022-04-29
Blog: clean up styling, add banner for guest posts

MarkLodato:
Context: https://github.com/slsa-framework/slsa/issues/309#issuecomment-1113559601

Add a "views expressed disclaimer" on guest posts. 

Clean up the styling of blog posts:
- Smaller header with white background.
- Format the date to look a bit nicer.

Preview: https://deploy-preview-379--slsa.netlify.app/blog/2022/04/slsa-is-no-free-lunch

top

[...]

bottom

# What it looked like before

old

### DIFF
--- a/docs/_layouts/post-list.html
+++ b/docs/_layouts/post-list.html
@@ -1,3 +1,6 @@
+---
+is_post: false
+---
 <!DOCTYPE html>
 <html lang="{{ page.lang | default: site.lang | default: "en" }}">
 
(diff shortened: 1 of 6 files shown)


inferno-chromium (approved):
Perfect, thank you so much. I would prefer the banner to come at end of post instead of front of every post, but open to other opinions.

MarkLodato:
> Perfect, thank you so much. I would prefer the banner to come at end of post instead of front of every post, but open to other opinions.

My feeling is that this is important to state up front. I don't think most readers will see a footer. Also official posts don't have this banner, e.g. for announcements.

Perhaps there's a way to make it less obtrusive?

inferno-chromium:
> > Perfect, thank you so much. I would prefer the banner to come at end of post instead of front of every post, but open to other opinions.
> 
> My feeling is that this is important to state up front. I don't think most readers will see a footer. Also official posts don't have this banner, e.g. for announcements.
> 
> Perhaps there's a way to make it less obtrusive?

Yes these kind of banners are not common, e.g. check https://blog.sigstore.dev/, https://kubernetes.io/blog/, https://openssf.org/blog/, that is why was suggesting at end or even skip. Lets get @slsa-framework/slsa-steering-committee and community opinion on this.

lumjjb:
+1 on notice at the bottom.

Other ideas:
- have an additional info button beside the "authored by" which will show text when hovered
- have a thinner SLSA blogpost banner on top on the blog title with a link to more about the SLSA blog or having the text there.

MarkLodato:
Sounds good, moved it to the bottom which is the simplest approach. I also updated the PR to include the new SBOM article and to simplify the implementation slightly.

lumjjb (approved):
LGTM

inferno-chromium (approved):
Thank you!

kimsterv:
lgtm!

PR #381, opened 2022-05-12
Re-label the threat diagram and add descriptions

MarkLodato:
This is PR contains a series of commits that simplify the threat diagram and better explain each category of threats. Each individual commit has a full explanation and readable diff. You may wish to review the commits individual rather than the PR overall.

Content changes:

* Use simpler and more consistent labels for threats (A)-(H).
* Add a description to each threat category.
* Split out Availability from Source Integrity since these are really two separate properties. I think this better explains the concept and highlights where there may still be gaps.
* Move threat (E) from the Dependencies box to the arrow above it, as originally intended (accidental change in #251). This threat is about the usage of a dependency, not the dependency itself.
* Move the build parameter threat to category (D), where I think it is a better fit.

Presentation changes:

* Move the definition of integrity to the threats page (from terminology). This way it is all explained on one page.
* Show threats (A)-(H) in the table of contents by increasing heading level.
* Reformat the example table to use HTML so that it's easier to maintain. (Not user visible.)

### Alternatives considered

For (C) and (F), I was originally looking for an adjective that means "has been tampered with", because "<X> that has been tampered with" is too verbose. In the end I chose "modified X" which gets the point across. It doesn't have the malicious intent connotation, but after some thought, I convinced myself that any modifications are a threat even if unintentional.

Here are the other adjectives considered, generated by polling some colleagues:

Adjective | My thoughts
--- | ---
tampered  | grammatically incorrect - "tampered X" sounds off
compromised | pretty good, but we already use it elsewhere so I was looking for a different adjective
adulterated | not bad
tainted | not bad, but I don't think it quite has the right connotation
inauthentic | not bad, but would need explanation; some readers might think signing is sufficient
contaminated | ugh
unverifiable | wrong meaning
bad | too generic


joshuagl (approved):
This is a nice set of improvements, thanks.

We might want to consider using a [data file](https://gohugo.io/templates/data-templates/) for the threats, separating presentation from data, to make reviewing changes easier?

joshuagl:
I appreciate the commit discipline here, it made the review much easier! :-)

PR #386, opened 2022-05-23
Refine 'build as code' requirement description

### LINKED ISSUE #368: Update "build as code" requirement to allow for build configuration stored in alternative locations
The ["build as code" requirement](https://slsa.dev/spec/v0.1/requirements#build-as-code) currently states that build definitions and configuration must be stored in version control, however there may be other ways to meet this requirement.

The spirit of the requirement seems to be that ultimately the definition should still be traceable back to version control but perhaps it can also be met by build definitions which are published from there to an intermediate location - for example Tekton has the concept of a [Tekton Bundle](https://github.com/tektoncd/pipeline/blob/main/docs/tekton-bundle-contracts.md) which is Tekton Task or Pipeline definitions that have been published to an OCI registry.

It would be great to update the "build as code" requirement to make it clear that while ultimately the origin of the build definition should be traceable back to version control (e.g. via provenance stored in the same OCI registry in the case of Tekton Bundles), not that the definition must have been fetched directly from version control.

(More context available in [SLSA + Tekton: Case Study](https://docs.google.com/document/d/12yYP0M_zz5Tc9ftGnLwhDZ_tdZBq4mEod70ngWtHmLs/edit#) - the doc is visible to anyone in mailing list slsa-discussion@googlegroups.com )

joshuagl:
Refine the text to make it clearer that the recipe and configuration must
be verifiably traceable to text files in an SCM, not necessarily fetched
from the SCM directly by the trusted control plane.

Fixes #368

### DIFF
--- a/docs/_spec/v0.1/requirements.md
+++ b/docs/_spec/v0.1/requirements.md
@@ -241,8 +241,8 @@ Examples: GitHub Actions, Google Cloud Build, Travis CI.
 <td>Build as code
 <td>
 
-The build definition and configuration which are executed by the build service
-is verifiably derived from text file definitions stored in a version control
+The build definition and configuration executed by the build service is
+verifiably derived from text file definitions stored in a version control
 system.
 
 Verifiably derived can mean either fetched directly through a trusted channel,
--- a/docs/_spec/v0.1/requirements.md
+++ b/docs/_spec/v0.1/requirements.md
@@ -241,10 +241,20 @@ Examples: GitHub Actions, Google Cloud Build, Travis CI.
 <td>Build as code
 <td>
 
-The build definition and configuration is defined in text files, stored in a
-version control system, and is executed by the build service.
+The build definition and configuration which are executed by the build service
+is verifiably derived from text file definitions stored in a version control
+system.
 
-Examples: cloudbuild.yaml, .github/workflows/build.yaml, zuul.yaml.
+Verifiably derived can mean either fetched directly through a trusted channel,
+or that the derived definition has some trustworthy provenance chain linking
+back to version control.
+
+Examples:
+
+-   .github/workflows/build.yaml stored in git
+-   Tekton bundles generated from text files by a SLSA compliant build process
+    and stored in an OCI registry with SLSA provenance metadata available.
+
 <td> <td> <td>✓<td>✓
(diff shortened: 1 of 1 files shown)


docs/_spec/v0.1/requirements.md
```diff
@@ -241,10 +241,20 @@ Examples: GitHub Actions, Google Cloud Build, Travis CI.
 <td>Build as code
 <td>
 
-The build definition and configuration is defined in text files, stored in a
-version control system, and is executed by the build service.
+The build definition and configuration which are executed by the build service
```
MarkLodato, line 244: nit: I think you can remove "which are".

MarkLodato (approved):
Thanks

docs/_spec/v0.1/requirements.md
joshuagl: Agreed, it makes sense to be more succinct. Removed in [eb89e57](https://github.com/slsa-framework/slsa/pull/386/commits/eb89e57c56d47f02f2efd6f9d94dc9319a205856)

docs/_spec/v0.1/requirements.md
```diff
@@ -241,10 +241,20 @@ Examples: GitHub Actions, Google Cloud Build, Travis CI.
...
+verifiably derived from text file definitions stored in a version control
+system.
+
+Verifiably derived can mean either fetched directly through a trusted channel,
+or that the derived definition has some trustworthy provenance chain linking
+back to version control.
+
+Examples:
+
+-   .github/workflows/build.yaml stored in git
```
adambkaplan, line 254: Would this require all information be "static" in the configuration, or are these definitions allowed to be parameterized?

For example - a typical GitHub action would include a checkout step. The checkout step receives (ex - through an environment variable) the commit SHA to clone. The SHA is not defined up front in `build.yaml`, but is implied through the context within the GitHub action (check out the main branch, a pull request branch, etc.).

joshuagl: No, this does not require all information to be static. The builder is trusted to record any variables from the event which triggers the build in the [`invocation`](https://slsa.dev/provenance/v0.2#invocation) object in the [provenance](https://slsa.dev/provenance/v0.2).

PR #399, opened 2022-06-20
✨ Recommend the slsa-github-generator for SLSA3

laurentsimon:
Signed-off-by: Laurent Simon [laurentsimon@github.com](laurentsimon@github.com)

/cc @asraa @MarkLodato  @ianlewis @joshuagl

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -89,7 +89,6 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
     </div>
 </section>
 
-
 <section class="section bg-white flex justify-center items-center">
     <div class="wrapper inner w-full">
         <div class="justify-between items-start md:-mr-10 md:-ml-10">
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -46,6 +46,42 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
         </div>
     </div>
 </section>
+
+<section class="section bg-pastel-green flex justify-center items-center">
+    <div class="wrapper inner w-full">
+        <div class="md:flex justify-between items-start mb-16">
+            <div class="text w-full md:w-1/3">
+<div class="h2 p-0">
+
+## Reaching SLSA Level 3
+
+</div>
+<p class="h4 font-semibold my-6 text-green-dark">Effort: Low</p>
+            </div>
+            <div class="w-full md:w-2/3">
+                <div class="bg-white h-full rounded-lg p-10">
+                    <p class="h4 font-bold mb-6">Overview<p>
+                    <p>
+                        This guide will help you achieve Level 3 on GitHub, and it should take less than a couple of hours for an individual project. The goals is to:
+                    <ul class="list-disc my-6 pl-6">
+                        <li>Use a GitHub trusted builder</li>
+                        <li>Verify the generate provenance</li>
+                    </ul>
+                    </p>
+                    <p class="h4 font-bold mb-6">Steps</p>
+                    <ul class="list-decimal mt-6 mb-10 pl-6">
+                        <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-github-generator">slsa-framework/slsa-github-generator</a> to use the vetted builder for your release</li>
+                        <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-github-generator">slsa-framework/slsa-github-generator</a> to use the vetted builder for your release
+                        </li>
+                        <li>You’re Level 3 for provenance integrity! Add the <a href="images/SLSA-Badge-full-level3.svg">SLSA Level 3 badge</a> to your project's readme.</li>
+                    </ul>
+                </div>
+            </div>
+        </div>
+    </div>
+</section>
+
+
 <section class="section bg-white flex justify-center items-center">
     <div class="wrapper inner w-full">
         <div class="justify-between items-start md:-mr-10 md:-ml-10">
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -80,7 +80,7 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
                     </ul>
                     <p class="h4 font-bold mb-6">Celebrate</p>
                     <ul class="list-decimal mt-6 mb-10 pl-6">
-                        <li>Update your documentation to let users know you generate provenance and encourage them to verify it when downloading your binaries. Also, consider contributing to the [SLSA blog](https://slsa.dev/blog) and let others know about your journey, or submit a [case study](https://github.com/slsa-framework/slsa/tree/main/case-studies)!</li>
+                        <li>Update your documentation to let users know you generate provenance and encourage them to verify it when downloading your binaries. Also, consider contributing to the <a href="https://slsa.dev/blog">SLSA blog</a> and let others know about your journey, or submit a <a href="https://github.com/slsa-framework/slsa/tree/main/case-studies">case study</a>!</li>
                         <li>You’re now generating non-forgeable <a href="https://slsa.dev/">SLSA provenance</a> that meets the <a href="https://slsa.dev/spec/v0.1/requirements#build-requirements">build</a> and <a href="https://slsa.dev/spec/v0.1/requirements#provenance-requirements">provenance</a> requirements for <a href="https://slsa.dev/spec/v0.1/levels">SLSA level 3 and above</a>! Add the <a href="images/SLSA-Badge-full-level3.svg">SLSA Level 3 badge</a> to your project's readme.</li>
                     </ul>
                 </div>
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -73,7 +73,7 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
                         <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-github-generator">slsa-framework/slsa-github-generator</a> to use the vetted builder for your release</li>
                         <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-github-generator">slsa-framework/slsa-github-generator</a> to use the vetted builder for your release
                         </li>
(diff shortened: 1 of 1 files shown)


docs/get-started.md
```diff
@@ -46,6 +46,42 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
...
+                    </ul>
+                    </p>
+                    <p class="h4 font-bold mb-6">Steps</p>
+                    <ul class="list-decimal mt-6 mb-10 pl-6">
+                        <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-github-generator">slsa-framework/slsa-github-generator</a> to use the vetted builder for your release</li>
+                        <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-framework">slsa-framework/slsa-verifier</a> to learn how to verify provenance for your release</li>
+                        <li>Update your documentation to tell users how to verify your provenance when downloading your binaries</li>
+                        <li>You’re now generating non-forgeable [SLSA provenance](https://slsa.dev/) on GitHub that meets the [build](https://slsa.dev/spec/v0.1/requirements#build-requirements) and [provenance](https://slsa.dev/spec/v0.1/requirements#provenance-requirements) requirements for [SLSA level 3 and above](https://slsa.dev/spec/v0.1/levels)! Add the <a href="images/SLSA-Badge-full-level3.svg">SLSA Level 3 badge</a> to your project's readme.</li>
```
asraa, line 76: nit: these links are markdown refs, not html so they don't render

docs/get-started.md
```diff
@@ -46,6 +46,42 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
...
+## Reaching SLSA Level 3
+
+</div>
+<p class="h4 font-semibold my-6 text-green-dark">Effort: Low</p>
+            </div>
+            <div class="w-full md:w-2/3">
+                <div class="bg-white h-full rounded-lg p-10">
+                    <p class="h4 font-bold mb-6">Overview<p>
+                    <p>
+                        This guide will help you achieve Level 3 on GitHub, and it should take less than a couple of hours for an individual project. The goals is to:
```
asraa, line 65: Is it worth direction to the generic or Golang ones?

docs/get-started.md
laurentsimon: ouch, i copied without reading. Thanks. Addressed

docs/get-started.md
laurentsimon: The generic projects have `This is a pre-release only and we will have the official release in July 2022.`, so I suppose we could keep the link as-is? (Simplifies maintenance)

Let's see if others have an opinion

docs/get-started.md
```diff
@@ -46,6 +46,42 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
...
+                    </p>
+                    <p class="h4 font-bold mb-6">Steps</p>
+                    <ul class="list-decimal mt-6 mb-10 pl-6">
+                        <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-github-generator">slsa-framework/slsa-github-generator</a> to use the vetted builder for your release</li>
+                        <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-framework">slsa-framework/slsa-verifier</a> to learn how to verify provenance for your release</li>
+                        <li>Update your documentation to tell users how to verify your provenance when downloading your binaries</li>
+                        <li>You’re now generating non-forgeable <a href="https://slsa.dev/">SLSA provenance</a> on GitHub that meets the <a href="https://slsa.dev/spec/v0.1/requirements#build-requirements">build</a> and <a href="https://slsa.dev/spec/v0.1/requirements#provenance-requirements">provenance</a> requirements for <a href="https://slsa.dev/spec/v0.1/levels">SLSA level 3 and above</a>! Add the <a href="images/SLSA-Badge-full-level3.svg">SLSA Level 3 badge</a> to your project's readme.</li>
```
mlieberman85, line 76: Does this only support v0.1 of the spec?

laurentsimon: good catch. We actually use v0.2 for our provenance, but I'm unable to get a link: https://slsa.dev/spec/v0.1 gives 404. Do you know the link?

mlieberman85: Oh, this is my mistake. The requirements spec might still be 0.1, but the in-toto predicate format spec for provenance might be 0.2? That's actually kind of confusing.

laurentsimon: You're correct. Sorry I was confused too :-). So we're good then. Resolving the issue.

docs/get-started.md
```diff
@@ -46,6 +46,42 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
...
+                        This guide will help you achieve Level 3 on GitHub, and it should take less than a couple of hours for an individual project. The goals is to:
+                    <ul class="list-disc my-6 pl-6">
+                        <li>Use a GitHub trusted builder</li>
+                        <li>Verify the generate provenance</li>
+                    </ul>
+                    </p>
+                    <p class="h4 font-bold mb-6">Steps</p>
+                    <ul class="list-decimal mt-6 mb-10 pl-6">
+                        <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-github-generator">slsa-framework/slsa-github-generator</a> to use the vetted builder for your release</li>
+                        <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-framework">slsa-framework/slsa-verifier</a> to learn how to verify provenance for your release</li>
```
ianlewis, line 74:
```suggestion
                        <li>Follow the steps in <a href="https://github.com/slsa-framework/slsa-verifier">slsa-framework/slsa-verifier</a> to learn how to verify provenance for your release.</li>
```

docs/get-started.md
ianlewis, line 76: Maybe add a step to evaluate other SLSA 3 requirements since using `slsa-github-generator` is not necessarily sufficient for SLSA 3.

docs/get-started.md
```diff
@@ -46,6 +46,42 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
...
     </div>
 </section>
+
+<section class="section bg-pastel-green flex justify-center items-center">
+    <div class="wrapper inner w-full">
+        <div class="md:flex justify-between items-start mb-16">
+            <div class="text w-full md:w-1/3">
+<div class="h2 p-0">
+
+## Reaching SLSA Level 3
```
ianlewis, line 56: Maybe we don't need to for this PR but I think it would be helpful to have different sections that cover different requirements.

For example, `slsa-github-generator` could fall under a "Build & Provenance requirements" section. While something like [gitsign](https://github.com/sigstore/gitsign) and/or GitHub commit verification could fall under "Source requirements".

docs/get-started.md
```diff
@@ -46,6 +46,42 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
...
+<p class="h4 font-semibold my-6 text-green-dark">Effort: Low</p>
+            </div>
+            <div class="w-full md:w-2/3">
+                <div class="bg-white h-full rounded-lg p-10">
+                    <p class="h4 font-bold mb-6">Overview<p>
+                    <p>
+                        This guide will help you achieve Level 3 on GitHub, and it should take less than a couple of hours for an individual project. The goals is to:
+                    <ul class="list-disc my-6 pl-6">
+                        <li>Use a GitHub trusted builder</li>
+                        <li>Verify the generate provenance</li>
```
MarkLodato, line 68: Let's remove this for now. Verification/policy is a big hole in SLSA today. I don't think we should just tell people to "verify" without further explanation of what they are trying to achieve or how to do so securely. Or if we do, add a caveat that it is not properly documented.

docs/get-started.md
```diff
@@ -46,6 +46,42 @@ hero_text: If you’re looking to jump straight in and try SLSA, here’s a quic
...
+</div>
+<p class="h4 font-semibold my-6 text-green-dark">Effort: Low</p>
+            </div>
+            <div class="w-full md:w-2/3">
+                <div class="bg-white h-full rounded-lg p-10">
+                    <p class="h4 font-bold mb-6">Overview<p>
+                    <p>
+                        This guide will help you achieve Level 3 on GitHub, and it should take less than a couple of hours for an individual project. The goals is to:
+                    <ul class="list-disc my-6 pl-6">
+                        <li>Use a GitHub trusted builder</li>
```
MarkLodato, line 67: This makes it sound like GitHub is required to reach SLSA 3, which is not true. Instead, this section needs to be more generic (similar to above) to say that the goal is to have a tamper-proof build process that generates trustworthy provenance, and suggest various ways to achieve that.

docs/get-started.md
MarkLodato: IMO this page needs a complete rewrite. It needs to be more focused on who the audience is, what they're trying to accomplish, and how they can do so.

MarkLodato (changes requested):
Let's get review from @slsa-framework/slsa-steering-committee. In particular, how do you think it best to suggest the various tools while being impartial as to vendor?

docs/get-started.md
laurentsimon: I was not sure whether you meant we need to re-write the entire get-started.md page or just the SLSA3 section of this PR. I've assumed the latter and updated this PR to address all other comments. PTAL

docs/get-started.md
laurentsimon: Done, thanks

docs/get-started.md
laurentsimon: Done, PTAL

docs/get-started.md
MarkLodato: Not for this PR. I was just agreeing with Ian that this page needs work.

MarkLodato (approved):

docs/get-started.md
laurentsimon: SG. I've created an issue for tracking https://github.com/slsa-framework/slsa/issues/407. Resolving this comment.

laurentsimon:
friendly ping https://github.com/orgs/slsa-framework/teams/slsa-steering-committee, can we get an additional LGTM?

kimsterv (approved):

PR #400, opened 2022-06-21
fix: code formatting color

asraa:
Signed-off-by: Asra Ali <asraa@google.com>

The code text over the code background does not have enough contrast.


asraa:
image


Text is readable now

laurentsimon (approved):

inferno-chromium (approved):
Minor color fix, fixing a user complaint, so merging.

inferno-chromium:
There is more color breakage with highlighted words seeimage

joshuagl:
We should probably revert this change until a fix is available.

inferno-chromium:
> We should probably revert this change until a fix is available.

https://github.com/slsa-framework/slsa/pull/402

PR #402, opened 2022-06-22
Fix code color regression, match code block background with code word

inferno-chromium:
Fixes regression from slsa-framework/slsa#400, match code block background with code word

### DIFF
--- a/docs/_sass/minima/_base.scss
+++ b/docs/_sass/minima/_base.scss
@@ -154,8 +154,8 @@ code {
 
 code {
     padding: 4px 6px;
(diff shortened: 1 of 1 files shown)


inferno-chromium:
See image after fix, both code block and code word backgrounds match.
image

dlorenc (approved):

MarkLodato:
This makes the inline code quite ugly. The dark green background is jarring. Is there a way to keep the light green for inline code but dark green for blocks? Or alternatively, switch to light green everywhere?

PR #411, opened 2022-06-29
Vendor all assets to fully self-host site

MarkLodato:
Previously we used CDNs to serve CSS, Javascript, and fonts. This made the site difficult to develop locally since it requires an internet connection. (I found this out while trying to develop on the train today.)

Now all external assets are copied into our git repository so that it can build and serve fully locally.

Sources:

- vendor/*: URLs found in previous head.html, resolving alpinejs 3.x.x to 3.10.2.

- fonts/ibm_plex: https://github.com/IBM/plex/releases/tag/v6.0.2, extracting Web/IBM-Plex-Mono/fonts/split/woff2/IBMPlexMono-Regular*

- fonts/inter: https://github.com/rsms/inter/releases/tag/v3.19

Note: Woff only adds [1% extra browser share](https://caniuse.com/?search=woff) over woff2, so it doesn't seem worth the effort to get those for IBM Plex Mono.


MarkLodato:
> You can still use a CDN if you have all assets self-hosted, you just have to access the site through a CDN you trust. We already do this with the best practices badge (we use Fastly; CloudFlare is another CDN often used used for this purpose).

Sorry, I don't understand. My goal is to be able to build and test the site without an internet connection. Do you suggest something other than what I've done in this PR?

trishankatdatadog (approved):

MarkLodato:
Ah, thanks! And thanks @trishankatdatadog for the quick review!

PR #415, opened 2022-07-07
Changes wording of non-falsfiability requirement

### LINKED ISSUE #414: SLSA should not implicitly require digital signatures
Raising a new issue as per https://github.com/slsa-framework/slsa/issues/191#issuecomment-1162388651

It appears that digital signing of attestations is almost impossible to avoid at SLSA 2+ even though the wording of the requirements seems to want to allow other forms of verification:

>The provenance’s authenticity and integrity can be verified by the consumer. This SHOULD be through a digital signature from a private key accessible only to the service generating the provenance.

at SLSA2 and then:

> Provenance cannot be falsified by the build service’s users.

> NOTE: This requirement is a stricter version of [Service Generated](https://slsa.dev/spec/v0.1/requirements#service-generated).

> The provenance signing key MUST be stored in a secure key management system accessible only to the build service account.
The provenance signing key MUST NOT be accessible to the environment running the user-defined build steps.

at SLSA 3. At no point do the requirements explicitly specify that provenance data be digitally signed to be non-falsifiable, but the two requirements regarding signing keys imply there must be a signing key and hence you must use a digital signature scheme. We should clarify that SLSA 3 explicitly requires digital signatures for provenance data if that is the intent, or we should alter the wording of these requirements to allow alternative mechanisms to ensure non-falsifiability.

shaunmlowry:
fixes #414 - don't accidentally force implementors to use digital
signatures to prove non-falsifiability.  Instead, leave things open
enough that other implementations are acceptable.

### DIFF
--- a/docs/_spec/v0.1/requirements.md
+++ b/docs/_spec/v0.1/requirements.md
@@ -400,9 +400,11 @@ Provenance cannot be falsified by the build service's users.
 
 NOTE: This requirement is a stricter version of [Service Generated](#service-generated).
 
--   The provenance signing key MUST be stored in a secure key management system
-    accessible only to the build service account.
--   The provenance signing key MUST NOT be accessible to the environment running
+-   Any secret material used to demonstrate the non-falsifiable nature of
+    the provenance, for example the signing key used to generate a digital
+    signature, MUST be stored in a secure management system appropriate for
+    such material and accessible only to the build service account.
+-   Such secret material MUST NOT be accessible to the environment running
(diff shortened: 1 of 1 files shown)


MarkLodato (approved):
Looks great, thanks for the fix!

MarkLodato:
Merging now since the new wording better reflects the original intention and it does not significantly change the meaning of the spec. I don't anticipate anyone disagreeing with this change. But if there is any disagreement, we can just revert or fix.

Issue #335, opened 2022-03-28
slsa.dev style: missing space between paragraphs


MarkLodato:
The CSS on the slsa.dev website removes virtual whitespace between paragraphs, which makes it really hard to read. This is because (at least) there are two different rules that set `p { margin: 0 }`, one in main.css and one in tailwind.min.css. Adding a margin after the paragraph will fix it, though care must be taken to not break anything else. (I wish we had an easy visual diffing tool.)

If anyone wants to send out a PR to fix this, it would be most appreciated!

# Example

https://slsa.dev/verification_summary/v0.1

Current:

![missing space](https://user-images.githubusercontent.com/58860/160475063-7a82604e-62fc-4cbd-b83b-cb535a82f4db.png)

Expected:

![has space](https://user-images.githubusercontent.com/58860/160475099-3702466f-37a8-4c25-93ea-3b78cb9f0f38.png)

MarkLodato:
Tailwind also ruins `<ol>` tags by setting `list-style: none`. I think we need to get rid of tailwind and instead use proper css.

MarkLodato:
If we use add the `markdown` class, it fixes most issues, but `blockquote` elements are formatted as horizontal boxes (flex model, different color, etc.). We'd need to fix that before applying that fix to the standards pages.

MarkLodato:
Also bulleted list spacing:

![image](https://user-images.githubusercontent.com/58860/200342536-0a4046aa-c9ab-45c8-9e97-53e0f057e545.png)

PR #420, opened 2022-07-11
Hack to fix paragraph spacing

MarkLodato:
The proper fix would be to fix the CSS, but for now the lack of inter-paragraph spacing (#335) is highly distracting so this is a cheap fix.

Preview: TODO

image


joshuagl (approved):
Thanks, this looks much better.

PR #422, opened 2022-07-13
Complete rewrite of use cases page

MarkLodato:
* Replace the old "use cases" content a more useful categorization: first party, open source, or vendor. This breakdown better reflects the different challenges of each category.
* Replace the old contrived "case studies" with links to real-world case studies, one per category.
* Remove extra margin at bottom of each box.

**Note to reviewers:** I'm not 100% confident in this, so I'm looking for feedback as to whether this is an improvement over the previous version, or if you have other ideas on how best to represent these ideas.


docs/use-cases.md
```diff
@@ -1,132 +1,135 @@
...
+</div>
+
+</div>
+<div class="md:w-2/3">
+
+<!--
+SLSA can be used entirely within an organization to reduce the risk of tampering
+by insiders or compromised insider credentials. In other words, SLSA protects
+an organization's first party software.
+-->
```
joshuagl, line 36: nit: should this be removed?

joshuagl (approved):
Thank you Mark. IMO this is an improvement. I like the much briefer page you have authored and very much prefer linking to real-world case studies, it's great that we have several to choose from now.

Aside, and only a minor nit: I have no idea if there's a matching number of opening and closing `<div>` elements. 😕

docs/use-cases.md
MarkLodato: Oops! Thanks for catching. Removed.

MarkLodato:
> Aside, and only a minor nit: I have no idea if there's a matching number of opening and closing `<div>` elements.

Yeah, this setup drives me insane. I really want to have a proper CSS so that we can write plain Markdown, rather than littering it with lots of HTML that does styling.

MarkLodato:
I'll leave this open for another day at least to give others time to comment, then merge if there's nothing more.

PR #437, opened 2022-07-22
Remove protocol and domain from internal links

joshuagl:
This makes disconnected local development more reliable


MarkLodato (approved):
Thanks!

PR #438, opened 2022-07-22
Move attestation model into the site

joshuagl:
This moves the attestation model, referenced from the Provenance and VSA pages, into the main site so we are not bouncing readers who follow the link to the GitHub repository.

Fixes #323

PR #444, opened 2022-07-25
RFC: Support markdown in hero_text for specification and standard?

joshuagl:
While playing around with layout for the Software attestations page in #438 I discovered that the `hero_text` did not support Markdown formatting. This PR enables support for Markdown in the `hero_text`. As I ultimately did not need it for #438 I pulled it out into a separate PR to see whether it's considered useful?


MarkLodato (approved):
LGTM. Thanks!

PR #446, opened 2022-07-25
Create verification summary specification v0.2

AdamZWu:
Introduces `resource_uri` field in predicate to hold the resource URI used for verification, instead of the statement field `subject.name`.

### DIFF
--- a/docs/verification_summary/v0.2-draft.md
+++ b/docs/verification_summary/v0.2-draft.md
@@ -3,6 +3,11 @@ title: SLSA Verification Summary Attestation (VSA)
 layout: standard
 hero_text: Verification summary attestations communicate that an artifact has been verified at a specific SLSA level and details about that verification.
 ---
+
+**IMPORTANT:** This is a DRAFT version of the VSA spec that is **subject to
+change.** Once this version is finalized, the "-draft" suffix will be removed.
+For the latest stable version, see [v0.1](0.1).
+
 <details class="mt-12">
 <summary>Note about 0.x versions</summary>
 
@@ -68,8 +73,8 @@ all of an artifact’s _transitive_ dependencies.  The verifier does this by
 either a) verifying the provenance of each non-source dependency listed in
 the [materials](/provenance/v0.2#materials) of the artifact
 being verified (recursively) or b) matching the non-source dependency
-listed in `materials` (by subject.digest == materials.digest and, ideally,
-subject.name == materials.uri) to a VSA _for that dependency_ and using
+listed in `materials` (by `subject.digest` == `materials.digest` and, ideally,
+`vsa.resource_uri` == `materials.uri`) to a VSA _for that dependency_ and using
 `vsa.policy_level` and `vsa.dependency_levels`.  Policy verifiers wishing
 to establish minimum requirements on dependencies SLSA levels may use
 `vsa.dependency_levels` to do so.
@@ -84,18 +89,19 @@ to establish minimum requirements on dependencies SLSA levels may use
 // Standard attestation fields:
 "_type": "https://in-toto.io/Statement/v0.1",
 "subject": [{
-  "name": <artifact-URI-in-request>,
+  "name": <NAME>,
   "digest": { <digest-in-request> }
 }],
 
 // Predicate
-"predicateType": "https://slsa.dev/verification_summary/v0.1",
+"predicateType": "https://slsa.dev/verification_summary/v0.2-draft",
 "predicate": {
   // Required
   "verifier": {
     "id": "<URI>"
   },
   "time_verified": <TIMESTAMP>,
+  "resource_uri": <artifact-URI-in-request>,
   "policy": {
     "uri": "<URI>",
     "digest": { /* DigestSet */ }
@@ -161,6 +167,11 @@ of the other top-level fields, such as `subject`, see [Statement]._
 
 > Timestamp indicating what time the verification occurred.
 
+<a id="resource_uri"></a>
+`resource_uri` _string ([ResourceURI]), required_
+
+> URI that identifies the resource associated with the artifact being verified.
+
 <a id="policy"></a>
 `policy` _object, required_
 
@@ -206,17 +217,18 @@ WARNING: This is just for demonstration purposes.
 ```jsonc
 "_type": "https://in-toto.io/Statement/v0.1",
 "subject": [{
-  "name": "https://example.com/example-1.2.3.tar.gz",
+  "name": "out/example-1.2.3.tar.gz",
   "digest": {"sha256": "5678..."}
 }],
 
 // Predicate
-"predicateType": "https://slsa.dev/verification_summary/v0.1",
+"predicateType": "https://slsa.dev/verification_summary/v0.2-draft",
 "predicate": {
   "verifier": {
     "id": "https://example.com/publication_verifier"
   },
   "time_verified": "1985-04-12T23:20:50.52Z",
+  "resource_uri": "https://example.com/example-1.2.3.tar.gz",
   "policy": {
     "uri": "https://example.com/example_tarball.policy",
     "digest": {"sha256": "1234..."}
@@ -255,6 +267,7 @@ Must be one of these values:
 
 </div>
 
+-   0.2-draft: Added `resource_uri` field.
 -   0.1: Initial version.
 
 [SlsaResult]: #slsaresult


docs/verification_summary/v0.2.md
```diff
@@ -0,0 +1,277 @@
...
+
+<a id="time_verified"></a>
+`time_verified` _string ([Timestamp]), required_
+
+> Timestamp indicating what time the verification occurred.
+
+<a id="resource_uri"></a>
+`resource_uri` _string ([ResourceURI]), required_
+
+> URI that identifies the resource associated with the artifact being verified.
```
joshuagl, line 170: It would be great if this was a little clearer, we might do that by being more terse, i.e.
```suggestion
> URI that identifies the artifact being verified.
```
or perhaps by being more verbose, maybe even including some design rationale?

joshuagl (changes requested):
I fully support the idea of an explicit field rather than overloading the `name` field in the `subject`. Do you think it's possible to provide more concrete guidance on what the `resource_uri` field should reference? Or is there ambiguity in what this might resolve to, depending on the type of artifact?

docs/verification_summary/v0.2.md
```diff
@@ -0,0 +1,277 @@
...
+in-toto attestation [Statement]) by evaluating the artifact and a `bundle`
+of attestations against some `policy`.  Users who trust the `verifier` may
+assume that the artifacts met the indicated SLSA level without themselves
+needing to evaluate the artifact or to have access to the attestations the
+`verifier` used to make its determination.
+
+The VSA also allows consumers to determine the verified levels of
+all of an artifact’s _transitive_ dependencies.  The verifier does this by
+either a) verifying the provenance of each non-source dependency listed in
+the [materials](https://slsa.dev/provenance/v0.2#materials) of the artifact
```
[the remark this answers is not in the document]
AdamZWu: Thanks, change applied.

docs/verification_summary/v0.2.md
```diff
@@ -0,0 +1,277 @@
...
+<div class="mt-16 mb-4">
+
+## Example
+
+</div>
+
+WARNING: This is just for demonstration purposes.
+
+```jsonc
+"_type": "https://in-toto.io/Statement/v0.2",
```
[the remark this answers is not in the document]
AdamZWu: Sorry, my bad. :P

docs/verification_summary/v0.2.md
AdamZWu: IMO the concept of "resource" is a bit different from that of the "artifact", from the supply-chain *internal* perspective.

- A resource identifies a SLSA protected "perimeter", e.g. a storage location, a well-known package name, a cryptographic key material, etc.;
- An artifact is a piece of data that wants to interact with a resource, e.g. stored in a location, claim a package name, signed by a key, etc., but may or may not qualify that interaction.

SLSA verification plays the role of checking the qualification of an artifact *before* it is allowed to interact with a resource. 

After the interaction took place, however, from the supply-chain *consumer's* perspective, "resource" and "artifact" are effectively the same thing, because the interaction above has established bindings between the two.

AdamZWu:
> Do you think it's possible to provide more concrete guidance on what the resource_uri field should reference? Or is there ambiguity in what this might resolve to, depending on the type of artifact?

Please see my above comments on "resource" vs. "artifact".

IMO the concept of resource is very much wide-open, so `resource_uri` really can be anything, depending on the scenario and needs of supply-chain protections.

MarkLodato:
Q: Do we need a new version for this, or can/should this be a backwards compatible change to v0.1? I know that old clients can just ignore the new field without chaning any information, but I guess the question is what to do if they were relying on the old `subject.name` field.

AdamZWu:
> Q: Do we need a new version for this, or can/should this be a backwards compatible change to v0.1?

Since it is not a pure addition of new fields, maybe it is appropriate to bump up the version?

docs/verification_summary/v0.2.md
```diff
@@ -0,0 +1,277 @@
...
+been included).
+
+<div class="mt-16 mb-4">
+
+## Prerequisite
+
+</div>
+
+Understanding of SLSA [Software
+Attestations](https://github.com/slsa-framework/slsa/blob/main/controls/attestations.md),
```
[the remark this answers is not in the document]
AdamZWu: Pulled upstream and applied link changes.

AdamZWu:
Rebased from the latest head and squashed previous revisions.

docs/verification_summary/v0.2.md
AdamZWu: Hi @joshuagl, I wonder whether this address your concern on "resource"?

Compared with VSA, I think the "resource" concept is more relevant to the topic of policy, maybe we can leave the more verbose explanations in there, and link to that doc when it becomes available?

joshuagl:
> Q: Do we need a new version for this, or can/should this be a backwards compatible change to v0.1? I know that old clients can just ignore the new field without chaning any information, but I guess the question is what to do if they were relying on the old `subject.name` field.

Somewhat related to this question, should we mark v0.2 as a draft? We did this with the core spec in the past so that we could collect several changes together before making a new release.

docs/verification_summary/v0.2.md
AdamZWu: Sounds good. Resolving for now.

AdamZWu:
> should we mark v0.2 as a draft?

Sounds good to me, will revise.

AdamZWu:
- Renamed to v0.2-draft
- Pulled the latest changes from master

docs/verification_summary/index.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en-US">
   <title>Redirecting&hellip;</title>
-  <meta http-equiv="refresh" content="0; url=v0.1">
-  <link rel="canonical" href="v0.1">
+  <meta http-equiv="refresh" content="0; url=v0.2-draft">
+  <link rel="canonical" href="v0.2-draft">
```
joshuagl, line 5: I think we should continue to redirect to 1.0 until this is non-draft, WDYT?

docs/verification_summary/index.md
```diff
@@ -1,5 +1,5 @@
 ---
 layout: redirect
-redirect_to_url: v0.1
+redirect_to_url: v0.2-draft
```
joshuagl, line 3: Same comment here, I think this change is premature.

docs/verification_summary/v0.1.md
```diff
@@ -89,7 +89,7 @@ to establish minimum requirements on dependencies SLSA levels may use
 }],
 
 // Predicate
-"predicateType": "/verification_summary/v0.1",
+"predicateType": "https://slsa.dev/verification_summary/v0.1",
```
joshuagl, line 92: Whoops! Thanks for fixing this.

joshuagl (changes requested):
Overall, this looks good to me as a draft. Thanks for the iterations! 

I don't think we should be redirecting to VSA v0.2 until it's no longer a draft, though.
I'd like to see a review from @TomHennen, as he's proposed as code owner for the VSA in #449.

I think we should file a follow-on issue to ensure we do provide more guidance around `resource_uri`.

TomHennen:
> I'd like to see a review from @TomHennen, as he's proposed as code owner for the VSA in #449.

Oh yes, I probably should have mentioned at the outset that I've been working pretty closely with Adam on this and suggested he submit this PR.  So I thought my review would be a conflict of interest in a way.  Next time I'll comment initially saying that I support it but that I'd like another review as well.

So yes, I approve of this change.  Keeping it as draft for perhaps the couple weeks makes sense to me.  We're starting to use this more internally and may identify some other helpful changes.

docs/verification_summary/index.html
AdamZWu: Ah sure, reverted.

docs/verification_summary/index.md
AdamZWu: Reverted.

docs/verification_summary/v0.1.md
AdamZWu: :D

AdamZWu:
Revised and rebased.

docs/verification_summary/v0.2-draft.md
```diff
@@ -0,0 +1,275 @@
+---
+title: SLSA Verification Summary Attestation (VSA)
+layout: standard
+hero_text: Verification summary attestations communicate that an artifact has been verified at a specific SLSA level and details about that verification.
+---
```
joshuagl, line 5: Let's try and ensure any readers are very aware that this is not yet final, we had text similar to this in the v0.2-draft provenance spec:
```suggestion
---

**IMPORTANT:** This is a DRAFT version of the VSA spec that is **subject to
change.** Once this version is finalized, the "-draft" suffix will be removed.
For the latest stable version, see [v0.1][0.1].

```

docs/verification_summary/v0.2.md
```diff
@@ -0,0 +1,5 @@
+---
+layout: redirect
+redirect_to_url: v0.2-draft
+sitemap: false
```
joshuagl, line 4: I'm not convinced it makes sense to redirect from a seeming final versioned url to a draft 🤔

joshuagl (changes requested):
Apologies for so much seemingly trivial back and forth on this, but I'd like to be very clear that this is not the final v0.2 so I've suggested an extra warning in the spec text itself and removing the redirect.

docs/verification_summary/v0.2-draft.md
AdamZWu: Sounds good. Thanks~

docs/verification_summary/v0.2.md
AdamZWu: This is for coherency to the "predicateType" URI, i.e. make it point to something that has meaningful content, instead of 404. I think with the latest additional disclaimer banner there should not be misunderstandings. WDYT?

joshuagl: For the 2.0-alpha of the provenance spec we had an alpha-specific TypeURI, i.e. https://github.com/slsa-framework/slsa/commit/a2786e93540d8ec5db20a19b96aad9886a843c3e. I think we should do the same here and have added some suggestions in-line in the PR, WDYT?

AdamZWu: Sure, sounds good.

AdamZWu:
Adopted suggestions and rebased.

TomHennen (approved):

joshuagl (approved):
Thanks for working through this!

PR #468, opened 2022-08-17
Add `input_attestations` to VSA and bump up to v0.2

### LINKED ISSUE #387: VSA: should reference input attestations
The Verification Summary Attestation should list the input attestations that it used as inputs to the decision. This gives us traceability and repeatability of the process.

I'm thinking something like:

```json
 "inputAttestations": [
    {
      "uri": "...(optional)...",
      "digest": { "sha256": "abcd1234..." }
    }
  ]
```

@TomHennen thoughts?

AdamZWu:
Adds `input_attestations` to fix #387.
Also takes VSA v0.2 out of draft status.

### DIFF
--- a/docs/verification_summary/v0.2.md
+++ b/docs/verification_summary/v0.2.md
@@ -4,10 +4,6 @@ layout: standard
 hero_text: Verification summary attestations communicate that an artifact has been verified at a specific SLSA level and details about that verification.
 ---
 
-**IMPORTANT:** This is a DRAFT version of the VSA spec that is **subject to
-change.** Once this version is finalized, the "-draft" suffix will be removed.
-For the latest stable version, see [v0.1](0.1).
-
 <details class="mt-12">
 <summary>Note about 0.x versions</summary>
 
@@ -94,7 +90,7 @@ to establish minimum requirements on dependencies SLSA levels may use
 }],
 
 // Predicate
-"predicateType": "https://slsa.dev/verification_summary/v0.2-draft",
+"predicateType": "https://slsa.dev/verification_summary/v0.2",
 "predicate": {
   // Required
   "verifier": {
@@ -106,6 +102,13 @@ to establish minimum requirements on dependencies SLSA levels may use
     "uri": "<URI>",
     "digest": { /* DigestSet */ }
   }
+  "input_attestations": [
+    {
+      "uri": "<URI>",
+      "digest": { <digest-of-attestation-data> }
+    },
+    ...
+  ],
   "verification_result": "<PASSED|FAILED>",
   "policy_level": "<SlsaResult>",
   "dependency_levels": {
@@ -187,6 +190,35 @@ of the other top-level fields, such as `subject`, see [Statement]._
 
 > Collection of cryptographic digests for the contents of the policy used to perform verification.
 
+<a id="input_attestations"></a>
+`input_attestations` _object, optional_
+
+> The collection of attestations that were used to perform verification.
+> Conceptually similar to the `materials` field in the [SLSA Provenance](/provenance).
+>
+> This field can be absent if the verifier does not support this feature;
+> If present, this field must contain information on _all_ the attestations
+> used to perform verification.
+> An empty collection indicates the decision was made without attestations.
+
+<a id="input_attestations.uri"></a>
+`input_attestations[*].uri` _string ([ResourceURI]), optional_
+
+> A URI that can be used (provided the requester has read permission) to fetch
+> the attestation.
+>
+> An absent field indicates the verifier does not know where to fetch the attestation
+> (e.g. it was supplied by the verification requester).
+>
+> The data fetched from the URI may contain more than one attestation (e.g. an
+> attestation bundle file). In that case, the attestation is identified by
+> `digest` field values.
+
+<a id="input_attestations.digest"></a>
+`input_attestations[*].digest` _object ([DigestSet]), required_
+
+> Collection of cryptographic digests for this attestation.
+
 <a id="verification_result"></a>
 `verification_result` _string, required_
 
@@ -222,7 +254,7 @@ WARNING: This is just for demonstration purposes.
 }],
 
 // Predicate
-"predicateType": "https://slsa.dev/verification_summary/v0.2-draft",
+"predicateType": "https://slsa.dev/verification_summary/v0.2",
 "predicate": {
   "verifier": {
     "id": "https://example.com/publication_verifier"
@@ -233,6 +265,12 @@ WARNING: This is just for demonstration purposes.
     "uri": "https://example.com/example_tarball.policy",
     "digest": {"sha256": "1234..."}
   },
+  "input_attestations": [
+    {
+      "uri": "https://example.com/provenances/example-1.2.3.tar.gz.intoto.jsonl",
+      "digest": {"sha256": "abcd..."}
+    }
+  ],
   "verification_result": "PASSED",
   "policy_level": "SLSA_LEVEL_3",
   "dependency_levels": {
@@ -267,7 +305,9 @@ Must be one of these values:
 
 </div>
 
--   0.2-draft: Added `resource_uri` field.
+-   0.2:
+    -   Added `resource_uri` field.
+    -   Added optional `input_attestations` field.
 -   0.1: Initial version.
 
 [SlsaResult]: #slsaresult
--- a/docs/verification_summary/index.html
+++ b/docs/verification_summary/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en-US">
   <title>Redirecting&hellip;</title>
-  <meta http-equiv="refresh" content="0; url=v0.1">
-  <link rel="canonical" href="v0.1">
+  <meta http-equiv="refresh" content="0; url=v0.2">
+  <link rel="canonical" href="v0.2">
   <meta name="robots" content="noindex">
 </html>
--- a/docs/verification_summary/index.md
+++ b/docs/verification_summary/index.md
@@ -1,5 +1,5 @@
 ---
 layout: redirect
-redirect_to_url: v0.1
+redirect_to_url: v0.2
 sitemap: false
 ---


docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,25 @@ of the other top-level fields, such as `subject`, see [Statement]._
 
 > Collection of cryptographic digests for the contents of the policy used to perform verification.
 
+<a id="input_attestations"></a>
+`input_attestations` _object, optional_
+
+> The collection of attestation bundles that are used to perform verification.
```
TomHennen, line 196: I wonder if instead of "The collection of attestation _bundles_" this should be the collection of "_attestations_"?

docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,25 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+> Conceptually similar to the `materials` field in the [SLSA Provenance](/provenance).
+>
+> A verifier MAY utilize this field to record information for facilitating location of the attestations,
+> e.g. for auditing, reproducing, or renewing the verification results.
+
+<a id="input_attestations.uri"></a>
+`input_attestations[*].uri` _string ([ResourceURI]), optional_
+
+> The URI of the attestation bundle used to perform verification.
```
TomHennen, line 205: Can you indicate what it means if this field is unset?  Probably just that the verifier doesn't know of a URI where the attestation is available?

docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,25 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+
+<a id="input_attestations.digest"></a>
+`input_attestations[*].digest` _object ([DigestSet]), optional_
```
TomHennen, line 208: Should the digest be required?  It should always be available if inputs_attestations is non-empty.  It seems like it would be odd to have attestations listed here without digests, What would be the point?

docs/verification_summary/v0.2.md
```diff
@@ -233,6 +255,12 @@ WARNING: This is just for demonstration purposes.
     "uri": "https://example.com/example_tarball.policy",
     "digest": {"sha256": "1234..."}
   },
+  "input_attestations": [
+    {
+      "uri": "https://example.com/provenances/example-1.2.3.tar.gz.intoto.jsonl",
```
TomHennen, line 260: I'm not sure if we want to address this here, but it's possible that example-1.2.3.tar.gz could be rebuilt with new attestations at some point.  If so the referenced digest below might not match.

We could:
1. Indicate that this is a possibility in the docs for the URI field.
2. Use the digest of the attestation in the URI to make it unambigious.
3. Something else?

TomHennen (changes requested):
Thanks for doing this Adam!

Can you update the PR description to reference the issues this PR addresses and maybe describe it a bit more?

I'd also like to get @joshuagl's thoughts.

docs/verification_summary/v0.2.md
AdamZWu: Since we don't dictate how the supply chain must provide attestation bundles, I think the best we can do is to provide sufficient mechanism to *allow* unambiguous identification of the data?

That is, in the current specs, we consider "a URI + a digest" should be sufficient for the systems that adopt this specs to resolve and locate any attestation bundles. How or whether they use them, we don't care.

If we get feedback that these two fields are insufficient, we can add more.

docs/verification_summary/v0.2.md
AdamZWu: Since all attestations are individually signed, having another checksum around them doesn't provide protection for their integrity or correctness of verification.

So, to me, the digests here seem only serve the purpose of providing a locator.

And if that is the case, if a URI already provides sufficient information (e.g. content addressable URI), digests are not necessary.

docs/verification_summary/v0.2.md
AdamZWu: This is a potential clarification point.

I think we are running into a limitation of current SLSA specs.
Since here we are dealing with data retrieval, it involves storage and transportation, which is not well defined yet: https://slsa.dev/attestation-model#recommended-suite

The furthest I can go is Bundle, and I feel this *may be* OK -- but I think if we go further up, e.g. use attestations, we'd run into more ambiguities and potentially could exert unnecessary practical restrictions.

For example, if each `input_attestations` entry is one attestation, would that mean every attestation should have its own URI? Would that pose toil for systems that store all attestations in a single place?

Let's say we allow multiple attestations to share a URI, I think this creates another toil -- in order to identify each attestation, they have to be individually digested and put into a separate entries, but who would use this information? The verifier already has its own internal logic to determine what attestation to use; And each attestation is already signed, so re-digest them won't have any security benefit either...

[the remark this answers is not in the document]
AdamZWu: Fixed.

docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,25 @@ of the other top-level fields, such as `subject`, see [Statement]._
 
 > Collection of cryptographic digests for the contents of the policy used to perform verification.
 
+<a id="input_attestations"></a>
+`input_attestations` _object, optional_
+
+> The collection of attestation bundles that are used to perform verification.
+> Conceptually similar to the `materials` field in the [SLSA Provenance](/provenance).
+>
+> A verifier MAY utilize this field to record information for facilitating location of the attestations,
```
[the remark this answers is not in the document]
AdamZWu: Good point. Adopted except the following:

> It SHOULD omit any attestations that were not relevant to making the verification decision.

There may be practical difficulties for attestations that come in bundles. We know that not all attestations in a bundle wre used during a verification. So unless a verifier reconstructs bundles after every verification to only contain attestations used (which probably has no security benefit), this may not be achievable.

docs/verification_summary/v0.2.md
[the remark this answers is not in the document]
AdamZWu: Good point. Done.

AdamZWu: Done.

AdamZWu:
Applies some suggested changes, squashed and rebased.

docs/verification_summary/v0.2.md
TomHennen: I'm fine with leaving it as referring to the bundle (but I'd like @MarkLodato and @joshuagl's thoughts).

If we do leave it as bundle can you be even more explicit about what the digest is of?  E.g. "the contents of the bundle as provided to the verifier" or something like that?  Not sure if that helps.  Perhaps it's clear enough as is?

docs/verification_summary/v0.2.md
TomHennen: Even though the URI could be content addressable that's no guarantee that it is and since we always have the ability to compute the digest of the thing we may as well require it.  That would allow users to verify if the bundle they got from the URI is the same one the verifier used.

Also, by requiring it we make it easier on people that want to use it.  They can always depend on the digest, even if URI isn't always present.

AdamZWu: I think your points are valid.

However, I am a bit concerned whether they may be unnecessarily restrictive -- the attestation bundle specs does not have requirements on the order of the items, but by requiring digest verification on all attestation bundle here, we implicitly added a restriction that is not part of the attestation bundle spec.

TomHennen: Let's chat when you're free. I think we can work this out.

joshuagl: Look forward to reading the outcome of that discussion here.

TomHennen: Perhaps it's more in the spirit of distributed development to try to do it all on GitHub?  We'd certainly document the result of the discussion here.  Also, FYI it will probably be a week before @AdamZWu is free.

docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,34 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+> The collection of attestation bundles that were used to perform verification.
+> Conceptually similar to the `materials` field in the [SLSA Provenance](/provenance).
+>
+> If this field is present and non-empty it must contain information on _all_ the attestations
+> used to perform verification.
+>
+> A present but empty field indicates the decision was made with no attestation.
+>
+> An absent field indicates the verifier does not support this feature,
+> and the decision may or may not be based on any attestations.
```
joshuagl, line 205: What is the expected value/presence of this field for this use-case as described in the purpose section?
> allows software publishers to keep the details of their build pipeline confidential while still communicating that some verification has taken place. 

In that scenario a decision is made with attestations but the attestations are not publicly available to users of the VSA.

docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,34 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+<a id="input_attestations"></a>
+`input_attestations` _object, optional_
+
+> The collection of attestation bundles that were used to perform verification.
+> Conceptually similar to the `materials` field in the [SLSA Provenance](/provenance).
+>
+> If this field is present and non-empty it must contain information on _all_ the attestations
+> used to perform verification.
+>
+> A present but empty field indicates the decision was made with no attestation.
```
joshuagl, line 202: Having three ways to define the `input_attestation` field feels prone to error in verifiers and generators. I wonder whether this option, that the decision was made with no attestation, should be an explicit field?

docs/verification_summary/v0.2.md
joshuagl: It's perhaps unreasonable to expect all discussion to happen here, but having any decision or results documented here would be most useful. Thanks.

docs/verification_summary/v0.2.md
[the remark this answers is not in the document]
AdamZWu: Discussed separately and agreed that we can leave out the "should omit" specification as implicit.

docs/verification_summary/v0.2.md
AdamZWu: Discussed with @TomHennen and agreed that we keep the specification at attestation level, and do not mention bundle -- practically attestations may be stored in a bundles, or in other forms, we don't need to be prescriptive.

The URI is simply associated with the data fetching, so if we have multiple attestations in a bundle, then there could be multiple entries with the same URI, and the digests are used as identifier to tell apart different attestations.

docs/verification_summary/v0.2.md
AdamZWu: Per discussion with @TomHennen, we took option 2.

docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,35 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+<a id="input_attestations"></a>
+`input_attestations` _object, optional_
+
+> The collection of attestations that were used to perform verification.
+> Conceptually similar to the `materials` field in the [SLSA Provenance](/provenance).
+>
+> This field can be absent if the verifier does not support this feature;
+> If present, this field must contain information on _all_ the attestations
+> used to perform verification.
+> (An empty collection indicates the decision was made without attestation.)
```
TomHennen, line 202: nit: ... made without attestations

docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,35 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+
+<a id="input_attestations.uri"></a>
+`input_attestations[*].uri` _string ([ResourceURI]), optional_
+
+> A URI that can be used (provided the requester has read permission) to fetch
+> an attestation.
```
TomHennen, line 208: nit: the attestation

docs/verification_summary/v0.2.md
AdamZWu: I think my writing made it feel more complicated. :P Rephrased the description and PTAL~

docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,35 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+>
+> An absent field indicates the verifier does not know how the attestation was fetched
```
TomHennen, line 210: nit: does not know where to fetch the attestation

TomHennen (changes requested):

docs/verification_summary/v0.2.md
AdamZWu: I feel it is an orthogonal matter.

VSA could always refer to the attestation used if the verifier supports this feature; As to whether those attestations are available to which group of users:

Option 1: The vendor can control attestation access via ACLs on the URI serving side;
Option 2: The vendor can choose to not disclose the URI (which is optional).

docs/verification_summary/v0.2.md
AdamZWu: Done.

docs/verification_summary/v0.2.md
AdamZWu: Done.

docs/verification_summary/v0.2.md
AdamZWu: Done.

TomHennen:
Thanks Adam!  This addresses all my issues.  We should still make sure @joshuagl is happy with this change before merging.

docs/verification_summary/v0.2.md
```diff
@@ -267,7 +305,9 @@ Must be one of these values:
 
 </div>
 
--   0.2-draft: Added `resource_uri` field.
+-   0.2:
+    -   Added `resource_uri` field;
+    -   Added optional `input_attestations` field;
```
joshuagl, line 310: Nit: let's use periods at the end of these lines to better match the change history of the provenance spec.
```suggestion
    -   Added `resource_uri` field.
    -   Added optional `input_attestations` field.
```

docs/verification_summary/v0.2.md
joshuagl: Thanks, that does make it less complicated. My mental model of what use-cases these different scenarios satisfy has also improved, LGTM.

docs/verification_summary/v0.2.md
joshuagl: Understood and I agree. Thanks.

docs/verification_summary/v0.2.md
```diff
@@ -187,6 +190,35 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+<a id="input_attestations"></a>
+`input_attestations` _object, optional_
+
+> The collection of attestations that were used to perform verification.
+> Conceptually similar to the `materials` field in the [SLSA Provenance](/provenance).
+>
+> This field can be absent if the verifier does not support this feature;
+> If present, this field must contain information on _all_ the attestations
+> used to perform verification.
+> (An empty collection indicates the decision was made without attestations.)
```
joshuagl, line 202: Nit: let's not make this a parenthetical?
```suggestion
> An empty collection indicates the decision was made without attestations.
```

joshuagl (approved):
I pointed out a couple of minor nits, but this looks good to me. Thanks for working through the review with me!

AdamZWu:
> I pointed out a couple of minor nits, but this looks good to me. Thanks for working through the review with me!

Thanks, Joshua! Accepted suggestions and squashed commits.

joshuagl:
@TomHennen I think we're waiting for your review on this one.

TomHennen (approved):
Looks good, thanks everyone!

PR #469, opened 2022-08-20
Update possible typo in doc

shafeeshafee:
Was just casually reading and thought this might be a typo. Apologies if it isn't! 🙏

Signed-off-by: shaf <68356865+shafeeshafee@users.noreply.github.com>


mlieberman85:
Thanks for the PR, I think it's correct already. I believe Kim was referring to SLSA resonating with different organizations, not the community resonating with different organizations. Let's just doublecheck. @kimsterv ?

shafeeshafee:
@mlieberman85 gotcha, that makes sense.

Issue #480, opened 2022-09-10
New badge icons are not propagated to the website 


godofredoc:
I landed a PR to add new SLSA badges for all the supported levels in https://github.com/slsa-framework/slsa/pull/479 but they don't seem to have propagated to the website. I was expecting to be able to get the icons from urls like https://slsa.dev/images/gh-badge-level2.svg in a similar fashion as https://slsa.dev/images/gh-badge-level3.svg.

Do I need to copy the badge images to a different place or is just a matter of waiting the new version of the website to be released on a predefined schedule?

godofredoc:
\cc @laurentsimon

inferno-chromium:
The reason it is not working is these are copied to the wrong dir. It needs to be moved under https://github.com/slsa-framework/slsa/blob/main/docs/images [site is deployed from docs/] and not https://github.com/slsa-framework/slsa/tree/main/resources/assets/badges. We should probably remove the all the badges from resources/assets/badges to avoid confusion.

inferno-chromium:
https://slsa.dev/images/gh-badge-level2.svg works now, closing.

PR #481, opened 2022-09-11
Move badge images to docs folder.

godofredoc:
The images are released from docs/images/ instead of resources/assets/badges/. This PR is moving the images to the expected location.

Bug: https://github.com/slsa-framework/slsa/issues/480


inferno-chromium (approved):

inferno-chromium:
Minor resources move of badges to right location, so merging.

PR #486, opened 2022-09-20
Create spec v1.0 as a copy of v0.1; add "draft" warning banner; fix version picker on VSA

MarkLodato:
This PR creates a new spec/v1.0 directory so that we can start making changes for v1.0.

Preview: https://deploy-preview-486--slsa.netlify.app/spec/v1.0

The new version of the spec:

*   Is hidden by default so that users do not stumble upon it. You need to manually type the URL to find it.

*   Has a warning banner at the bottom of the page to indicate that it is an unpublished draft.

*   Is copied verbatim from v0.1 for now, which you can verify with `git ls-tree`:

    ```
    $ git ls-tree HEAD docs/spec/v{0.1,1.0}
    040000 tree 2c3924e63122e1a557abb71d15255189d6dd6c64    docs/spec/v0.1
    040000 tree 2c3924e63122e1a557abb71d15255189d6dd6c64    docs/spec/v1.0
    ```

Other bundled changes:

*   Unify the version pickers into a single template. This adds the version picker to the VSA page and supports draft/hidden for any attestation spec (not just the core levels spec).

To review this PR:

*   I suggest viewing the individual commits, which have more detail and are easier to review.
*   Please see whether the banner is sufficiently visible.
*   Please check whether other pages work, e.g. FAQ (not versioned, so no banner), index, sub-pages.


MarkLodato:
Note to reviewers: I just pushed an additional commit (4cfd689) that is beyond the scope of the original PR but the implementation is related, so I thought it might be easier for you to review as a batch. If you want me to undo, just let me know.

joshuagl (commented):
Some really nice cleanups in here, especially unifying the version pickers, thanks Mark! 

I worry that the banner is easy to miss, especially given that it sits in the same space as most cookie banners. I wonder if it would be more apparent at the top of the page, as folks tend to read from the top?

MarkLodato:
> I worry that the banner is easy to miss, especially given that it sits in the same space as most cookie banners. I wonder if it would be more apparent at the top of the page, as folks tend to read from the top?

Agreed. I just moved it to the top. Please take another look.

joshuagl (approved):

PR #489, opened 2022-09-26
Blog post for EO and the recent OMB memo

hepwori:
Adds a blog post with an educational plain-english summary of the implications of the recent OMB memo following the Executive Order on Improving the Nation’s Cybersecurity. 

Signed-off-by: Isaac Hepworth <isaac.hepworth@gmail.com>

PR #490, opened 2022-09-27
Fix rendering of numbered lists

MarkLodato:
Before (from #489):

before image

After:

after image


joshuagl (approved):
I am not a CSS person, but the results look good and the CSS is understandable. LGTM.

PR #492, opened 2022-09-28
Disable smooth scrolling

MarkLodato:
I personally find this smooth scrolling very distracting, especially during development.

Things to test:
- Try clicking a link in the table of contents. Previously it would animate the scroll. Now it jumps straight there.
- Go to a particular section of the page then reload. Previously it would jump to the top then animate the scroll, which was distracting. Now it just reloads quickly.


joshuagl (approved):
Given that it's quite common to link to parts of the spec (i.e., a specific requirement) this change makes sense to me.

Issue #187, opened 2021-10-19
KMS requirement for SLSA 3/4 is too prescriptive


mattmoor:
This is the wording in particular that I'm looking at:
> The provenance signing key MUST be stored in a secure key management system accessible only to the build service account.

In particular, this precludes models that take advantage of ephemeral keys and CAs (e.g. Sigstore's Fulcio), which are not KMS systems.  However, I'd argue that despite something like Fulcio *not* satisfying this as-written, it is a strictly stronger model because long-lived key material is never exposed to the build service.

As a sort of meta-comment, SLSA doesn't really touch on the exclusive use of [very] short-lived credentials, despite the fact that exfiltrating long-lived creds is a key attack vector.  I was mentioning to @kimsterv that the use of exclusively short-lived credentials (e.g. Workload Identity) and keys (e.g. Fulcio) might make a nice basis for SLSA 5.

dlorenc:
I think the language around all of the signing stuff needs to be improved a bit. Key management is only one part of the puzzle, without rotation/revocation you're really not left in a much better place than not signing things at all.

trishankatdatadog:
Mark: I think a bit of both: generalize the description to include both use cases, and also list some concrete options.

kimsterv:
@trishankatdatadog flagging that this issue ties into the [doc](https://docs.google.com/document/d/1z_9C13hvGFcBx3l7ehvXmiCcU5Ywf8KROPDia7Nhymw/edit?usp=sharing) discussed today in the wg

awwad:
Current text:
> This SHOULD be through a digital signature from a private key accessible only to the service generating the provenance.

' nice to know this has been highlighted before.  ^_^

Preston and I happened to stumble over this requirement and brought it up this week because the existing text is vague enough to reasonably be seen _not_ to fit very _good_ implementations and also to reasonably be seen to fit very _bad_ implementations.

Access to a private key can be seen as meaning visibility of the raw key value (exfiltration risk) or ability to use the key during compromise (use risk).  It's easy to think of this as a requirement for reducing the odds of key exfiltration, but I think it would be best to make this about key access in the broad sense of ability to use the key.

shaunmlowry:
related https://github.com/slsa-framework/slsa/issues/414

PR #494, opened 2022-09-29
Use vs Exfiltration: Provenance - Authenticated

pkmooreanaconda:
First swing at adding some text that separates illegitimate use of authentication-related material from exfiltration of said material.  Sebastien and I have been bouncing this back and forth a bit and wanted to get something out to drive discussion.

I think one thing that makes this text difficult is accounting for different mechanisms use different means to protect the provenance.  Were we just talking about digital signatures, it would be straightforward to say something that boils down to "don't let people steal or misuse the signing materials."  Keeping things generic means inventing a term like "authentication-related materials."  

This is related to issue #187

### DIFF
--- a/docs/spec/v0.1/requirements.md
+++ b/docs/spec/v0.1/requirements.md
@@ -366,8 +366,15 @@ all the other requirements.
 <td>
 
 The provenance's authenticity and integrity can be verified by the consumer.
-This SHOULD be through a digital signature from a private key accessible only to
-the service generating the provenance.
+This SHOULD be through a digital signature from a private key that may only
+be accessed and used by the build service.  
+
+Regardless of mechanism, there MUST be controls in place to prevent exfiltration of
+private authentication-related material and illegitimate use of said material through
+intended channels.  It must not be possible to read or alter the private authentication materials
+(Other than for maintenance operations, such as key rotation)
+nor should it be possible to use the mechanism to authenticate artifacts outside of the build process.
+
 
 <td> <td>✓<td>✓<td>✓
 <tr id="service-generated">


mlieberman85:
Can you clarify the goal a bit?

awwad:
I think there are two different conversations here.  Preston and I are primarily concerned with the signing scenario.  Existing text was written in such a way as to permit other means of data authentication.  (I suppose that that would include things like published hashes of provenance information, or transport authentication like TLS 🫤.)  Some of the ambiguous language in this PR is there to try not to interfere with this prior flexibility.

Our actual goal, though, is to clarity what seemed to us to be unclear when we were deciding on whether or not what we were up to met this criterion.  The existing text seems like it could **_include_** a system with completely unreachable keys but wide open access to make signing requests to those well-hidden keys (i.e. just signs whatever you throw at it) and **_exclude_** a system with reasonable key rotation mechanisms but better use-related protections.

awwad:
I think there are two different conversations here. Preston and I are primarily concerned with the signing scenario. Existing text was written in such a way as to permit other means of data authentication. (I suppose that that would include things like published hashes of provenance information, or transport authentication like TLS 🫤).  Some of the ambiguous language in this PR is there to try not to interfere with this prior flexibility.

Our actual goal, though, is to clarify what seemed to us to be unclear when we were deciding on whether we meet this criterion.  The existing text seems like it could include a system with completely unreachable keys but wide open access to make signing requests to those well-hidden keys (i.e. just signs whatever anyone throws at it) and exclude a system with reasonable key rotation mechanisms but better use-related protections.

arewm:
I feel like a lot of the comments here have already been addressed in the v1 specification as it currently is (i.e. provenance authenticity and accuracy with build system isolation). I think that questions/clarification around "best effort provenance completeness" and what should be included for the threat model at a build L3 are related to this and might be able to arrive at similarly interesting details and refinement.

MarkLodato:
I'll close it for now to reduce clutter on the "open pull requests" page. Anyone feel free to pick this up and either reopen or create a new PR when ready for review.

PR #495, opened 2022-09-30
Add label to guest blog posts

lehors:
This change adds the label "Guest Post" to the author line of guest blog posts, both to the post itself and the blog page.

### DIFF
--- a/docs/_layouts/post.html
+++ b/docs/_layouts/post.html
@@ -3,7 +3,7 @@ layout: post-list
 is_post: true
 ---
 
-<p class="h3 post-author">{% if page.is_guest_post -%}A Guest Post{%- endif %} by {{ page.author }}</p>
+<p class="h3 post-author">{% if page.is_guest_post -%}Guest post{%- endif %} by {{ page.author }}</p>
 
 <p class="post-date"><time datetime="{{ page.date | date: "%Y-%m-%d" }}">{{ page.date | date_to_string }}</time></p>
 
--- a/docs/_layouts/post.html
+++ b/docs/_layouts/post.html
@@ -2,7 +2,9 @@
 layout: post-list
 is_post: true
 ---
-<p class="h3 post-author">by {{ page.author }}</p>
+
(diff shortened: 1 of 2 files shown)


abacchi:
I think this is fair to include. The question I have is: when is a post not a guest post? I think we have talked about that before but it would be good to have that clearly defined.

lehors:
> I think this is fair to include. The question I have is: when is a post not a guest post? I think we have talked about that before but it would be good to have that clearly defined.

I think this has been done:

1) The governance docs introduce the two types of posts: See sections 1.1.1.4. Blog Posts and 4.6 Blog Post Specific in https://github.com/slsa-framework/governance/blob/main/6._Contributing.md

2) The blog posts source files do carry a boolean flag "is_guest_post" 

3) issue #309 also provides some additional info.

melba-lopez:
Thanks @lehors; yes I would prefer this to be more visible, but we also need to make sure we have due diligence to make OFFICIAL posts not marked as guest. _(There was one marked as guest post recently that was a community post)_

melba-lopez (approved):
Approve changes.  Helps users to better see when its a guest.

docs/_layouts/post.html
```diff
@@ -2,7 +2,9 @@
 layout: post-list
 is_post: true
 ---
-<p class="h3 post-author">by {{ page.author }}</p>
+
+<p class="h3 post-author">{% if page.is_guest_post -%}A Guest Post{%- endif %} by {{ page.author }}</p>
```
MarkLodato, line 6: suggestion: "Guest post" (instead of "A Guest Post") would make it easier to read at a glance. (The "A" and capital "P" make it harder to visually parse.) Same on other file.

Compare:

image

image

MarkLodato (approved):
Thanks! Looks good, though I have one tiny suggestion.

MarkLodato (approved):

abacchi (approved):

inferno-chromium:
> Oops, I wish GitHub reminded us about pull requests that are ready to merge. Merging now!

I think you can set Auto-merge on simple ones ?

Issue #337, opened 2022-03-29
Less reductive SLSA leveling 


anvega:
I've been trying to rationalize SLSA levels but struggle when I see three discrete separate domains of provenance, integrity, compliance conflated as one that lumps together criteria across the three for the different SLSA levels in a single axis, rather than aggregating and computing across three different axes for the three different areas of guarantees. 

Borrowing from the docs, here is where I see this reductiveness become problematic. I don't see how this makes sense except as theater:

> The SLSA level is not transitive ([see our FAQs](https://slsa.dev/spec/faq)). This makes each artifact’s SLSA rating independent from one another, allowing parallel progress and prioritization based on risk. The level describes the integrity protections of an artifact’s build process and top-level source, but nothing about the artifact’s dependencies. Dependencies have their own SLSA ratings, and it is possible for a SLSA 4 artifact to be built from SLSA 0 dependencies.

I understand why requiring levels to be transitive makes it all a much harder problem to solve,  It's almost as to say: 'we know it’s all mystery meat but hey at least we double-checked and signed off on it' 

One approach here would be to introduce interim levels such as 3.5 and a level 5, retaining the same guarantees but adding more levels to the progression, for instance, level 4 would require two-person reviews as a guarantee of compliance while getting to level 5 would require attaining hermetic builds as the ultimate integrity guarantee. 

Now, a much better solution would be to separate compliance from provenance and integrity altogether and introduce a split stack of levels for compliance while enforcing that the supply chain levels need to be transitive.

dlorenc:
> It's almost as to say: 'we know it’s all mystery meat but hey at least we double-checked and signed off on it'

To put this a bit more charitably, I think of it as, "I did my part! Yes, we're still all connected in the supply chain, but I did all I could to make my portion safe.* :)

mlieberman85:
It's probably worthwhile a writeup on the transitive problem.

There are still plans for a "transitive" SLSA level but where does it end? What is that bottom turtle, do we go back to the kernel and all the compilers? What are individual projects and organizations comfortable with?

To @dlorenc 's point SLSA's initial releases are in that piece of getting individual projects to start attesting that they did their part.

Nothing will get done if no project can start generating some metadata without attesting to all their upstream dependencies. This would be much easier if everyone from OS on down were already generating SLSA attestations, but we're not there yet.

Separating compliance out is a good point.

shaunmlowry:
I think there's 2 potential transitive axes here:

[...]

PR #502, opened 2022-10-11
SLSA v1.0: rewrite levels.md to split Build vs Source, undefine L4, and better explain

### LINKED ISSUE #111: Clarify the goals / use-cases for each level

### LINKED ISSUE #268: Document intention at each SLSA level
The original SLSA authors at Google (cc @MarkLodato @TomHennen @msuozzo) often refer to the original intentions of the SLSA requirements at each level, i.e. from https://github.com/slsa-framework/slsa/issues/261#issue-1093774677

> The original intention was as follows:
> 
> L1 can be tacked on to any existing process with no changes to the build service. Provenance can be trivially forged.
> L2 requires relatively inexpensive changes to the build service. Provenance should not be trivially exploitable.
> Open question: Should it still be strong against key theft?
> L3 may require significant changes to the build service. Provenance must not be exploitable even for a determined adversary.

This finger in the air metric for the implementation effort required and protection offered at each level is useful information for contributors and adopters.

Can we capture this explicitly, either in the public-facing documentation or as part of the contributor documentation?

MarkLodato:
**Preview:** https://deploy-preview-502--slsa.netlify.app/spec/v1.0/levels

Completely rewrite the levels.md file for v1.0 as per the [proposal](https://github.com/slsa-framework/slsa-proposals/blob/main/0003/README.md):

- Split Build vs Source levels, now called "tracks"
- Undefined Build L4 and the Source levels
- Better explain the summary, purpose, and requirements of each level

Otherwise no changes to the meaning of each level.

There are still some TODOs, and equivalent changes are also necessary for the rest of the spec. This PR is intended to be an incremental improvement.

Fixes #111.
Fixes #268.
Partially addresses #337.

#### Note to reviewers

See individual commits for more detailed explanations and an easier-to-review diff.

I'm interested if these changes are indeed an improvement. If you have suggestions, please let me know!


docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,286 @@
...
+<dl class="as-table">
+<dt>Summary<dd>
+
+No requirements---L0 represents the lack of SLSA.
+
+<dt>Intended for<dd>
+
+Software that does not need SLSA, such as:
+
+-   Binaries built and executed on the same machine, including continuous
```
mlieberman85, line 95: I would probably change this to "Software" generally as it could be something other than Binaries. I would also maybe change it from continuous integration tests to local dev builds or something.

Separately I'm not sure we even want to say it's intended for anything. I just want folks to understand that pretty much any software used internally or externally should have SLSA attestations outside of: "I am building software inside a dev workspace/workstation"

docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,286 @@
...
+-   On each build, the release process automatically generates and distributes a
+    [provenance attestation] describing how the artifact was *actually* built,
+    including: who built the package (person or system), what process/command
+    was used, and what the input artifacts were.
+
+<dt>Benefits<dd>
+
+For consumers:
+
+-   Allows the consumer to known what the exact source code was for a given
```
mlieberman85, line 139: nit: typo - known -> know

docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,286 @@
...
+    artifact so that they can audit it or make decisions based on it.
+-   Makes it easier for consumers to build the software themselves, by knowing
+    what the official process and sources were.
+
+For large organizations and enterprises:
```
mlieberman85, line 144: This read a bit weird. Above you have "for consumers" and then for here you have "large organizations and enterprises" -- it's hard to know if it's expected that enterprises are also consumers in this context.

docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,286 @@
...
+<dt>Benefits<dd>
+
+For package maintainers:
+
+-   Prevents tampering of the artifact after the build.
+-   In some cases, reduces impact of compromised package upload credentials.
+
+For consumers:
+
+-   Strongly identifies what build service was used.
```
mlieberman85, line 188: We want to make it clearer that it's not just "build service" but some identity associated with the project, maintainer, org, etc. running the build service. In most cases with SLSA provenance at least we're seeing stuff signed not just indicating that it was built by something like "github actions" but also indicating that it is signed with info based on the project, maintainer, etc.

docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,286 @@
...
+-   Build service implements strong controls to prevent runs from influencing
+    one another, even within the same project.
+
+> **TODO:** Add requirement about survey and audit as per the v1.0 proposal.
+
+<dt>Benefits<dd>
+
+For package maintainers:
+
+-   Prevents tampering of the artifact during or after the build.
```
mlieberman85, line 223: assuming the build service itself hasn't been compromised.

mlieberman85: I think the intent we want to get across is: Prevents non-build service tampering of the artifact build or the artifact being tampered with after the build

...Just in a nicer way.

MarkLodato (commented):
Thanks for the feedback! In addition to the other comments, I updated the text to hopefully address #268.

trishankatdatadog (commented):
Interesting. This makes a lot of sense to me.

Is the thinking to make it clear that, for right now at least, SLSA is focused on hardening build, and not source, dependencies, and packages?

MarkLodato:
> Is the thinking to make it clear that, for right now at least, SLSA is focused on hardening build, and not source, dependencies, and packages?

Yes! More rationale in the [proposal](https://github.com/slsa-framework/slsa-proposals/blob/main/0003/README.md#separate-build-levels-vs-source-levels).

trishankatdatadog:
> Yes! More rationale in the [proposal](https://github.com/slsa-framework/slsa-proposals/blob/main/0003/README.md#separate-build-levels-vs-source-levels).

I like this proposal. All in all, it has more pros than cons: it makes sense to separate out fixing each of the 4 tracks (build, source, dependencies, packages).

docs/spec/v1.0/levels.md
```diff
@@ -38,6 +38,24 @@ an artifact is "safe".
 > own SLSA determinations, though in practice they may delegate to some
 > authority.
 
+## Who is SLSA for?
+
+There are three main audiences for SLSA:
```
joshuagl, line 43: A small but important audience are the folks creating build services, should we be calling them out here?

joshuagl (commented):
I've left some minor comments and suggestions in-line, but this is already a vast improvement.

docs/_sass/minima.scss
```diff
@@ -21,6 +21,7 @@ $light-green: #58DA95;
 $green-darkest: #102B2D;
```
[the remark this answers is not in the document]
MarkLodato: I'll send a separate PR to fix :-)

docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,293 @@
...
+the specification.
 
-The SLSA level is not transitive ([see our FAQs](../faq.md)). This makes each artifact’s SLSA rating independent from one another, allowing parallel progress and prioritization based on risk. The level describes the integrity protections of an artifact’s build process and top-level source, but nothing about the artifact’s dependencies. Dependencies have their own SLSA ratings, and it is possible for a SLSA 4 artifact to be built from SLSA 0 dependencies.
+For reference, the [previous version] of SLSA included the following
+requirements, which *may or may not* be part of a future Source track:
 
-## Limitations
+-   Strong authentication of author and reviewer identities to resist account
+    and credential compromise, such as 2-factor authentication using a hardware
+    security key.
```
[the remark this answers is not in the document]
MarkLodato: Done.

docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,293 @@
...
+All of [Build L1], plus:
+
+-   Builds run on a hosted build service that generates and signs the
+    provenance itself, outside the control of the users of the build service.
+
+<dt>Benefits<dd>
+
+All of [Build L1], plus:
+
+-   Prevents tampering after the build through digital signatures.
```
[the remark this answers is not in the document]
joshuagl: I agree that this is tricky and I haven't come up with a good, concise, alternative. I'm OK leaving this as a TODO and figuring it out in a follow-on.

docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,301 @@
...
+    is authentic and has not been tampered with. They are concerned about rogue
+    maintainers, compromised credentials, and compromised infrastructure.
+
+-   **Organizations,** who are both producers and consumers of software. In
+    addition to the goals above, organizations also want to broadly understand
+    and reduce risk across the organization.
+
+-   **Infrastructure providers,** such as build services and packaging
+    ecosystems, who are critical to achieving SLSA. While not the primary
+    beneficiary of SLSA, they may benefit from increased security and customer
```
joshuagl, line 62: Nit: can we say users, not customers, here?

joshuagl (approved):
LGTM!

docs/spec/v1.0/levels.md
MarkLodato: Done.

MarkLodato:
Thanks for the review! I'll leave this open for other reviewers and then merge on Monday if I don't hear anything else.

docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,301 @@
...
+    and reduce risk across the organization.
+
+-   **Infrastructure providers,** such as build services and packaging
+    ecosystems, who are critical to achieving SLSA. While not the primary
+    beneficiary of SLSA, they may benefit from increased security and user
+    trust.
+
+## Levels and tracks
+
+SLSA levels are split into *tracks*. Each track has own set of levels that
```
abacchi, line 67: small nit- "Each track has **it's** own set of levels..."

docs/spec/v1.0/levels.md
```diff
@@ -1,60 +1,301 @@
...
+Projects and organizations wanting to gain moderate security benefits of SLSA by
+switching to a hosted build service, while waiting for changes to the build
+service itself required by [Build L3].
+
+<dt>Requirements<dd>
+
+All of [Build L1], plus:
+
+-   Builds run on a hosted build service that generates and signs the
+    provenance itself, outside the control of the users of the build service.
```
abacchi, line 193: I may have missed a discussion inside the spec meetings- however the existing level 2 only requires the build service to generate the provenance, not sign it. Its nice to have the build service sign it but that's a hard requirement in level 3. You could still have signed provenance by the maintainer but not the build service. I saw it as a nice step between level 2 and 3- where 3 was much more solid because it required that.

docs/spec/v1.0/levels.md
abacchi: I like the re-edit referencing unit tests locally. I think its worth leaving in there now to address the type of FAQ.

docs/spec/v1.0/levels.md
MarkLodato: No change was intended from v0.1. The existing L2 [Authenticated](https://slsa.dev/spec/v0.1/requirements#authenticated) requirement says:

> The provenance’s authenticity and integrity can be verified by the consumer. This SHOULD be through a digital signature from a private key accessible only to the service generating the provenance.

So I think that precludes having a maintainer sign it because then the maintainer's machine and account would be in the trust base, which we want to reduce at L2.

But I agree that the wording is too prescriptive and narrow. I'm thinking we want something like:

L1 - no authenticity requirement
L2 - weak guarantee that the provenance came from a service and not a person, but exploitation may be easy
L3 - strong guarantee that the provenance came from a service, and exploitation is difficult

trishankatdatadog (approved):

MarkLodato:
In the interest in moving the v1.0 process forward, since I have two LGTMs and no outstanding requests I'm going to submit this now. Please feel free to add more comments and I can address in a future PR.

docs/spec/v1.0/levels.md
abacchi: I suppose I should have been more specific- I didn't mean that the maintainer would be signing provenance from their workstation, but rather a job or step in the build process - still not great because the maintainer has the ability to meddle with the provenance (comparted to L3 where the service holds the signing material) - I'll take a stab at the wording you outlined above

MarkLodato: Thanks! For anyone reading this in the future, this is #509.

Issue #371, opened 2022-04-16
Are SLSA levels self-assessed ?


sbs2001:
SLSA noob here, so maybe it's a stupid question.

From reading the docs I didn't see a tool which could help with determining the SLSA level of some project. So I want to know whether folks just compare their project state with SLSA spec and determine their SLSA level.

tmsteen:
I am not sure there is a tool that would provide enough coverage to provide an automated assessment in a broad sense. That being said, the requirements are much more detailed that you see from frameworks like the NIST CSF of CIS Controls. 

Another option is to have an external entity attest to the alignment against SLSA requirements is self-assessment is not rigorous enough.

nris-jingu:
You may see the compliance matrix by DevOps tool vender. It seem to help self-assesment.

https://about.gitlab.com/solutions/supply-chain/

I think that it's difficult to assess automatically.

mlieberman85:
Sorry, this issue appears to have fallen through the cracks there. So currently SLSA is self assessed, however there is nothing to stop a third party audit firm  or tooling to do the assessment for you. It's up to the consumer though to validate those certification. e.g. if audit person X says that your automated assessment done via tool Y is suitable for SLSA 3 builder certification then the end user would validate those identities and the attestations being made.

@krol3 Chain-bench is great and definitely one of the tools we've been using to hit some of the CIS stuff for another OpenSSF tool Frsca which should be SLSA compliant and also an implementation of the CNCF's secure software factory ref arch. It would be cool though to also include SLSA requirements in the chain-bench benchmarking tool as well.

krol3:
@mlieberman85 I opened an issue, please add your comments about this topic. https://github.com/aquasecurity/chain-bench/issues/63

PR #503, opened 2022-10-11
SLSA v1.0: shift emphasis to expectations + verification in levels.md

MarkLodato:
**Preview:** https://deploy-preview-503--slsa.netlify.app/spec/v1.0/levels

(As discussed in #130 and the Specification SIG meeting)

Previously the specification only required the publication of provenance but did not say anything about its verification. The latter is what actually detects or prevents attacks, so this was a big gap. Futhermore, the previous "scripted build" requirement did not have a clear reason why it was included.

Now there is explicit language around:

- Defining an expectation of how the package should be built, replacing the previous "scripted build" requirement.
- Verifying that the provenance meets expectations.

NOTE: This PR only changes levels.md. A future commit will make an equivalent change to the rest of the spec, e.g. requirements.md.

Related issues (not fully fixed): #46, #130, #371


docs/spec/v1.0/levels.md
```diff
@@ -49,7 +49,7 @@ tracks without invaliding previous levels.
 | Track/Level | Requirements | Benefits |
 | ----------- | ------------ | -------- |
 | [Build L0]  | (none)       | (n/a)    |
-| [Build L1]  | Attestation showing how the package was built | Documentation, inventorying |
+| [Build L1]  | Attestation showing that the package was built as expected/documented | Documentation, mistake prevention, inventorying |
```
joshuagl, line 52: Can we just say "as expected", given we use that term in the track summary and the L1 requirements?

docs/spec/v1.0/levels.md
```diff
@@ -114,8 +118,8 @@ n/a
 <dl class="as-table">
 <dt>Summary<dd>
 
-Package has a provenance attestation showing how it was built, though without
-tamper protection.
+Package has a provenance attestation showing that it was built as
+expected/documented, though without tamper protection.
```
joshuagl, line 122: same here, can we just use "as expected"?

joshuagl (commented):
Great to see this being fleshed out. I made suggestions inline to consistently use "as expected" rather than "as documented/expected".

MarkLodato:
Thanks. I updated the language as suggested and rebased onto the latest version of #502.

joshuagl (approved):
LGTM, thanks. Do you think we should provide guidance on how to verify provenance meets expectations (and is signed) for users of the SLSA provenance format?

MarkLodato:
> LGTM, thanks. Do you think we should provide guidance on how to verify provenance meets expectations (and is signed) for users of the SLSA provenance format?

Yes. I assume that would be a necessary part of #46. This PR here is only a first step.

mlieberman85 (approved):
LGTM

abacchi (approved):

PR #506, opened 2022-10-14
Layer "draft" banner under the nav menu

MarkLodato:
Previously the "draft" banner was on top of the nav menu, which made it impossible to navigate unless you dismissed the banner. We don't actually need a z-index on the banner since it's defined earlier in the document than the rest of the page, but the nav menu needs a z-index to make it appear above the banner.

**Preview:** https://deploy-preview-506--slsa.netlify.app/spec/v1.0/levels

#### Desktop - Before

before - desktop

#### Desktop - After

after - desktop

#### Mobile - Before

before - mobile

#### Mobile - After

after - mobile

### DIFF
--- a/docs/_sass/minima/_custom.scss
+++ b/docs/_sass/minima/_custom.scss
@@ -499,7 +499,6 @@ details {
   right: 0;
   pointer-events: none;
   margin: 80px 10px 10px;
-  z-index: 800;
 
   > input[type=checkbox]:checked + div {
     display: none;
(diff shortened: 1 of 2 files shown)


abacchi:
I tested and it looks good! Thank you. (I'd approve if I had the permissions to do so)

abacchi (approved):
Looks good- Approved

mlieberman85 (approved):
LGTM.

PR #507, opened 2022-10-17
Use consistent color names in .scss source files

MarkLodato:
As noted by @joshuagl, the old code did not name color variables consistently. This PR fixes that.


joshuagl (approved):
Thanks for doing this!

PR #516, opened 2022-10-20
SLSA v1.0: delete most of index.md

MarkLodato:
The previous content on spec/\*/index.md was redundant with other pages and fell into an awkward middle ground: it was slightly more detailed than the top-level splash page but much less detailed than the individual specification pages. Furthermore, the main thing that people wanted on this page was the table of contents, but that was all the way at the bottom.

Instead of keeping it up-to-date for v1.0, let's drop it entirely. Now the page contains only a table of contents plus a short blurb informing readers about versioning of the spec and the presence of other specifications, which is a common point of confusion.

Note: The `<span>` tags are to used to direct the reader to the version chooser: bottom of page on mobile and right on desktop. You can test this by resizing your window or using the Chrome developer tools.

### DIFF
--- a/docs/spec/v1.0/index.md
+++ b/docs/spec/v1.0/index.md
@@ -1,202 +1,25 @@
----
-title: Introduction
-layout: standard
-hero_text: SLSA is a set of standards and technical controls you can adopt to improve artifact integrity, and build towards completely resilient systems. It’s not a single tool, but a step-by-step outline to prevent artifacts being tampered with and tampered artifacts from being used, and at the higher levels, hardening up the platforms that make up a supply chain. These requirements are explained below, along with the rest of the essential specifications.
-stages:
-    - 1:
-        title: Build integrity
-        text: SLSA starts at the build, the last step before an artifact’s released. This makes sure software’s built from the correct sources and dependencies, and hasn’t been modified. More resilient build integrity means protection from modifying code after source control, compromised build platforms or bypassing CI/CD.
-    - 2:
-        title: Source integrity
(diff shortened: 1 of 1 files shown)


melba-lopez (commented):
I checked the links to make sure they all work within the page and other external pages.  I noticed some go to a v0.2 link.  Will that always happen?  Would someone get confused if they started on a 1.0 page and somehow wound up on a v0.2 page?

MarkLodato:
> I checked the links to make sure they all work within the page and other external pages. I noticed some go to a v0.2 link. Will that always happen? Would someone get confused if they started on a 1.0 page and somehow wound up on a v0.2 page?

There are two types of links:

- Navigation links at the top of the page. They all point to v0.2 because that is the latest stable version, but I agree that this is super confusing. Google is hiring a vendor to fix this.
- Links within the page:
  - All links outside `/spec/<version>` should point to the latest stable version, v0.2
  - All links within `/spec/<version>` should point to that version (unless explicitly referring to another version).

If you find something that doesn't agree with the description above, it was unintentional.

mlieberman85 (approved):
LGTM, on a related note is there an easy way to consolidate all this spec info across the multiple pages for the purposes of exporting like a SLSA 1.0 spec PDF?

MarkLodato:
> LGTM, on a related note is there an easy way to consolidate all this spec info across the multiple pages for the purposes of exporting like a SLSA 1.0 spec PDF?

Not that I'm aware of with our current tooling, but it might not be too hard to script together.

joshuagl (approved):

PR #522, opened 2022-10-31
Change link in getting started

ianlewis:
The [`github-actions-demo`](https://github.com/slsa-framework/github-actions-demo) is old and unmaintained. This PR updates the link to point to the newer and better maintained [`slsa-github-generator`](https://github.com/slsa-framework/slsa-github-generator) repository.

Signed-off-by: Ian Lewis <ianmlewis@gmail.com>

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -3,6 +3,7 @@ title: Get started
 layout: standard
 hero_text: If you’re looking to jump straight in and try SLSA, here’s a quick start guide for the steps to take to reach the first SLSA level. Level 1 ensures that you’re setting up the foundation of trust in a system and that all your applications are generating appropriate provenance data. It also sets a baseline to achieve higher SLSA compliance later, which we explain in detail below.
 ---
+
 <section class="section bg-pastel-green flex justify-center items-center">
     <div class="wrapper inner w-full">
         <div class="md:flex justify-between items-start mb-16">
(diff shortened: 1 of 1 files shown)


diogoteles08:
The deprecated link is also being referred on the [SLSA website](https://slsa.dev/get-started). Can you also change there?

Also, this repo will still be visible and findable at GitHub, and possibly also being referred in some other documentations. So I'd suggest to also change the repo's readme to say that it's deprecated and point to the new repo that should be used. This would ensure people won't think it's the official SLSA generator Repo.

ianlewis:
> The deprecated link is also being referred on the [SLSA website](https://slsa.dev/get-started). Can you also change there?

IIUC, this repo is the source for the website. So that's exactly what I'm doing :) Though I noticed that we already had a link to slsa-github-generator in the SLSA 3 section so I opted to just remove the link from the SLSA 1 section.

The preview for this PR can be viewed here:
https://deploy-preview-522--slsa.netlify.app/get-started

> Also, this repo will still be visible and findable at GitHub, and possibly also being referred in some other documentations. So I'd suggest to also change the repo's readme to say that it's deprecated and point to the new repo that should be used. This would ensure people won't think it's the official SLSA generator Repo.

Thanks, I went ahead and created https://github.com/slsa-framework/github-actions-demo/pull/33 to add a message to the README on that repo and propose archiving it.

mlieberman85 (approved):
LGTM

diogoteles08:
Though I understand your point of redundancy, I think at the SLSA 1 section we should make clear that the [provenance-only generators](https://github.com/slsa-framework/slsa-github-generator#provenance-only-generators) are also applicable.
My concern on this solution is the case of SLSA starters looking for easy implementations, which most probably would mean SLSA 1. I guess the GitHub generators are a solution applicable for the context of a big group of repos, and some of maintainers might give up SLSA adoption before looking up to SLSA 3 to see this tool. 

To differentiate from the link on SLSA 3 section, we could also have linked only the provenance-only generators, instead of the whole builders+generator repo

ianlewis:
> Though I understand your point of redundancy, I think at the SLSA 1 section we should make clear that the [provenance-only generators](https://github.com/slsa-framework/slsa-github-generator#provenance-only-generators) are also applicable. My concern on this solution is the case of SLSA starters looking for easy implementations, which most probably would mean SLSA 1. I guess the GitHub generators are a solution applicable for the context of a big group of repos, and some of maintainers might give up SLSA adoption before looking up to SLSA 3 to see this tool.
> 
> To differentiate from the link on SLSA 3 section, we could also have linked only the provenance-only generators, instead of the whole builders+generator repo

@diogoteles08 That sounds reasonable but wasn't obvious to me at the time given the page structure. I think the overall page structure could use a rework but adding the link to the SLSA 1 section as well would be a good stopgap. Do you want to create the PR?

diogoteles08:
Sure, I can do that

PR #524, opened 2022-11-01
cleanup: remove divs around headings

MarkLodato:
Previously `layout: standard` had no margins around headings, so we had to add a `<div>` with margins around every single heading. This was error prone and distracting. Now the margins are in the style sheet so no extra `<div>` is needed.

This layout is used by two completely different types of pages:

- Provenance and VSA specs: these now have `<div>` tags removed so they are pure Markdown.

- Community, get-started, and use-cases: these have a bunch of Tailwind HTML and really ought to use a different layout. The change in default style added undesired extra margins. To fix, we now have negative margins to compensate. It's ugly, but the code was already ugly. It seems to be no worse than before.

There should be no visual differences except as follows:

- use-cases: adds space after `<h3>` elements, which was missing before
- get-started: unintended 1-pixel difference for "Leave a GitHub issue" on desktop, which is baffling but not worth fixing

### Note to reviewers

To ensure that there are no other visual differences, use diffsite on both mobile (iPhone) and desktop (iPad):

https://pianomister.github.io/diffsite

**TODO:** I'll update the above link with clickable links for each page once the preview is ready.


TomHennen (approved):
Looks good!

mlieberman85 (approved):
LGTM

Issue #508, opened 2022-10-17
v1.0: Define how systems are verified (manually)


MarkLodato:
For v1.0, we need to define how systems (particularly build systems) are verified to meet the SLSA requirements. These are the requirements that cannot be automated because they depends on the design of the system.

Sub-issues:

* [ ] #515 
* [x] #478 
* [x] #187

Context:

* Parent issue: #130 
* Proposal: [evidence of security claims](https://github.com/slsa-framework/slsa-proposals/blob/main/0003/README.md)

marcelamelara:
Taking a stab at this. It seems that we're really trying to answer two questions here: 
(1) How can each build requirement be assessed?
(2) Which, if any, evidence can we obtain to demonstrate a given requirement is met?

Here are all the build requirements, as of right now:
* Scripted build (All build steps were fully defined in some sort of “build script”. The only manual command, if any, was to invoke the build script.)
* Build service (All build steps ran using some build service, not on a developer’s workstation.)
* Build as code (The build definition and configuration executed by the build service is verifiably derived from text file definitions stored in a version control system.) -- this is essentially scripted build + build service
* Ephemeral environment (The build service ensured that the build steps ran in an ephemeral environment, such as a container or VM, provisioned solely for this build, and not reused from a prior build.)
* Isolated (The build service ensured that the build steps ran in an isolated environment free of influence from other build instances, whether prior or concurrent.)

Answering the questions above, looking at the requirements in order of increasing difficulty to assess automatically:
* For "scripted build" and "build as code", the _existence_ of these files can be assessed and demonstrated by including a pointer/URI to the appropriate Github repo or file. The `buildConfig` field in the SLSA provenance seems appropriate for this, and could be described as a requirement for L1-L3 builds.
 
* "Build service" is a bit tougher to assess, but I think including (a pointer to) the corresponding GHA or Jenkins logs may demonstrate that the specified build script was indeed run by a build service. There are two challenges here. First, the CI logs tend to be private, so project owners may be hesitant to expose these. Second, CI logs get huge quite quickly, so they cannot simply be downloaded and included in SLSA provenance as is. That said, I don't know if it should be within the scope of SLSA to address these issues, only to require/recommend that the build service logs be made available somehow for L2+ builds.

[...]

Issue #454, opened 2022-07-28
Parameterless requirement for Level 4 prevents re-use of build scripts


shaunmlowry:
For build systems which are capable of building many millions of individual packages and versions, it's not practical to require that every individual package version has its own unique build script especially if those scripts are the interface between a build service and the native build system for a package's ecosystem.  For example, a build service might have a script which is generically capable of building Python wheels but needs to be aware of the specific build backend it needs to invoke for that module (e.g. flit, poetry, etc.).  A script designed to build autotools-based C libraries may need to know the location or configuration of a dependency to feed that information to the `configure` script.  Whilst it would be possible to create a script which wrapped the generic script along with the specific options required, that seems a little disingenuous as it's hiding details of the build and would seem to exist only as a solution to the 'Parameterless' requirement.

My understanding of the intent of the Parameterless requirement is that is is to prevent end users of the build system injecting their own parameters at build time.  Instead of allowing this, would it instead be an option to allow only service-generated parameters?

For example, the ActiveState Platform maintains a database of packages which not only acts as a revision control system for the packages, but also includes a revision-controlled mechanism for composing command line flags for build scripts.  This allows us to maintain a few build scripts per ecosystem and have them perform builds for many thousands of packages and versions, customized at build time via parameters generated by rules in the build orchestrator.  These builds are then not strictly parameterless and we should report the parameters supplied to the build scripts in the appropriate attestations (at the very least for reproducibility), however there's no way for a user of the build system to influence those parameters so it meets the intent.

MarkLodato:
Should we merge this with #278?

So I understand correctly, are you saying that effectively the only allowed parameters are booleans or enums, not arbitrary strings? (Follow-on question: what about integers? floats? anything else generally considered safe?)

If so, that sounds good to me. (If not, I'll need to first understand.)

mlieberman85:
Closing as inactive — no activity since this was filed. If it's still worth looking at, feel free to reopen or file a new ticket with updated context.

Issue #278, opened 2022-01-26
Clarification on scope of the requirement for parameterless builds


shaunmlowry:
At ActiveState, we're producing a build system which is capable of building arbitrary sets of open source components.  Even when considering just the Perl and Python ecosystems, there are tens of millions of individual components and versions thereof so providing a parameterless build script for every one isn't practical.  When you also consider that users sometimes want variants of builds (static vs dynamic linking, debug vs optimised, threading, GPU acceleration etc. etc.) creating a matrix of build scripts for every component for every combination of variations is completely intractable.  In ActiveState's case, we have a few dozens of unique build scripts which can be applied to many components to build them along with some constrained sets of parameters for individual component quirks and/or different build variants.  It is, however, an anti-goal to allow end users direct access to the parameters for the build of any given component, so for example we'd want to allow our users to specify they want a debuggable build across their entire runtime, but we don't want them to be able to specifically set CFLAGS=-g to do that - the ActiveState platform and its constrained set of parameters to build scripts are designed to manage that expression of the user's intent.

Can we clarify the wording on the requirement for parameterless builds to at least give guidance on the scope at which we can consider builds to be parameterless and some detail on the spirit of the requirement?

MarkLodato:
Agreed on the need to address this by updating the spec with more nuance.

FWIW, we have the same issue internally within Google. Our build system allows parameters for the same reasons, e.g. to support debug vs optimized builds. We need to figure out how to handle this. It's impractical for us to just say, "No more parameters."

As mentioned in yesterday's bi-weekly meeting, the motivation behind this requirement is to make it practice for the consumer to determine what parameters are acceptable. For example, is it OK for an artifact to be built with debug flags? Maybe, since that doesn't significantly change behavior. Is it OK for an artifact to be built with `CFLAGS=-Dsomething=else`? Maybe not, since that does change behavior. But it's not yet clear to me how to encode this into the spec.

A good first step would be to explain this motivation in the spec, and maybe link to this issue?

david-a-wheeler:
E.g., an allowlist of parameters for the build system as a whole.

Note: Individual components may have more capabilities without invalidating this. E.g., a makefile used in a build, may have a huge number of parameters, but I think this should apply to the build system as a whole (e.g., the build system / script that invokes make, not the makefile in isolation).

shaunmlowry:
[...]

PR #525, opened 2022-11-01
Provenance v1.0: initial draft

MarkLodato:
This is an initial draft v1.0 for the provenance format. It attempts to address the major sources of confusion with the current format (v0.2), particularly the separation of concerns between top-level external inputs, resolved dependencies, and per-run details.

There are still many TODOs. I hope to get this merged sooner rather than later so that we can get feedback on the schema and substance, then send further PRs to address the remaining TODOs.

(I'll rewrite the history before merging so that it's not full of all my "WIP" commits.)

Preview: https://deploy-preview-525--slsa.netlify.app/provenance/v1/


docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,228 @@
...
+
+  // Other information necessary to perform the build but either derived from
+  // `topLevelInputs` or provided by the build system.
+  //
+  // The values here SHOULD NOT be under external control unless they are
+  // constrained and unlikely to significantly affect the build. For example,
+  // the username who triggered the build might be listed here, rather than
+  // `topLevelInputs`, because it is a small input and unlikely to be affect the
+  // behavior of the output, aside from simple inclusion.
+  BuildDependencies buildDependencies = 2;
```
joshuagl, line 69: The term dependency is accurate, but doesn't quite gel for me, I wonder if we might call this `serviceInputs` or similar. That's where we _expect_ these inputs to come from at higher SLSA levels, but might not make sense at lower levels. Perhaps even `controlledInputs`? 🤔 

Apologies for bike shedding.

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,228 @@
...
+  // debugging, incident response, and vulnerability management. The values here
+  // MAY be necessary for reproducing the build, but ideally this would not be
+  // the case.
+  google.protobuf.Struct environment = 2;
+}
+
+message ArtifactReference {
+  // [URI] describing where this artifact came from. When possible, this SHOULD
+  // be a universal and stable identifier, such as a source location or Package
+  // URL.
```
joshuagl, line 156:
```suggestion
  // URL (purl).
```
to help unfamiliar readers find it

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,228 @@
...
+  // Collection of cryptographic digests for the contents of this artifact.
+  DigestSet digest = 2;
+
+  // The name for this artifact local to the build.
+  //
+  // Example: `PyYAML-6.0.tar.gz`
+  string localName = 3;
+
+  // [URI] identifying the location that this artifact was downloaded from, if
+  // different and not derivable from `canonicalId`.
```
joshuagl, line 170:
```suggestion
  // different and not derivable from `uri`.
```
?

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,228 @@
...
+  // -   Logs generated during the build.
+  // -   Fully evaluated build configuration.
+  //
+  // In most cases, this SHOULD NOT contain all intermediate files generated
+  // during the build. Instead, this should only contain files that are likely
+  // to be useful later and that cannot be easily reproduced.
+  //
+  // TODO: Do we need some recommendation for how to distinguish between
+  // byproducts? For example, should we recommend using `localName`?
+  repeated ArtifactReference byproducts = 3;
```
joshuagl, line 207: are we in favour of defining this here and _not_ in the statement layer? https://github.com/in-toto/attestation/issues/114

joshuagl (commented):
This is looking really good. I did a first pass review and added some minor observations and comments.

docs/provenance/v1.0.proto
MarkLodato: Done. I added `[]`'s so that, once it's Markdownified, it can link to the PURL website.

docs/provenance/v1.0.proto
MarkLodato: Yup, thanks for catching.

docs/provenance/v1.0.proto
MarkLodato: Either way works for me, so long as it's clear to readers how to use it. My reason for adding it at all (whether here or in statement) is as a replacement for some use cases of `buildConfig`, which I hopefully better explained in the other comment.

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,353 @@
...
+}
+```
+
+The following fields from v0.2 are no longer present in v1.0:
+
+-   `buildConfig`: Replacement depends on the use case:
+    -   If the build configuration is an independent input, hash it
+        deterministically and include it as `inputArtifacts.config`.
+    -   Else if there is a known use case for the resolved build configuration,
+        hash it deterministically and include it in `byproducts`.
```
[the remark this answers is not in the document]
MarkLodato: Added, PTAL.

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,359 @@
...
+                // The top-level binary that orchestrates the build within the
+                // container.
+                // TODO: Is that right?
+                "builderBinary": {
+                    "uri": "git+https://github.com/slsa-framework/slsa-github-generator@refs/tags/v1.2.0",
+                    "digest": { "sha1": "bdd89e60dc5387d8f819bebc702987956bcd4913" }
+                }
+            },
+            // Some config file within the source that has more options.
+            "entryPoint": "path/to/config.file",
```
laurentsimon, line 113: if this is generated by a GitHub builder, should it not be the workflow.yml to be consistent with provenance from other GitHub generators? The config file would be part of the `parameters` then?

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,359 @@
...
+                },
+                // The binary that runs under the reusable workflow.
+                // TODO: is this actually a top-level input, or is this really
+                // the `builder.id` and a `resolvedDependencies`?
+                "builderBinary": {
+                    "uri": "git+https://github.com/slsa-framework/slsa-github-generator@refs/tags/v1.2.0",
+                    "digest": { "sha1": "bdd89e60dc5387d8f819bebc702987956bcd4913" }
+                }
+            },
+            "entryPoint": ".github/workflow/release.yml",
```
laurentsimon, line 155: I wonder whether we should think about differentiating the CI's entryPoint from the lower-level "command" entryPoint. My reasoning is that CI systems / builders' entryPoint are not friendly to re-builders. For example:
- A GitHub reusable workflow cannot easily be re-run without building a complicated system.
- A GCB entryPoint (cloudbuild.yml) can only be re-built thru GCB or a re-builder needs to extract the commands to run from the cloudbuild config file. Even then, there are peculiarities like substitution variables, etc.

I feel like providing a lower-level entryPoint / commands that is (ideally) *agnostic* to the builder would be useful. In this Go example, it could be the go compiler commands to run (or the builder's binary that wraps it, or both). In the case of GCB, it could be a go command. For dockerfile builders, it could be a docker command. In the case of tekton, it could be a tekton command (unless something else is also possible). 

You could imagine builders generating additional docker image that helps rebuilder - which would pack all the binaries needed.

I think this would lower the bar to re-build, and force builders to add information about compiler version, hashes, etc. which would also pave the way to possible reproducibility.

Wdut?

docs/provenance/v1.0.proto
joshuagl: Thanks. Note, the purl folks never capitalise – it's purl not PURL.

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,353 @@
...
+hero_text: To trace software back to the source and define the moving parts in a complex supply chain, provenance needs to be there from the very beginning. It’s the verifiable information about software artifacts describing where, when and how something was produced. For higher SLSA levels and more resilient integrity guarantees, provenance requirements are stricter and need a deeper, more technical understanding of the predicate.
+---
+
+## Purpose
+
+Describe how an artifact or set of artifacts was produced so that:
+
+1.  Consumers of the provenance can verify that the artifact was built according
+    to expectations.
+2.  Others can rebuild the artifact, if desired.
```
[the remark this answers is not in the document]
joshuagl: Thank you.

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,353 @@
...
+
+> **WARNING:** This is not yet finalized.
+
+> **TODO:** Move this to a separate file with a full description.
+
+```jsonc
+"predicate": {
+    "buildDefinition": {
+        "topLevelInputs": {
+            "buildType": "https://slsa.dev/container-based-build/v0.1-draft",
```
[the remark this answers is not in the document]
joshuagl: Thanks for the additional text!

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,353 @@
...
+To migrate from [version 0.2][0.2] (`old`), use the following pseudocode. The
+meaning of each field is unchanged unless otherwise noted.
+
+```javascript
+{
+    "buildDefinition": {
+        "topLevelInputs": {
+            // The `buildType` MUST be updated for v1.0 to describe how to
+            // interpret `inputArtifacts`.
+            "buildType": old.buildType,
```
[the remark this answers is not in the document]
joshuagl: This change addresses my concern, thanks!

docs/provenance/v1.0.md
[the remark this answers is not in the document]
joshuagl: Much clearer with the updates, thanks for doing that!

docs/provenance/v1.0.md
MarkLodato: Maybe I should remove this from this PR until it's done. The idea is that it's irrelevant what triggered the build, GitHub Actions or whatever.

rbehjati: The documentation for `entryPoint` says that
```
  // This field SHOULD be omitted if it is implicit from `buildType`. Being
  // empty simplifies verification of the provenance, thus is preferred when
  // possible.
```

Perhaps in this case, the `entryPoint` should be omitted, and instead be included as a parameter?

> should it not be the workflow.yml 

I thought about it too, but then I am not sure if this would conform to the definition of `entryPoint`.

In my discussions with Asra (cc @asraa), we concluded that the builder tool (or actually a parser tool) should generate the `buildDefinition` and use that for actually building the binary. In my opinion, it is more sutable to include `workflow.yml` in `runDetails`.

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,228 @@
...
+  // target label within that file. The syntax and interpretation are defined by
+  // `buildType`.
+  //
+  // This field SHOULD be omitted if it is implicit from `buildType`. Being
+  // empty simplifies verification of the provenance, thus is preferred when
+  // possible.
+  //
+  // Design rationale: The `entryPoint` is distinct from `parameters` to make it
+  // easier to verify. Most builds need an entry point but not extra parameters.
+  string entryPoint = 3;
```
rbehjati, line 115: I am struggling to understand the `entryPoint`. Do you have an example? For instance, in case of docker, the entrypoint is basically the command to run. The `entryPoint` here does not seem to have the same semantics. Is it a file (config, or script) that is expected to be used as input to a build command?

docs/provenance/v1.0.md
MarkLodato: My feeling is that this would go in the wrong direction. That is kind of what we had with the old `buildConfig`, and I think it led to confusion.

It all comes down to trust boundaries. What is the interface to the build? All input artifacts and parameters that cross that boundary must be verified. To be practical, we want to minimize the number of inputs that cross that boundary.

Either:

- The CI/CD config is the top-level input and has to be verified, e.g. that the source repo doesn't change from one build to the next. In this case, the lower-level commands are implementation details and thus irrelevant to SLSA. Yes, this has all of the downsides you mention, but that's where we are today. If the canonical list of commands to run is in a cloudbuild.yml file, the only option is to verify that file.
- The lower-level entry point is the top-level input. Everything defined in the CI/CD config that gets passed into the lower-level command has to be verifier. This has two challenges:
  - The builder needs to design the build system such that this entry point is an actual security boundary. For most builders, that is not the case today.
  - The lower-level entry point is often a more complex interface with more parameters, and some of these are subject to change between builds. Thus it becomes difficult to verify. For example, if one version runs `make foo` then the next runs `make -j3 foo` then another does a refactor and does `make -j3 foo-binary`, they all look like meaningful changes that a downstream system would need to make sure are safe. The only way to make this work would be to define a stable entry point that is unexpected to change.

I do think it's desirable to define a lower-level entry point as the interface, and that is in fact what the docker-based rebuilder is aiming to do. But it needs to be designed explicitly for this purpose. I'm worried that listing both in the same provenance would lead to misunderstanding and mistakes.

docs/provenance/v1.0.proto
MarkLodato: I'm wondering whether we should just get rid of `entryPoint`.

The idea was that the `entryPoint` is the primary parameter to the build, whether a file path, a "target", a command, or really anything. The thinking was that:
- Virtually every build has such a parameter, which has the majority impact on the build.
- Making it a standard field makes verification simpler and allows us to build requirements for SLSA L3. In particular, the builder must record the entry point and the verifier must check that it matches expectations.
- Other parameters are assumed to have less of an effect and are only subject to recording/verifying at SLSA L4, because doing so would be more difficult than the entry point.

Maybe we should drop this idea. The notion of "entry point" seems to be generating more confusion than it is solving, and the "parameterless" requirement of v0.1 was too strict anyway (#278, #454).

Instead, perhaps we should simplify this by saying that L3 requires all build inputs to be recorded and matched to expectations.

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,359 @@
...
+                // The repo containing the top-level workflow.
+                "source": {
+                    "uri": "git+https://github.com/laurentsimon/slsa-verifier-test-gen@refs/heads/main",
+                    "digest": {
+                        "sha1": "15bf79ea9c89fffbf5dd02c6b5b686b291bfcbd2"
+                    }
+                },
+                // The binary that runs under the reusable workflow.
+                // TODO: is this actually a top-level input, or is this really
+                // the `builder.id` and a `resolvedDependencies`?
```
asraa, line 149: @rbehjati mentioned this in the comment https://github.com/slsa-framework/slsa/pull/525#discussion_r1015372561
But I want to address it to this TODO. Technically, we can have different versions of the `builderBinary`, and still produce the same artifact (e.g. this is the "orchestrator" or the execution, and is not a factor of the artifact ultimately produced).

 I think this is either a `resolvedDependency` or belongs in `runDetails`. My argument for why it's NOT `resolvedDependency` is because it's NOT inferred from topLevelInputs, but actually from the `builder` in the `runDetails`. It sort-of feels like there's no place for it: I don't think it's a `byProduct`, but actually it's a resolved dependency within the builder. E.g. a tool used by the builder.

docs/provenance/v1.0.proto
MarkLodato: Any inputs to the build (actions or binaries) would be listed in `resolvedDependencies`. The `byproducts` is for intermediate outputs, such as log files or a structured list of steps taken. Or am I misunderstanding your command, @asraa?

docs/provenance/v1.0.md
MarkLodato: Ah! Thanks for explaining! Yes, that was missing - dependencies used by the orchestrator. I just added `builderDependencies` as a placeholder and to spur discussions.

This is further evidence that we need a better model and terminology for the builder, with clearly defined trust zones / security boundaries.

docs/provenance/v1.0.proto
asraa: Got it: byproducts is probably not the right thing. I just noticed a last commit including `builderDependencies`: I think this would be the right thing

docs/provenance/v1.0.proto
MarkLodato: Agreed (both here and the new `builderDependencies`). Question: would it help to separate out things provided by the builder vs things reference by the user's config? Would that make naming easier?

joshuagl: I think it would help to separate out things provided by the builder vs. things referenced by the recipe. That would help this naming discussion, but could also help clarify requirements around requiring all inputs be recorded.

docs/provenance/v1.0.proto
rbehjati: Based on Oak use case and the docker-based builder, I vote for removing `entryPoint` :) I am not familiar with other use cases.

> Instead, perhaps we should simplify this by saying that L3 requires all build inputs to be recorded and matched to expectations.

This sounds like a clear and reasonable requirement to me.

docs/provenance/v1.0.md
rbehjati: Thanks for bring this up @asraa. I like the new `builderDependencies`.

docs/provenance/v1.0.proto
MarkLodato: Done.

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,234 @@
...
+  // different and not derivable from `uri`.
+  //
+  // Example: `https://files.pythonhosted.org/packages/36/2b/61d51a2c4f25ef062ae3f74576b01638bebad5e045f747ff12643df63844/PyYAML-6.0.tar.gz`
+  string downloadLocation = 4;
+
+  // [Media Type] (aka MIME type) of this artifact.
+  string mediaType = 5;
+}
+
+message DigestSet {
```
kommendorkapten, line 179: How about this:
```protobuf
enum HashAlgorithm {
    HASH_ALGORITHM_UNSPECIFIED = 0;
    SHA1 = 1;
    SHA256 = 2;
    SHA512 = 2;
}

message MessageDigest {
    HashAlgorithm algorithm = 1;
    bytes digest = 2;
} 

message DigestSet {
    repeated MessageDigest digests = 1;
}
```

The benefit of this is the the digests can be iterated over, and for clients who wishes to rely on the proto-JSON serialialzer don't need to think about binary marshalling as proto-JSON defines bytes to be base64 encoded strings when JSON encoded, small thing but makes the developer experience better as one failure mode (explicit base64 decoding) can be omitted.

docs/provenance/v1.0.proto
kommendorkapten: From the perspective of GitHub actions, I think it makes sense to keep it, as there can be many workflow files defined in the repository. My reasoning is that this file would contain `/path/to/file.yaml:$JOB_NAME` (for GitHub Actions that is) as that will uniquely define the real job that initiated the build. 

Having this value dependant on the `buildType` as defined today feels flexible, build systems that does not need it can simply ignore it.

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,234 @@
...
+  // TODO: Should this be a map instead of an array? Then each MUST be named
+  // explicitly, which would be less ambiguous but more difficult.
+  repeated ArtifactReference resolvedDependencies = 1;
+
+  // Parameters of the build environment that were provided by the `builder` and
+  // not under external control. The primary intention of this field is for
+  // debugging, incident response, and vulnerability management. The values here
+  // MAY be necessary for reproducing the build, but ideally this would not be
+  // the case.
+  google.protobuf.Struct environment = 2;
```
kommendorkapten, line 150: Do we really need a `Struct` here? Would it be to limiting to require `map<string, string> environment`?

docs/provenance/v1.0.proto
TomHennen: If specifying the algorithm as an enum would users still be able to specify algorithm types of their own?  Some users might have custom immutable identifiers that they want/need to use instead of the traditional ones.

TomHennen: I suppose this might also an issue with the current proposal, even without the enum.

kommendorkapten: Yeah, we can of course list "all known hash algorithms" too, or switch to a string. Risk with a free form string is that there can be variations like `sha-256` and `sha256`. Casing is of course easy to resolve. 
We can even have an enum `CUSTOM` where the verifier _should_ know how to interpret the bytes, either by oob knowledge, or by inspecting the number of bytes.

MarkLodato: Let's move this to a separate issue: #533. It's too complex to discuss here, and it goes beyond just this PR.

docs/provenance/v1.0.proto
MarkLodato: The alternative would be to have that be a parameter. So the two options are:

```json
{
  "entryPoint": ".github/workflows/build.yaml:my_job",
  "parameters": {}
}
```

```json
{
  "parameters": {
    "workflow": ".github/workflows/build.yaml",
    "job": "my_job"
  }
}
```

I'm personally on the fence.

docs/provenance/v1.0.proto
MarkLodato: I'm open to it. It would make verification simpler at the cost of some edge cases being more difficult to express. Probably worth looking through real-world builders to see what their parameters are.

docs/provenance/v1.0.proto
ianlewis: I'm not super familiar with the goals of the docker-based builder but I would think then that this would be the image used. The command to run would be a parameter. Docker images can contain default commands in their metadata for example which would correspond to an image entrypoint and no parameters.

kommendorkapten: > The alternative would be to have that be a parameter. So the two options are:

Having the parameters containing two attributes `workflow` and `job` is preferred I think, as that removes the need for any bespoke composition/parsing of workflow and job.

rbehjati: > I'm not super familiar with the goals of the docker-based builder but I would think then that this would be the image used. The command to run would be a parameter. Docker images can contain default commands in their metadata for example which would correspond to an image entrypoint and no parameters.

A docker image has to be specified by a URI and digest, the same way materials are. A single string is not good enough for specifying the image.
The builder image is an artifact, and is included as an `inputArtifacts` in the `buildDefinition`.

rbehjati: > > The alternative would be to have that be a parameter. So the two options are:
> 
> Having the parameters containing two attributes `workflow` and `job` is preferred I think, as that removes the need for any bespoke composition/parsing of workflow and job.

All this feels more like something that belongs to `runDetails` in my opinion. That way the `buildDefinition` can be independent of builder (e.g., GitHub and Google Cloud Build), which I thought was the point about separating `buildDefinition` and `runDetails`. That way the same builds on GitHub and GCB will have the same `buildDefinition` and only differ in `runDetails`.

asraa: > That way the same builds on GitHub and GCB will have the same buildDefinition and only differ in runDetails.

I think this is definitely the part I like about this format. 

But I too agree that entrypoint in the case of GitHub is pretty critical information to differentiate the workflow in case there are multiple workflows defined and a user wants to ensure/pinpoint it. To me, it would be nice if this was specified in `runDetails`, since it's not a property of the build definition, but more of the run/invocation that users are interested in.

docs/provenance/v1.0.proto
MarkLodato: Renamed, PTAL.

MarkLodato (commented):
Hey all, thanks for the reviews and suggestions! I did some major refactoring again to hopefully address all concerns. I also added a diagram and expanded the model, with lots of help from the SLSA Specification group today. Please take a look and let me know what you think! As always, feedback is very welcome, both positive and constructive!

docs/provenance/v1.0.cue
```diff
@@ -0,0 +1,51 @@
...
+        "buildDefinition": {
+            "buildType": string,
+            "externalParameters": {
+                "artifacts": { [string]: #ArtifactReference },
+                "values": { [string]: string },
+            }
+            "systemParameters": {
+                "artifacts": { [string]: #ArtifactReference },
+                "values": { [string]: string },
+            }
```
MarkLodato, line 18: I'm not in love with this design. It seems like we have four properties we want to express:
- disposition (external vs system)
- type (artifact vs string)
- name (which must be unique across all four fields!)
- value

Designs I considered:
- `externalParameters.artifacts["name"] = <value>` (i.e. status quo). The outer grouping by disposition makes it easy to say "only verify the external parameters, don't worry so much about system parameters", and then grouping by type makes the data more compact. But it's really awkward to document, particularly having to ensure unique names across four fields. It makes me think it's too confusing.
- `externalParameters["name"] = {"artifact": <object>}` / `{"value": <string>}`. More logical, though it has more repeated keys (everyone has to say `"artifact"` or `"value"`). Still have to say not to duplicated names across fields, but now it's only two and not four (and only one level of nesting instead of two.)
- `parameters["name"] = {"disposition": "external", "artifact": <object>}`. Avoids the name duplication issue, but I think that is harder to understand and harder to apply a policy against.

Any other thoughts or ideas? I'm leaning towards the second one but I'm still not thrilled with it.

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,400 @@
...
+                // The repo containing the top-level workflow.
+                "source": {
+                    "uri": "git+https://github.com/laurentsimon/slsa-verifier-test-gen@refs/heads/main",
+                    "digest": {
+                        "sha1": "15bf79ea9c89fffbf5dd02c6b5b686b291bfcbd2"
+                    }
+                }
+            },
+            "values": {
+                "workflow": ".github/workflow/release.yml",
```
laurentsimon, line 181: Should these not be similar to the docker example? The dockerfile example may be built by a GitHub reusable workflow, yet the release.yaml does not show in its provenance. If we follow the same structure, there should be no release.yaml and we should add a `config: <>` instead, for https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_go_slsa3.yml#L47? And probably this argument should also be declared https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_go_slsa3.yml#LL52C6-L52C21 because it's necessary to define the build fully?

Recording this project workflow seems useful, but I'm not sure it belongs in the `values`.

docs/provenance/v1.0.proto
laurentsimon: What would the interface be for a GitHub reusable workflow? The interface it attests to is the set of inputs: https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_go_slsa3.yml#L37-L65, IIUC? But it also knows about the "inside" and can theoretically attest to it. I think the "docker-based builder" really is an "inside" builder. The majority of builders have an inside builder which is closer to the compiler or is the compiler - in the case of docker-based builder, the compiler *is* the container image.

Why is the "docker-based" inner builder treated differently from a "Go binary" inner builder? Is it because it's harder to define the "interface" for a Go builder?

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,250 @@
...
+  // REQUIRED for SLSA Build L1.
+  ParameterCollection externalParameters = 2;
+
+  // Parameters of the build environment that were provided by the `builder` and
+  // not under external control. The primary intention of this field is for
+  // debugging, incident response, and vulnerability management. The values here
+  // MAY be necessary for reproducing the build.
+  //
+  // OPTIONAL.
+  ParameterCollection systemParameters = 3;
```
kommendorkapten, line 91: For all the attributes which are optional, wouldn't it make sense to declare them as `optional` too in the protobuf spec? As clients using protobuf parsers can then query for the existence of a value? 
Also by having them as optional in the protobuf, the formal specification states that they are optional, without the need of relying on a comment, which I think is a good thing.

MarkLodato: I don't think so. The `optional` keyword enables differentiation between "set-but-empty" and "unset", but we don't want that behavior. Our parsing rules say that the two should be considered the same, which is generally the less confusing route.

Also the lack of "optional" to mean "required" is too subtle, so we'll still need a comment. I think it's easier to read if everything uses the same convention to document optional vs required.

docs/provenance/v1.0.proto
MarkLodato: Sorry about these hastily constructed examples. They seem to be generating more confusion than they are helping. Really each needs a proper specification of its own. I just haven't had time to do this.

By "GitHub Actions" I meant the public top-level interface provided by the platform, which is the workflow. I don't want to over-index on the "SLSA GitHub reusable workflows" because that model is not widely applicable to other builders. It's really a hack, and ideally GitHub itself would generate the provenance. If GitHub generates the provenance, then it can only attest to the workflow or possibly the job, because that is the security boundary.

You're right about the "builder_go_slsa3" being similar to the docker-based builder in that both have a narrower interface. That was my mistake - I meant the "builder_generic_slsa3" here. Fixed.

docs/provenance/v1.0.cue
```diff
@@ -0,0 +1,51 @@
...
+                "startedOn": #Timestamp,
+                "finishedOn": #Timestamp,
+            },
+            "byproducts": [ ...#ArtifactReference ],
+        }
+    }
+}
+
+#ArtifactReference: {
+    "uri": string,
```
mlieberman85, line 38: Just note: https://github.com/cue-lang/cue/issues/1433 This should hopefully be a type eventually

docs/provenance/v1.0.cue
```diff
@@ -0,0 +1,51 @@
...
+        "sha1": string,
+        // TODO: list the other standard algorithms
+        [string]: string,
+    },
+    "localName": string,
+    "downloadLocation": string,
+    "mediaType": string,
+}
+
+#Timestamp: string  // <YYYY>-<MM>-<DD>T<hh>:<mm>:<ss>Z
```
mlieberman85, line 51: There is no need for us, in CUE you can use `time.Time`

MarkLodato: I was trying to strike a balance in terms of human readability vs machine processing. My intent is for this to be primarily documentation without the reader even knowing it's CUE, but it just so happens to be a CUE file as well since the syntax is easy to read.

I'm worried that `time.Time` will be a hit in terms of readability. It requires an `import` statement at the top (which is distracting) and forces the reader to understand what `time.Time` means. I thought perhaps a simple string was good enough.

mlieberman85: I see. It might make sense to highlight that in a comment at the top so folks who might look at the file and not recognize it's under docs/provenance know it's intended mostly as a document for reference, not as something that should be used for automation.

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,250 @@
...
+  // identify the tenant project?
+  //
+  // REQUIRED for SLSA Build L1 unless the id is implicit from the attestation
+  // envelope (e.g. public key).
+  string id = 1;
+
+  // TODO: Do we want to add this field? (#319)
+  //
+  // OPTIONAL
+  string version = 2;
```
MarkLodato, line 229: todo: `map<string, string>`

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,404 @@
...
+
+Understanding of SLSA [Software Attestations](/attestation-model)
+and the larger [in-toto attestation] framework.
+
+## Model
+
+Provenance is an attestation that the `builder` produced the `subject` software
+artifacts through execution of the `buildDefinition`.
+
+![Build Model](v1_model.svg)
```
MarkLodato, line 28: TODO: Add caches to the model.

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,404 @@
...
+> **TODO:** Limit the size of the diagram - it's too big on desktop (but fine on
+> mobile). Also reimplement it in Figma so that others can edit it.
+
+## Schema
+
+> **TODO:** Get proper syntax highlighting for cue, and explain that this is a
+> cue schema.
+
+```javascript
+{% include_relative v1.0.cue %}
```
MarkLodato, line 77: TODO: Add requirements for each level. I think something like:

- Build L1: SHOULD set all fields, but there is some amount of leeway to allow implementations to avoid expensive bits. For example, if not all input artifacts are known, that's OK. But if everything is technically optional, something could trivially pass L1 by simply outputting an empty provenance, which is obviously no good. So maybe have some requirement that at most 1-2 of the following things could be absent: {...}
- Build L2: Same guidance, but perhaps a bit stronger?
- Build L3: All fields required?

docs/provenance/v1.0.cue
joshuagl: Nice succinct description of the four properties. 👍 

I'm torn between options two (separate fields for system and external, keyed by name) and five (array of objects).

One variant would be to index on name as in your fourth option but to have defaults for disposition (system) and type (artifact). I am not advocating for this though, I still prefer option two or five.

docs/provenance/v1.0.md
```diff
@@ -0,0 +1,404 @@
...
+            ]
+        }
+    }
+}
+```
+
+### GitHub Actions
+
+```jsonc
+"predicate": {
```
kommendorkapten, line 165: When comparing this example with the protobuf file and the documentation in it, I see some ambiguities. I'm not sure if it's due to my misunderstanding, or because this is still a draft/in motion.
In particular the comments for the `ParameterCollection`:
```
  // `source`        | The primary input to the build.
  // `config`        | The build configuration, if different from `source`.
  // `builderBinary` | The top-level binary that runs the build.
  // `buildImage`    | The container or VM image in which the build occurred.
```

In this example, my understanding is that the `buildImage` is instead recorded under `buildDefinition.resolvedDefinitions` and the `builderBinary` at `runDetails.builder.builderDependencies`.

My feeling is that the `builderBinary` and `builderImage` could as well be part of the `systemParameters`?. 
When reading the documentation for `runDetails.builder.builderDependencies` and `resolvedDependencies` it too makes sense to have the data in those fields, but it feels a bit unclear to what's the correct approach.

kommendorkapten, line 165: When comparing this example with the protobuf file and the documentation in it, I see some ambiguities. I'm not sure if it's due to my misunderstanding, or because this is still a draft/in motion.
In particular the comments for the `ParameterCollection`:
```
  // `source`        | The primary input to the build.
  // `config`        | The build configuration, if different from `source`.
  // `builderBinary` | The top-level binary that runs the build.
  // `buildImage`    | The container or VM image in which the build occurred.
```

In this example, my understanding is that the `buildImage` is instead recorded under `buildDefinition.resolvedDefinitions` and the `builderBinary` at `runDetails.builder.builderDependencies`.

My feeling is that the `builderBinary` and `builderImage` could as well be part of the `systemParameters`?. 
When reading the documentation for `runDetails.builder.builderDependencies` and `resolvedDependencies` it too makes sense to have the data in those fields, but it feels a bit unclear to what's the correct approach.

MarkLodato: Agreed. Maybe we should just define the ones we *know* are common (namely `source` and `config`), then expand it over time with more conventions as they appear?

kommendorkapten: Yes, I think that's a good approach. 

A follow up question, how to distinguish what goes into the `buildDefinition.resolvedDefinitions` vs `runDetails.builder.builderDependencies`. With the current example, the OS image goes into the build definition whereas the build image goes into builder dependencies, I'm not sure that totally obvious to me. Let's say I would need to create a new build type just by reading the specs.

The current comments are:
Build definition/Resolved dependencies: `Resolved dependencies needed at build time. For example, if the build script fetches and executes "example.com/foo.sh", which in turn fetches "example.com/bar.tar.gz", then both "foo.sh" and "bar.tar.gz" should be listed here.`
Run details/Builder dependencies: `Dependencies used by the orchestrator that are not run within the workload and that should not affect the build, but may affect the provenance generation or security guarantees.` 

I guess in retrospective, It's clear the the OS image is a resolved dependency? But I'm thinking for when a new build type is being defined (say another CI/CD system). Will it be clear how to pack all details? Can we offer more concrete examples as I do think it's desirable to keep the builders as similar as possible as that would make it easier to adopt new ones.

docs/provenance/v1.0.cue
TomHennen: IIRC array's of objects are much easier to deal with if you're using protos, but I don't know how much that matters?

From a policy perspective I suspect I'd be much more interested in checking external parameters (and would treat them differently).  System Parameters are likely going to be evaluated during the initial onboarding of the builder so they'd need less scrutiny (and end-users are probably much less likely to know what they should be anyways!).  As a result I'm guessing it would be nicer to have externalParamters and systemParameters as a separate field.

Similarly, the code paths to verify artifacts is going to differ quite a bit from ones verifying 'values' so having them in separate buckets would be handy, but isn't strictly required.

MarkLodato: OK, I wrote up both options:

- 9ca7346 = `externalParameters = { <name>: {"artifact": <value> } }`
- d63f6ba = `externalParameters = [ { "name": <name>, "artifact": <value> } }`

Let me know what you think.

docs/provenance/v1.0.cue
```diff
@@ -0,0 +1,53 @@
...
+    "subject": [...],
+
+    // Predicate:
+    "predicateType": "https://slsa.dev/provenance/v1.0?draft",
+    "predicate": {
+        "buildDefinition": {
+            "buildType": string,
+            "externalParameters": [ ...#Parameter ],
+            "systemParameters": [ ...#Parameter ],
+            "resolvedDependencies": [ ...#ArtifactReference ],
```
marcelamelara, line 13: Trying to map the discussion in #508 back to this PR and the updated requirements. We had recommended adding fields for evidence for the "scripted build" and "build as code" requirements. Even though the requirements have been updated, we still want to make sure we include a field in the provenance where the producer can attest to having used a specific build script and/or service addressing the `Producer` requirements, even if they do not want to reveal the build service or script in the provenance.

docs/provenance/v1.0.cue
joshuagl: thanks for writing up both approaches. I've looked a couple of times now and prefer https://github.com/slsa-framework/slsa/commit/9ca734683bf6163f54968c8d065d1336d2805208. It feels more human-readable (important for spec readers) and easier to work with (important for tool authors).

docs/provenance/v1.0.cue
MarkLodato: Added as a TODO

docs/provenance/v1.0.md
MarkLodato: Added a TODO.

docs/provenance/v1.0.cue
MarkLodato: Based on the discussion, I think we do not want such a field. There is an out-of-band process for certifying that a  given `builder.id` implies that various requirements have been met. It is not a property per build.

docs/provenance/v1.0.md
MarkLodato: Not sure how to implement this, so I'll just close this thread.

docs/provenance/v1.0.proto
MarkLodato: I removed all examples except the GitHub Actions one, which I fully fleshed out. Please take a look.

MarkLodato (commented):
Hi everyone, I think this is ready for proper review. It still has a bunch of TODOs for documentation, but I think the schema is ready.

NOTE: If you previously reviewed, I changed the filename so that will mess up diffs. use the GitHub "changes from" selector in the upper left to navigate the changes. Sorry about that.

MarkLodato:
> Will `systemParameters` include host details like the platform the build is running on? A build on an arm64 system will probably have different output from an amd64 host. This will need to be balanced between enough data to document the build without exposing potentially private data (like internal hostnames and IP addresses).

@sudo-bmitch I would assume that is a parameter?

sudo-bmitch:
> > Will `systemParameters` include host details like the platform the build is running on? A build on an arm64 system will probably have different output from an amd64 host. This will need to be balanced between enough data to document the build without exposing potentially private data (like internal hostnames and IP addresses).
> 
> @sudo-bmitch I would assume that is a parameter?

@MarkLodato I'm reading these as a user learning the spec, so no fair asking me questions. 😁 I'm happy if you say it's already covered.

docs/provenance/v1.0.cue
marcelamelara: Ok, so the `builder.id` would establish the binding between the provenance and the infrastructure provider questionnaire? However, presumably, in the case of a third-party verifier service who certified the build level, does this mean that the provenance consumer needs to go separately query the verifier service for the certification?

My main concern is that consumers of the provenance won't be able to (easily) link/connect the provenance to the evidence (i.e., questionnaire) for the claimed build level, so I'm still wondering why the provenance might not provide a hint for where to find the certification of the `builder.id`. I'm assuming that there isn't a single, centralized verifier/certification service; maybe I'm missing some crucial details about how the slsa-verifier fits here.

marcelamelara: Put differently, does this current approach prevent attackers from producing "fake" provenance as `builder.id` that a consumer/verifier would deem valid (based on the questionnaire)?

laurentsimon: IIUC, the concern is that a `builder.id` *inside* the provenance can be forged by anyone, and a builder is defined not just by its ID, but by its keys. So you're asking how a verifier knows *how* to verify, i.e., not just the builder.id verification but also the signature verification.
If there is an official tool like the slsa-verifier that supports the builder, users should know about it. if there is none, what do we do? I suppose the idea here is that corporate builders will have their own tool...? Or shall we encourage folks to add support in the same tool (e.g., slsa-verifier) and work on a plug-in support https://github.com/slsa-framework/slsa-verifier/issues/275?

mlieberman85: This might be just me but I had always assumed that a lot of the verification assumed you in the least trusted the identity of the signature that it's reporting accurate information. I had also assumed the certification would be an external certification something like: Alice certified on this date that Bob's builder service is SLSA 3.

MarkLodato: Yes, as @laurentsimon said, the questionnaire is tied to signature verification plus an **optional** `builder.id`.

* By "signature verification" this really means some sort of public key infrastructure that identifies the attestor, i.e. who signed the attestation. It could be a simple public key, or an X.509 cert chain like Sigstore, or something like [SPIFFE](https://spiffe.io/).

* The `builder.id` exists to allow a single attestor identity to represent two different logical build systems, with two different SLSA properties. For example, suppose Awesome Builder meets SLSA Build L3 but allows users to disable isolation between builds, which allows tampering between builds but does not allow forging of the provenance. Awesome Builder could represent this in two ways:

    * (a) As two different attestor identities, e.g. two different public keys.
    * (b) As two different `builder.id` values, signed by the same attestor identity.

I'm not 100% sure if (b) is needed. If not, we could drop `builder.id` which would simplify things a lot, I think. But I hesitate to do that until we have confidence it's not needed. Maybe just a bigger warning around it for now?

> If there is an official tool like the slsa-verifier that supports the builder, users should know about it. If there is none, what do we do? Do we expect corporate builders to provide their own verifying tool...? Or shall we encourage folks to add support in the same tool (e.g., slsa-verifier) and work on a plug-in support [slsa-framework/slsa-verifier#275](https://github.com/slsa-framework/slsa-verifier/issues/275) to scale that?

Sorry, I'm not sure who this comment was addressed to, or if you'd like to see anything in this PR about it.

mlieberman85: That second bullet point I think we just need to be clear about. I'm thinking about implementation issues where having the same builder use different identities in different contexts could complicate things. I think (b) probably makes the most sense though it could also lead to a compromised builder claiming a higher SLSA level.

docs/github-actions-workflow/v0.1/example.json
```diff
@@ -0,0 +1,44 @@
...
+                "source": {
+                    "artifact": {
+                        "uri": "git+https://github.com/octocat/hello-world@refs/heads/main",
+                        "digest": { "sha1": "c27d339ee6075c1f744c5d4b200f7901aad2c369" }
+                    }
+                },
+                "workflow_path": { "value": ".github/workflow/release.yml" }
+            },
+            "systemParameters": {
+                "github_actor": { "value": "MarkLodato" },
```
laurentsimon, line 19: nit: should keep these field name camel case like the rest of format

docs/github-actions-workflow/v0.1/example.json
```diff
@@ -0,0 +1,44 @@
...
+                "inputs_build_id": { "value": "123456768" },
+                "inputs_deploy_target": { "value": "deployment_sys_1a" },
+                "inputs_perform_deploy": { "value": "true" },
+                "source": {
+                    "artifact": {
+                        "uri": "git+https://github.com/octocat/hello-world@refs/heads/main",
+                        "digest": { "sha1": "c27d339ee6075c1f744c5d4b200f7901aad2c369" }
+                    }
+                },
+                "workflow_path": { "value": ".github/workflow/release.yml" }
```
laurentsimon, line 16: for a the GitHub runner that runs a GitHub workflow, these inputs will be the workflow_dispatch inputs.
For a re-usable workflow, the inputs will be the inputs passed to the re-usable workflow interface (e.g., https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml#L31), correct?

docs/provenance/v1.0.cue
```diff
@@ -0,0 +1,51 @@
...
+            },
+            "byproducts": [ ...#ArtifactReference ],
+        }
+    }
+}
+
+#ParameterValue: {
+    "artifact": #ArtifactReference
+} | {
+    "value": string
```
asraa, line 34: I've been doing an exercise migrating out v0.2 provenance to v1.0 and encourntered this issue: 

For things in systemParameters, some systems of builders may define their own custom object to define parameters. For example, GCB defines an "options" field https://cloud.google.com/build/docs/securing-builds/view-build-provenance#view_provenance_using_the, GitHub Actions defines a custom webhook payload (https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads) that details information that differs according to the event. This info is useful for determining the context of the run: for example, which branch it may have actually been running from may depend on whether this was a push event or a workflow_dispatch.

It would be useful to extend the flat-list of `map[string]: ParameterValue` to something that could represent a GitHub object, or to represent nested/namespaced mappings of parameter values as well (for e.g. the webhook payload could exist in its own map, rather than a flattened list).

RE flattened lists: I noticed this was done for the GitHub inputs (prefix all inputs with `inputs_` into the `externalParmeters`. I wonder if namespacing like above would clean this up, as it's a lot easier to say "does this have an inputs object? Then "is there any params that start with `inputs_` in order to determine whether something had user defined inputs. It also may not distinguish between types of inputs (CI inputs, substitution inputs, reusable workflow inputs).

melba-lopez:
@MarkLodato curious about something.  A while ago the Positioning group mentioned about SLSA L1-4 (for Build) and ensuring that anything about “source” was minimum or excluded (since its still in draft/being thought about).   Is that true for provenance as well?  I'm reading this _(with a biased view of prior conversations)_ as Provenance for the **BUILD Environment**'s dependencies (not of the package/product/application of which you are building)

docs/provenance/v1.0.cue
MarkLodato: Thanks for the feedback! I had switched to the simple map based on https://github.com/slsa-framework/slsa/pull/525#discussion_r1019250456 (@kommendorkapten) but it seems like added complexity of JSON objects is worth it.

I'm not sure how best to do this. A few ideas come to mind:

* Add a third option to specify an arbitrary JSON object: `"object": {}` in cue, `google.protobuf.Struct object = 3;` in proto
* Replace "value" with any JSON value: `google.protobuf.Value value = 2;` in proto (not sure how to this in cue)
* Replace `systemParameters` with a JSON object (`google.protobuf.Struct`) and use duck typing for artifact references. That is, any parameter that has a `uri` and/or `digest` field is considered an artifact reference.

Overall I'm still not thrilled with how parameters are specified. I find the `artifact`/`value` thing awkward, though I don't love the other options we've considered either. I wish we had a better idea for how to do this.

docs/github-actions-workflow/v0.1/example.json
laurentsimon: the `workflow_path` is the field that is a bit awkward: the GitHub runner runs the workflow, but a re-usable workflow jut runs some code based on the input provided by the caller. The rest seems to work just fine.

docs/github-actions-workflow/v0.1/example.json
laurentsimon: SGMT to use an object

docs/provenance/v1.0.cue
kommendorkapten: I'm not opposing a JSON object in general, but I do feel like there could be a risk of just showing in "too much" data and so possibly causing noise? With a more strict data type I feel like we would be more cautious about what data we actually put in the provenance. As an example, the GitHub object, do we really need the entire object in the provenance?

A potential middle ground that is a bit more flexible and kind of gives the name-spacing @asraa is looking for:
Define the `ParameterValue` like this:

```protobuf
message ParameterValue {
  // REQUIRED.
  oneof value {
    // A parameter that is a reference to an artifact.
    ArtifactReference artifact = 1;
    // A parameter that is a scalar value. For simplicity, only string values
    // are supported.
    string scalar_value = 2;
    // A parameter that is a map structure. For simplicity, only string values
    // are supported.
    map<string, string> map_value = 3;
    // A parameter that is a list value. For simplicity, only string values
    // are supported.
    repeated string array_value = 4;
  }
}
```

With this construct, we could group all the `input`s together.

(This is a simplified example, I do think `repeated` values are forbibbedn in oneofs, maybe `map` too, but that could easily be fixed with two new message types, but omitted now for simplicity.)

docs/github-actions-workflow/v0.1/example.json
kommendorkapten: If we generalize the `ParameterValue` as proposed here https://github.com/slsa-framework/slsa/pull/525/files#r1066962916

Would it make sense to have two "namespaces", `invocation` (where the `workflow_path` would reside) and an `input` where the inputs to the reusable workflow could be in. Or does this just complicate things too much?

docs/provenance/v1.0.cue
joshuagl: > Overall I'm still not thrilled with how parameters are specified. I find the artifact/value thing awkward, though I don't love the other options we've considered either. I wish we had a better idea for how to do this.

Perhaps we can pull this out into a separate issue and seek additional suggestions (from, i.e., tooling and specification groups?) and give ourselves a deadline (end of Jan? two weeks from today?) to pick an option?

docs/github-actions-workflow/v0.1/example.json
```diff
@@ -0,0 +1,44 @@
+{
+    "predicateType": "https://slsa.dev/provenance/v1?draft",
+    "predicate": {
+        "buildDefinition": {
+            "buildType": "https://slsa.dev/github-actions-workflow/v0.1?draft",
+            "externalParameters": {
+                "inputs_build_id": { "value": "123456768" },
```
laurentsimon, line 7: inputs may not be strings. Example Actions support other sort of arguments types https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/
More may be supported in the future, like arrays, etc

The discussion for `systemParameters` seem to be relevant for externalParameters too

docs/provenance/v1.0.cue
laurentsimon: It's a bit dangerous to restrict fields arbitrarily if we don't know what other builders out there *may* want. IMO, the input types are builder-specific, so they should be able to decide *how* they record information. We could complement this by have guidelines that encourage simple data structures (+ conformance / accreditation for builders). IMO, we should avoid a situation where we force builders to "canonicalize" their inputs to fit the specs, because this will lead to mistakes and friction. It would be unfortunate to have to bump the specs version *just* to accommodate a different data structure further down the line.

re: noise. Provenance is not often shown to human, and will be verified by automated tools. What we found in the slsa-github-generator is that cherry-picking the *right* fields from the GH context is pretty hard: the context format varies depending on the trigger event, etc. Several times, we've had failed verification bugs for branch verification because of subtle differences in the format, e.g., a release from the latest commit in a branch is different from a release from an older commit in the same branch. Saving the entire GH context allowed us to quickly push a fix in the verifier without the need for users to update their builders. and re-run their build. Asking users to re-run builds is worse in terms of user experience, I think. I can't recall the issue where we had this discussion (@joshuagl was on the issue as well)

Saving the entire context *may* be useful for re-running a build given a provenance. (maybe that's low priority and not all builders may want to support such feature)

docs/github-actions-workflow/v0.1/example.json
kommendorkapten: Looking at the migration plan it's mixed, some data from v0.2 invocation goes into `externalParameters` and some into `systemParemeters`. Only the `invocationId` goes into `runDetails`.
Assuming the migration plan is updated :)

docs/provenance/v1.0.cue
kommendorkapten: Yes, those are good points @laurentsimon, especially on being able to fix the verifier instead of rerunning the build 👍

asraa: Let's say a buildType is something like a GCB builder, and depending one user of GCB they can provide different named inputs/substitutions than a different user of GCB. This was the reasoning that I think having a flat, buildType dictated `externalParameters` doesn't work. That being said -- verification would likely be of the form "does this have any inputs at all?", in which case serializing the payload into a string to fit into the string-string map is totally fine.

asraa: Serializing it also means that if you *are* interested in its content, you'd be buildType/builder aware. Which, is, what this format is promoting.

docs/provenance/v1.0.cue
```diff
@@ -0,0 +1,51 @@
...
+            "buildType": string,
+            "externalParameters": { [string]: #ParameterValue },
+            "systemParameters": { [string]: #ParameterValue },
+            "resolvedDependencies": [ ...#ArtifactReference ],
+        },
+        "runDetails": {
+            "builder": {
+                "id": string,
+                "version": string,
+                "builderDependencies": [ ...#ArtifactReference ],
```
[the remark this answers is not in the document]
asraa: These may be builder tools, compilers, binaries, etc that it uses that may be interesting either for reproducibility or debugging.

docs/github-actions-workflow/v0.1/example.json
MarkLodato: > the `workflow_path` is the field that is a bit awkward: the GitHub runner runs the workflow so it is an input, but a re-usable workflow runs some code based on the input provided by the caller, and the `workflow_path` is not an input. It's fine to keep it, but it has a slightly different meaning.

Really `source` + `workflow_path` combined specifies which workflow is being run, while `inputs` specifies the extra parameters passed in (plus `vars` - see other thread). I think that's the same for both top-level and reusable workflows. The difference is that a reusable workflow _also_ gets the repo + path of the top-level workflow that triggered the reusable workflow. Given that it has two repo + path pairs, I'm not sure what to name them such that it makes sense for both cases.

In other words, top-level workflow has:

- `source` + `workflow_path` specifies which workflow was run
- `inputs` specifies the workflow_dispatch inputs
- `vars` specifies the [variables](https://docs.github.com/en/actions/learn-github-actions/variables#defining-variables-for-multiple-workflows)

While reusable workflows have:

- One repo + path for the reusable workflow (i.e. the thing being executed)
- Another repo + path for the top-level workflow (i.e. the thing that triggered the reusable workflow; this gets passed in via the context so it's a parameter.)
- `inputs` of the reusable workflow + top-level workflow (mashed together!)
- `vars` (same as top-level)

It's awkward.

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,259 @@
...
+  repeated ArtifactReference byproducts = 3;
+}
+
+message Builder {
+  // [URI] ... (same as v0.2)
+  // TODO: In most cases this is implicit from the envelope layer (e.g. the
+  // public key or x.509 certificate), which is just one more thing to mess up.
+  // Can we rescope this to avoid the duplication and thus the security concern?
+  // For example, if the envelope identifies the build system, this might
+  // identify the tenant project?
```
jagathprakash, line 227: Would we need different ids to identify the tenant project and the build system?
The tenant project is the actual user (id: group1@company1.com) who is making the build using the build system (www.build_company.com).

The signing key could then be associated with actual user (group1@company1.com) . With that we have a way of identifying builds by a particular tenant or a source code owner.

docs/provenance/v1.0.cue
rbehjati: > A potential middle ground that is a bit more flexible and kind of gives the name-spacing @asraa is looking for: Define the `ParameterValue` like this:
> 
> ```proto
> message ParameterValue {
>   // REQUIRED.
>   oneof value {
>     // A parameter that is a reference to an artifact.
>     ArtifactReference artifact = 1;
>     // A parameter that is a scalar value. For simplicity, only string values
>     // are supported.
>     string scalar_value = 2;
>     // A parameter that is a map structure. For simplicity, only string values
>     // are supported.
>     map<string, string> map_value = 3;
>     // A parameter that is a list value. For simplicity, only string values
>     // are supported.
>     repeated string array_value = 4;
>   }
> }
> ```

I like this suggestion, as I am currently struggling with serializing an array into a parameter value. I could encode the array as a string, but it is not easy to make it work currently in the general case (e.g., escaping quotes, etc.).

docs/github-actions-workflow/v0.1/example.json
asraa, line 7: Is there a rationale behind not camelcasing the key values in the parameters?

docs/provenance/v1.0.cue
tiziano88: +1 for a format that allows specifying objects / arrays natively without having to encode them as strings

docs/_data/versions.yml
```diff
@@ -29,6 +29,9 @@ provenance:
       name: Version 0.1
     v0.2:
       name: Version 0.2
+    v1.0:
```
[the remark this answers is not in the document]
MarkLodato: Converted so that it's ready for review.

docs/github-actions-workflow/v0.1/example.json
MarkLodato: Done. (Previously it was snake case because that's what the contexts use. Now that we have `mapValue` it's more natural.)

docs/github-actions-workflow/v0.1/example.json
MarkLodato: Is this thread OK to resolve, or is there a specific request here (hard to follow thread)?

docs/github-actions-workflow/v0.1/example.json
MarkLodato: Done.

docs/provenance/v1.0.cue
MarkLodato: Is this thread OK to resolve, or is there a specific request here (hard to follow thread)?

docs/provenance/v1.0.cue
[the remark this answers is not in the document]
MarkLodato: Added a TODO. (Please reopen if the TODO is not sufficient to capture the comment.)

docs/provenance/v1.0.cue
MarkLodato: Went with the suggestion from @kommendorkapten for now, with an "RFC" comment that we may revisit this in the future. I'm actually leaning towards allowing arbitrary objects, but let's leave that for a future PR.

I'm closing this thread because I think it now captures what all the commenters were saying. Could you please open a new comment in the new file if there are still outstanding issues? Thanks!

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,259 @@
...
+  // Example: `pkg:pypi/pyyaml@6.0`
+  //
+  // REQUIRED unless `digest` is set.
+  string uri = 1;
+
+  // [DigestSet] of cryptographic digests for the contents of this artifact.
+  // TODO: Decide on hex vs base64 in #533 then document it here.
+  //
+  // REQUIRED unless `uri` is set.
+  map<string, string> digest = 2;
```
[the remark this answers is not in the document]
MarkLodato: Yes. Added an example.

docs/provenance/v1.0.proto
MarkLodato: Good question. In the interest of getting this merged, let's defer this question until later. I added an "RFC" paragraph asking readers to weigh in. We'll decide this before releasing 1.0.

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,259 @@
...
+  // REQUIRED at SLSA Build L1 unless it is implicit from the attestation
+  // envelope (e.g. public key).
+  string id = 1;
+
+  // TODO: Do we want to add this field? (#319)
+  // TODO: Should we merge this with builderDependencies into a combined
+  // "builderParameters"? Then arbitrary information can be stored.
+  //
+  // OPTIONAL.
+  map<string, string> version = 2;
```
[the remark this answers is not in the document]
MarkLodato: Good point. I added an "RFC", but if that does not fully capture the gist, please let me know.

docs/provenance/v1.0.proto
```diff
@@ -0,0 +1,259 @@
...
+
+  // Dependencies used by the orchestrator that are not run within the workload
+  // and that should not affect the build, but may affect the provenance
+  // generation or security guarantees.
+  // TODO: Flesh out this model more.
+  //
+  // OPTIONAL.
+  repeated ArtifactReference builderDependencies = 3;
```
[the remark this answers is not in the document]
MarkLodato: This should only be (1). (2) should go in `resolvedDependencies`. Do you have a suggestion to make that more clear?

docs/provenance/v1.0.proto
MarkLodato: Marking as resolved. Please open new comments in the updated files if there are outstanding issues. Thanks!

MarkLodato (commented):
OK everyone, thanks for the very helpful feedback! I'm hoping to get this merged this week, so if you could review end-of-day Friday (US Eastern), that would be appreciated!

There are still many TODOs and outstanding issues, particularly on the design of the parameters. The intention is that we submit this and iterate. By having it checked in, we can more easily get feedback from potential implementers.

Preview: https://deploy-preview-525--slsa.netlify.app/provenance/v1/

NOTES:

1. I aggressively marked GitHub comment threads as "resolved" in order to reduce noise. I did my best to try to capture the intent either as a TODO or actually implementing it. Feel free to reopen threads, though the UI doesn't make that particularly easy...

2. I reorganized a bit and changed around the documentation format. Please take a look. I'm hoping this new format is now easier to follow. (The reason for moving under a directory is to make it easier to have related files, and the reason for dropping the ".0" from the URL is that is what our documentation says we would do, since minor version bumps are backwards compatible.)

Thanks all!

trishankatdatadog (commented):
Looks good to me, thanks very much! Could we please ask another CI provider (e.g., GitLab) to contribute their own Build Type to make sure that we have designed a general enough Build Provenance?

MarkLodato:
> Looks good to me, thanks very much! Could we please ask another CI provider (e.g., GitLab) to contribute their own Build Type to make sure that we have designed a general enough Build Provenance?

Good point. Added TODO noting that it's a blocker for calling it stable.

docs/provenance/v1.0.cue
MarkLodato: > I think so too. Do you know what timeline that would be?

Early next week? Basically I just want to unblock getting this merged.

docs/github-actions-workflow/v0.1/example.json
laurentsimon: I think it's fine, yes

docs/provenance/v1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+[0.1]: v0.1.md
+[0.2]: v0.2.md
+[DigestSet]: https://github.com/in-toto/attestation/blob/main/spec/field_types.md#DigestSet
+[ResourceURI]: https://github.com/in-toto/attestation/blob/main/spec/field_types.md#ResourceURI
+[Statement]: https://github.com/in-toto/attestation/blob/main/spec/README.md#statement
+[Timestamp]: https://github.com/in-toto/attestation/blob/main/spec/field_types.md#Timestamp
+[TypeURI]: https://github.com/in-toto/attestation/blob/main/spec/field_types.md#TypeURI
+[in-toto attestation]: https://github.com/in-toto/attestation
+[parsing rules]: https://github.com/in-toto/attestation/blob/main/spec/README.md#parsing-rules
```
joshuagl, line 609:
```suggestion
[parsing rules]: https://github.com/in-toto/attestation/blob/main/spec/README.md#parsing-rules
[purl]: https://github.com/package-url/purl-spec
```

docs/provenance/v1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+<table>
+<tr><th>Field<th>Type<th>Description
+
+<tr id="uri"><td><code>uri</code>
+<td>string (URI)<td>
+
+URI describing where this artifact came from. When possible, this SHOULD
+be a universal and stable identifier, such as a source location or Package
+URL ([purl]). Example: `pkg:pypi/pyyaml@6.0`
```
joshuagl, line 292: Missing a link destination, I added a suggestion to the bottom of the file. Aside: we should document a convention on links and when the destination lives alongside the text.

docs/provenance/v1/index.md
```diff
@@ -0,0 +1,609 @@
...
+> -   Allow strings, maps of strings, or arrays of strings (current design)
+> -   Allow arbitrary JSON (challenge: how do we do [ArtifactReference]?)
+
+### ArtifactReference
+
+[ArtifactReference]: #artifactreference
+
+REQUIRED: at least one of `uri` or `digest`
+
+<table>
```
joshuagl, line 284: I'm not sure why, but this table (and this table only) is rendered wider than my browser window 😕 
It seems to be the `uri` and `downloadLocation` rows only:
Screenshot 2023-01-20 at 15 09 34

trishankatdatadog (approved):

trishankatdatadog:
> Good point. Added TODO noting that it's a blocker for calling it stable.

Approved, but shouldn't hang on my opinion alone 🙂

docs/provenance/v1/index.md
```diff
@@ -0,0 +1,609 @@
...
+The name for this artifact local to the build. Example: `PyYAML-6.0.tar.gz`
+
+<tr id="downloadLocation"><td><code>downloadLocation</code>
+<td>string (URI)<td>
+
+URI identifying the location that this artifact was downloaded from, if
+different and not derivable from `uri`. Example:
+`https://files.pythonhosted.org/packages/36/2b/61d51a2c4f25ef062ae3f74576b01638bebad5e045f747ff12643df63844/PyYAML-6.0.tar.gz`
+
+<tr id="MediaType"><td><code>MediaType</code>
```
joshuagl, line 314: Shouldn't this field name start with a lower case like the others?
```suggestion
<tr id="mediaType"><td><code>MediaType</code>
```

docs/provenance/v1/index.md
```diff
@@ -0,0 +1,609 @@
...
+[RunDetails]: #rundetails
+
+REQUIRED for SLSA Build L1: `builder` (unless `id` is implicit from the
+attestation envelope)
+
+<table>
+<tr><th>Field<th>Type<th>Description
+
+<tr id="builder"><td><code>builder</code>
+<td><a href="builder">Builder</a><td>
```
joshuagl, line 335:
```suggestion
<td><a href="#builder">Builder</a><td>
```

docs/provenance/v1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+Identifies the entity that executed the invocation, which is trusted to have
+correctly performed the operation and populated this provenance.
+
+<tr id="metadata"><td><code>metadata</code>
+<td><a href="buildmetadata">BuildMetadata</a><td>
```
joshuagl, line 341:
```suggestion
<td><a href="#buildmetadata">BuildMetadata</a><td>
```

docs/provenance/v1/index.md
```diff
@@ -0,0 +1,609 @@
...
+
+Metadata about this particular execution of the build.
+
+<tr id="byproducts"><td><code>byproducts</code>
+<td>array (<a href="artifactreference">ArtifactReference</a>)<td>
```
joshuagl, line 346:
```suggestion
<td>array (<a href="#artifactreference">ArtifactReference</a>)<td>
```

docs/provenance/v1/index.md
```diff
@@ -0,0 +1,609 @@
...
+**TODO:** Provide guidance on how to choose a URI, what scope it should have,
+stability, how [verification] works, etc.
+
+<tr id="builder.version"><td><code>version</code>
+<td>map (string→string)<td>
+
+Version numbers of components of the builder.
+
+<tr id="builderDependencies"><td><code>builderDependencies</code>
+<td>array (<a href="artifactreference">ArtifactReference</a>)<td>
```
joshuagl, line 391:
```suggestion
<td>array (<a href="#artifactreference">ArtifactReference</a>)<td>
```

joshuagl (approved):
This is a fantastic evolution of the provenance format. Thanks for calling out the bits that would especially benefit from feedback with the "RFC" markers.

I fully support merging this as-is and iterating in tree. I have made a few minor suggestions to fix some of the links, otherwise LGTM.

docs/provenance/v1/provenance.cue
```diff
@@ -0,0 +1,55 @@
+{
+    // Standard attestation fields:
+    "_type": "https://in-toto.io/Statement/v0.1",
+    "subject": [...],
+
+    // Predicate:
+    "predicateType": "https://slsa.dev/provenance/v1.0?draft",
```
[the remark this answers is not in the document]
MarkLodato: Oops! Thanks, done.

MarkLodato:
Thanks. I fixed all those issues, I think. I'll do a "squash and merge", but I'll leave this open for a bit for you to verify if you have time.

docs/provenance/v1/index.md
```diff
@@ -480,8 +487,8 @@ validate that the design is general enough to apply to other builders.
 
 ## Migrating from 0.2
 
-To migrate from [version 0.2][0.2] (`old`), use the following pseudocode. The
-meaning of each field is unchanged unless otherwise noted.
+To migrate from [version 0.2](../v0.2.md) (`old`), use the following pseudocode.
```
joshuagl, line 490: Nit: earlier edits in this patch remove the `.md` file suffix from links.

joshuagl (approved):
LGTM

docs/github-actions-workflow/v0.1/example.json
```diff
@@ -0,0 +1,59 @@
...
+                    "uri": "https://github.com/actions/virtual-environments/releases/tag/ubuntu20/20220515.1"
+                }
+            ]
+        },
+        "runDetails": {
+            "builder": {
+                "id": "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_go_slsa3.yml@refs/tags/v0.0.1"
+            },
+            "metadata": {
+                "invocationId": "https://github.com/octocat/hello-world/actions/runs/1536140711/attempts/1",
```
feelepxyz, line 48: Love having the full URL in here!

PR #528, opened 2022-11-03
SLSA v1.0: add Key Principles page

### LINKED ISSUE #106: Explain why we prefer attestations over ACLs
SLSA implicitly prefers provenance attestations instead of carefully configured ACLs. We should explain why. Something like:

Theoretically one could configure ACLs properly such that a CI/CD system *only* pulls from the proper source, signing keys to *only* allow access to the CI/CD system, etc. In practice, that is almost impossible to get right and keep right. There are often over-provisioning, or confused deputy problems, or mistakes.

For these reasons, we prefer attestations of the properties we want to verify.

MarkLodato:
Addresses the following issues for v1.0:

-   Fixes #106.
-   Partially addresses #130. Future PR needs to integrate this concept into levels.md and requirements.md. (I'm waiting for the outstanding PRs for those files to be submitted first.)

Preview: https://deploy-preview-528--slsa.netlify.app/spec/v1.0/principles

### DIFF
--- a/docs/spec/v1.0/principles.md
+++ b/docs/spec/v1.0/principles.md
@@ -1,0 +1,80 @@
+# Key principles
+
+<div class="subtitle">
+
+This page lays out the key guiding principles behind SLSA.
+
+</div>
+
+## Automatically verify artifacts, manually verify systems
+
+One of the main ideas behind SLSA is to manually verify that a few *systems*,
+and then automatically verify all *artifacts* produced by those systems. In this
+way, SLSA is able to scale to support hundreds of thousands of software projects
+with a near-constant amount of central work.
+
+We have no option but to manually verify the behavior of some number of trusted
+*systems*, such as change management, build, and packaging systems. There will
+always be some trusted computing base on which we rely for security. We need to
+analyze and reason about this computing base to convince ourselves that it meets
+our security needs. Such analysis is difficult and expensive. We cannot possibly
+harden and audit thousands of bespoke build systems. We can, however, do this
+for a small number of general purpose systems.
+
+Once we have a set of trusted systems, we can verify the supply chain of
+arbitrarily many *artifacts* by ensuring that artifacts were produced by trusted
+systems. This allows SLSA to scale to entire ecosystems and organizations
+regardless of size.
+
+For example, a security engineer may analyze the architecture and implementation
+of a build system to ensure that it meets the SLSA build requirements. Once the
+analysis is complete, the public keys used by the build system to sign
+provenance can then be "trusted" up to a given SLSA level. Downstream systems
+can then automatically determine that an artifact meets a given level by
+verifying that it has provenance signed by that public key.
+
+### Corollary: Minimize the number of trusted systems
+
+A corollary to the previous principle is to minimize the size of the trusted
+computing base. Every system we trust adds attack surface and increases the need
+for manual security analysis. Where possible:
+
+-   Concentrate trust in shared infrastructure. For example, instead of each
+    team maintaining their own build system, use a shared build system. This way
+    hardening work can be shared across all teams.
+
+-   Remove the need to trust components whenever possible. For example, use
+    end-to-end signing to avoid the need to trust intermediate distribution
+    systems.
+
+## Place trust in code, not individuals
+
+Another idea behind SLSA is to place our trust in source code rather than
+individual maintainers who have write access to package registries. Code is
+static and analyzable. People, on the other hand, are prone to mistakes,
+credential compromise, and sometimes malicious action. Therefore, SLSA aims to
+securely trace all software back to source code, without the possibility for an
+individual maintainer to tamper with it after it has been committed.
+
+## Prefer explicit attestations over inferences from configuration
+
+SLSA places a strong emphasis on explicit attestations about security
+properties, particularly an artifact's provenance, rather than inferring
+security properties from how a system is configured.
+
+Theoretically one could configure access control such that the only path from
+source to release is through the official channels: the CI/CD system only pulls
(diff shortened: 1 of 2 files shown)


docs/spec/v1.0/principles.md
```diff
@@ -0,0 +1,80 @@
+# Key principles
+
+<div class="subtitle">
+
+This page lays out the key guiding principles behind SLSA.
+
+</div>
+
+## Automatically verify artifacts, manually verify systems
```
MarkLodato, line 9: Alternative title: "Trust systems, [automatically] verify artifacts" (With or without "automatically".)

Any thoughts? That's a bit shorter and might better convey the point.

docs/spec/v1.0/principles.md
```diff
@@ -0,0 +1,80 @@
...
+
+One of the main ideas behind SLSA is to manually verify that a few *systems*,
```
joshuagl, line 11:
```suggestion
One of the main ideas behind SLSA is to manually verify a few *systems*,
```

docs/spec/v1.0/principles.md
```diff
@@ -0,0 +1,80 @@
...
+### Corollary: Minimize the number of trusted systems
+
+A corollary to the previous principle is to minimize the size of the trusted
+computing base. Every system we trust adds attack surface and increases the need
+for manual security analysis. Where possible:
+
+-   Concentrate trust in shared infrastructure. For example, instead of each
+    team maintaining their own build system, use a shared build system. This way
+    hardening work can be shared across all teams.
+
```
joshuagl, line 45:
```suggestion
```
There's a lot of whitespace between these bullets in the preview:
Screenshot 2022-11-04 at 15 18 43

docs/spec/v1.0/principles.md
```diff
@@ -0,0 +1,80 @@
+# Key principles
+
+<div class="subtitle">
+
+This page lays out the key guiding principles behind SLSA.
```
joshuagl, line 5: I think it might be worth enumerating the key principles at the top of the page and expanding on them further down, something like:
```suggestion
The key guiding principles behind SLSA are:

- Trust systems, verify artifacts
- Place trust in code, not individuals 
- Prefer explicit attestations over inferences from configuration

These principles are explained below.
```
It somewhat duplicates the navigation bar but may provide a smoother reading experience. WDYT?

docs/spec/v1.0/principles.md
joshuagl: I much prefer the shorter, alternative title. I would go with "Trust systems, verify artifacts".

docs/spec/v1.0/principles.md
```diff
@@ -0,0 +1,80 @@
...
+
+SLSA places a strong emphasis on explicit attestations about security
+properties, particularly an artifact's provenance, rather than inferring
+security properties from how a system is configured.
+
+Theoretically one could configure access control such that the only path from
+source to release is through the official channels: the CI/CD system only pulls
+from the proper source, package registry only allows access to the CI/CD system,
+and so on. In practice, that is almost impossible to get right and keep right.
+There are often over-provisioning, or confused deputy problems, or mistakes.
```
joshuagl, line 69:
```suggestion
There are often over-provisioning, confused deputy problems, or mistakes.
```
?

docs/spec/v1.0/principles.md
```diff
@@ -0,0 +1,80 @@
...
+Even if it is configured properly at one moment, there is no guarantee that it
+will continue to stay so configured. And humans almost always end up getting in
```
joshuagl, line 71:
```suggestion
will continue to stay so. And humans almost always end up getting in
```

docs/spec/v1.0/principles.md
```diff
@@ -0,0 +1,80 @@
...
+the access control lists.
+
+Instead, SLSA prefers explicit attestations about an artifact's provenance.
+Access control is still important, but SLSA provides a defense in depth: it is
+not sufficient to be granted access to publish a package, but one must also have
```
joshuagl, line 76:
```suggestion
not sufficient to be granted access to publish a package, one must also have
```

joshuagl (commented):
This is great!

docs/spec/v1.0/principles.md
MarkLodato: Personally I find that less readable, and it's also a bit harder to maintain (easy to get out-of-sync).

Thoughts from other reviewers?

docs/spec/v1.0/principles.md
MarkLodato: It's the same as the regular paragraph spacing. I switched to "tight" lists - does that look better to you?

MarkLodato (commented):
Thanks for the review!

docs/spec/v1.0/principles.md
joshuagl: Yes, looks much better. Thanks.

docs/spec/v1.0/principles.md
joshuagl: LGTM!

docs/spec/v1.0/principles.md
joshuagl: I get it. The current layout looks a bit awkward with a single short sentence followed by a large white space, but perhaps we can do something about that later? I'm approving the PR as-is.

joshuagl (approved):

MarkLodato:
Thanks for the review!

docs/spec/v1.0/principles.md
MarkLodato: Yup, let's try to improve it in a future PR.

PR #529, opened 2022-11-03
SLSA v1.0: frame levels around tamper protection

### LINKED ISSUE #504: SLSA v1.0: reproducible builds should satisfy SLSA Build Levels
The current wording states that a "hosted build service" is required. One way to implement this would be to have a system of reproducible builds where the original build happens anywhere and then some (hosted) rebuilders reproduce the build bit-for-bit and generate their own provenance. I want to make sure that the wording of the requirements makes it clear that this is an acceptable implementation. Right now I don't think that is clear.

MarkLodato:
Reframe the meaning of each level in terms of how much tamper protection they afford, rather than prescribing specific implementation. In concrete terms, this means Build L1 = mistake prevention, L2 = weak tamper protection (fireable offense), L3 = strong tamper protection. This framing seems to resonate well with various parties I've talked to.

In particular, this makes it more clear that reproducible builds or other types of systems can still satisfy SLSA.

Fixes #504.

### DIFF
--- a/docs/spec/v1.0/levels.md
+++ b/docs/spec/v1.0/levels.md
@@ -96,15 +96,19 @@ Summary of the build track:
 
 -   Set **project-specific expectations** for how the package should be built.
 -   Generate a **provenance attestation** automatically during each build.
--   Build on **hardened services** that have been manually audited.
 -   **Automatically verify** that each package's provenance meets expectations
     before allowing its publication and/or consumption.
 
-Exactly how this is implemented is up to the packaging ecosystem (for open
-source) or organization (for closed source), including: means of defining
-expectations, what provenance format is accepted, whether reproducible builds
-are used, how provenance is distributed, when verification happens, and what
-happens on failure. Guidelines for implementers can be found in the
+What sets the levels apart is how much trust there is in the accuracy of the
+provenance and the degree to which adversaries are detected or prevented from
+tampering with the package. Higher levels require hardened builds and protection
+against more sophisticated adversaries.
+
+Each ecosystem (for open source) or organization (for closed source) defines
+exactly how this is implemented, including: means of defining expectations, what
+provenance format is accepted, whether reproducible builds are used, how
+provenance is distributed, when verification happens, and what happens on
+failure. Guidelines for implementers can be found in the
 [requirements](requirements.md).
 
 <section id="build-l0">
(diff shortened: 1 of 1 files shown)


docs/spec/v1.0/levels.md
```diff
@@ -138,8 +142,9 @@ n/a
 <dl class="as-table">
 <dt>Summary<dd>
 
-Package has a provenance attestation showing that it was built as expected, but
-the provenance is trivial to forge.
+Package has a provenance attestation showing how it was built, and a downstream
+system automatically verifies that packages were built as expected. Prevents
+mistakes but is trivial to bypass or forge.
```
joshuagl, line 147: I suggest we use produced, not built, here. [Provenance v0.2](https://slsa.dev/provenance/v0.2) describes itself as "where, when and how something was produced".
```suggestion
Package has a provenance attestation showing how it was produced, and a downstream
system automatically verifies that packages were produced as expected. Prevents
mistakes but is trivial to bypass or forge.
```

docs/spec/v1.0/levels.md
```diff
@@ -203,8 +212,10 @@ service itself required by [Build L3].
 
 All of [Build L1], plus:
 
--   Builds run on a hosted build service that generates and signs the
-    provenance itself.
+-   The build runs on a hosted build service that generates and signs the
+    provenance itself. This may be the original build, an after-the-fact
+    reproducible build, or some equivalent system that ensures the accuracy of
+    the provenance.
```
joshuagl, line 218: Is it worth being a little more explicit about what accurate provenance means? I'm not sure, but maybe:
```suggestion
    reproducible build, or some equivalent system that ensures the provenance
    accurately describes the produced artifacts.
```

joshuagl (approved):
Solid improvements here, thanks!

docs/spec/v1.0/levels.md
MarkLodato: We use "build" throughout this doc. Is it odd to change it here but not elsewhere? Should we instead change the provenance spec to match this language? Or if we change it here, do you have a suggestion on when to use build vs produce?

docs/spec/v1.0/levels.md
MarkLodato: Thanks for calling this out. I didn't like "accurate" either. What about "trustworthiness" of the provenance?

For context, I added this clause because I thought there might be some other way to satisfy the requirements that doesn't neatly fit into "original build" or "reproducible build".

PR #530, opened 2022-11-04
css: distinguish multi-line entries in ToC

MarkLodato:
Previously the table-of-contents did not have any visual distinction between two different entries vs one entry with multiple lines - they all ran together.

Now there is slightly more whitespace between entries and slightly less whitespace between lines within an entry.

before | after
--- | ---
before | after
before | after
before | after


joshuagl (approved):
Looks much better in the attached screenshots, thanks.

PR #531, opened 2022-11-07
Address unresolved review comment in #529

joshuagl:
The additional requirements at build L2 help improve trustworthiness, not accuracy per discussion 
https://github.com/slsa-framework/slsa/pull/529/commits/d3b931f366ef2e5d765f3a17c02c94d979c35938#r1014411226


MarkLodato (approved):
Oops! Thanks for the fix!

Issue #321, opened 2022-03-15
How strong of a guarantee is required for Isolated and Ephemeral Environment at SLSA 3?


MarkLodato:
A major difference between SLSA 2 and 3 is that the build steps run in an [Isolated], [Ephemeral] environment. But how strong of a guarantee do we need to make to call something SLSA 3, given that network access is still allowed and thus some build steps could execute remotely? This can be solved at SLSA 4 by requiring a [Hermetic] environment with no network access, but at SLSA 3 it seems impossible to make a strong claim in general. Thus we need to figure out what is "good enough" for SLSA 3.

Motivation: GitHub Actions. A [GitHub-hosted runner] executes each job within an ephemeral, isolated VM, which would normally meet the corresponding SLSA 3 requirements. But there are several ways this can fail:

* The workflow may use a self-hosted runner for some of its jobs, which GitHub cannot make any guarantees about.
* More generally, build steps may call remote execution services, such as Google Cloud Build, which again GitHub cannot make any guarantees about.
* Build steps may open services that unintentionally allow influence from other jobs, breaking the Isolated guarantee.

In the face of this, if we have a non-falsifiable provenance that an artifact really was produced by a particular GitHub Actions workflow run, can we call it SLSA 3? Do we need to disallow self-hosted runners, for example? In general, how do we model this and what are the specific requirements?

I'm inclined to say that these requirements are "best effort" at SLSA 3, similar to the existing language for Hermetic. The controls and expectations should be clear such that a "normal" build is isolated and hermetic, and only a user doing something out of the ordinary would violate it.

So in GitHub Action's case, I'm leaning towards:

* Disallow self-hosted runners.
* Instruct users not to call out to remote execution.
* Instruct users not to open services that allow remote influence.

Any thoughts?

/cc @laurentsimon @asraa

[Isolated]: https://slsa.dev/spec/v0.1/requirements#isolated
[Ephemeral]: https://slsa.dev/spec/v0.1/requirements#ephemeral-environment
[Hermetic]: https://slsa.dev/spec/v0.1/requirements#hermetic
[GitHub-hosted runner]: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

MarkLodato:
This was presented at today's meeting (2021-03-17).

joshuagl:
Agree that we should label these as best effort at SLSA level 3. 

> So in GitHub Action's case, I'm leaning towards:
> 
> Disallow self-hosted runners.
[...]

PR #532, opened 2022-11-11
SLSA v1.0: Combine requirements- ephemeral and isolated

abacchi:
Related to issue https://github.com/slsa-framework/slsa/issues/321
Looking at combining ephemeral and isolated l3 req into one. 
Appended two "Should not" items as recommendations (not requirements)

### DIFF
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -115,7 +115,7 @@ instances, whether prior or concurrent.
 -   It MUST NOT be possible for two builds that overlap in time to influence one another.
 -   It MUST NOT be possible for one build to persist or influence the build environment of a subsequent build.
 -   Build caches, if used, MUST be purely content-addressable to prevent tampering.
--   The build SHOULD NOT call out to remote execution.
+-   The build SHOULD NOT call out to remote execution unless it's considered part of the "builder" within the trust boundary.
 -   The build SHOULD NOT open services that allow for remote influence.
 
 <td> <td> <td>✓
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -38,8 +38,7 @@ showing what the options are:
 | Build - [Scripted build]         | ✓        | ✓        | ✓        |
 | Build - [Build service]          |          | ✓        | ✓        |
 | Build - [Build as code]          |          |          | ✓        |
-| Build - [Ephemeral environment]  |          |          | ✓        |
-| Build - [Isolated]               |          |          | ✓        |
+| Build - [Ephemeral and isolated]               |          |          | ✓        |
 | Provenance - [Available]         | ✓        | ✓        | ✓        |
 | Provenance - [Authenticated]     |          | ✓        | ✓        |
 | Provenance - [Service generated] |          | ✓        | ✓        |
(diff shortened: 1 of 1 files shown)


docs/spec/v1.0/requirements.md
```diff
@@ -105,26 +103,20 @@ Examples:
...
-free of influence from other build instances, whether prior or concurrent.
+The build service ensured that the build steps ran in an ephemeral and isolated
+environment provisioned solely for this build, free of influence from other build
+instances, whether prior or concurrent.
 
 -   It MUST NOT be possible for a build to access any secrets of the build service, such as the provenance signing key.
 -   It MUST NOT be possible for two builds that overlap in time to influence one another.
 -   It MUST NOT be possible for one build to persist or influence the build environment of a subsequent build.
 -   Build caches, if used, MUST be purely content-addressable to prevent tampering.
+-   The build SHOULD NOT call out to remote execution.
```
MarkLodato, line 118: I suspect this may be too strict. Some build services use remote execution internally and that's not a problem, so long as it's considered part of the trust boundary of the builder. So maybe something like:

> Remote execution services, if used, SHOULD be considered as part of the "builder" and thus within the trust boundary.

That might be confusing though...

MarkLodato (commented):
Overall merging SGTM.

docs/spec/v1.0/requirements.md
abacchi: What about something like:

> The build SHOULD NOT call out to remote execution unless it's considered part of the "builder" within the trust boundary.

MarkLodato: 👍 yes, that's much better!

docs/spec/v1.0/requirements.md
```diff
@@ -105,26 +103,20 @@ Examples:
...
+The build service ensured that the build steps ran in an ephemeral and isolated
+environment provisioned solely for this build, free of influence from other build
+instances, whether prior or concurrent.
 
 -   It MUST NOT be possible for a build to access any secrets of the build service, such as the provenance signing key.
 -   It MUST NOT be possible for two builds that overlap in time to influence one another.
 -   It MUST NOT be possible for one build to persist or influence the build environment of a subsequent build.
 -   Build caches, if used, MUST be purely content-addressable to prevent tampering.
+-   The build SHOULD NOT call out to remote execution unless it's considered part of the "builder" within the trust boundary.
+-   The build SHOULD NOT open services that allow for remote influence.
```
marcelamelara, line 119: I wonder if we can make these last two recommendations stronger (and maybe combine them?) by stating something like "The build MUST explicitly define allowed incoming and outgoing network connections"

MarkLodato (approved):

joshuagl (approved):
Thanks for working on this! This is an improvement and moves 1.0 in the right direction, so I am in favour of merging and iterating and improving in-tree.

docs/spec/v1.0/requirements.md
MarkLodato: I'm not sure that this is feasible for most builders. I suggest that we merge as-is and then perhaps file a separate issue if there is something further to tweak?

MarkLodato:
Agreed, merging now.

docs/spec/v1.0/requirements.md
marcelamelara: I see your point. Maybe we can reconsider this for L4.

PR #552, opened 2022-12-16
Add status section to specifications

lehors:
This change makes use of the specification stages that have been defined to predominently display the status of the specification on every page, with a pointer to the definition and a link to the superseding version if there is one.

### DIFF
--- a/docs/_data/versions.yml
+++ b/docs/_data/versions.yml
@@ -16,6 +16,7 @@ spec:
   versions:
     v0.1:
       name: Version 0.1
+      status: Stable
     v1.0:
       name: Version 1.0 (DRAFT)
       status: Working Draft
--- a/docs/_data/versions.yml
+++ b/docs/_data/versions.yml
@@ -39,7 +39,7 @@ verification_summary:
   versions:
     v0.1:
       name: Version 0.1
-      status: Working Draft
+      status: Retired
     v0.2:
       name: Version 0.2
       status: Working Draft
--- a/docs/_data/versions.yml
+++ b/docs/_data/versions.yml
@@ -3,6 +3,7 @@
 #       versions:
 #         <version>:
 #           name: <string> # display name
+#           status: <string> # one of: Working Draft, In Review, Stable, Retired
 #           draft: <bool, optional> # if true, display a "draft" warning banner
 #           hidden: <bool, optional> # if true, do not show in dropdown
 #       current: <version>
@@ -17,6 +18,7 @@ spec:
       name: Version 0.1
     v1.0:
       name: Version 1.0 (DRAFT)
+      status: Working Draft
       draft: true
       # We will hide 1.0 until it is ready for review. Viewers can still type
       # the URL manually, but we will not link to it from the main page.
@@ -27,14 +29,18 @@ provenance:
   versions:
     v0.1:
       name: Version 0.1
+      status: Working Draft
     v0.2:
       name: Version 0.2
+      status: Working Draft
   current: v0.2
 
 verification_summary:
   versions:
     v0.1:
       name: Version 0.1
+      status: Working Draft
     v0.2:
       name: Version 0.2
+      status: Working Draft
   current: v0.2
--- a/docs/_data/versions.yml
+++ b/docs/_data/versions.yml
@@ -29,7 +29,7 @@ provenance:
   versions:
     v0.1:
       name: Version 0.1
-      status: Working Draft
+      status: Retired
     v0.2:
       name: Version 0.2
       status: Working Draft
--- a/docs/_data/versions.yml
+++ b/docs/_data/versions.yml
@@ -33,7 +33,7 @@ provenance:
       status: Retired
     v0.2:
       name: Version 0.2
-      status: Working Draft
+      status: Stable
   current: v0.2
 
 verification_summary:
@@ -43,5 +43,5 @@ verification_summary:
       status: Retired
     v0.2:
       name: Version 0.2
-      status: Working Draft
+      status: Stable
   current: v0.2
--- a/docs/_includes/status.html
+++ b/docs/_includes/status.html
@@ -10,7 +10,7 @@
 {%- if page.layout == "specifications" %}
 <div class="rounded-lg p-4 border border-gray-400 mt-8 md:mt-0 mb-4">
 {%- else %} <!-- page.layout == "standard" -->
-<div class="rounded-lg p-4 border border-green-900 mt-8 md:mt-0 mb-4">
+<div class="rounded-lg p-4 border border-green-900 mb-4">
 {%- endif %}
 
   Status: <a href="/spec-stages" style="display: inline">{{ site.data.versions[spec_name].versions[spec_version].status }}</a>
--- a/docs/_includes/status.html
+++ b/docs/_includes/status.html
@@ -7,12 +7,14 @@
 
 {%- if site.data.versions[spec_name].versions[spec_version].status %}
 
-## Status: {{ site.data.versions[spec_name].versions[spec_version].status }}
-
-{%- if spec_version_num < current_version_num %}
-This draft has been superseded. [View the latest version]({{ page.url | replace: spec_version, site.data.versions[spec_name].current | relative_url}}).
+{%- if page.layout == "specifications" %}
+<div class="rounded-lg p-4 border border-gray-400 mt-8 md:mt-0 mb-4">
+{%- else %} <!-- page.layout == "standard" -->
+<div class="rounded-lg p-4 border border-green-900 mt-8 md:mt-0 mb-4">
 {%- endif %}
 
-For more information regarding the meaning of this status, see the [Specification Stages definition]({{ "/spec-stages" | relative_url }}).
+  Status: <a href="/spec-stages" style="display: inline">{{ site.data.versions[spec_name].versions[spec_version].status }}</a>
+
+</div>
 
 {%- endif %}
--- a/docs/_includes/status.html
+++ b/docs/_includes/status.html
@@ -8,9 +8,9 @@
 {%- if site.data.versions[spec_name].versions[spec_version].status %}
 
 {%- if page.layout == "specifications" %}
-<div class="rounded-lg p-4 border border-gray-400 mt-8 md:mt-0 mb-4">
+<div class="rounded-lg p-4 border border-gray-400 mb-4">
 {%- else %} <!-- page.layout == "standard" -->
-<div class="rounded-lg p-4 border border-green-900 mb-4">
+<div class="rounded-lg p-4 border border-green-900 mt-4 md:mt-0 mb-4">
 {%- endif %}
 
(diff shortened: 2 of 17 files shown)


lehors:
I think it's worth pointing out that I didn't remove the floating banner that pops up over the slsa v1.0 spec but I'm actually not sure this is still needed.
Also I didn't set a status for the v0.1 spec because I'm not sure how you would want to label it. For this reason, the status section is optional and will only get displayed if a status is defined. I'm happy to change any of that if something else is preferred.

docs/_data/versions.yml
```diff
@@ -27,14 +29,18 @@ provenance:
   versions:
     v0.1:
       name: Version 0.1
+      status: Working Draft
```
MarkLodato, line 32:
```suggestion
      status: Retired
```

docs/_data/versions.yml
```diff
@@ -27,14 +29,18 @@ provenance:
...
     v0.2:
       name: Version 0.2
+      status: Working Draft
   current: v0.2
 
 verification_summary:
   versions:
     v0.1:
       name: Version 0.1
+      status: Working Draft
```
MarkLodato, line 42:
```suggestion
      status: Retired
```

docs/_data/versions.yml
```diff
@@ -27,14 +29,18 @@ provenance:
...
     v0.2:
       name: Version 0.2
+      status: Working Draft
```
MarkLodato, line 45:
```suggestion
      status: Stable
```

docs/_includes/status.md
```diff
@@ -0,0 +1,19 @@
+{%- assign url_parts = page.url | split: '/' %}
+{%- assign spec_name = url_parts[1] %}
+{%- assign spec_version = url_parts[2] %}
+{%- assign spec_version_num = spec_version | remove: 'v' %}
+{%- assign spec_version_num = spec_version | remove: 'v' %}
```
MarkLodato, line 5: Remove duplicate line

```suggestion
{%- assign spec_version_num = spec_version | remove: 'v' %}
```

docs/provenance/v0.2.md
```diff
@@ -3,6 +3,9 @@ title: Provenance
 layout: standard
 hero_text: To trace software back to the source and define the moving parts in a complex supply chain, provenance needs to be there from the very beginning. It’s the verifiable information about software artifacts describing where, when and how something was produced. For higher SLSA levels and more resilient integrity guarantees, provenance requirements are stricter and need a deeper, more technical understanding of the predicate.
 ---
+
+{% include status.md %}
```
MarkLodato, line 7: Instead of adding this to every page, what do you think about adding it to the version picker template? Then it would show up on all pages (including the sub-pages of /spec/v.../). Here's a mock:

image

docs/_data/versions.yml
```diff
@@ -17,6 +18,7 @@ spec:
       name: Version 0.1
```
MarkLodato, line 18: I'd call this stable, no?

```suggestion
      name: Version 0.1
      status: Stable
```

docs/_includes/status.md
```diff
@@ -0,0 +1,19 @@
...
+
+{%- if site.data.versions[spec_name].versions[spec_version].status %}
+
+## Status: {{ site.data.versions[spec_name].versions[spec_version].status }}
+
+{%- if spec_version_num < current_version_num %}
+This draft has been superseded. [View the latest version]({{ page.url | replace: spec_version, site.data.versions[spec_name].current | relative_url}}).
+{%- endif %}
+
+For more information regarding the meaning of this status, see the [Specification Stages definition](../../spec-stages).
```
MarkLodato, line 17: suggestion: use an absolute path so that it works from all pages.

(I think the suggested diff works but I haven't tested it.)

```suggestion
For more information regarding the meaning of this status, see the [Specification Stages definition]({{ "/spec-stages" | relative_url }}).
```

MarkLodato (commented):
Thanks! I think we can mark things as Stable/Retired, not just Working Draft, no?

docs/_data/versions.yml
```diff
@@ -27,14 +29,18 @@ provenance:
   versions:
     v0.1:
       name: Version 0.1
+      status: Working Draft
     v0.2:
       name: Version 0.2
+      status: Working Draft
```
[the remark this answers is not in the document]
lehors: > Thanks! I think we can mark things as Stable/Retired, not just Working Draft, no?

Yes, however it seems odd to me to have a v0.x marked as stable. The use of 0.x implies it's not really finished and something else is expected to come next, so labeling it stable seems contradictory to me.

I could understand this being done for the SLSA v0.1 though given that it has been published for a while and is already used in production (for better and for worse).

But it's all a matter of convention so anything is possible. :-)

docs/_data/versions.yml
lehors: As said earlier, I think that's reasonable given that it is in use.

docs/provenance/v0.2.md
lehors: I quite like the look of this! The downside is that it doesn't provide for space to add any additional text but that could still be added in the content if necessary.
Let me see if I can produce that output.

lehors: So, it took me a bit of time because of the varying templates and styles but the latest commit implements this layout. I like it.

docs/_data/versions.yml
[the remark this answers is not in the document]
MarkLodato: Yeah, I thought the same thing. Maybe "Latest" would be a better word than "Stable"? (If that's the case, then "Outdated" might be better than "Retired?") Either way, the current state does not match "Working draft", which is the stage *before* "In review".

docs/_includes/status.html
```diff
@@ -0,0 +1,20 @@
...
+{%- assign spec_version_num = spec_version | remove: 'v' %}
+{%- assign current_version = site.data.versions[spec_name].current %}
+{%- assign current_version_num = current_version | remove: 'v' %}
+
+{%- if site.data.versions[spec_name].versions[spec_version].status %}
+
+{%- if page.layout == "specifications" %}
+<div class="rounded-lg p-4 border border-gray-400 mt-8 md:mt-0 mb-4">
+{%- else %} <!-- page.layout == "standard" -->
+<div class="rounded-lg p-4 border border-green-900 mt-8 md:mt-0 mb-4">
```
MarkLodato, line 13: No need for the `mt` thing - it just adds extra space on mobile, which looks bad.

```suggestion
<div class="rounded-lg p-4 border border-green-900 mb-4">
```

docs/_layouts/specifications.html
```diff
@@ -27,7 +27,8 @@ <h1 class="pr-32 pb-16 pt-32">{{ page.title }}</h1>
                     </div>
                     <div class="table-of-contents w-full md:w-1/3 md:sticky md:top-32 pb-12 md:pl-12">
                       {% include versions-dropdown.html %}
-                        <div class="rounded-lg p-4 border border-gray-400">
+                      {% include status.html %}
+                        <div class="rounded-lg p-4 border border-gray-400 mt-4">
```
MarkLodato, line 31: No need for the `mt-4`. Can revert to the original version.

```suggestion
                        <div class="rounded-lg p-4 border border-gray-400">
```

MarkLodato (commented):
I like it!

docs/provenance/v0.2.md
MarkLodato: 👍

docs/_data/versions.yml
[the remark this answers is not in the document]
lehors: Ok, I just realized that we have a different mental model because while we agreed on the different stages we/I failed to define the lifecycle - the order in which a spec is expected to go through the different stages.

I take it that you are thinking of something like the following, which we would just keep repeating:

WD -> In Review -> Stable -> Retired

Is that right?

For what it's worth, my initial mental model was:

- Normal path: WD -> WD -> ... -> WD -> In Review -> Stable -> Retired
- Alternatively we could go back to WD after In Review in case of major changes we are trying to work out or have another In Review (or more) before finally reaching Stable and eventually Retired.
- Another alternative is to go from WD to Retired in case a spec is abandoned.

Admittedly this is highly influenced by the W3C Recommendation track (although simpler). One key aspect is that in this model the "In Review" stage is meant to be an invitation for review by a broader audience than the WG which is expected to do so on an on-going basis against the WDs.

docs/_includes/status.html
lehors: Ah. Good to know what these are about. Thanks.

docs/_data/versions.yml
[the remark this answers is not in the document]
MarkLodato: Sorry, I don't understand the difference. I agree with everything you wrote under "my initial mental model was", so it must be something else.

Perhaps it's an issue of versioning? In my mental model, each version (0.1, 0.2, 1.0, 1.1, 1.2, etc.) undergoes one of those paths. In practice, at any given time, we will have:

- The latest released version is called "Stable" (or "Latest"). Right now this is v0.1 for the levels spec and v0.2 for provenance and VSA.
- All older versions are marked "Retired" (or "Outdated"). Right now this is v0.1 for provenance and VSA.
- If there is an upcoming version being developed, it is called "Working Draft" while it is still undergoing heavy edits, then "In Review" during the community review period. Right now this is v1.0 for the levels spec and provenance.

Once the new version (v1.0) is marked stable, the previous stable version (v0.1/v0.2) is marked as retired.

Is that what you were thinking?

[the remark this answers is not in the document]
lehors: So, having thought about this some more I realized that the problem comes from the overlap we have with what is being conveyed by the status and by the version number.

Specifically, the current definition for "stable" reads: "It is suitable for reference and implementation in production." And it doesn't seem right to me to label a 0.x version as something suitable for production.

The solution is simply to remove this piece from the definition of "stable" and simply leave it as:

"At this stage the document is considered stable. Changes may only happen if deemed necessary and will be avoided as much as possible if they might adversely impact implementations."

This way, the status doesn't interfere with the version number. Whether a given spec is suitable for production can be conveyed through its version number alone.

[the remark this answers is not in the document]
MarkLodato: Ah, yes, good catch! That change sounds good to me!

In that case, should we also remove the "Retired" stage? Then we'd have: Working Draft -> In Review -> Stable.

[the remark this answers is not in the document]
lehors: No, I think we should keep Retired because it's useful to communicate that a spec is no longer maintain. The version number alone can't do that.
But so, I'll update the definition of the stable stage and accept the rest of the changes you suggested on this PR. This should get everything straight.
Thanks.

lehors:
@MarkLodato with PR #561 merged, I think this one is ready for merging.

MarkLodato (approved):
Thanks!

PR #553, opened 2022-12-16
Update requirements.md

Pierre-Gronau-ndaal:
The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14),  [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119),  [RFC 8174](https://tools.ietf.org/html/rfc8174) when, and only when,  they appear in all capitals, as shown here.

Signed-off-by: Pierre-Gronau-ndaal <72132223+Pierre-Gronau-ndaal@users.noreply.github.com>

### DIFF
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -15,7 +15,10 @@ mitigations](threats.md).
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
-interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
+interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14), 
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119), 
+[RFC 8174](https://tools.ietf.org/html/rfc8174) when, and only when, 
+they appear in all capitals, as shown here.
 
 ## TODO
 


docs/spec/v1.0/requirements.md
```diff
@@ -15,7 +15,10 @@ mitigations](threats.md).
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
-interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
+interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14), 
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119), 
+[RFC 8174](https://tools.ietf.org/html/rfc8174) when, and only when, 
+they appear in all capitals, as shown here.
```
MarkLodato, line 21: Thanks for the PR, but I find the update wording to be harder to read (much too wordy) and follow (caps vs not is a subtle distinction).

Instead, what do you think about having a lint test to disallow any use of any use of non-capitalized keywords on any page mentioning "RFC 2119"? That would avoid ambiguity without adding extra rules for the reader.

This could then suggest the author either:

- Make the word capitalized if it follows RFC 2119 meaning.
- Choose an alternate word if that meaning is not intended, such as using present tense instead of "must" or "should", or using "might" or "could" instead of "may".

We should also clean up existing uses.

Pierre-Gronau-ndaal: This document aims to reduce the ambiguity by clarifying that only UPPERCASE usage of the key words have the defined special meanings.

Pierre-Gronau-ndaal: all together:

The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14), 
[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119), 
[RFC 8174](https://tools.ietf.org/html/rfc8174). 
This document aims to reduce the ambiguity by clarifying that only UPPERCASE usage of the key words have the defined special meanings.

MarkLodato: See #555 for what I meant by an alternate suggestion. If we never use lowercase, there's no need to be more specific in this blurb here.

If we do keep the present PR, I suggest removing the other two links and just linking to RFC 2119. That document is the only one defining the keywords. The others aren't really useful for the reader.

Pierre-Gronau-ndaal: maybe we SHOULD follow the more actual RFC represented by RFC 8174 in combination of ... and then we use it in the text

di:
Is this obsoleted by #555? Should we close it?

MarkLodato:
Yes, I'll close it. Thanks.

PR #555, opened 2022-12-19
Only use caps for RFC 2119 keywords

MarkLodato:
Usage of lowercase RFC 2119 keywords is ambiguous. PR #553 suggestes switching to [RFC 8174](https://www.rfc-editor.org/rfc/rfc8174) phrasing, which says that only uppercase words have that meaning. Personally I find that to be more work for the reader, and it also means that authors may accidentally use lowercase keywords when they should have been uppercase.

Instead, this PR fixes our specs so that we only ever use those keywords in uppercase, and adds a lint test to prevent regressions. In some cases this means switching to uppercase, while in other cases it means using an alternate phrase. The downside to this approach is that sometimes the alternate phrasing is a bit awkward, but overall this might be the best compromise.

/cc @Pierre-Gronau-ndaal (author of #553) what do you think?


Pierre-Gronau-ndaal:
if this clarification of UPPERCASE is not accepted then a need of UPPERCASE will not  arise. It is important to keep the two others because they use the tweak for RC 2119. Just to say it  in other words only 2119 and 8174 together you have  the actual interpretation what  the RFC is all about. Please keep my proposal The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
interpreted as described in [BCP 14](https://tools.ietf.org/html/bcp14), [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119), [RFC 8174](https://tools.ietf.org/html/rfc8174).
This document aims to reduce the ambiguity by clarifying that only UPPERCASE usage of the key words have the defined special meanings.
The consequence is later on to update some text if required

abacchi:
Should we name `lint.sh` something more descriptive like `rfc2119-lint.sh` ?

lehors (approved):
LGTM. @Pierre-Gronau-ndaal 's suggestion is good but if you prefer this lighter approach I think that's reasonable. This is what a lot of people did during the 20 years in between RFC2119 and RFC8174. :-)

MarkLodato:
The lint check works whether or not we accept #553.

PR #558, opened 2022-12-28
Fix relative path to example.md

s-spindler:
The link currently returns 404.


abacchi:
Good find- I do see that it is a 404 in prod right now: https://slsa.dev/spec/example.md#vision-case-study

Am I missing something with your edit? It seems to also lead to 404 - https://deploy-preview-558--slsa.netlify.app/example.md#vision-case-study

s-spindler:
Weird. I can't find other places in the repo where one md file links to another outside of the same directory. This might not be how jekyll works, though. I found this: https://jekyllrb.com/docs/liquid/tags/#links

I'll try those...

s-spindler:
Yea, that wasn't it... One more try - if it doesn't work I'll close here, open an issue and let someone take over that actually knows jekyll.

s-spindler:
Well, it works, but no idea if this is how you want to structure the site. Feel free to close if it doesn't make sense.

lehors:
I agree the link should be fixed but I think we should keep the file outside a version specific folder as long as it applies to all versions.

s-spindler:
> I went ahead and submitted PR #604.

Sorry, I didn't get around to replying earlier. Thanks for fixing it!

PR #561, opened 2023-01-06
Fix definition of Stable stage

lehors:
Remove reference to use in production left to the version number in line with discussion on PR #552 at: https://github.com/slsa-framework/slsa/pull/552/files/a040cafa72fed25b718f52a0b288d552868faae7#r1059504377

### DIFF
--- a/docs/spec-stages.md
+++ b/docs/spec-stages.md
@@ -35,9 +35,10 @@ anymore.
 
 ## Stable
 
-At this stage the document is considered stable. Changes may only
-happen if deemed necessary and will be avoided as much as possible if
-they might adversely impact implementations.
+At this stage the document is considered stable. No changes that would
+constitute a significant departure from the existing specification are
+expected although changes to address ambiguities and edge cases may
+still occur.
 
 ## Retired
 
--- a/docs/spec-stages.md
+++ b/docs/spec-stages.md
@@ -35,10 +35,9 @@ anymore.
 
 ## Stable
 
-At this stage the document is considered stable. It is suitable for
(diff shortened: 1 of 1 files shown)


docs/spec-stages.md
```diff
@@ -35,10 +35,9 @@ anymore.
 
 ## Stable
 
-At this stage the document is considered stable. It is suitable for
-reference and implementation in production. Changes may only happen if
-deemed necessary and will be avoided as much as possible if they might
-adversely impact implementations.
+At this stage the document is considered stable. Changes may only
+happen if deemed necessary and will be avoided as much as possible if
+they might adversely impact implementations.
```
MarkLodato, line 40: While we're at it, I wonder if we should soften the second sentence. I'm not sure that we intend to "avoid [changes] as much as possible" but rather that we guarantee that the *meaning* of the spec won't change. We do plan to continue to revise the specification's wording to clarify or address edge cases. I don't think we plan to do incremental point releases for every change.

lehors: I agree the wording is a bit too strong given the other change. Formally speaking I think we could refer to "normative changes" and "editorial changes" that respectively have bearings and no bearings on compliance but to keep it light I think we could simply make this something like:
"At this stage the document is considered stable. No changes that would constitute a significant departure from the existing specification are expected although changes to address ambiguities and edges cases may still occur."

MarkLodato: That wording LGTM

lehors: Done

MarkLodato (approved):

PR #565, opened 2023-01-09
SLSA v1.0: Rename "Key principles" to "Guiding principles"

MarkLodato:
The new name better reflects that it is background material rather than an important document that ever reader must read.

Also explain the intent of the page at the top so that readers can quickly determine whether or not to continue reading.


mlieberman85 (approved):
LGTM

Issue #515, opened 2022-10-20
SLSA Conformance Program


JoshuaMulliken:
**Doc:** https://docs.google.com/document/d/1iWjO4UGcGm0PeCm9mbqeT-PiD4z4S7qXMaZsGIFUn0s/edit
**Presentation:** https://docs.google.com/presentation/u/0/d/1oQoJYy9aCGvnEi43NtgSEfuw3IZbYRuapKFrwSceudA/edit

## Background

The Supply Chain Levels for Software Artifacts (SLSA) framework provides a check-list of standards and controls to prevent tampering, improve integrity, and secure packages and infrastructure in your projects, businesses or enterprises. A conformance program that gives explicit permission for organizations to assert their compliance with the SLSA program will allow companies to utilize their security efforts in marketing and allow consumers to evaluate their suppliers effectively

The SLSA Conformance Program utilizes several common industry patterns to give consumers and businesses a transparent understanding of their rights and obligations when asserting compliance with the SLSA framework. This is provided through a self-assessment attestation program and a third-party certification program that are structured into tiers.

## Next Steps

- [ ] Define the scope of the program
  - [ ] Let’s clarify what types of things would be certified: projects, repositories, practices, artifacts, suppliers, toolchains, closed source, etc.
- [ ] Discuss in the specification SIG

MarkLodato:
👍. We also need to figure out how to incorporate it into the specification.

joshuagl:
Thank you for working on this!

I like the two-tiered proposal, getting folks (and process) up and running with a self assessment while we figure out third-party feels like a sound approach.

The self-assessment process reminds me of the OpenSSF Best Practices badge and their model of a web application to complete the form may make sense? https://bestpractices.coreinfrastructure.org/en

I also wanted to link to a discussion around an attestation predicate for human reviews of artefacts happening in the in-toto attestation repository 
* https://github.com/in-toto/attestation/issues/77

tracymiranda:
For reference here is the [SLSA Compliance Assessment](https://docs.google.com/spreadsheets/d/1fNdJJZZdEJAJU0ZaCnYcIlZA3L-bwT7vY7ZesqW3WEs/edit#gid=0) that is in use by the Kubernetes community for their SLSA efforts https://github.com/kubernetes/enhancements/issues/3027

JoshuaMulliken:
[...]

PR #572, opened 2023-01-18
Add SLSA conformance to requirements page

kpk47:
#515

Signed-off-by: kpk47 <1079282+kpk47@users.noreply.github.com>

### DIFF
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -459,7 +459,7 @@ An auditor SHOULD periodically reassess build systems for conformance.
 
 If an auditor is willing to attest to a build system's SLSA conformance,
 then it MUST provide the build system with a non-forgeable attestation for
-the build system to embed provenance it generates.
+the build system to embed in the provenance it generates.
   
 The attestation MUST include the build system's identity, the auditor's
 identity, and an expiration date. The attestation MAY also include one or
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -441,7 +441,7 @@ package ecosystem.
   
 ## Auditor
 
-An <dfn>auditor</dfn> is an organization or individual that assesses build
+An <dfn>auditor</dfn> is an organization or individual that certifies build
 systems for conformance with the SLSA requirements.
   
 A consumer MAY act as their own auditor.
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -268,7 +268,7 @@ The build system is responsible for isolating between builds, even within the
 same tenant project. In other words, how strong of a guarantee do we have that
 the build really executed correctly, without external influence?
 
-The SLSA Build level describes the minimum bar for isolation strength.
+The SLSA Build level describes the minimum bar for isolation strength. For more information on assessing a build system's isolation strength, see [Verifying build systems](verifying-systems.md).
 
 <table>
 <tr><th>Requirement<th>Description<th>L1<th>L2<th>L3
@@ -436,12 +436,8 @@ package.
 The consumer MAY have to opt-in to enable SLSA verification, depending on the
 package ecosystem.
   
-The consumer MUST explicitly trust that the build system meets the SLSA requirements.
-That trust SHOULD be derived from evidence provided by the build system. The
-consumer MAY delegate their trust to an auditor.
-
 > **TODO:** Anything else? Do they need to make risk-based decisions? Respond to
-> errors/warnings?
+> errors/warnings? Do consumers trust builders, or is that up to the package ecosystem?
   
 ## Auditor
 
@@ -454,33 +450,6 @@ An auditor SHOULD use the prompts in [verifying systems](verifying-systems.md)
 when assessing build systems. Auditors MAY go beyond these prompts.
   
 An auditor SHOULD periodically reassess build systems for conformance.
-  
-### SLSA conformance attestations
-
-If an auditor is willing to attest to a build system's SLSA conformance,
-then it MUST provide the build system with a non-forgeable attestation for
-the build system to embed in the provenance it generates.
-  
-The attestation MUST include:
-
--   the build system's identity
--   the auditor's identity
--   an expiration date
--   the SLSA level(s) that the build system satisifies
-The attestation MAY also include one or
-more URIs that resolve to evidence of the build system's conformance.
-  
-TODO: Where does this requirement go? Consumer, package ecosystem?
-  
-<Party> MUST verify the auditor's attestation for validity. An attestation
-is valid if and only if all of the following are true:
-
--   the build system identity is equal to the builder identity in the enclosing
-  provenance
--   the expiration date is strictly later than the time of verification
--   the verifier knows that any secret material used to demonstrate the attestation's
-  non-forgability belongs to the auditor specified in the attestation
--   any secret material used to demonstrate the attestation's non-forgeability has not expired or been revoked at the time of verification
   
 ## Source control
 
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -433,12 +433,47 @@ showing what the options are:
 A package's <dfn>consumer</dfn> is the organization or individual that uses the
 package.
 
-The only requirement on the consumer is that they MAY have to opt-in to enable
-SLSA verification, depending on the package ecosystem.
+The consumer MAY have to opt-in to enable SLSA verification, depending on the
+package ecosystem.
+  
+The consumer MUST explicitly trust that the build system meets the SLSA requirements.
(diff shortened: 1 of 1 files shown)


docs/spec/v1.0/requirements.md
```diff
@@ -433,12 +433,49 @@ showing what the options are:
...
+more URIs that resolve to evidence of the build system's conformance.
+  
+TODO: Where does this requirement go? Consumer, package ecosystem?
+  
+<Party> MUST verify the auditor's attestation for validity. An attestation
+is valid if and only if all of the following are true:
+
+-   the build system identity is equal to the builder identity in the enclosing
+  provenance
+-   the expiration date is strictly later than the time of validation
```
laurentsimon, line 475: I don't think this works with for systems that supports key rotation, e.g. SIgstore.

docs/spec/v1.0/requirements.md
```diff
@@ -433,12 +433,49 @@ showing what the options are:
...
+
+An <dfn>auditor</dfn> is an organization or individual that assesses build
+systems for conformance with the SLSA requirements.
+  
+A consumer MAY act as their own auditor.
+  
+An auditor SHOULD use the prompts in [verifying systems](verifying-systems.md)
+when assessing build systems. Auditors MAY go beyond these prompts.
+  
+An auditor SHOULD periodically reassess build systes for conformance.
```
lehors, line 456:
```suggestion
An auditor SHOULD periodically reassess build systems for conformance.
```

docs/spec/v1.0/requirements.md
```diff
@@ -433,12 +433,49 @@ showing what the options are:
...
+  
+### SLSA conformance attestations
+
+If an auditor is willing to attest to a build system's SLSA conformance,
+then it MUST provide the build system with a non-forgeable attestation for
+the build system to embed provenance it generates.
```
JoshuaMulliken, line 462: small grammar issue. I think it should read as follows:

```
the build system to embed in the provenance it generates.
```

JoshuaMulliken (changes requested):
I'm thrilled with the progress we've made so far. There are a few areas we can refine to make things even better.

Thanks @kpk47!

JoshuaMulliken (approved):

docs/spec/v1.0/requirements.md
```diff
@@ -433,12 +433,49 @@ showing what the options are:
...
+TODO: Where does this requirement go? Consumer, package ecosystem?
+  
+<Party> MUST verify the auditor's attestation for validity. An attestation
+is valid if and only if all of the following are true:
+
+-   the build system identity is equal to the builder identity in the enclosing
+  provenance
+-   the expiration date is strictly later than the time of validation
+-   the verifier knows any secret material used to prove the attestation's
+  non-forgability to belong to the auditor named in the attestation
```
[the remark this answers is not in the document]
kpk47: Done.

docs/spec/v1.0/requirements.md
```diff
@@ -433,12 +433,55 @@ showing what the options are:
...
 package.
 
-The only requirement on the consumer is that they MAY have to opt-in to enable
-SLSA verification, depending on the package ecosystem.
+The consumer MAY have to opt-in to enable SLSA verification, depending on the
+package ecosystem.
+  
+The consumer MUST explicitly trust that the build system meets the SLSA requirements.
+That trust SHOULD be derived from evidence provided by the build system. The
+consumer MAY delegate their trust to an auditor.
```
MarkLodato, line 441: I wonder if we should be a bit less prescriptive here. Ideally, consumers wouldn't have to do anything to get some amount of protection by default. I imagine most would use whatever trust roots (build system) come by default, or maybe whatever their organization has configured. In some cases consumers might opt-in for higher protection, e.g. only allow SLSA Build L3 unless granted an exception, or configure custom trust roots.

For example, imagine a system that only does verification at publication time. As long as you trust the packaging ecosystem, this would protect all users automatically without any client-side validation. This could be particularly valuable for closed-source applications, where provenance is less meaningful. In this case, I think this should still be considered SLSA even though the consumer does nothing at all.

Maybe this part really should go to the packaging ecosystem section after #571 is implemented? And in that case, I'm not sure "explicitly trusted" is the right word, but more that someone made a security decision that a particular build system (as identified by the provenance) is trusted at a particular level?

docs/spec/v1.0/requirements.md
laurentsimon: first question: does "attestation" refers to the auditor's attestation or the "SLSA provenance" attestation?

laurentsimon: I would say something  to the effect of: "the validity period of the enclosing provenance is within the validity period of the auditor's attestation". Then maybe iterate on what this means for static keys vs ephemeral / rotated keys and leaked keys.

docs/spec/v1.0/requirements.md
```diff
@@ -433,12 +433,55 @@ showing what the options are:
...
+systems for conformance with the SLSA requirements.
+  
+A consumer MAY act as their own auditor.
+  
+An auditor SHOULD use the prompts in [verifying systems](verifying-systems.md)
+when assessing build systems. Auditors MAY go beyond these prompts.
+  
+An auditor SHOULD periodically reassess build systems for conformance.
+  
+### SLSA conformance attestations
```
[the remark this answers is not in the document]
kpk47: I've stashed this subheading and will reintroduce it on the 'certification' page when it's ready.

docs/spec/v1.0/requirements.md
kpk47: Done

docs/spec/v1.0/requirements.md
```diff
@@ -433,12 +433,55 @@ showing what the options are:
 A package's <dfn>consumer</dfn> is the organization or individual that uses the
 package.
```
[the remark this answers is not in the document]
kpk47: Done.

It's just a link to the verifying systems page for now, but I don't mind expanding if you don't think it would be duplicative.

docs/spec/v1.0/requirements.md
```diff
@@ -268,7 +268,7 @@ The build system is responsible for isolating between builds, even within the
 same tenant project. In other words, how strong of a guarantee do we have that
 the build really executed correctly, without external influence?
 
-The SLSA Build level describes the minimum bar for isolation strength.
+The SLSA Build level describes the minimum bar for isolation strength. For more information on assessing a build system's isolation strength, see [Verifying build systems](verifying-systems.md).
```
MarkLodato, line 271: nit: wrap at 80 columns

docs/spec/v1.0/requirements.md
```diff
@@ -433,12 +433,24 @@ showing what the options are:
...
+package ecosystem.
+  
 > **TODO:** Anything else? Do they need to make risk-based decisions? Respond to
-> errors/warnings?
-
+> errors/warnings? Do consumers trust builders, or is that up to the package ecosystem?
+  
+## Auditor
+
+An <dfn>auditor</dfn> is an organization or individual that assesses build
```
MarkLodato, line 444: optional: should we use "certifies" here? Perhaps we want to rename this role to "certifier" to avoid proliferation of terms?

```suggestion
An <dfn>auditor</dfn> is an organization or individual that certifies build
```

MarkLodato (approved):
I'd like to hear @lehors's opinion as well.

docs/spec/v1.0/requirements.md
joshuagl: Certifier is a great suggestion. I anticipate that in many cases the auditor/certifier is the same organisation that produces the build system. Self-certification seems much more reasonable than self-auditing.

lehors: Actually, although I support the goal of limiting the number of different terms being used, I don't think certifier is the right term here because certification is for the benefit of someone else. Self-certification is about telling others that based on your own audit you claim that your implementation is conformant. It is not about telling yourself that someone else's implementation is conformant. It doesn't really make sense to say that a consumer may act as their own certifier.

lehors (commented):
I must admit that I don't see much value in the AUDITOR section. It merely contains a MAY and a couple of SHOULDs that don't buy us much.

I would much prefer that we scrap this off and instead add that consumers MUST ensure that the build systems they depend on are conformant and that to do so they can either do their own audit leveraging the prompts in the verifying systems or leverage the certification program which lists conformant systems.

We can leave it to the certification program to get into the details of what certifiers are required to do.

docs/spec/v1.0/requirements.md
kpk47: Done

lehors:
@kpk47 I think it was premature to merge this. I said that I thought the AUDITOR section should be deleted and @MarkLodato agreed.

kpk47:
I'm so sorry. I'm still not used to tracking code reviews in GitHub and thought that I'd addressed all the open comments. I'll send another PR.

PR #575, opened 2023-01-22
Reorg specifications menu

lehors:
Separate the SLSA spec pages from the others within their own submenu.

### DIFF
--- a/docs/_data/nav.yml
+++ b/docs/_data/nav.yml
@@ -5,28 +5,26 @@
 - title: Specifications
   children:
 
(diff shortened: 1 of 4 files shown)


lehors:
I find the current layout confusing so although this may not be the exact layout I'd like to propose something that would be a bit clearer. Happy to modify if something else is preferred.

MarkLodato (approved):
Thanks! This is an improvement.

joshuagl (approved):
Agreed, this is an improvement. Thanks!

PR #582, opened 2023-01-25
Provenance v1: allow arbitrary JSON for params, refine github-actions-workflow

MarkLodato:
This PR refines the provenance v1 schema and the accompanying github-actions-workflow buildType, based on feedback.

### Allow arbitrary JSON for params

Generalize `externalParameters` and `systemParameters` to allow arbitrary JSON objects rather than only specific types. This makes it easier to represent real-world systems. It also avoids the needs for the extra layer of `artifactRef`, `scalarValue`, etc., which was awkward.

This necessitated the removal of the "artifact reference" type of parameter, which combined the actual parameter (usually represented as a URI) with the resolved digest of the artifact that the parameter pointed to. Now the resolved digest is moved to `resolvedDependencies` (with the URI repeated there to allow joining with the parameters.) The impetus for the removal was because there was no good way to represent the concept of an artifact reference parameter, but upon further reflection, the removal actually seems to simplify things.

### Refine github-actions-workflow

Limit to specific event types for which we have fully analyzed the semantics, and add new external parameters for deployment and release events.

Replace the `source` and `workflow` fields with a tuple of `repository`, `ref`, and `path`. This is hopefully easier for everyone to understand and use correctly. It comes at the cost of making it harder to join with `resolvedDependencies`, but this seems like a worthwhile tradeoff.

Expand the documentation to reduce ambiguity.

### Other changes

Expand the guidance for BuildDefinition.

## Preview links

https://deploy-preview-582--slsa.netlify.app/provenance/v1/
https://deploy-preview-582--slsa.netlify.app/github-actions-workflow/v0.1/

/cc @feelepxyz @ianlewis  @tiziano88  @mlieberman85  @asraa  @TomHennen  @kommendorkapten  @rbehjati  @jagathprakash  @laurentsimon  @marcelamelara  @joshuagl  @trishankatdatadog


docs/github-actions-workflow/v0.1/example.json
```diff
@@ -5,36 +5,27 @@
...
+                    "build_id": 123456768,
+                    "deploy_target": "deployment_sys_1a",
+                    "perform_deploy": "true"
                 },
-                "source": {
-                    "artifactRef": {
-                        "uri": "git+https://github.com/octocat/hello-world@refs/heads/main",
-                        "digest": { "sha1": "c27d339ee6075c1f744c5d4b200f7901aad2c369" }
-                    }
+                "source": "git+https://github.com/octocat/hello-world@refs/heads/main",
```
feelepxyz, line 12: 👍  on flattening these structures but does leave a lot up the implementor?

Somewhat tangential feedback on `source`, it's currently a bit painful to parse this URI, as the branch name might also contain`@`. It's also limited, as it currently only contains the source uri and ref.

Would this be flexible and could be expanded with extra props or should this go in `resolvedDependencies` instead?

```json
"source": {
  "uri": "git+https://github.com/octocat/hello-world",
  "digest": { "sha1": "c27d339ee6075c1f744c5d4b200f7901aad2c369" },
  "ref": "refs/heads/main",
  "ref_type": "branch",
  // for completeness I could also imagine adding more things like
  "resource_id": "<repo id>",
  "resource_owner_id": "<repo owner id>",
  "ref_protected": true,
  "head_ref": "",
  "base_ref": ""
}
```

feelepxyz: To clarify, the parsing bit is less of an issue in my mind, I care more about being able to express more details about the source and unsure where this would go atm.

docs/github-actions-workflow/v0.1/example.json
```diff
@@ -5,36 +5,27 @@
...
                 },
-                "inputs": {
-                    "mapValue": {
-                        "mascot": "Mona"
-                    }
-                },
-                "workflowPath": {
-                    "scalarValue": ".github/workflow/release.yml"
-                }
+                "workflow": ".github/workflow/release.yml"
```
feelepxyz, line 16: There's a new claim in the GitHub Actions OIDC token: `workflow_ref` which has the following format (and always points to the triggering/initiating workflow):

```suggestion
                "workflow": "octocat/hello-world/.github/workflow/release.yml@refs/heads/main"
```

rbehjati:
Thanks for this nice change. Is the colour coding for the digram in `docs/provenance/v1/model.svg` described anywhere? Are the boxes in green the "trusted" parts that do not require verification?

docs/github-actions-workflow/v0.1/index.md
```diff
@@ -64,16 +62,16 @@ the path. See [getEntryPoint] for one possible implementation.
...
-The `github` map SHOULD contains the following elements:
+The `github` object SHOULD contains the following elements:
 
-| GitHub Context Parameter        | Description |
-| ------------------------------- | ----------- |
-| `github.mapValue["actor"]`      | The username of the user that triggered the initial workflow run. |
-| `github.mapValue["event_name"]` | The name of the event that triggered the workflow run. |
+| GitHub Context Parameter | Type   | Description |
+| ------------------------ | ------ | ----------- |
+| `github.actor`           | string |The username of the user that triggered the initial workflow run. |
```
asraa, line 73: I would suggest updating this to be `actor_id`. We've had concerns around leaking usernames in public attestations, and the ID better helps allow detection of account compromise or takeover.

docs/github-actions-workflow/v0.1/example.json
asraa: it seems like maybe the details go in the linked `ArtifactReference` in `resolvedDependencies`, but this means that `ArtifactReference` should allow for additional details

docs/provenance/v1/index.md
```diff
@@ -200,13 +197,24 @@ all the information necessary and sufficient to initialize the build and begin
...
-following conventional names are RECOMMENDED when appropriate:
+template, meaning inputs not derived from another input. Each is an arbitrary
+JSON object, though it is RECOMMENDED to keep the structure simple with string
+values to aid verification. The same field name SHOULD NOT be used for both
+`externalParameters` and `systemParameters`.
 
--   `source`: The primary input to the build.
--   `config`: The build configuration, if different from `source`.
+When a parameter is a references to an artifact, such as a git commit, the
+parameter SHOULD be a URI and the `resolvedDependencies` SHOULD record the same
```
asraa, line 206: Since this is just a SHOULD -- if it's not, and say, a `localName`, how can we identify what param the linkage should be against in `resolvedDependencies`?

Should instead `resolvedDependencies` also be a map from `externalParameters.source` to an object?

At the same time then, I'm unsure why it's necessary to separate them. In my mind, the reason I could separate them is because `externalParameters` can reflect a simple recording of top-level inputs, while the trusted builder can populate the `resolvedDependencies` later.

docs/github-actions-workflow/v0.1/example.json
feelepxyz: Another thought, we also expose `workflow_sha`, which will be different to the materials digest sha when using `pull_request_target` in github actions. Using this event to release/publish is an anti-pattern but seems good to capture nonetheless so we can spot this pattern.

What if we changed this to an object?
```suggestion
                "workflow": {
                  "uri": "git+https://github.com/octocat/hello-world/.github/workflow/release.yml",
                  "digest": { "sha1": "c27d339ee6075c1f744c5d4b200f7901aad2c369" }
                }
```

Looking ahead, I could see a world where it will be possible to reference initiating workflows from a different repo to the source with some kind of org policy, at which point we will need to track the uri and digest (btw, this isn't in the works as far as I know).

asraa: +1000! I wonder if this then falls into the pattern of referencing a complete `ArtifactReference` like in `resolvedDependencies` (I'm not sure this is a dependency though, but I like the idea of resolving the ref int oa digest)

docs/github-actions-workflow/v0.1/example.json
MarkLodato: > it seems like maybe the details go in the linked `ArtifactReference` in `resolvedDependencies`, but this means that `ArtifactReference` should allow for additional details

Yes, that was my intent. All these extra fields are not actually parameters but rather annotations about the parameter (or the artifact that the parameter refers to). Maybe we just open this up to allow arbitrary fields.

IIUC, the only two true parameters in the example above is `uri` and `ref`, which I had merged into one URI. The digest is *not* a parameter. You don't tell GitHub Actions to build "refs/heads/main @ c27d339...". You either tell it to build "refs/heads/mains" *or* "c27d339...". (For parsing, you just go up to the first `@`; to be clear, this is the existing [SPDX Download Location](https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field) format.)

The reason I had it as a single URI was to make it easier to join with `resolvedDependencies` or any other source. Joining on a single string is easier than multiple fields. But I don't feel too strongly.

@feelepxyz what do you think?

docs/github-actions-workflow/v0.1/example.json
MarkLodato: What is the actual external interface to GitHub Actions? That is what I'm trying to capture here. As I mentioned in the other thread, the sha is not part of that interface AFAICT so it would go in `resolvedDependencies` instead.

Can the `github.repo` (called `source` here) ever differ from the the top-level workflow (aside from reusable workflows)? Does this happen for `pull_request_target`, or are you just saying it might happen in the future?

We could split this in two, though it would add a bit of complexity:

```jsonc
"source": "git+https://github.com/repo/repo@refs/heads/main",  // maybe omit this if the same as `workflow.repo`?
"workflow": {
  "repo": "git+https://github.com/workflow/workflow@refs/heads/main",
  "path": ".github/workflow/release.yml"
}
```

Regarding the format of the URI and path, my thinking was this:

- The `uri` is the locator of the git commit. It contains both the repository location (https://github.com/octocat/hello-world) and the ref that was evaluated (refs/heads/main). We use a single URI here so that we can join `externalParameters` and `resolvedDependencies`.
- The `workflow` is the path to the workflow within that commit. It seems redundant and error-prone to repeat parts of the `uri` here if it always has to agree. Alternatively, if it can ever disagree, then the workflow actually came from a different commit, and we should have a separate URI for that commit (again, for joining purposes).

docs/provenance/v1/index.md
MarkLodato: > Thanks for this nice change. Is the colour coding for the digram in `docs/provenance/v1/model.svg` described anywhere? Are the boxes in green the "trusted" parts that do not require verification?

No, it's not documented. I was wondering when someone would call me on this. 😄 

This is what I was thinking, though I'd like a more rigorous approach if anyone has ideas:

- Green = trusted, does not require verification
- Orange = external
- Grey = logical, doesn't really exist
- Red = the build process itself, red because it's under the tenant's control

MarkLodato:
> Thanks for this nice change. Is the colour coding for the digram in `docs/provenance/v1/model.svg` described anywhere? Are the boxes in green the "trusted" parts that do not require verification?

No, it's not documented. I was wondering when someone would call me on this. 😄 

This is what I was thinking, though I'd like a more rigorous approach if anyone has ideas:

- Green = trusted, does not require verification
- Orange = external
- Grey = logical, doesn't really exist
- Red = the build process itself, red because it's under the tenant's control

docs/github-actions-workflow/v0.1/index.md
MarkLodato: Done.

docs/provenance/v1/index.md
MarkLodato: > Since this is just a SHOULD -- if it's not, and say, a `localName`, how can we identify what param the linkage should be against in `resolvedDependencies`?

Good question. There are actually two SHOULDs here:

- You SHOULD record the digest in `resolvedDependencies`.
- You SHOULD use the same identifier in both so that you can join them, preferably a URI.

IIUC you're suggestion (1) to explain why so that the reader can make an informed decision on the interpretation of SHOULD, and (2) more clearly state that there can be alternate ways of joining to `resolvedDependencies`. Is that correct?

> Should instead `resolvedDependencies` also be a map from `externalParameters.source` to an object?

Sorry, I don't understand what you mean here. Could you rephrase?

> In my mind, the reason I could separate them is because `externalParameters` can reflect a simple recording of top-level inputs, while the trusted builder can populate the `resolvedDependencies` later.

Yes, this is the reason for the separation. Separating them makes verification more straightforward because you only verify the `externalParameters`. (By the way, the verification still needs to be written!)

docs/github-actions-workflow/v0.1/example.json
laurentsimon: Note that for a reusable workflow, the `workflow` is actually not relevant and we could omit it entirely, ie only report it as part of `systemParameters`.

docs/provenance/v1/index.md
```diff
@@ -200,13 +197,24 @@ all the information necessary and sufficient to initialize the build and begin
...
-name MUST be unique across `externalParameters` and `systemParameters`. The
-following conventional names are RECOMMENDED when appropriate:
+template, meaning inputs not derived from another input. Each is an arbitrary
+JSON object, though it is RECOMMENDED to keep the structure simple with string
+values to aid verification. The same field name SHOULD NOT be used for both
+`externalParameters` and `systemParameters`.
 
--   `source`: The primary input to the build.
--   `config`: The build configuration, if different from `source`.
+When a parameter is a references to an artifact, such as a git commit, the
```
marcelamelara, line 205:
```suggestion
When a parameter is a reference to an artifact, such as a git commit, the
```

marcelamelara (commented):
I think the changes in this PR are generally going to be quite helpful and agree that they simplify the provenance format.

docs/github-actions-workflow/v0.1/example.json
kommendorkapten: I the current proposal makes sense, as the details are in the `resolvedDependencies`.

docs/github-actions-workflow/v0.1/example.json
kommendorkapten: Even if `workflow` would be irrelevant when using a reusable workflow, I think it make sense to keep it in `externalParameters` for consistency? I would think that it's up the verifier to expect certain values. 
When using a reusable workflow, that would go into the `reseolvedDependencies` right? We can get the hash via the `job_workflow_sha` claim from the id token.

MarkLodato: PTAL. I just updated the schema as follows and I hope it is more understandable now.

1. Only support the specific event types that have clear semantics that align with the model. There are so many different event types, and new ones will crop up, that I think it's safer to take an allowlist approach. That avoids issues like `pull_request` and `pull_request_trigger` that seem error prone.
2. Rename to `workflowPath`.
3. Clarify documentation.

I originally tried to expand based on the idea of having a separate URI for workflow, particularly to support reusable workflows, but it made everything much harder to understand.

docs/github-actions-workflow/v0.1/index.md
```diff
@@ -9,12 +9,39 @@ hero_text: |
...
+
+Only the following [event types] are supported:
+
+| Supported event type  | Event description |
+| --------------------- | ----------------- |
+| [`create`]            | Creation of a git tag or branch. |
+| [`push`]              | Creation or update of a git tag or branch. |
+| [`workflow_dispatch`] | Manual trigger of a workflow. |
+
+[`create`]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#create
```
laurentsimon, line 22: let's add the release trigger, which is commonly used as well https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release?

docs/github-actions-workflow/v0.1/example.json
kommendorkapten: Explicitly dropping support for `pull_request` and especially `pull_request_target` is IMHO a very good idea.

docs/github-actions-workflow/v0.1/example.json
feelepxyz: Ah yes this makes sense! Thanks for clarifying.

TIL about the [SPDX Download Location](https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field) format 👌

docs/github-actions-workflow/v0.1/index.md
```diff
@@ -64,18 +62,18 @@ the path. See [getEntryPoint] for one possible implementation.
...
-The `github` map SHOULD contains the following elements:
+The `github` object SHOULD contains the following elements:
 
-| GitHub Context Parameter        | Description |
-| ------------------------------- | ----------- |
-| `github.mapValue["actor"]`      | The username of the user that triggered the initial workflow run. |
-| `github.mapValue["event_name"]` | The name of the event that triggered the workflow run. |
+| GitHub Context Parameter | Type   | Description |
+| ------------------------ | ------ | ----------- |
+| `github.actor_id`        | string | The numeric ID of the user that triggered the initial workflow run. |
```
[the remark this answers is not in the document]
feelepxyz: hm yes does not look like `triggering_actor_id` is exposed, but will look at getting this added to match `actor_id` 👍

docs/github-actions-workflow/v0.1/example.json
feelepxyz: 👍  on only supporting a subset of event types.

What are your thoughts on aligning the value with what you get from the current claim: `workflow_ref`/`GITHUB_WORKFLOW_REF` env var so we don't need to restructure it?

`workflow_ref` has the following format: `octocat/hello-world/.github/workflow/release.yml@refs/heads/main`. Should we align it to the `source` URI format?

```
"workflow": "git+https://github.com/octocat/hello-world/.github/workflow/release.yml@refs/heads/main"
or
"workflowRef": "octocat/hello-world/.github/workflow/release.yml@refs/heads/main"
```

Although looks like [spdx](https://spdx.github.io/spdx-spec/v2.3/package-information/#773-examples) has a different idea about how this should be formatted.. `git+https://github.com/octocat/hello-world@master#.github/workflow/release.yml`

docs/github-actions-workflow/v0.1/index.md
MarkLodato: Done. This took me several hours to fully analyze and capture here, since I had to understand what the actual external interface is. I tried to emphasize this in the intro to set the expectation for readers that any future addition (or other build type) should do the same. PTAL.

@feelepxyz Please take a look to make sure that I properly captures what the external interface is.

docs/github-actions-workflow/v0.1/index.md
```diff
@@ -9,85 +9,137 @@ hero_text: |
...
-workflow, such as an action, a job, or a reusable workflow.
-
+workflow that builds a software artifact.
+
+Only the following [event types] are supported:
+
+| Supported event type  | Event description |
+| --------------------- | ----------------- |
+| [`create`]            | Creation of a git tag or branch. |
+| [`deployment`]        | Creation of a deployment. NOTE: This contains |
```
feelepxyz, line 19: Looks like the note is missing some text

rbehjati:
> > Thanks for this nice change. Is the colour coding for the digram in `docs/provenance/v1/model.svg` described anywhere? Are the boxes in green the "trusted" parts that do not require verification?
> 
> No, it's not documented. I was wondering when someone would call me on this. 😄
> 
> This is what I was thinking, though I'd like a more rigorous approach if anyone has ideas:
> 
> * Green = trusted, does not require verification
> * Orange = external
> * Grey = logical, doesn't really exist
> * Red = the build process itself, red because it's under the tenant's control

Thanks for clarifying. This looks good to me. Perhaps this legend could be added to a corner of the image, and then we can iterate on it later. As an idea for a future improvement, I think it would be nice to clarify how the diagram and particularly the "Build Process" relate to the `builder`. The description in https://deploy-preview-582--slsa.netlify.app/provenance/v1/ talks about `builder`, but is not shown in the diagram.

docs/github-actions-workflow/v0.1/index.md
feelepxyz: @MarkLodato was there are a reason for not including the [release](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) event here? I'm pretty sure there will be quite a few use cases where publishing depends on creating a gh release, which then kicks of package ecosystem specific publishing.

docs/github-actions-workflow/v0.1/example.json
feelepxyz: Ah just spotted https://github.com/slsa-framework/slsa/pull/582/files#diff-bbdc17f357d443804cf506038eba199aed6be071c25af1b7307e699bf9ad69c1R104 - seems doable but fwiw I would definitely prefer just copying in the `workflow_ref` in here for ease of use 😄

docs/github-actions-workflow/v0.1/index.md
MarkLodato: Oops! I added `deployment` thinking it was `release`! Luckily doing the same thing again for release was much faster. Added. Please take a look.

docs/github-actions-workflow/v0.1/index.md
MarkLodato: Thanks for catching! Fixed.

MarkLodato:
> Thanks for clarifying. This looks good to me. Perhaps this legend could be added to a corner of the image, and then we can iterate on it later. As an idea for a future improvement, I think it would be nice to clarify how the diagram and particularly the "Build Process" relate to the `builder`. The description in https://deploy-preview-582--slsa.netlify.app/provenance/v1/ talks about `builder`, but is not shown in the diagram.

Great points! I filed #593 to track since it will take some time and it's a bit orthogonal to the original intent of this PR.

docs/github-actions-workflow/v0.1/example.json
MarkLodato: In my opinion, copying `workflow_ref` makes it easier to *generate* the provenance but harder to *verify* it. We only need to implement generation once or twice, while we need to implement verification and write down expectations many times. Therefore I think we should be optimizing for verification, not generation.

In particular, I'm worried about having duplicate information in `workflow_ref` that is already present in `source` because either it makes expectations be more verbose than necessary or it bakes in an assumption that they're always the same, which could change in the future.

I imagine we have three basic options:

- Three fields: `repo`, `ref`, `workflowPath`
- Two fields: `source` URI (repo@ref) and `workflowPath`
- One field: `workflow` URI (repo@ref#path)

I originally went with the two-field option because that matched our original idea of always using URIs to identify commits, and because you can join with `resolvedDependencies` using the same URI. But maybe that's not the right thing to optimize for. It does require parsing a URI even for expectations, since people will probably say "allow this particular repo but at any tag".

The one-field option is kind of nice since it's just a URI, but probably harder to work with in practice. It doesn't map 1:1 with `resolvedDependencies` (need to strip off the path, since the dependency is the full commit, not just that one file).

So maybe the three-field option is the best to go with? What do you think of that?

docs/github-actions-workflow/v0.1/index.md
```diff
@@ -9,85 +9,153 @@ hero_text: |
...
+top-level and reusable would make the schema too error-prone.
+
+[SLSA Issues]: https://github.com/slsa-framework/slsa/issues
 [GitHub Actions]: https://docs.github.com/en/actions
+[event types]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
+[isolation]: /spec/v1.0/requirements#isolation-strength
 
 ## Build Definition
 
 ### External parameters
```
MarkLodato, line 55: Any thoughts on grouping these into:

- `workflow`, containing `repository`, `ref`, and `path`
- `event`, containing `dispatch`, `inputs`, and `release` (at most one set)
- `vars`

MarkLodato: I tried it out and like grouping by `workflow` but not by `event`. Done.

docs/github-actions-workflow/v0.1/index.md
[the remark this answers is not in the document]
MarkLodato: Added a TODO.

docs/github-actions-workflow/v0.1/index.md
```diff
@@ -25,24 +25,22 @@ All external parameters are REQUIRED unless empty.
 <table>
 <tr><th>Parameter<th>Type<th>Description
 
-<tr id="inputs"><td><code>inputs</code><td>mapValue<td>
+<tr id="inputs"><td><code>inputs</code><td>object<td>
 
-The [inputs context], with each value converted to string. Every non-empty input
-value MUST be recorded. Empty values SHOULD be omitted.
+The [inputs][inputs context] to the top-level workflow. Every non-empty input
```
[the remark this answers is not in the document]
MarkLodato: In order to get this PR submitted, I forked this as #594. Marking this thread as resolved, but please reopen if you disagree.

docs/provenance/v1/index.md
MarkLodato: Please take a look at the updated wording. Does this address your concern?

> The parameters SHOULD only contain the actual values passed in through the interface to the build system. Metadata about those parameter values, particularly digests of artifacts referenced by those parameters, SHOULD instead go in resolvedDependencies. The documentation for buildType SHOULD explain how to convert from a parameter to the dependency uri.

docs/github-actions-workflow/v0.1/example.json
MarkLodato: After more thought, I split into three fields, grouped under a common `workflow` object. Please take a look.

docs/github-actions-workflow/v0.1/example.json
MarkLodato: Marking this as resolved since the other thread also covers these same concepts, and it's easier to have one thread instead of two.

docs/github-actions-workflow/v0.1/example.json
feelepxyz: Good call out on verification! I like the three field workflow object, I think it will be easier to work with and clear how they are pointing to the same source now 👍

docs/github-actions-workflow/v0.1/index.md
```diff
@@ -9,87 +9,174 @@ hero_text: |
...
 ### Resolved dependencies
 
-The resolved dependencies MAY contain any artifacts known to be input to the
-workflow, such as the specific versions of the virtual environments used.
+The `resolvedDependencies` SHOULD contain an entry identifying the resolved the
+git commit ID corresponding to `externalParameters.workflow`. The dependency's
+`uri` MUST be in [SPDX Download Location] format, i.e.
+`"git+" + workflow.uri + "@" + workflow.ref`. See [Example].
+
+The `resolvedDependencies` MAY contain additional artifacts known to be input to
```
laurentsimon, line 176: nit: maybe make it clearer that if there are additional entries, the *first* entry MUST be the one corresponding to `externalParameters.workflow`

docs/github-actions-workflow/v0.1/index.md
```diff
@@ -9,87 +9,174 @@ hero_text: |
...
-| ------------------------------- | ----------- |
-| `github.mapValue["actor"]`      | The username of the user that triggered the initial workflow run. |
-| `github.mapValue["event_name"]` | The name of the event that triggered the workflow run. |
+| GitHub Context Parameter | Type   | Description |
+| ------------------------ | ------ | ----------- |
+| `github.actor_id`        | string | The numeric ID of the user that triggered the initial workflow run. |
+| `github.event_name`      | string | The name of the event that triggered the workflow run. |
 
-> TODO: What about `actor_id`, `repository_id`, and `repository_owner_id`? Those
+> TODO: What about `repository_id`, and `repository_owner_id`? Those
```
feelepxyz, line 159: These ids seem pretty important to detect transfers and resurrected repos.

They are part of the ENV so straight forward to extract but can also look into adding these to the github.context object for parity.

ianlewis: Yes we [include these](https://slsa.dev/blog/2022/06/slsa-github-workflows#detect-recycled-github-accounts) for slsa-github-generator. It probably would have been nicer if it was in the github context. I don't think we put the full ENV in the provenance for various reasons.

docs/github-actions-workflow/v0.1/index.md
MarkLodato: The `resolvedDependencies` is currently documented unordered. I suggest keeping it that way.

MarkLodato:
Hey all, I would like to submit this ASAP so that we can continue to iterate. Any approvals would be appreciated. I'm happy to make small fixes and corrections, but any other significant change I'd like to defer via a TODO so that this PR gets unblocked. Thanks!

docs/github-actions-workflow/v0.1/index.md
MarkLodato: I added them to the doc, even though technically they're not in the github context yet. I think it's clear enough. Please reopen if you disagree.

abacchi (approved):

kommendorkapten (approved):
Good work all!

One minor comment (can be resolved later). The section on how to migrate from `0.2` format still relies on `externalParameters` such as `entryPoint` and `source` which is not covered any more here.

joshuagl (approved):

docs/provenance/v1/index.md
```diff
@@ -606,6 +603,7 @@ Renamed to "slsa.dev/provenance".
 Initial version, named "in-toto.io/Provenance"
 
 [Statement]: https://github.com/in-toto/attestation/blob/main/spec/README.md#statement
+[DigestSet]: https://github.com/in-toto/attestation/blob/main/spec/field_types.md#DigestSet
```
ianlewis, line 606: This doesn't seem to be used. I think you meant to replace the direct url reference elsewhere in the page?

ianlewis (approved):
Sorry for the late comments.

rbehjati (approved):
Thanks for making this change.

docs/provenance/v1/index.md
MarkLodato: Fixed. Thanks for catching!

docs/github-actions-workflow/v0.1/index.md
```diff
@@ -9,87 +9,170 @@ hero_text: |
...
+specification is first updated. This ensures that the [external parameters] are
+fully captured and that the semantics are unambiguous. Other event types are
+irrelevant for software builds (such as `issues`) or have complex semantics that
+may be error prone or require further analysis (such as `pull_request` or
+`repository_dispatch`). To add support for another event type, please open a
+[GitHub Issue][SLSA Issues].
+
+Note: Consumers are REQUIRED to reject unrecognized external parameters, so new
+event types can be added without a major version bump as long as they do not
+change the semantics of existing external parameters.
```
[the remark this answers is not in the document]
MarkLodato: Ooh, good catch! This completely contradicts the provenance doc, which says "Consumers MUST **ignore** unrecognized fields."!

I'm going to add a TODO for now so that I can get this submitted, but this is a really important point that needs to be fixed.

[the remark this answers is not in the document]
MarkLodato: And regarding knowing buildType, I'll cover that in the verification section. I added a TODO for that as well.

MarkLodato:
Thanks, everyone! Tons of great contributions! I'll send out more PRs to address the TODOs and iterate on the design.

PR #584, opened 2023-01-25
Fix switching between provenance spec versions

joshuagl:
The v1 spec is in a child directory, whereas earlier (0.1 and 0.2) versions are direct children of the provenance folder. This means that when the versions-dropdown replaces the v1 version part with 0.1 or 0.2 in the replacement url includes a trailing '/'. Stripping the trailing url for all items in the version selector fixes navigating between provenance specification versions.


MarkLodato (approved):
Thanks!

Separately, we might also want to move v0.1 and v0.2 to a directory, both for consistency and also to allow us to keep the versioned images alongside the text.

PR #585, opened 2023-01-25
Add requirement on using best practices

lehors:
This adds a small section indicating that all implementations MUST use industry security best practices.

### DIFF
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -75,6 +75,19 @@ Responsibility to implement SLSA is spread across the following parties.
   <td>✓<td>✓<td>✓
 </table>
 
+### Security Best Practices
+
+While the exact definition of what constitutes a secure system is beyond the
+scope of this specification, to be conformant all implementations MUST use
+industry security best practices. This includes, but is not limited to, using
+proper access controls, securing communications, implementing proper management
+of cryptographic secrets, doing frequent updates, and promptly fixing known
+vulnerabilities.
+
+Various relevant standards and guides can be consulted for that matter such as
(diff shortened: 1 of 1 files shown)


lehors:
My goal is to address the gaping hole we currently have in the spec which fails to define a baseline for what implementations are required to do to ensure basic security and be able to claim compliance.
I propose to use the CIS controls as a non-prescriptive reference while recognizing that many other references could be used instead. Other people, like @david-a-wheeler , may have other ideas.

docs/spec/v1.0/requirements.md
```diff
@@ -75,6 +75,17 @@ Responsibility to implement SLSA is spread across the following parties.
...
+### Security Best Practices
+
+While the exact definition of what constitutes a secure system is beyond the
+scope of this specification, to be conformant all implementations MUST use
+industry security best practices. This includes, but is not limited to, using
+proper access controls, securing communications, implementing proper management
+of cryptographic secrets, doing frequent updates, and promptly fixing known
+vulnerabilities.
+
+Various relevant standards and guides can be consulted for that matter such as the [CIS Critical Security Controls](https://www.cisecurity.org/controls/cis-controls-list).
```
MarkLodato, line 87: nit: wrap at 80 columns for consistency

MarkLodato (approved):
Thanks!

docs/spec/v1.0/requirements.md
lehors: Sorry, I had missed that one. Fixed.

joshuagl (approved):

kpk47 (approved):

PR #588, opened 2023-01-26
Fix consumer opt-in requirement

lehors:
The use of MAY in the consumer requirement is incorrect and fails to capture the intended requirement. MAY means that something is optional and can be done or not without any impact on conformance. This commit changes the requirement into a MUST using a different formulation.

### DIFF
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -443,8 +443,8 @@ showing what the options are:
 A package's <dfn>consumer</dfn> is the organization or individual that uses the
 package.
 
-The only requirement on the consumer is that they MAY have to opt-in to enable
-SLSA verification, depending on the package ecosystem.
+The consumer MUST enable SLSA verification when the package ecosystem requires
+to opt-in for the verification to be enabled.
 
 > **TODO:** Anything else? Do they need to make risk-based decisions? Respond to
 > errors/warnings?


docs/spec/v1.0/requirements.md
```diff
@@ -443,8 +443,8 @@ showing what the options are:
 A package's <dfn>consumer</dfn> is the organization or individual that uses the
 package.
 
-The only requirement on the consumer is that they MAY have to opt-in to enable
-SLSA verification, depending on the package ecosystem.
+The consumer MUST enable SLSA verification when the package ecosystem requires
+to opt-in for the verification to be enabled.
```
MarkLodato, line 447: I agree that the old usage of MAY was incorrect, but this wording is a bit hard to parse, and I'm also not sure how to reconcile this with the notion that we would consider a monitoring-only mode still SLSA conformant.\* An alternative could be to avoid a keyword here, such as:

"Depending on the ecosystem, the consumer might need to opt-in to enable SLSA enforcement."? You could go stronger as "If the package ecosystem has opt-in SLSA enforcement, the consumer MUST opt-in it to receive the benefits of SLSA enforcement" but I'm not sure that buys us.

What do you think?

\* See Line 495 of https://github.com/slsa-framework/slsa/pull/571/files#diff-3a161534a67508acd917c97c9e5d57ed648c50c3a7da23acd58aede57fafbdb2R495

lehors: Yeah, I must admit I struggled a bit to find the right way to say it so I'm happy to change to something else.

Now, my understanding is that the reason we even say anything about the consumer is because the consumer has some responsibility and we need to define what that is. The table above defines the consumer's requirement as "Opt-in to verification".
If this is not always require we need to change the table to reflect that.

I think we need to decide what is actually required before spending more time on the exact wording. Maybe we can discuss this on Monday's call?

MarkLodato: I added it to the agenda, though there are other topics on there as well. In the meantime, we can try to make some progress here. I created #591 to discuss since it's a big topic and it's hard to follow PR threads.

PR #590, opened 2023-01-26
compliance: certification program

SecKatie:
- adds `spec/v1.0/certification.md` to document the process and tiers that define trust
- adds `spec/v1.0/certification-registry.md` to hold self-certifications
- adds `_data/spec_v1-0/certification-registry.yml` to hold a machine readable format for builder self-certifications

Addresses #515


docs/_data/spec_v1-0/certification-registry.yml
```diff
@@ -0,0 +1,5 @@
+- name: "[Example] Foo Build System"
+  slsa_level: 2
+  website: https://example.com/foo-build-system
+  self_attestation: https://example.com/foo-build-system/attestation
```
laurentsimon, line 4: should this attestation link be immutable? 
Alternatively, could the compliance repo store it? This way there's a track record even if the links disappear

docs/_data/spec_v1-0/certification-registry.yml
```diff
@@ -0,0 +1,5 @@
...
+  public_key: https://example.com/foo-build-system/public-key
```
laurentsimon, line 5: this key needs to be referred to immutably. Maybe add a sha256 field?
Alternatively, could the compliance repo store it? This way you get all the ifo in one call, instead of querying multiple links for their attestations

laurentsimon, line 5: We may also want to say that a key should contain not just the raw bytes but also the key type (RSA, ECDSA, etc). Most format support this out of the box, I think

docs/spec/v1.0/certification.md
```diff
@@ -0,0 +1,63 @@
...
+#### Process
+
+The self-certification process includes the following steps:
+
+1. The build system owner fills out the SLSA Self-Certification Questionnaire to
+   document their build system's conformance to a particular SLSA level.
+   **[TODO] provide link to questionnaire**
+2. The build system owner attests to the accuracy of the questionnaire responses
+   and publishes it on their website.
+3. The build system owner publishes their public key to a public key server.
```
laurentsimon, line 56: the key should be stored in a versioning system?

docs/spec/v1.0/certification-registry.md
```diff
@@ -0,0 +1,21 @@
+# Certification Registry
+
+## **TODO**
+
+- [ ] Add a link to the SLSA Self-Certification Questionnaire.
+
+This page lists build systems that have certified conformance to the
+[SLSA Framework Version 1.0](index.md) by following the
```
abacchi, line 8: Would we create a clone of these pages for version 1.1 (& above)-  build the T1 +T2 list from scratch as certifications come in?

docs/spec/v1.0/certification.md
abacchi: Would it make any sense for the SLSA.dev website to accept these attestations and keys in this repository for storage + reference? (Maybe instruct the build system owner to open a PR here?)

docs/spec/v1.0/certification-registry.md
```diff
@@ -0,0 +1,21 @@
+# Certification Registry
+
+## **TODO**
+
+- [ ] Add a link to the SLSA Self-Certification Questionnaire.
+
+This page lists build systems that have certified conformance to the
```
abacchi, line 7: I think you mentioned we should replace the word conformance with compliance in the spec meeting today

docs/spec/v1.0/certification.md
laurentsimon: should ask. Will the collection of metadata require versioning? Will `main` be the ground truth or will folks cut release every X months to release? (have to be careful how this plays with revocation)

docs/spec/v1.0/certification-registry.md
lehors: Although the two terms conformance and compliance are often used interchangeably the right term to use is conformance (along with conformant).
I suggest use this consistently throughout our documentation.

https://wikidiff.com/conformant/compliant for instance states that:
_the difference between compliant and conformant is that compliant is willing to comply; [...] while conformant is in accordance with a set of specifications._

lehors:
I don't think the registry should be stored along with the specification. Running a certification program requires quite a bit of infrastructure and people dedicated to the task. I don't think it is practical to have PRs related to registry updates mixed in with specification work.

We ought to have a separate repo for the certification program that holds the registry, with its own set of maintainers, against which PRs related to applications can be submitted.

This means that certification entries ought to specify which version of SLSA it relates to, in addition to the specific version of the product/offering.

docs/_data/spec_v1-0/certification-registry.yml
ianlewis: +1 to the compliance repo storing the attestation and public key

docs/_data/spec_v1-0/certification-registry.yml
ianlewis: If it's stored in the compliance repo we can add tests to make sure this is the case.

JoshuaMulliken: Removed the registry from this PR

docs/spec/v1.0/certification.md
```diff
@@ -0,0 +1,63 @@
...
+4. The build system owner submits a pull request to add their build system to
```
kpk47, line 58: The registry should also support keyless provenance signing with Sigstore. I think the OIDC issuer + Subject Alternate Name should be enough to identify a builder.

docs/spec/v1.0/certification.md
```diff
@@ -0,0 +1,63 @@
...
+#### Process
+
+The self-certification process includes the following steps:
+
+1. The build system owner fills out the SLSA Self-Certification Questionnaire to
+   document their build system's conformance to a particular SLSA level.
+   **[TODO] provide link to questionnaire**
+2. The build system owner attests to the accuracy of the questionnaire responses
+   and publishes it on their website.
+3. The build system owner publishes their public key to a public key server.
```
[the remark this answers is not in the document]
JoshuaMulliken: I think using x509 certs is the cleanest option (from a user perspective) but it is definitely more difficult from an implementation perspective.

I would love to be able to point to a "root CA" for lack of a better term that each build service owner maintains.

docs/spec/v1.0/certification.md
JoshuaMulliken: ^ that is a great idea! Do you think that this would scale better than the x509 idea?

Should we standardize on one of these options or offer both as options?

docs/spec/v1.0/certification.md
JoshuaMulliken: What do you think of the ideas stated here around keys: https://github.com/slsa-framework/slsa/pull/590#discussion_r1094733574 https://github.com/slsa-framework/slsa/pull/590#discussion_r1093849945

Do you think that either of these options would help to satisfy the need for immutable data but also revocation? I hate the idea of a build system owner having to come to our repo and update the version of their key if it get's revoked or expires. It seems like it is a process that is highly vulnerable to human error.

docs/spec/v1.0/certification.md
laurentsimon: Sigstore *is* a CA (Fulcio) using x509 certs for users. It's basically an "instantiation" of the idea

docs/spec/v1.0/certification.md
laurentsimon: PKI is definitely better since it pushes the problem to a CA instead of us. One thing to think about, though, is verification. PKI was built for "online" verification, where the client wants to verify that the cert is valid at the time of the connection.
Provenance verification happens "offline", ie a client wants to verify that the provenance was signed at a time where the certificate (and the builder attestation) was valid. This requires to have a log of "time of a signature" like Rekor provides. If you want to support key rotation, you need this feature or a client would always reject provenance signed with a certificate that is expired at verification time.

docs/spec/v1.0/certification.md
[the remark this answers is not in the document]
kpk47: Yes, that's exactly the model I had in mind. Each build service (or maybe build service type if an owner operates more than one) is the root CA, and they mint certs for each worker they spin up. The problem is that it feels unreasonable to ask a build system operator also to run a CA, or at least it limits the PKI option to organizations of a certain size. That's why we should also have a keyless option with Sigstore.

docs/spec/v1.0/certification.md
kpk47: We should support both Sigstore and more conventional PKI. Sigstore is easier to use, but I don't like the idea of forcing SLSA builders to use Sigstore.

docs/spec/v1.0/certification.md
kpk47: I must admit that I don't understand certificate transparency all that well. It would help prove a certificate hadn't expired at the time of signature but not whether the certificate had been revoked, right? IIUC, Sigstore uses very short-lived certificates by design, so revocation isn't part of the design, so even if we require CT for the build system PKI we need a way of handling revocation. Is that right? Or could we just punt revocation to the CAs in the case where builders don't use Sigstore?

laurentsimon: CT in general just stores certs, not time of signatures: this works fine for web PKI / TLS, because verification happens "online". Rekor does store time of signatures, which enables "offline" revocation + key rotation (ephemeral keys are just short-lived rotated keys).

You're correct that Sigstore does not support revocation today, I think there are efforts to build some mechanism "on top" of Sigstore.

docs/spec/v1.0/certification.md
abacchi: I think that's a fair point- maybe its to the extent that we recommend Sigstore and show how to do the steps- but also allow/support other options as well.

MarkLodato:
Given the age of this PR, it appears that this effort has stalled. I'm closing this PR to keep the list of open PRs clean. If this gets picked up, we can reopen or create a new PR.

PR #595, opened 2023-01-30
Fix RFC2119 lint: Sentence Case, "Recommended Suite"

MarkLodato:
-   Catch Sentence Case, not just full lowercase, and fix existing uses of sentence case.
-   Allowlist "recommended suite" from attestation-model.md. This was previously causing the linter to fail. I can't think of a better way to phrase that avoids the term "recommended", and use of caps doesn't work because we use that title in a link.


lehors (approved):
LGTM

PR #598, opened 2023-01-31
Remove auditor section from requirements. Add certification to consum…

kpk47:
…er section.

Signed-off-by: kpk47 <kkris@google.com>


lehors (approved):
Thanks for following up on this. We currently have a discussion on the exact requirements we want to have on the consumer (captured in Issue#591) and this section will need updating but this can be done later on as part of PR#588 which is in progress.

Issue #601, opened 2023-02-02
"Identifies artifact" requirement no longer explicit


joshuagl:
The 1.0 requirements and provenance format no longer explicitly state that provenance "provenance MUST identify the output artifact via at least one cryptographic hash" per the 0.1 [Identified artifact](https://slsa.dev/spec/v0.1/requirements#identifies-artifact) requirement.

This is implicit if you're following the specifications recommendation to use the SLSA provenance format and the in-toto attestation suite, but what about for those who are not? Should we make this an explicit requirement again?

MarkLodato:
Hey @joshuagl what do you consider the priority of this? Is it a blocker for v1, can we address it later without a version bump, or is it just lower enough priority to close?

joshuagl:
I think we can address this later without a version bump. This is covered in the spec today. In fact, we could probably even close this and add the clarification if requested, because:
* In #602 we added additional text to state that "The build process MUST generate provenance that *unambiguously identifies* the output package" (emphasis added)
* We refer to the provenance's subject in the spec, i.e, in step 2 of [How to verify](https://slsa.dev/spec/v1.0/verifying-artifacts#step-1-check-slsa-build-level).
* _possibly other places we talk about `subject`_

MarkLodato:
Thanks. That's helpful. Re-reading the text, I still think "unambiguously identifies" is too ambiguous 😝  so I sent out #768 to add "by cryptographic digest". With that I think this issue is closed.

joshuagl:
That changes removes ambiguity, thanks Mark!

PR #602, opened 2023-02-02
Make the need to identify artifacts more explicit

joshuagl:
Per issue #601 the v1 spec doesn't make it explicit that provenance metadata should unambiguously identify the output artifact. This minor change makes that more apparent (though perhaps not as explicit as it should be).

### DIFF
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -191,16 +191,16 @@ The SLSA Build level describes the minimum bound for:
 
 <tr id="provenance-exists"><td>Provenance Exists<td>
 
-The build process MUST generate provenance describing how the package was
-produced.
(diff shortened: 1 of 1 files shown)


docs/spec/v1.0/requirements.md
```diff
@@ -191,16 +191,16 @@ The SLSA Build level describes the minimum bound for:
 
 <tr id="provenance-exists"><td>Provenance Exists<td>
 
-The build process MUST generate provenance describing how the package was
-produced.
+The build process MUST generate provenance which unambiguously identifies the
```
MarkLodato, line 194:
```suggestion
The build process MUST generate provenance that unambiguously identifies the
```

(that == restrictive clause, which == informational clause)

MarkLodato (approved):
Thanks!

PR #607, opened 2023-02-06
Prefer 'authentic'/'authenticity' over 'authenticated'

di:
This resolves a TODO and some confusion around the use of the word "authenticated":

> **TODO:** The term "Authenticated" is not quite right, more "authenticatable"
> but that's not a word.

by preferring to speak to the verifiable integrity of the SLSA provenance file instead.

We still retain the use of the word "authenticated", but only when talking about authentication of author/reviewer identities (e.g. with 2FA).


docs/spec/v1.0/requirements.md
```diff
@@ -182,7 +182,7 @@ package was produced.
 The SLSA Build level describes the minimum bound for:
 
 -   *Completeness:* What information is contained in the provenance?
--   *Authenticity:* How strongly can the provenance be tied back to the builder?
+-   *Integrity:* How strongly can the provenance be tied back to the builder?
```
olivekl, line 185: Isn't this the definition of authenticity, though? Whether something can be traced back to its origin? Whereas integrity would be whether something has been tampered with.

olivekl (commented):
I agree that the section as written ("provenance is authenticated") is confusing and another word that captures the idea of "authenticatable" would be better. 

Will switching to "integrity" throughout this section introduce new confusion, though? Most of the suggested changes make sense, but in line 185, "integrity" is used with a definition that sounds more like "authenticity" (see in-line comment). 

Could "verifiable" replace "authenticated," or would that open a different can of worms?

docs/spec/v1.0/requirements.md
di: My perspective is that "authentication" is really about digital signing of the attestation file, which ensures the integrity of everything in that file, as the builder created it. There are two levels of integrity:

- the signature is invalid/nonexistent, you can't trust anything in the provenance file
- the signature checks out, you can trust everything in the file that has been signed over, including who it says the builder was

Additionally, if signing includes a notion of identity (e.g. Sigstore, or some other way to tie keys to builders) then the signature itself _also_ can tie the provenance back to the builder.

So while this definition still holds, maybe we want to be less specific about what "integrity" gets us? (because it's not just tying back to the builder)

di:
> Could "verifiable" replace "authenticated," or would that open a different can of worms?

I think we should think of integrity as a prerequisite for "verification". I would say that "verification" is making sure that some provenance file aligns with some policy for consuming the corresponding artifact, and this aligns with [the current description of the verification process](https://slsa.dev/spec/v1.0/terminology#verification-model):

> artifacts are verified to ensure they meet the producer defined expectations

whereas "integrity" is mostly abut a digital signature, which is what "authentication" was mostly referring to originally.

docs/spec/v1.0/requirements.md
joshuagl: The authenticity requirement for provenance provides two features:
1. identifying the build system — which entities are necessary to trust
2. ensuring integrity of the provenance — that it was not tampered with

The proposed changes here retain 2, but lose 1. It's not sufficient for the provenance to retain integrity, we must also what entity generated the provenance and whether that entity is one that we trust.

di: Isn't that just "[verification](https://slsa.dev/spec/v1.0/terminology#verification-model)" though? If not, what's the difference or relationship between "authentication" and "verification"?

MarkLodato:
To clarify, the TODO meant specifically the part of speech "Authenticated". The terms in the ["Degree" column](https://deploy-preview-607--slsa.netlify.app/spec/v1.0/requirements#build-levels) are currently adjective-like descriptors: "Exists", "Authenticated", and "Non-forgeable". The intention was to describe the degree to which the provenance generation is trustworthy. "Authenticated" fits as a part of speech but meaning-wise is technically incorrect because it's not actually authenticat**ed** by the builder (just signed). We could use "Authenticity", but then it would be the wrong part of speech (as would "Integrity"). Other options might be "Able to be authenticated" (too wordy) or "Signed" (overly prescriptive).

Though I do wonder if there is a better term that is neither "integrity" nor "authenticity", since the requirements here are more than just about signatures. In my mind, L1 = untrusted, L2 = partially trusted, L3 = fully trusted. Any ideas for terms here?

As for "integrity" vs "authenticity", I do think "authenticity" is slightly more accurate but I'm not strongly opposed to "integrity" if that is the consensus. Might be worth asking for more input.

olivekl:
> We could use "Authenticity", but then it would be the wrong part of speech (as would "Integrity"). Other options might be "Able to be authenticated" (too wordy) or "Signed" (overly prescriptive).

Authentic? Confirmable?

di:
OK, I rolled back some of these changes a bit. I agree with @MarkLodato that "authenticated" is technically incorrect and generally a confusing word to use here, but I think "authentic" and "authenticity" should be fine. PTAL!

docs/spec/v1.0/requirements.md
```diff
@@ -214,18 +214,26 @@ bi-directionally translatable to SLSA Provenance.
...
--   Identify the transitive closure of the build system, i.e. which entities
-    are necessary to trust.
--   Verify the integrity of the provenance, i.e. that it was not tampered with
-    after the build.
+-   *Ensure integrity:* Verify that the digital signature of the provenance
+    attestation is valid and the provenance was not tampered with after the
+    build.
+-   *Define trust:* The provenance attestation identifies the build system or
+    other entities that are necessary to trust in order to trust the artifact
+    they produced.
```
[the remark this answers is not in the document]
di: Yes, when reviewing with @olivekl we identified this as a term that's probably too technical for the average audience and could probably be better explained in plain english. Since I was rewriting this section anyways, I went ahead and did that, but let me know if this could be more clear.

joshuagl (approved):
I like how this has shaped up, thanks!

olivekl (approved):
LGTM, this is clearer! 
One point that comes out more now: "Provenance Authentic" and "Provenance Non-forgeable" would be more easily understood as "Authentic Provenance" and "Non-forgeable Provenance." I don't think it's a problem if the order isn't parallel with "Provenance Exists."

di:
> "Provenance Authentic" and "Provenance Non-forgeable" would be more easily understood as "Authentic Provenance" and "Non-forgeable Provenance." 

I agree these read a bit oddly. What about "Provenance is Authentic" and "Provenance is Non-forgeable"?

docs/spec/v1.0/requirements.md
```diff
@@ -460,6 +459,6 @@ The consumer MUST ensure that artifacts they consume are built on
 SLSA-conformant build systems. Consumers MUST either audit the build systems
 themselves using the prompts in [verifying systems](verifying-systems.md) or
 rely on the [SLSA certification program](certification.md).
-  
+
 > **TODO:** Anything else? Do they need to make risk-based decisions? Respond to
-> errors/warnings? Do consumers trust builders, or is that up to the package ecosystem?
+> errors/warnings? Do consumers trust builders, or is that up to the package ecosystem?
```
di, line 464:
```suggestion
> errors/warnings? Do consumers trust builders, or is that up to the package ecosystem?

```

olivekl:
> > "Provenance Authentic" and "Provenance Non-forgeable" would be more easily understood as "Authentic Provenance" and "Non-forgeable Provenance."
> 
> I agree these read a bit oddly. What about "Provenance is Authentic" and "Provenance is Non-forgeable"?

That's even clearer, +1.

MarkLodato:
Thanks @di and @olivekl!

PR #611, opened 2023-02-08
Fix version selector

lehors:
The selector currently offers to switch between versions even when the page does not exist in the other versions, which leads to a 404. This change fixes that by disabling the option if the other version of the page doesn't exist.

To replicate the problem simply go to: https://slsa.dev/spec/v1.0/principles then select "Version 0.1" -> 404

With this fix, the version 0.1 is disabled in the selector.

### DIFF
--- a/docs/_includes/versions-dropdown.html
+++ b/docs/_includes/versions-dropdown.html
@@ -9,16 +9,19 @@
   {%- endif %}">
 {%- for item in versions reversed %}
 {%- unless item[1].hidden and spec_version != item[0] %}
-  {% assign page_url_parts = page.url | split: '/' %}
-  {% assign page_url = page_url_parts | join: '/' %}
+  {%- assign page_url_parts = page.url | split: '/' %}
+  {%- assign page_url = page_url_parts | join: '/' %}
   {% comment %}
     The url of index pages don't end with index but with /.
     Beware, the contains test is string based so it could give a false positive
     in case of a page name collision, such as if you have "foo" and "foobar"
   {% endcomment %}
-  {% assign terminator = page.url | slice: -1 %}
+  {%- assign terminator = page.url | slice: -1 %}
   {%- if terminator == "/" or item[1].pages == nil or item[1].pages contains page_url_parts.last %}
     <option value="{{ page_url | replace: spec_version, item[0] | relative_url }}" {% if spec_version == item[0] %}selected{% endif %} class="inline-block">{{item[1].name}}</option>
+  {%- elsif terminator != "/" and item[1].pages != nil %}
+    {% comment %}when the page doesn't exist disable the option{% endcomment %}
+    <option disabled class="inline-block">{{item[1].name}}</option>
   {% endif %}
 {%- endunless %}
 {%- endfor %}
--- a/docs/_includes/versions-dropdown.html
+++ b/docs/_includes/versions-dropdown.html
@@ -12,16 +12,18 @@
   {%- assign page_url_parts = page.url | split: '/' %}
   {%- assign page_url = page_url_parts | join: '/' %}
   {%- assign other_version = page_url | replace: spec_version, item[0] | relative_url %}
-  {% comment %}
+  {%- comment %}
     If the other version of the page doesn't exist disable the option
-  {% endcomment %}
+  {%- endcomment %}
   {%- assign p = site.pages | where_exp: "page", "page.url contains other_version" | first %}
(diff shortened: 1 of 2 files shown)


lehors:
> > With this fix, only version 1.0 is offered in the selector.
> 
> Is that desirable? As a user, I'd prefer to know that there exists another version. Maybe just redirect the user to the index page in this case? Or just add a placeholder file for the other versions that say the page doesn't exist in this version?

If we want to display the other version anyway, for awareness, then I think the only reasonable thing to do is indeed to point to the index page. It might be a bit confusing but definitely better than a 404. :-)
Let me give it a try.

MarkLodato:
optional: Rather than having to maintain the versions.yml file with every page, and writing code to do the redirection if the page doesn't exist, it might be easier to just ensure that every page name exists in every version. You can use the `redirect` layout to redirect non-existent pages to the index.

di:
The selector could just set the `disabled` attribute on options for versions for which the page doesn't exist:

image

lehors:
> The selector could just set the `disabled` attribute on options for versions for which the page doesn't exist:

That's an interesting idea!

lehors:
> optional: Rather than having to maintain the versions.yml file with every page, and writing code to do the redirection if the page doesn't exist, it might be easier to just ensure that every page name exists in every version. You can use the `redirect` layout to redirect non-existent pages to the index.

I don't like the idea of having to maintain more data as part of versions.yml either for that matter but, I think having redirects for files that never existed all over the place is a hack and not good practice.

I'm still finding my way around Jekyll but I think there is a way to gather the list of pages without having to manually put it into the versions.yml file. That would be even better.

lehors:
Ok, so I implemented @di 's idea. Let me know what you think of it. You can directly go to:
https://deploy-preview-611--slsa.netlify.app/spec/v1.0/principles

di (approved):
LGTM!

lehors:
All right, I found a way to simply test the existence of a page without adding any additional data. (Slowly getting to know Jekyll...)

I think this is it!

docs/_includes/versions-dropdown.html
```diff
@@ -9,9 +9,20 @@
...
+  {% comment %}
+    If the other version of the page doesn't exist disable the option
+  {% endcomment %}
+  {%- assign p = site.pages | where_exp: "page", "page.url contains other_version" | first %}
+  {%- if p %}
+    {%- assign disabled = "" %}
+  {%- else %}
+    {%- assign disabled = "disabled " %}
+  {%- endif %}
+<option {{ disabled }} value="{{ other_version }}" {% if spec_version == item[0] %}selected{% endif %} class="inline-block">{{item[1].name}}</option>
```
MarkLodato, line 24: I like your change! It made it much easier to read.

Here's a suggestion to go one step further, since selected and disabled are mutually exclusive. (The `p and` is probably unnecessary, but it probably makes it more clear.)

```suggestion
  {%- if p and spec_version == item[0] %}
    {%- assign state = "selected" %}
  {%- elsif p %}
    {%- assign state = "" %}  
  {%- else %}
    {%- assign state = "disabled" %}
  {%- endif %}
<option {{ state }} value="{{ other_version }}" class="inline-block">{{item[1].name}}</option>
```

MarkLodato (approved):
Nice!!!

docs/_includes/versions-dropdown.html
```diff
@@ -9,9 +9,20 @@
...
 {%- unless item[1].hidden and spec_version != item[0] %}
-  {% assign page_url_parts = page.url | split: '/' %}
-  {% assign page_url = page_url_parts | join: '/' %}
-  <option value="{{ page_url | replace: spec_version, item[0] | relative_url }}" {% if spec_version == item[0] %}selected{% endif %} class="inline-block">{{item[1].name}}</option>
+  {%- assign page_url_parts = page.url | split: '/' %}
+  {%- assign page_url = page_url_parts | join: '/' %}
+  {%- assign other_version = page_url | replace: spec_version, item[0] | relative_url %}
+  {% comment %}
+    If the other version of the page doesn't exist disable the option
+  {% endcomment %}
```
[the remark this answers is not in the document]
lehors: I thought this didn't matter for the comment tag but I was wrong... Thanks!

docs/_includes/versions-dropdown.html
lehors: Yes, that's even nicer. Thanks!

PR #615, opened 2023-02-08
Remove buildType/predicateType URL trailing slash

### LINKED ISSUE #609: Provenance: canonicalize trailing slashes
Previously, the `predicateType` and `buildType` URIs omitted the trailing slash in the JSON file, and this matched the URL in the browser because we used files like `v0.1.md`. Now that we have moved to directories, the server redirects to a URL containing a trailing slash, which disagrees with the URI in the JSON. We need to resolve this somehow.

A few ideas (not exhaustive):

- Do something with our server (either in Jekyll or by switching host) so that the URL always omits the trailing slash, redirecting slash to no-slash.
- Use a slash in the URI.
- Have some documentation explaining which to use. (Right now https://slsa.dev/github-actions-workflow/v0.1 does not explicitly document the `buildType` string, though it should.)
- Instruct that either must be accepted (seems brittle).

Example: The spec says `predicateType: "https://slsa.dev/provenance/v1"` (no slash), but if you go to that URL, it redirects to https://slsa.dev/provenance/v1/ (with slash).

di:
Fixes #609.


di:
The following URLs no longer have a trailing slash:

- before: https://slsa.dev/github-actions-workflow/v0.1/
    - after: https://deploy-preview-615--slsa.netlify.app/github-actions-workflow/v0.1
- before: https://slsa.dev/provenance/v1/
    - after: https://deploy-preview-615--slsa.netlify.app/provenance/v1

docs/provenance/v1.md
```diff
@@ -33,7 +33,7 @@ interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
 Provenance is an attestation that the `builder` produced the `subject` software
 artifacts through execution of the `buildDefinition`.
 
-![Build Model](model.svg)
+![Build Model](../images/v1/model.svg)
```
MarkLodato, line 36: I think the name is too generic. Maybe "provenance_v1_model.svg"? Alternatively could keep it in the same location and just refer to it as "v1/model.svg".

MarkLodato (approved):
Thanks!

docs/provenance/v1.md
di: Which would you prefer? Not sure I understand the second option, is that just leaving it as-is?

MarkLodato: Sorry for the confusion. The second option is leaving it as-is. Yes, I think that's the easiest thing if you agree.

MarkLodato: Sorry I was again ambiguous. I meant the original location (provenance/v1/model.svg) seems OK to me, so long as having v1.md and v1/ at the same time works. (I think it does but worth checking.)

di: To close the loop: creating a `v1.md` file and a `v1` directory in the same parent directory causes the slash to return, so I opted to move the image to a filepath that includes "provenance" to be more clear.

PR #619, opened 2023-02-16
Split 'Requirements' page into 'Producing artifacts' and 'Verifying a…

kpk47:
…rtifacts'


kpk47:
@olivekl I'm doing some restructuring here and would like your review if you have time.

lehors (changes requested):
I think this is going in the right direction. Thanks!

docs/spec/v1.0/requirements.md
```diff
@@ -92,20 +80,33 @@ A package's <dfn>producer</dfn> is the organization that owns and releases the
...
+
+### Distribute provenance
+
+The producer MUST distribute provenance to artifact consumers. The producer
+MAY delegate this responsibility to the package ecosystem, provided that the
+package ecosystem is capable of distributing provenance.
+
+### How a producer implements SLSA
+>
+> **TODO:** Do we still need this content?
```
olivekl, line 104: I don't think so? Or, if needed, does it belong on this page? I think it would do better under "Getting Started."

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
+# Verifying Artifacts
+
+This page describes how to verify artifacts and their SLSA provenance. The
+intended audience is system implementers.
+
+## Overview
+
+Although there is no formal requirement that consumers verify SLSA provenance,
```
olivekl, line 8: Multiple points:

- The previous section said "the intended audience is system implementers" but this first sentence is about consumers, which is confusing. 
- Is bringing up the idea of "formal requirement" even necessary? 
-  "Third-party verification" is introduced but not included in the table below. 
- The table says the consumer should opt in to verification, but it's not clear from the text what that means or how they'd do it. 

Overall, this section could be clearer about who is actually responsible for what.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+<tr>
+  <td colspan=2><a href="#verifying-expectations-for-artifacts">
+    Verifying expectations for artifacts</a>
+  <td>✓<td>✓<td>✓
+<tr>
+  <td rowspan=1><a href="#consumer">Consumer</a>
+  <td colspan=2>Opt-in to verification
+  <td>✓<td>✓<td>✓
+  </table>
+
```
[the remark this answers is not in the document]
olivekl: +1, are three column needed if all three columns have check marks for all rows?

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+distribution. Every package has an ecosystem, whether it is formal or ad-hoc.
+Some ecosystems are formal, such as language distribution (e.g.
+[Python/PyPA](https://www.pypa.io)), operating system distribution (e.g.
+[Debian/Apt](https://wiki.debian.org/DebianRepository/Format)), or artifact
+distribution (e.g. [OCI](https://github.com/opencontainers/distribution-spec)).
+Other ecosystems are informal, such as a convention used within a company. Even
+ad-hoc distribution of software, such as through a link on a website, is
+considered an "ecosystem".
+
+The package ecosystem is responsible for ensuring that consumers only use
```
olivekl, line 50: A package ecosystem is defined as a set of conventions and tooling. But here a package ecosystem is being given a responsibility. That seems to assume humans behind the ecosystem have the responsibility, not the tooling and conventions.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+artifacts that meet the producers' stated expectations. The pacakge ecosystem
+may also be responsible for providing provenance to consumers.
+
+### Setting Expectations
+
+<dfn>Expectations</dfn> define the allowed values for
+[`buildType`](/provenance/v1#buildType) and
+[`externalParameters`](/provenance/v1#externalParameters)
+for a given package (assuming the SLSA provenance format) in order to address
```
olivekl, line 59: assuming it uses the SLSA provenance format?

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+the [build integrity threats](threats#build-integrity-threats).
+> **TODO:** link to more concrete guidance once it's available.
+
+It is important to note that expectations are tied to a *package name*, whereas
+provenance is tied to an *artifact*. Package ecosystem's using the
```
olivekl, line 64:
```suggestion
provenance is tied to an *artifact*. Package ecosystems using the
```

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+### Setting Expectations
+
+<dfn>Expectations</dfn> define the allowed values for
+[`buildType`](/provenance/v1#buildType) and
+[`externalParameters`](/provenance/v1#externalParameters)
+for a given package (assuming the SLSA provenance format) in order to address
+the [build integrity threats](threats#build-integrity-threats).
+> **TODO:** link to more concrete guidance once it's available.
+
+It is important to note that expectations are tied to a *package name*, whereas
```
olivekl, line 63: Will the reader understand the distinction being made here? My best guess is that it means that the expectations for the same artifact can be different in different ecosystems (i.e., with different package names)?

docs/spec/v1.0/verifying-artifacts.md
olivekl: Having read to the end, I think this section could also be more useful if it introduced at a high level the "pipeline" of verification that's in the table below (explaining that the process involves provenance distribution, setting expectations, verifying, and optionally opt-ing in (for consumers).

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+distribution (e.g. [OCI](https://github.com/opencontainers/distribution-spec)).
+Other ecosystems are informal, such as a convention used within a company. Even
+ad-hoc distribution of software, such as through a link on a website, is
+considered an "ecosystem".
+
+The package ecosystem is responsible for ensuring that consumers only use
+artifacts that meet the producers' stated expectations. The pacakge ecosystem
+may also be responsible for providing provenance to consumers.
+
+### Setting Expectations
```
olivekl, line 54: Why is "Provenance distribution" not included as a section?

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+
+<dfn>Expectations</dfn> define the allowed values for
```
olivekl, line 56: I'm having a hard time grasping this section. Would it be possible to add a high-level summary of what "setting expectations" means in more general terms, either here or at the start?

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+<td><td>✓<td>✓
+</table>
+
+### Verifying expectations for artifacts
+
+It is a critical responsibility of the package ecosystem to verify that the
+provenance for a package matches the expectations defined for that package.
+In our threat model, the adversary has ability to invoke a build and to publish
+to the registry but not to write to the source repository, nor do they have
+insider access to any trusted systems. Expectations MUST be sufficient to detect
```
olivekl, line 112: `Expectations MUST be sufficient to detect
or prevent this adversary from injecting unofficial behavior into the package.
Example threats in this category include building from an unofficial fork or
abusing a build parameter to modify the build. Usually expectations identify
the canonical source repository (which is the main external parameter) and
any other security-relevant external parameters.`

This seems more relevant to the "Setting Expectations" section?

MarkLodato (commented):
I think this split is a good start, but the verifying-artifacts page needs much more work. Hopefully we can align it with https://slsa.dev/provenance/v1#verification?

I'm fine merging something that is a work in progress - just let us know what the bar is for merging and i think we can review to that standard.

docs/spec/v1.0/requirements.md
```diff
@@ -92,20 +80,33 @@ A package's <dfn>producer</dfn> is the organization that owns and releases the
...
 company, or even an individual.
 
-The producer is responsible for setting expectations on how the package ought to
-be built and then using that process on every release. How to set expectations
-and what build systems are acceptable is determined by the [package ecosystem]
-and/or [consumer]. Higher SLSA Build levels can be reached by choosing a build
-system that supports that level.
+### Define expectations
+
+The producer MUST set expectations on how the package ought to be built and then
```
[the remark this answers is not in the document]
kpk47: I've reworded this to take implicit vs explicit expectations into account. I'm not sure we can get around the fact that the verifier controls the expectations format. Maybe we should create an expectations format for v2.0.

docs/spec/v1.0/verifying-artifacts.md

docs/spec/v1.0/verifying-artifacts.md
kpk47: Done

docs/spec/v1.0/verifying-artifacts.md

docs/spec/v1.0/verifying-artifacts.md
kpk47: I added some elaboration. Do you think it makes more sense now?

docs/spec/v1.0/verifying-artifacts.md

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,188 @@
...
+There are no formal requirements for how artifact consumers interact with SLSA,
+but they will benefit from verifying an artifact's provenance and the build
+system used to produce the artifact.
+
+The consumer may have to opt-in to enable SLSA verification, depending on the
+package ecosystem.
+
+Consumers may either audit the build systems
+themselves using the prompts in [verifying systems](verifying-systems.md) or
+rely on the [SLSA certification program](certification.md).
```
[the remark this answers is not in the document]

docs/spec/v1.0/verifying-artifacts.md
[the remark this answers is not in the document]
kpk47: I removed the table in favor of more detailed text.

docs/spec/v1.0/verifying-artifacts.md
kpk47: I rewrote this section as a high-level overview of process. Thanks for the suggestion!

lehors (approved):

docs/spec/v1.0/index.md
```diff
@@ -6,6 +6,18 @@ SLSA is a specification for describing and incrementally improving supply chain
 security, established by industry consensus. It is organized into a series of
 levels that describe increasing security guarantees.
 
+> **TODO:** Is there a better place for this paragraph?
+
+Using SLSA is a bit like using nutrition labels. It is the software producer's
```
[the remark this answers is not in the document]
olivekl: Yeah, a more universal analogy would be ideal, but the art provenance maps so much more cleanly. I think the way Mark explains it makes it clear even to someone who doesn't interact with the art world. The idea of Sotheby's = package ecosystem, for example, clarifies things more than the grocery store. +1 to moving the Provenance analogy to the Overview.

docs/spec/v1.0/requirements.md
```diff
@@ -92,56 +80,27 @@ A package's <dfn>producer</dfn> is the organization that owns and releases the
...
-
-The process differs whether or not the path to SLSA is
-through reproducible builds.
+### Define expectations
 
-> **TODO:** Reorganize to match the [Overview](#overview) table, i.e. "define
-> expectations" and "Use an acceptable builds and release process". There might
-> also need to be a first "decide on a build system" step?
+Verifying provenance requires having expectations about the provenance values for
+an authentic artifact. The verifier can set those expectations either implicitly (e.g.
```
di, line 86:
```suggestion
an authentic artifact. The producer can set those expectations either implicitly (e.g.
```

I think? Since we're talking about setting expectations in the "producing artifacts" page, and later we say "the producer MUST meet any expectations they provided to the artifact verifier"

MarkLodato (approved):
I'm good with merging this and iterating. More work is needed, but this seems like a good first step.

di:
I have a review incoming!

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,191 @@
...
+[Python/PyPA](https://www.pypa.io)), operating system distribution (e.g.
+[Debian/Apt](https://wiki.debian.org/DebianRepository/Format)), or artifact
+distribution (e.g. [OCI](https://github.com/opencontainers/distribution-spec)).
+Other ecosystems are informal, such as a convention used within a company. Even
+ad-hoc distribution of software, such as through a link on a website, is
+considered an "ecosystem".
+
+The package ecosystem's maintainers are responsible for ensuring that consumers only
+use artifacts that meet the producers' stated expectations. The package ecosystem
+may also be responsible for providing provenance to consumers.
```
di, line 41: Not sure I agree with this. I think the responsibility of the ecosystem is to a) reliably redistribute artifacts/provenance, b) make the producers expectations known c) provide tooling that makes safe consumption possible. Other than that, I don't think the ecosystem has any responsibility (or ability) to ensure users only consumer artifacts that meet their expectations.

We might want to soften this to something like the above.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -0,0 +1,191 @@
...
+
+-   Create a more concrete guide on how to do expectations
+-   Whether provenance is generated during the initial build and/or
+    after-the-fact using reproducible builds
+-   How provenance is distributed
+-   What happens on failure: blocking, warning, and/or asynchronous notification
+
+### Provenance Distribution
+
+The package ecosystems MUST distribute provenance for the packages they distribute.
```
di, line 169: I think the language here is probably too strong -- ecosystem's won't be able to distribute provenance for everything. Should this say something like "must support distribution of provenance if provided", or something simlar?

docs/spec/v1.0/verifying-artifacts.md
olivekl: Much clearer!

docs/spec/v1.0/requirements.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Requirements
+title: Producing artifacts
```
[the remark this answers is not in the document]
kpk47: Done

docs/spec/v1.0/verifying-artifacts.md

docs/spec/v1.0/verifying-artifacts.md

docs/spec/v1.0/verifying-artifacts.md

di (approved):

PR #621, opened 2023-02-16
Update get-started.md

jonvnadelberg:
Changing scope of document from achieving SLSA 1 to various SLSA levels depending on build system.

Removed Google Cloud Build as a recommended build system

Added table to end containing known build systems and levels they can obtain.

Reformatted document so it is in simple markdown with extra HTML removed.


docs/get-started.md
```diff
@@ -1,136 +1,109 @@
...
+
+SLSA 1 requires that the build process is documented. Some tools suggested below also support signed provenance. Though not required for SLSA 1, signing your provenance increases trust in the document by showing that it has not been tampered with.
+
+### Tooling
+
+A build configuration file (i.e., GitHub workflow) qualifies for SLSA 1. It would be considered unsigned, unformatted provenance.
+
+### Build service plugins or extensions
+
+These options work with your build system to produce unsigned, formatted provenance. They do not qualify for SLSA 2 because they are unsigned and not run by the hosted server:
```
lehors, line 52:
```suggestion
The following options work with your build system to produce unsigned, formatted provenance. They do not qualify for SLSA 2 because they are unsigned and not run by the hosted server:
```

docs/get-started.md
```diff
@@ -1,136 +1,109 @@
...
+
+-   [Azure DevOps extension](https://github.com/slsa-framework/azure-devops-demo)
+-   [Jenkins SLSA generator](https://github.com/slsa-framework/slsa-jenkins-generator)
+-   [Jenkins plugin](https://plugins.jenkins.io/in-toto/)
+
+Downstream users may verify the provenance with [Cue Policies](https://cuelang.org/docs/).
+
+### Build observers with hosted services
+
+These options are user-configured inside a hosted service. They observe the build process and produce signed, formatted provenance. These options do not qualify for SLSA 2 because they are configured by users, not the hosted service.
```
lehors, line 62:
```suggestion
The following options are user-configured inside a hosted service. They observe the build process and produce signed, formatted provenance. These options do not qualify for SLSA 2 because they are configured by users, not the hosted service.
```

docs/get-started.md
```diff
@@ -1,136 +1,109 @@
...
+## Choosing your SLSA level
+
+For all [SLSA levels](https://slsa.dev/spec/v0.1/levels), you follow the same steps:
+
+1)  Generate provenance, i.e., document your build process
+2)  Make the provenance available, to allow downstream users to verify it
+
+What differs for each level is the robustness of the build and provenance. For more information about provenance see [https://slsa.dev/provenance/v0.2](https://slsa.dev/provenance/v0.2).
+
+The builders suggested in this guide meet SLSA expectations. If you believe a build option is missing or misclassified, please [open an issue](https://github.com/slsa-framework/slsa/pull/589).
```
lehors, line 16: We need to soften this because there is no way to test it yet and no certification program and we should point out that the reason we are discussing specific tools is because they are freely available. Besides, the link is wrong. I suggest something like:
```suggestion
The tools discussed in this guide are freely available and believed to meet SLSA expectations. The [Builder SLSA levels section](#builder-slsa-levels) provides a more complete list. If you think a build option is misclassified or want to add one, please [open an issue](https://github.com/slsa-framework/slsa/issues) or [submit a PR](https://github.com/slsa-framework/slsa/pulls) against this page.
```

docs/get-started.md
```diff
@@ -1,136 +1,109 @@
...
+-   Ensure that build runs cannot influence each other
+-   Produce signed provenance that can be verified as authentic
+
+### GitHub Actions
+
+If you are building on GitHub Actions, adopt the [Language-agnostic GitHub provenance generator / builder](https://github.com/slsa-framework/slsa-github-generator) to make your build qualify for SLSA 3. Consumers can use the [Generic SLSA Verifier](https://github.com/slsa-framework/slsa-verifier) for provenance verification.
+
+## Builder SLSA levels
+
+The following table shows known build software packages and the potential SLSA level they qualify for.  Potential levels are reached with proper provenance.
```
lehors, line 102: We really need to be careful here because this may involve some legal liability.
```suggestion
The following table shows known build software packages and the potential SLSA level they are believed to qualify for.  Potential levels are reached with proper provenance. Again, if you think a build option is misclassified or want to add one, please [open an issue](https://github.com/slsa-framework/slsa/issues) or [submit a PR](https://github.com/slsa-framework/slsa/pulls) against this page.

Note that this list is provided "as is". OpenSSF makes no claim as to the reliability of this information. A certification program is under development to provide a more definitive list.
```

lehors (changes requested):
Thanks, this is better than the previous version but still needs some work. Please, consider the various suggestions I made.

docs/get-started.md
```diff
@@ -1,136 +1,109 @@
...
+For all [SLSA levels](https://slsa.dev/spec/v0.1/levels), you follow the same steps:
+
+1)  Generate provenance, i.e., document your build process
+2)  Make the provenance available, to allow downstream users to verify it
+
+What differs for each level is the robustness of the build and provenance. For more information about provenance see [https://slsa.dev/provenance/v0.2](https://slsa.dev/provenance/v0.2).
+
+The builders suggested in this guide meet SLSA expectations. If you believe a build option is missing or misclassified, please [open an issue](https://github.com/slsa-framework/slsa/pull/589).
+
+SLSA levels are progressive: SLSA 3 includes all the guarantees of SLSA 2, and SLSA 2 includes all the guarantees of SLSA 1. Currently, though, the work required to achieve lower SLSA levels will not necessarily accrue toward the work needed to achieve higher SLSA levels because lower SLSA levels may not be compatible with future implementations. For that reason, **you should start with the highest level that’s possible for your project or organization to avoid wasted work**.
```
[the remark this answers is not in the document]
olivekl: I don't think this is talking about the guarantees of lower levels, but rather the work involved. The current problem is that SLSA 1 requires quite a bit of effort and then you hit a dead end. To move on to SLSA 2, none of the effort from SLSA 1 carries over. And the same for SLSA 3 --- if you adopt FRSCA to achieve SLSA 2, then to meet SLSA 3 you (currently) have to leave FRSCA and adopt GitHub Actions. There's no path from 2 to 3 that would be easier than just starting at 3. 

There's the possibility that users will misunderstand the SLSA levels as steps, and assume that you must start at SLSA 1. Yet in actuality, the easiest option currently is to jump directly to SLSA 3, which gives all the benefits of SLSA 1 and 2, and requires adoption effort only once. It's good to include instructions that warn against that misunderstanding. 

Instead of "future implementations," what about: "will not necessarily accrue toward the work needed for higher levels, since they currently involve adopting different build systems."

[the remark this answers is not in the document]
lehors: All right, now I understand what you're trying to say. Thanks @olivekl 

I agree that this change would make it clearer but I would go a bit further and try to make it a bit less time sensitive by saying something like:

"will not necessarily accrue toward the work needed for higher levels, because achieving a higher level may require migrating to a different build system altogether."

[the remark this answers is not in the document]
jonvnadelberg: Hi @lehors 

I've made the requested changes.  Please check.

docs/get-started.md
```diff
@@ -1,136 +1,111 @@
...
+
+Various build methods require different methods to verify provenance. The SLSA community already has some verification methods and is working on additional solutions for provenance verification. Even if there is currently no simple verification method for a particular build system, adopting these builders now means you will be “SLSA ready” when more ecosystem tooling is released. Note that when you create provenance, you must supply to users a method for interpreting what you have created.
+
+### Provenance formatting
+
+Your provenance format depends on who will be consuming it. See [provenance](/provenance) for an explanation of which format to choose.
+
+### Provenance storage
+
+Containers have a standard place to put the provenance in the OCI container registry. With time, the SLSA community hopes to create standard ecosystem-based repositories for provenance. For now, the convention is to keep the provenance attestation with your artifact. Though the sigstore bundle that's used currently is semi-standard, the format of the provenance is currently tool-specific.
```
lehors, line 36: I don't really know what it means to say that "the sigstore bundle that's used currently is semi-standard".
Used by whom? What's a semi-standard? Old standards geezers like myself will wonder. :-)
I suggest avoiding making any claims related to standards which would be hard to backup. And I think it would make sense to capitalized Sigstore and put a link to sigstore.dev:

`Containers have a standard place to put the provenance in the OCI container registry. With time, the SLSA community hopes to create standard ecosystem-based repositories for provenance. For now, the convention is to keep the provenance attestation with your artifact. Though [Sigstore](https://www.sigstore.dev/) is becoming more and more popular, the format of the provenance is currently tool-specific.`

lehors (changes requested):
Thank you, much better!
I just have a final nit - sorry I didn't see it earlier.

docs/get-started.md
jonvnadelberg: @lehors ok, i changed the sentence.   Would that be all?

docs/get-started.md
```diff
@@ -1,136 +1,111 @@
...
-    </div>
-</section>
+## Choosing your SLSA level
+
+For all [SLSA levels](https://slsa.dev/spec/v1.0/levels), you follow the same steps:
+
+1)  Generate provenance, i.e., document your build process
+2)  Make the provenance available, to allow downstream users to verify it
+
+What differs for each level is the robustness of the build and provenance. For more information about provenance see [https://slsa.dev/provenance/v0.2](https://slsa.dev/provenance/v0.2).
```
di, line 14:
```suggestion
What differs for each level is the robustness of the build and provenance. For more information about provenance see [https://slsa.dev/provenance/](https://slsa.dev/provenance/).
```

I think this should probably just link to the latest version of the spec?

docs/get-started.md
lehors: Yes, that's it for me. Thank you!

lehors (approved):

docs/get-started.md
lehors: Good catch.

di (approved):

PR #624, opened 2023-02-17
Rename github-actions-workflow/v0.1 to v1

MarkLodato:
No one is relying on this string yet, and everything else is v1, so it makes sense to just call it v1. We can rename to v2 if we have a breaking change.


joshuagl (approved):
Seems reasonable 👍

PR #626, opened 2023-02-20
Add the SLSA spec as a single page

lehors:
More work can be done on this but with this change one can access the whole SLSA spec on a single page. All the internal links are fixed to be relative to that page, headers are shifted one level done to get the proper TOC.
I primarily focused on the v1.0 version but it also works on the v0.1 version although some headers appear on several pages so when put on the same page it seems a bit odd.
Let me know what you think.

The page is found at /spec/v1.0/onepage and /spec/v0.1/onepage. The version selector even works across! :-)

### DIFF
--- a/docs/_includes/onepage.liquid
+++ b/docs/_includes/onepage.liquid
@@ -1,0 +1,25 @@
+{% comment %}
+This code turns a set of pages contained in a directory into one single page.
+{% endcomment %}
+
+{% assign filenames = include.filenames %}
+{% assign dir = include.dir %}
+
+{% assign files = filenames | split: "," %}
+{% capture pattern %}page.url contains "{{dir}}"{% endcapture %}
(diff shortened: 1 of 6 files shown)


lehors:
You access the preview page directly at: https://deploy-preview-626--slsa.netlify.app/spec/v1.0/onepage

mlieberman85 (approved):
This is awesome! A lot of folks have been asking for something like this.

MarkLodato (approved):
Nice!

joshuagl (approved):
Very nice! 👏

ianlewis:
The sidebar on the right hand isn't scrollable so it's a bit hard to click on sections that are farther down the page.

lehors:
@ianlewis Yes, that is correct. I find this annoying too but this is totally independent from what this PR is about. There is separate work underway to revamp the menus and sidebar which I hope will address that!

Issue #520, opened 2022-10-26
Better define "package" and associated terminology


MarkLodato:
Draft proposal: https://docs.google.com/document/d/1yHc3Bcq8fV5sthCcOhhQBd5UC6xfH9UV-NTry56ocD0/edit

Ideally this would be done in sync with the OpenSSF Best Practices WG, which is also defining similar terminology.

joshuagl:
The draft proposal looks reasonable to me. I agree with the priorities of unambiguous and minimal think the proposal meets those priorities and the goal.

Further, the proposal mostly matches my mental model (from the Linux distribution world) and does not seem to contradict the [Package URL specification](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst).

The only thing which may be missing from the draft is the notion of a versioned package, which we might consider to be the human readable artifact name or package identifier. However, I can't convince myself that such a definition is necessary and thus, per the minimal priority, we should not define it. :-)

MarkLodato:
> The only thing which may be missing from the draft is the notion of a versioned package, which we might consider to be the human readable artifact name or package identifier. However, I can't convince myself that such a definition is necessary and thus, per the minimal priority, we should not define it. :-)

My thoughts exactly. :-)

MarkLodato:
> The only thing which may be missing from the draft is the notion of a versioned package, which we might consider to be the human readable artifact name or package identifier. However, I can't convince myself that such a definition is necessary and thus, per the minimal priority, we should not define it. :-)

My thoughts exactly. :-)

joshuagl:
> Ideally this would be done in sync with the OpenSSF Best Practices WG, which is also defining similar terminology.

I couldn't find where this is happening, do you have any pointers? Has this draft document been shared with that WG?

PR #627, opened 2023-02-20
Terminology to do

jonvnadelberg:
Added TODO entries and definitions.  Alphabetized table entries.

### DIFF
--- a/docs/spec/v1.0/terminology.md
+++ b/docs/spec/v1.0/terminology.md
@@ -22,8 +22,8 @@ supply chains plus its own sources and builds.
 
 | Term | Description | Example |
 | --- | --- | --- |
-| Attestation | An authenticated statement (metadata) about a software artifact or collection of software artifacts. | A feed into automated policy engines, such as in-toto and Binary Authorization. |
 | Artifact | An immutable blob of data; primarily refers to software, but SLSA can be used for any artifact. | A file, a git commit, a directory of files (serialized in some way), a container image, a firmware image. |
+| Attestation | An authenticated statement (metadata) about a software artifact or collection of software artifacts. | A feed into automated policy engines, such as in-toto and Binary Authorization. |
 | [Build] | Process that transforms a set of input artifacts into a set of output artifacts. The inputs may be sources, dependencies, or ephemeral build outputs. | .travis.yml (process) run by Travis CI (platform). |
 | Dependency | Artifact that is an input to a build process but that is not a source. In the model, it is always a package. | Alpine package (package) distributed on Alpine Linux (platform). |
 | Ecosystem | Environment where the software is tightly integrated with other systems, technologies, infrastructure, and domains (people, organizations and external environments) and where it is configured to support local clinical and business processes. | The end-to-end framework for supply chain integrity. |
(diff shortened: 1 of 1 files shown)


jonvnadelberg:
@olivekl Pull request for terminology page.

joshuagl:
Hi @jonvnadelberg, I appreciate your efforts here! I haven't reviewed your changes in depth, because I think we might want to take a different approach to updating the terminology page.

The SLSA team have taken the position that if we can avoid defining a term, we should. In some cases we think we can remove a term by changing (throughout SLSA, not just on the terminology page) to use a term which is more widely recognised. 

A good example of this is package vs. artifact. We've had some discussion about this in https://github.com/slsa-framework/slsa/issues/520 and the Google Doc linked to from that issue. The conclusion we came to was that we should use "artifact" for the output of a build and "package name" (not just "package") only when "artifact" is not clear. This kind of terminology update/review is a much larger task than just ensuring everything is covered, but should greatly simplify things for readers of the SLSA spec.

When we do need to introduce new terms in the terminology page, it could make sense to introduce them in a sequential order that builds on prior defined terms. That way someone can get a complete understanding of new/special terms by reading the page (or section of the page) top-to-bottom.

The definition of all terms into a glossary-like page here makes me question the table layout, but perhaps that's less of a problem if we implement the two suggestions above.

What do you think?

jonvnadelberg:
> Hi @jonvnadelberg, I appreciate your efforts here! I haven't reviewed your changes in depth, because I think we might want to take a different approach to updating the terminology page.
> 
> The SLSA team have taken the position that if we can avoid defining a term, we should. In some cases we think we can remove a term by changing (throughout SLSA, not just on the terminology page) to use a term which is more widely recognised.
> 
> A good example of this is package vs. artifact. We've had some discussion about this in #520 and the Google Doc linked to from that issue. The conclusion we came to was that we should use "artifact" for the output of a build and "package name" (not just "package") only when "artifact" is not clear. This kind of terminology update/review is a much larger task than just ensuring everything is covered, but should greatly simplify things for readers of the SLSA spec.
> 
> When we do need to introduce new terms in the terminology page, it could make sense to introduce them in a sequential order that builds on prior defined terms. That way someone can get a complete understanding of new/special terms by reading the page (or section of the page) top-to-bottom.
> 
> The definition of all terms into a glossary-like page here makes me question the table layout, but perhaps that's less of a problem if we implement the two suggestions above.
> 
> What do you think?

Hi @joshuagl 

I think if we are going to have a list of terms we wish to define on a page, that we put them in alphabetical order instead of sequential order because if people are going to a terminology page, they want the definition of a word or phrase, and the easiest way to find that is alphabetically.


A terminology page would be helpful to users when trying to understand what they are reading.  By not defining terms that we are using, it seems to me that we are obfuscating what we are talking about, making implementation less understandable and unclear.


Maybe instead of having the page as it currently exists, and glossary-like page would be helpful to users.  We should avoid explaining and re-explaining the same information on different pages where possible.

MarkLodato:
I just sent out #629 which covers a lot of the changes here, except for "ecosystem" and "project." For those, I think we need a separate 'Package model' section with a bit more terms, rather than expanding the Overview section.

MarkLodato:
Since this PR is out of date, I'll close it to avoid clutter. Feel free to open as separate PR.

PR #628, opened 2023-02-21
Provenance v1: Clarify when to ignore unrecognized fields

MarkLodato:
Previously there was an inconsistency:

- The "Parsing Rules" said that consumers MUST **ignore** unrecognized fields. The intention is that implementations, or future minor versions of the spec, can add extra fields without invalidating what's already there.
- However, the design of `externalParameters` and the corresponding verification instructions assumed that consumers would **reject** unrecognized parameters. The idea here is that an unrecognized parameter cannot be expected, thus it's err on the side of safety an reject.

This PR attempts to fix this issue by stating that consumers MUST ignore unrecognized fields **unless otherwise noted** and noting that `externalParameters` is such an exception.

### Questions for reviewers

Is this sufficient to resolve the issue? Are there still any concerns about confusion or incorrect implementations?

### DIFF
--- a/docs/github-actions-workflow/v1.md
+++ b/docs/github-actions-workflow/v1.md
@@ -39,9 +39,9 @@ may be error prone or require further analysis (such as `pull_request` or
 `repository_dispatch`). To add support for another event type, please open a
 [GitHub Issue][SLSA Issues].
 
-Note: Consumers are REQUIRED to reject unrecognized external parameters, so new
-event types can be added without a major version bump as long as they do not
-change the semantics of existing external parameters.
+Note: Consumers SHOULD reject unrecognized external parameters, so new event
+types can be added without a major version bump as long as they do not change
+the semantics of existing external parameters.
 
 Note: This build type is **not** meant to describe execution of a subset of a
 top-level workflow, such as an action, job, or reusable workflow. Only workflows
--- a/docs/provenance/v1.md
+++ b/docs/provenance/v1.md
@@ -85,18 +85,13 @@ For concrete examples, see [index of build types](#index-of-build-types).
 
 This predicate follows the in-toto attestation [parsing rules]. Summary:
 
--   Consumers MUST ignore unrecognized fields.
+-   Consumers MUST ignore unrecognized fields unless otherwise noted.
 -   The `predicateType` URI includes the major version number and will always
     change whenever there is a backwards incompatible change.
 -   Minor version changes are always backwards compatible and "monotonic." Such
     changes do not update the `predicateType`.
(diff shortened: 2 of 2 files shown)


TomHennen:
Is there any more context on this change?  I'm not sure I fully understand the implications.

MarkLodato:
Sorry about that. I updated the top comment to describe. Does that answer?

docs/github-actions-workflow/v1.md
```diff
@@ -39,9 +39,9 @@ may be error prone or require further analysis (such as `pull_request` or
 `repository_dispatch`). To add support for another event type, please open a
 [GitHub Issue][SLSA Issues].
 
-Note: Consumers are REQUIRED to reject unrecognized external parameters, so new
-event types can be added without a major version bump as long as they do not
-change the semantics of existing external parameters.
+Note: Consumers SHOULD reject unrecognized external parameters, so new event
```
TomHennen, line 42: What's the rationale for changing this MUST to SHOULD?

MarkLodato: Superficially, because the old MUST referred to a non-existent requirement, while the new SHOULD refers to the new SHOULD in provenance/v1.md line 176:

> Verifiers SHOULD reject unrecognized or unexpected fields within `externalParameters`.

Now, why is *that* a SHOULD instead of a MUST? I'm not super confident, but my litmus test was whether there exists a plausible reason why a verifier might ignore unrecognized fields. Two ideas come to mind:

1. I can see this below L3, though perhaps that just means we should say "MUST (at least at L3)".
2. I could believe that there exists a situation where unrecognized parameters are deemed safe, though maybe you could phrase this as "I explicitly allow arbitrary parameters and it's OK" rather than "unrecognized".

So I could be convinced to turn this into a MUST.

@lehors Any opinions? I feel like you have good experience here.

TomHennen: Thanks, that sounds fine to me.  I'll defer to others.

joshuagl: Perhaps we could leave this as a MUST and follow an "explicit beats general" rule? In general we ignore unrecognised fields, but we explicitly state that `externalParameters` they are to be rejected.

joshuagl (approved):
LGTM. We might even resolve SHOULD vs. MUST after the RC?

docs/github-actions-workflow/v1.md
MarkLodato: My inclination is to leave this a SHOULD for now until we really solidify expectations and verification in the spec. So I'll submit this but we can always fix in a future PR.

PR #630, opened 2023-02-22
Add a scrollbar to the TOC sidebar when getting long

lehors:
It's been bugging me for a while and @ianlewis's comment on PR #626 gave me the incentive to try and fix it.
Someone more fluent than me in modern CSS and tailwindcss can probably do better than me (my CSS days go back to CSS2 circa 2008! :-) but I think this is already an improvement.

### DIFF
--- a/docs/_layouts/specifications.html
+++ b/docs/_layouts/specifications.html
@@ -28,7 +28,7 @@
                     <div class="table-of-contents w-full md:w-1/3 md:sticky md:top-32 pb-12 md:pl-12">
                       {% include versions-dropdown.html %}
                       {% include status.html %}
-                        <div class="rounded-lg p-4 border border-gray-400">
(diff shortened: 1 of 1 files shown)


lehors:
The effect is especially visible on the onepage version of the spec: https://deploy-preview-630--slsa.netlify.app/spec/v1.0/onepage

MarkLodato:
This makes the ToC really small on larger screens (max height 384px). Should we wait for Psycle to fix this?

lehors:
Ok, see if this is better now.

MarkLodato (approved):
I'm OK with this change as a temporary hack until Psycle has a proper fix. It improves desktop so that you can actually see the full TOC, while degrading mobile (causes unnecessary and confusing scrolling), which seems like an OK trade.

PR #635, opened 2023-02-23
SLSA v1: update threats for v1

MarkLodato:
- Update Build Lx terminology, links, and concepts. Update some text where the model or concepts changed.
- Add "expectations" and "verification" where policy checks apply.
- Remove references to Source requirements but leave recommended mitigations in place, with an intro paragraph explaining this.
- Update some TODOs, replacing "TODO" with "RFC" to indicate we are asking for feedback.
- Formatting:
  - Replace `<sup>` references with inline links.
  - Remove the "(mitigation)" tags for threats that are not addressed by v1.0.


kpk47 (approved):

docs/spec/v1.0/threats.md
```diff
@@ -449,33 +447,39 @@ Solution: Policy rejects because the build steps did not come from the expected
...
+branch/tag, but using a parameter that injects unofficial behavior.
 
-*Mitigation:* Policy requires that the provenance's build entry point matches an
-expected value.
+*Mitigation:* Policy requires that the provenance's external parameters all
+match expected values.
+
+*Example 1:* MyPackage is supposed to be built from the `release.yml` workflow.
+Adversary builds from the `debug.yml` workflow. Solution: Policy rejects because
+the workflow parameter does not match.
```
joshuagl, line 460: Rather than "policy" and introducing a new, ambiguous, term to the spec we could say:

> Solution: verifying expectations for artifacts fails because the workflow parameter does not match"?

docs/spec/v1.0/threats.md
```diff
@@ -449,33 +447,39 @@ Solution: Policy rejects because the build steps did not come from the expected
...
 
-*Example:* MyPackage is supposed to be built from the `release` workflow.
-Adversary builds from the `debug` workflow. Solution: Policy rejects because the
-entry point does not match.
+*Example 2:* MyPackage's GitHub Actions Workflow uses `github.event.inputs` to
+allow users to specify custom compiler flags per invocation. Adversary sets a
+compiler flag that overrides a macro to inject malicious behavior into the
+output binary. Solution: Policy rejects because it does not allow any `inputs`.
```
joshuagl, line 465: similar comment here re: policy

docs/spec/v1.0/threats.md
```diff
@@ -612,44 +607,43 @@ correct way.
 expected value.
 
 *Example:* MyPackage is expected to be built on Google Cloud Build, which is
-trusted up to SLSA 4. Adversary builds on SomeOtherBuildService, which is only
-trusted up to SLSA 2, and then exploits SomeOtherBuildService to inject bad
+trusted up to Build L3. Adversary builds on SomeOtherBuildService, which is only
+trusted up to Build L2, and then exploits SomeOtherBuildService to inject bad
 behavior. Solution: Policy rejects because builder is not as expected.
```
joshuagl, line 612: similar comment on policy here (even though this line was not changed in this PR)

docs/spec/v1.0/threats.md
```diff
@@ -612,44 +607,43 @@ correct way.
...
 </details>
-<details><summary>Tamper with provenance <span>(SLSA 2)</span></summary>
+<details><summary>Tamper with provenance <span>(Build L2)</span></summary>
 
 *Threat:* Perform a build that would not otherwise pass the policy, then modify
 the provenance to make the policy checks pass.
 
-*Mitigation:* Policy only accepts provenance that was cryptographically signed
-by the public key corresponding to an acceptable builder.
+*Mitigation:* Policy only accepts provenance that was [cryptographically
```
joshuagl, line 645: Similar comment re: policy, perhaps instead

> Provenance verification ensures that the provenance was [cryptographically signed][authentic] by the public key corresponding to an acceptable builder.

docs/spec/v1.0/threats.md
```diff
@@ -721,34 +716,34 @@ availability.
 Threats that can compromise the ability to prevent or detect the supply chain
 security threats above but that do not fall cleanly into any one category.
 
-<details><summary>Tamper with policy <span>(TBD)</span></summary>
+<details><summary>Tamper with policy</summary>
```
joshuagl, line 719: Is "policy" here larger than our notion of expectations?

joshuagl (commented):
This looks good, thanks. I made a few comments on trying to remove/clarify use of the term "policy", because we don't use that term elsewhere in the spec.

I'm also torn on whether we should be including threats in the spec which SLSA does not currently address. 🤔

MarkLodato:
Thanks! I just pushed a change to replace "policy" with "verifier" in most cases. A few others required a bit more wording changes.

MarkLodato:
> I'm also torn on whether we should be including threats in the spec which SLSA does not currently address. 🤔

Yeah, I wasn't sure either but I erred on the side of keeping the change smaller. My inclination is to leave it for the RC but consider removing or altering for the final release. Maybe more clear wording to say that this is prospective or future?

docs/spec/v1.0/threats.md
```diff
@@ -628,27 +630,28 @@ missing.
...
 <details><summary>Tamper with provenance <span>(Build L2)</span></summary>
 
-*Threat:* Perform a build that would not otherwise pass the policy, then modify
-the provenance to make the policy checks pass.
+*Threat:* Perform a build that would not pass expectations, then modify the
+provenance to make the expectations checks pass.
 
-*Mitigation:* Policy only accepts provenance that was [cryptographically
-signed][authentic] by the public key corresponding to an acceptable builder.
+*Mitigation:* Verifier only accepts provenance with a valid [cryptographically
```
joshuagl, line 647: cryptographically signature --> cryptographic signature

joshuagl (approved):
> My inclination is to leave it for the RC but consider removing or altering for the final release. Maybe more clear wording to say that this is prospective or future?

If we have more clear wording that could work. Part of me wants to suggest that these threats become part of the future directions section, but that reduces the value of the threats page so I am in favour of just adding more explicit delineation between currently handled threats and threats we want to include in future.

MarkLodato:
I'm going to merge this to lock in incremental progress. If we have time, let's try to iterate on the wording to make it more clear.

PR #637, opened 2023-02-24
Clarify producer's requirement for setting expectations

kpk47:


docs/spec/v1.0/requirements.md
```diff
@@ -83,12 +83,13 @@ company, or even an individual.
...
-Verifying provenance requires having expectations about the provenance values for
-an authentic artifact. The producer can set those expectations either implicitly (e.g.
-trust on first use) or explicitly (e.g. valid builds come from a known source repo).
-If the expectations are set implicitly, then the producer doesn't need to do anything.
-If the expectations are set explicitly, then the producer MUST provide the
-expectations.
+> ⚠ **RFC:** Is this requirement confusing or unnecessary?
+
+Verifying provenance requires having expectations about the provenance values
+for an authentic artifact. The verifier is ultimately responsible for setting
```
di, line 89: I think the _producer_ sets expectations here, not the verifier, right?

kpk47: No, and that's why we're thinking of deleting this requirement entirely. What we're trying to say is more or less "if the verifier asks you for your source repo, please tell them", but by doing so we've created a weird split responsibility for setting expectations. Maybe it's best to leave it solely to the verifier.

di (approved):

PR #638, opened 2023-02-24
Provenance v1: make `builder.id` required again

### LINKED ISSUE #622: Provenance v1.1+: remove `builder.id`?
### Summary

**Proposal:** Remove the `builder.id` field.

The envelope _mostly_ identifies the builder (through the public key, x.509 SAN, or similar). The question is, why do we need the `builder.id`? It generates a lot of confusion. Can we get rid of it?

### Background

The purpose of the `builder.id` is twofold:

1. If a build system has multiple modes, only some of which satisfy a particular SLSA Build level, and they are all signed by the same identity, then the `builder.id` can be used to differentiate between these modes.
2. Similarly, if an external provenance generator can support multiple build systems (say by reading logs or by the build service calling it) and it signs everything with the same key, the `builder.id` can be used to differentiate between these build systems / callers.

However, this has a few problems:

- This is confusing and not well documented. Several implementers have misunderstood this.
- It makes verification more complex. Instead of just checking the signer, you *also* have to check the `builder.id` and make sure they are consistent. If you forget, it's a security hole.
- If implemented poorly, it can be a security vulnerability.

Relatedly, v1.0 is moving towards having the builder's identity be the sole indicator of SLSA build level; see #618. Prior versions required checking various fields (particularly the "parameterless" requirement of SLSA 4) but v1.0 simplifies this by (a) shifting the burden to make a claim about the SLSA Build level and (b) replacing the "parameterless" with expectations on parameters. This we now have more clear footing to say explicitly that the signer identity is a claim by the builder that the build meets a certain level.

### Proposal

Remove the `builder.id` field entirely and instruct implementers to use different signer identities to represent different builder modes or build levels. This simplifies the spec and avoids a common point of confusion.

### Open questions

Are there any remaining use cases where `builder.id` would still be needed, i.e. where the signer identity is not sufficient?

### Alternatives considered

- Keep the `builder.id` field.
  - Pro: Flexible.
  - Con: Confusing and error prone.
- Replace `builder.id` with a more narrow "mode of operation" or similar, to differentiate between modes that have fundamentally different security properties. The verifier would use this the same as they use `builder.id` today.
  - Pro: Flexible.
  - Pro: Somewhat less confusing and error prone.
  - Con: Still confusing and error prone.
- Replace `builder.id` with an explicit "Claimed SLSA Build level" field that the builder populates.
  - Pro: Explicit, clear.
  - Con: Too easy for verifiers to blindly accept this without "roots of trust" (#618).
  - Con: Confusing if the claim is X but the roots of trust say Y (< X).
  - Con: Cannot differentiate between different modes that have the same SLSA Build level.
- Replace `builder.id` with a list of SLSA requirements that are met.
  - Pro: Explicit, clear.
  - Pro: More flexible.
  - Con: Much more complex for verifiers. The roots of trust now list not just level but also properties that are believed.
  - Con: Confusing if claims disagree with roots of trust.

### Next steps

If you have an opinion one way or the other, please comment here.

I'll send out a pull request that implements this, to see how it looks.

MarkLodato:
- Make `builder.id` required, which is how it was in v0.2. (Before this commit, v1 said that the `id` MAY be implicit from the signer's identity.)
- Replace uses of `builder` with `builder.id`. The latter is the identity, while the former includes more information about the builder.
- Remove TODOs about `id` being optional.

Fixes #622.

### DIFF
--- a/docs/provenance/v1.md
+++ b/docs/provenance/v1.md
@@ -38,7 +38,7 @@ artifacts through execution of the `buildDefinition`.
 The model is as follows:
 
 -   Each build runs as an independent process on a multi-tenant platform. The
-    `builder` is the identity of this platform, representing the transitive
+    `builder.id` identifies this platform, representing the transitive
     closure of all entities that are [trusted] to faithfully run the build and
     record the provenance. (Note: The same model can be used for platform-less
     or single-tenant build systems.)
@@ -69,7 +69,7 @@ The model is as follows:
 
 -   During execution, the build process might communicate with the build
     platform's control plane and/or build caches. This communication is not
-    captured directly in the provenance, but is instead implied by the `builder`
+    captured directly in the provenance, but is instead implied by `builder.id`
     and subject to [SLSA Requirements](/spec/v1.0/requirements). Such
     communication SHOULD NOT influence the definition of the build; if it does,
     it SHOULD go in `resolvedDependencies` instead.
@@ -128,7 +128,8 @@ REQUIRED for SLSA Build L1: `buildDefinition`, `runDetails`
 <tr id="buildDefinition"><td><code>buildDefinition</code>
 <td><a href="#builddefinition">BuildDefinition</a><td>
 
-The input to the build. The accuracy and completeness are implied by `runDetails.builder.id`.
+The input to the build. The accuracy and completeness are implied by
+`runDetails.builder.id`.
 
 <tr id="runDetails"><td><code>runDetails</code>
 <td><a href="#rundetails">RunDetails</a><td>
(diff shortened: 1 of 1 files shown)


docs/provenance/v1.md
```diff
@@ -372,24 +369,21 @@ later and that cannot be easily reproduced.
...
-
-**TODO:** In most cases this is implicit from the envelope layer (e.g. the
-public key or x.509 certificate), which is just one more thing to mess up. Can
-we rescope this to avoid the duplication and thus the security concern? For
-example, if the envelope identifies the build system, this might identify the
-tenant project?
+URI indicating the transitive closure of the trusted builder. This is
+[intended][Step 1] to be the sole determiner of the SLSA Build level.
 
-**TODO:** Provide guidance on how to choose a URI, what scope it SHOULD have,
```
kpk47, line 391: You didn't resolve this TODO. Should we keep it?

MarkLodato: Turned it into an RFC.

PR #652, opened 2023-03-01
Redirect between "v1" and "v1.0"

MarkLodato:
I keep mistyping `/spec/v1` when it's really `/spec/v1.0` and similarly `/provenance/v1.0` when it's really `/provenance/v1`. This commit creates simple redirects to jump to the right one.


TomHennen (approved):

mlieberman85 (approved):
lgtm

PR #664, opened 2023-03-06
SLSA v1 terminology: add Package Model

### LINKED ISSUE #520: Better define "package" and associated terminology

MarkLodato:
This defines the minimal terminology needed for Verifying Artifacts. It is based on the [draft proposal](https://docs.google.com/document/d/1yHc3Bcq8fV5sthCcOhhQBd5UC6xfH9UV-NTry56ocD0/edit) that several people contributed to, with some updated wording and a new introduction.

Design notes:

-   Originally we wanted to use "Project" instead of "Package name", since the former is the concept and the latter is an identifier for the concept. Unfortunately only Python gives this concept a proper name, and some readers misunderstood "Project" to mean more than just the thing within the package ecosystem (e.g. they thought the source repo would also be part of the project, whereas that was not the intent). So "Package name" seems like the best option among our choices.

Fixes #520.

Preview: https://deploy-preview-664--slsa.netlify.app/spec/v1.0/terminology#package-model


TomHennen (changes requested):
The "Example: Open source language distribution" could use some cleaning up.  It looks like it currently using some ambiguous terms ('package') and some terms to avoid ('package manager').

Otherwise this looks good!

MarkLodato:
Good point on "package manager". We do need a term for the client-side tooling. Since I can't come up with a better name, I used "package manager client" to avoid ambiguity, but a shorter name would be preferred.

We'll leave cleanup of existing usage to a separate PR.

TomHennen (approved):

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
 of build types](/provenance/v1#index-of-build-types).
 
+### Package model
+
+Software is distributed in identifiable units called <dfn>packages</dfn>
+according the the rules and conventions of a <dfn>package ecosystem</dfn>.
+Examples of formal ecosystems include [Python/PyPA](https://www.pypa.io),
+[Debian/Apt](https://wiki.debian.org/DebianRepository/Format), and
+[OCI](https://github.com/opencontainers/distribution-spec), while examples of
+informal ecosystems include a linking to a file on a website or distributing
```
joshuagl, line 77:
```suggestion
informal ecosystems include linking to a file on a website or distributing
```
"a linking to ..." reads a bit awkwardly

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
+    <th>Package artifact
+  <tr>
+    <td colspan=4><em>Languages</em>
+  <tr>
+    <td><a href="https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-name-convention">Python</a>
+    <td><a href="https://packaging.python.org/en/latest/glossary/#term-Package-Index">Index</a>
+    <td><a href="https://packaging.python.org/en/latest/glossary/#term-Project">Project</a>
+    <td><a href="https://packaging.python.org/en/latest/glossary/#term-Distribution-Package">Distribution</a>
+  <tr>
+    <td><a href="https://www.npmjs.com/">npm</a> (Javascript)
```
joshuagl, line 132:
```suggestion
    <td><a href="https://www.npmjs.com/">npm</a> (JavaScript)
```

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
+  <tr>
+    <td colspan=4><em>Operating systems</em>
+  <tr>
+    <td><a href="https://wiki.debian.org/Teams/Dpkg">Dpkg </a> (e.g. Debian)
+    <td><em>?</em>
+    <td>Package name
+    <td>Package
+  <tr>
+    <td><a href="https://rpm.org">RPM</a> (e.g. Red Hat)
+    <td>Repositry
```
joshuagl, line 170:
```suggestion
    <td>Repository
```

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
+    <td><a href="https://packaging.python.org/en/latest/glossary/#term-Package-Index">Index</a>
+    <td><a href="https://packaging.python.org/en/latest/glossary/#term-Project">Project</a>
+    <td><a href="https://packaging.python.org/en/latest/glossary/#term-Distribution-Package">Distribution</a>
+  <tr>
+    <td><a href="https://www.npmjs.com/">npm</a> (Javascript)
+    <td><a href="https://docs.npmjs.com/about-the-public-npm-registry">Registry</a>
+    <td><a href="https://docs.npmjs.com/package-name-guidelines">Package Name</a>
+    <td><a href="https://docs.npmjs.com/about-packages-and-modules">Package</a>
+  <tr>
+    <td><a href="https://go.dev/ref/mod">go</a>
```
joshuagl, line 137:
```suggestion
    <td><a href="https://go.dev/ref/mod">Go</a>
```

joshuagl (approved):
This looks good to me. I'm curious for wider feedback.

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
+
+| Term | Description |
+| ---- | ----------- |
+| Package | An identifiable unit of software intended for distribution, ambiguously meaning either an "artifact" or a "package name". Only use this term when the ambiguity is acceptable or desirable. |
+| Package artifact | A file or other immutable object that is intended for distribution. |
+| Package ecosystem | A set of rules and conventions governing how packages are distributed, including how clients resolve a package name into one or more specific artifacts. |
+| Package manager client | Client-side tooling to interact with a package ecosystem. |
+| Package name | <p>A mutable collection of artifacts that all represent different versions of the same software. This is the primary identifier that consumers use to obtain the software. Therefore, it is the thing being protected by SLSA.<p>A package name is specific to an ecosystem + registry, has an owner/maintainer, is more general than a specific hash or version, and has a “correct” source location. A package ecosystem may group package names into some sort of hierarchy, such as the Group ID in Maven, though SLSA does not have a special term for this. |
```
olivekl, line 99: `Therefore, it is the thing being protected by SLSA.`

Seems like one of the most useful points to emphasize. Is there anything to say that could spell this out a bit more? Or is here not the place?

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
+| Package registry | An entity responsible for mapping package names to artifacts within a packaging ecosystem. Most ecosystems support multiple registries, usually a single global registry and multiple private registries. |
+| Publish [a package] | Make an artifact available for use by registering it with the package registry. In technical terms, this means associating an artifact to a package name. This does not necessarily mean fully public; an artifact may be published for only a subset of users, such as internal testing or a closed beta. |
```
olivekl, line 101:
```suggestion
| Publish [a package] | Make an artifact available for use by registering it with the package registry. In technical terms, this means associating an artifact to a package name. This does not necessarily mean making the artifact fully public; an artifact may be published for only a subset of users, such as internal testing or a closed beta. |
```

olivekl (approved):
A few minor comments, but LGTM.

docs/spec/v1.0/terminology.md
kpk47: I'd suggest "a link".

docs/spec/v1.0/terminology.md
kpk47, line 99:
```suggestion
| Package name | <p>The primary identifier referring to a mutable collection of artifacts that all represent different versions of the same software. SLSA protects this identifier from [impersonation? squatting? reuse?].
<p>A package name is specific to an ecosystem + registry and has an official owner/maintainer. It is more general than a specific hash or version and is associated with a “correct” source location. A package ecosystem may group package names into some sort of hierarchy, such as the Group ID in Maven, though SLSA does not have a special term for this. |
```

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
+  <tr>
+    <td><a href="https://maven.apache.org/glossary">Maven</a> (Java)
+    <td>Repository
+    <td>Group ID + Artifact ID
+    <td>Artifact
+  <tr>
+    <td><a href="https://doc.rust-lang.org/cargo/appendix/glossary.html">Cargo</a> (Rust)
+    <td><a href="https://doc.rust-lang.org/cargo/appendix/glossary.html#registry">Registry</a>
+    <td>Crate name
+    <td><a href="https://doc.rust-lang.org/cargo/appendix/glossary.html#crate">Crate</a> (or <a href="https://doc.rust-lang.org/cargo/appendix/glossary.html#artifact">Artifact</a>?)
```
kpk47, line 150: If I understand your definitions correctly, then this entry should be Artifact. Artifact refers specifically to the compiled binary, whereas Crate is ambiguous and may refer to either the compiled binary or its (bundled?) source code.

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
+</table>
+
+Notes:
+
+-   Go: Uses a significantly different model than any other ecosystem, where the
+    package name is a git repository URL. Clients can fetch directly from the
+    git repository, in which case there is no "package". But clients usually
+    instead fetch a zip file from a module mirror that proxies the git
+    repository, and consult with a checksum database to detect tampering. It is
+    not yet clear which of these constitute the "registry", if either.
```
kpk47, line 226: How will we clarify which is the registry? Do we lack information about Go, or is the problem that we need to form an opinion on our definitions or model?

kpk47 (approved):

docs/spec/v1.0/terminology.md
di: I think in this case, the registry would be the (one or more) git repositories, because they are the ones that permit/refuse changes to a given package namespace (in this case, the repo itself) based on ownership/access.

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
+    <th>Package ecosystem
+    <th>Package registry
+    <th>Package name
+    <th>Package artifact
+  <tr>
+    <td colspan=4><em>Languages</em>
+  <tr>
+    <td><a href="https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-name-convention">Python</a>
+    <td><a href="https://packaging.python.org/en/latest/glossary/#term-Package-Index">Index</a>
+    <td><a href="https://packaging.python.org/en/latest/glossary/#term-Project">Project</a>
```
di, line 129:
```suggestion
    <td><a href="https://packaging.python.org/en/latest/glossary/#term-Project">Project Name</a>
```
The "project" is really the entire thing. The "project name" is just the mutable name.

docs/spec/v1.0/terminology.md
```diff
@@ -73,6 +67,164 @@ dependency.
...
+  <tr>
+    <td><a href="https://nixos.org/guides/how-nix-works.html">nix</a> (e.g. <a href="https://nixos.org/">NixOS</a>)
+    <td><em>?</em>
+    <td><a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-store-object">Store Object</a>?
+    <td>Package or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-derivation">Derivation</a>
+  <tr>
+    <td colspan=4><em>Storage systems</em>
+  <tr>
+    <td><a href="https://cloud.google.com/storage/docs/key-terms">GCS</a>
+    <td><em>n/a</em>
```
di, line 197:
```suggestion
    <td><em><a href="https://cloud.google.com/storage/docs/buckets">Bucket</a>.</em>
```
Maybe? Seems to fit to me.

docs/spec/v1.0/terminology.md
di: I agree, I think this is `Artifact`.

docs/spec/v1.0/terminology.md
MarkLodato: Reworded the first half and removed the second half (see other comment).

docs/spec/v1.0/terminology.md
MarkLodato: I don't think so? I think Google (or "GCS") might be the registry here, since it is the entity with control. A bucket is just a unit of the package name, isn't it?

docs/spec/v1.0/terminology.md
MarkLodato: After more thought, I argue that it's the module proxy. I updated the text accordingly. PTAL.

MarkLodato (commented):
Addressed comments. PTAL.

kpk47 (approved):

MarkLodato:
I'll merge this now since I already got several approvals, but if you have further comments, please let me know and I'm happy to fix in a future PR.

docs/spec/v1.0/terminology.md
```diff
@@ -11,13 +11,6 @@ next_page:
 Before diving into the [SLSA Levels](levels.md), we need to establish a core set
 of terminology and models to describe what we're protecting.
 
-## TODO: Terms we still need to define
-
-> **TODO:** Define these terms before the v1.0 release.
-
--   Ecosystem
--   Project
-
```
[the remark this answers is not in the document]
MarkLodato: Oh, thanks for spotting this! Sent out #718 to fix.

PR #667, opened 2023-03-07
Create v1 of verification summaries

### LINKED ISSUE #663: Update VSA with v1?
Has any thought been given to updating the vsa to v1 with SLSA v1?

I think my biggest area of concern might be the [SlsaResult](https://slsa.dev/verification_summary/v0.2#slsaresult) type, it includes SLSA L4 (removed in SLSA v1) and says it "Must be one of these values".

We have an internal implementation of SLSA that still meets the spirit of the old SLSA L4.  It would be nice if it were allowed to include additional values here.  "SHOULD be one of these values" would be good enough for us.

Additionally it could probably be updated to use @marcelamelara's [ResourceDescriptor](https://github.com/in-toto/attestation/pull/143) in a couple places.

Any thoughts?

TomHennen:
Copied from v0.2.md
* materials -> resolvedDependencies
* Relax SlsaResult to allow other fields

Can update to use ResourceDescriptor or AritfactReference if/when desired.

fixes #663

### DIFF
--- a/docs/verification_summary/v1.0.md
+++ b/docs/verification_summary/v1.0.md
@@ -1,0 +1,5 @@
+---
+layout: redirect
+redirect_to_url: /verification_summary/v1
+sitemap: false
+---
--- a/docs/verification_summary/v1.md
+++ b/docs/verification_summary/v1.md
@@ -1,0 +1,274 @@
+---
+title: SLSA Verification Summary Attestation (VSA)
+layout: standard
+hero_text: Verification summary attestations communicate that an artifact has been verified at a specific SLSA level and details about that verification.
+---
+
+## Purpose
+
+Describe what SLSA level an artifact or set of artifacts was verified at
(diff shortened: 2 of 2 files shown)


TomHennen:
@AdamZWu PTAL

AdamZWu:
Looks great. Thanks Tom!

Just a small nit: I thought v1.md would be pointing to v1.0.md, instead of the other way around.

TomHennen:
> Just a small nit: I thought v1.md would be pointing to v1.0.md, instead of the other way around.

This mirrors what /provenance/v1.0.md is doing.  I think the idea is that minor version changes aren't breaking.

joshuagl (approved):
LGTM, thanks!

TomHennen:
Thanks!  Let me know if we need any additional changes prior to SLSA 1.0.

PR #671, opened 2023-03-07
Edit guiding principles

olivekl:
Copyedit to make the important principles, reasoning behind them, and benefits stand out more for readers


docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
-provenance can then be "trusted" up to a given SLSA level. Downstream systems
-can then automatically determine that an artifact meets a given level by
-verifying that it has provenance signed by that public key.
+## Principle 1: Trust systems, verify artifacts
+
+Establish trust in a small number of systems (such as change management, build, and packaging systems) and then automatically verify the many artifacts produced by those systems.
+
+Reason: Trusted computing bases are unavoidable, but analysis to establish that trust is
+difficult and expensive. It's infeasible to harden and audit thousands of
+bespoke build systems.
```
hayleycd, line 16:
```suggestion
bespoke build systems.
```

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
+
+Establish trust in a small number of systems (such as change management, build, and packaging systems) and then automatically verify the many artifacts produced by those systems.
+
+Reason: Trusted computing bases are unavoidable, but analysis to establish that trust is
+difficult and expensive. It's infeasible to harden and audit thousands of
+bespoke build systems.   
+
+We can, however, do this for a small number of general purpose systems. **Once
+those systems are trusted, we can verify the supply chain of _artifacts_ by
+ensuring that they were produced by trusted systems**.
```
hayleycd, line 20:
```suggestion
ensuring that they were produced by trusted systems**.
```

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
+Reason: Trusted computing bases are unavoidable, but analysis to establish that trust is
+difficult and expensive. It's infeasible to harden and audit thousands of
+bespoke build systems.   
+
+We can, however, do this for a small number of general purpose systems. **Once
+those systems are trusted, we can verify the supply chain of _artifacts_ by
+ensuring that they were produced by trusted systems**. 
+
+Benefits: Allows SLSA to scale to entire ecosystems or organizations with a near-constant
+amount of central work.
```
hayleycd, line 23:
```suggestion
amount of central work.
```

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
+    shared build system. Hardening work can be shared across all teams.
+-   Remove the need to trust components. For example, use end-to-end signing
+    to avoid the need to trust intermediate distribution systems.
+
+## Principle 2: Trust code, not individuals
+
+Securely trace all software back to source code.  
+
+Reason: Code is static and analyzable. People, on the other hand, are prone to mistakes,
+credential compromise, and sometimes malicious action.
```
hayleycd, line 51:
```suggestion
credential compromise, and sometimes malicious action.
```

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
+## Principle 3: Prefer attestations over inferences
+
+Require explicit attestations about an artifact's provenance; do not infer
+security properties from a system's configurations.
+
+Reason: Theoretically, access control can be configured so that the only path from
+source to release is through the official channels: the CI/CD system pulls only
+from the proper source, package registry allows access only to the CI/CD system,
+and so on. We might infer that we can trust artifacts produced by these systems
+based on the system's configuration.
```
hayleycd, line 65:
```suggestion
based on the system's configuration.
```

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
+based on the system's configuration. 
+
+In practice, though, these configurations are almost impossible to get right and
+keep right. There are often over-provisioning, confused deputy problems, or
+mistakes. Even if a system is configured properly at one moment, it might not
+stay that way, and humans almost always end up getting in the access control
+lists.  
+
+Access control is still important, but SLSA goes further and **requires proof in
+the form of attestations that the package was built correctly**.
```
hayleycd, line 74:
```suggestion
the form of attestations that the package was built correctly**.
```

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
+based on the system's configuration.
+
+In practice, though, these configurations are almost impossible to get right and
+keep right. There are often over-provisioning, confused deputy problems, or
+mistakes. Even if a system is configured properly at one moment, it might not
+stay that way, and humans almost always end up getting in the access control
+lists.  
+
+Access control is still important, but SLSA goes further and **requires proof in
+the form of attestations that the package was built correctly**.
```
MarkLodato, line 74: Can we keep the term "defense in depth"? This is a widely used security term that many readers will instantly understand.

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
+
+Benefits: The attestation removes intermediate systems from the trust base and ensures
+individuals do not have sufficient permission to tamper with the package.
```
MarkLodato, line 77: The original said "individuals that are accidentally granted access" because the intent of "configured so that the only path from source to release is through the official channels" is not to grant individuals such access. I think it's probably still valuable to keep that.

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
-that are accidentally granted access do not have sufficient permission to tamper
-with the package.
+    team within an organization maintaining their own build system, use a
+    shared build system. Hardening work can be shared across all teams.
+-   Remove the need to trust components. For example, use end-to-end signing
+    to avoid the need to trust intermediate distribution systems.
+
+## Principle 2: Trust code, not individuals
+
+Securely trace all software back to source code.
```
MarkLodato, line 48: This drops an important concept:

> rather than individual maintainers who have write access to package registries

Without SLSA, we solely trust the individual maintainers (or their credentials) that the binary we receive matches the source code that we expect and review. SLSA addresses this by removing (or reducing) the need to trust these maintainers.

Could we add this back somehow?

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
-
-For example, a security engineer may analyze the architecture and implementation
-of a build system to ensure that it meets the SLSA build requirements. Once the
-analysis is complete, the public keys used by the build system to sign
-provenance can then be "trusted" up to a given SLSA level. Downstream systems
-can then automatically determine that an artifact meets a given level by
-verifying that it has provenance signed by that public key.
+## Principle 1: Trust systems, verify artifacts
+
+Establish trust in a small number of systems (such as change management, build, and packaging systems) and then automatically verify the many artifacts produced by those systems.
```
MarkLodato, line 12:
```suggestion
Establish trust in a small number of systems---such as change management, build,
and packaging systems---and then automatically verify the many artifacts
produced by those systems.
```

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
+
+Reason: Trusted computing bases are unavoidable, but analysis to establish that trust is
+difficult and expensive. It's infeasible to harden and audit thousands of
+bespoke build systems.  
+
+We can, however, do this for a small number of general purpose systems. **Once
+those systems are trusted, we can verify the supply chain of _artifacts_ by
+ensuring that they were produced by trusted systems**.
```
MarkLodato, line 20: nit: Could we be a bit more verbose here? I find this new version a bit harder to read because it's so terse.

In my mind, the gist of this principle is:

- We have no choice but to trust some number systems.
- It is hard, manual work to harden and verify each such system. (Plus, each system we trust increases attack surface == increases the chance of compromise.)
- It is easy, automatic work to verify that an artifact was produced by such as system.
- Therefore, to scale and to "reduce attack surface", we want to trust a small, limited number of systems and then verify arbitrarily many artifacts (or packages?) produced by those systems. This gives us a near-constant amount of central work (and have a near-constant amount of attack surface) regardless of how many packages we support.
- Contrast this to the alternative, where each package uses its own trusted systems. This would result in a roughly linear scaling of work and attack surface.

Technically all that is there, but I am not sure if it comes across super clearly. Though maybe my worry is unfounded. Might be valuable to get other opinions here.

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
 
 <div class="subtitle">
 
-This page provides a background on the guiding principles behind SLSA. It is
-intended to help the reader better understand SLSA's design decisions.
+This page describes the three guiding principles behind SLSA's design
+decisions.
```
MarkLodato, line 6: nit: "the three" makes it sound like there are exactly three, official principles, especially when they're numbered 1-3. I don't think this list is necessarily complete or immutable. Any thoughts on softening this to remove "three" and the numbering below?

docs/spec/v1.0/principles.md
```diff
@@ -2,79 +2,76 @@
...
-of a build system to ensure that it meets the SLSA build requirements. Once the
-analysis is complete, the public keys used by the build system to sign
-provenance can then be "trusted" up to a given SLSA level. Downstream systems
-can then automatically determine that an artifact meets a given level by
-verifying that it has provenance signed by that public key.
+## Principle 1: Trust systems, verify artifacts
+
+Establish trust in a small number of systems (such as change management, build, and packaging systems) and then automatically verify the many artifacts produced by those systems.
+
+Reason: Trusted computing bases are unavoidable, but analysis to establish that trust is
```
MarkLodato, line 14: suggestion: make "Reason:" and "Benefit:" bold so that it stands out more, and maybe consider renaming to "Reasoning" to emphasize it's more about the thinking than an iron-clad argument?

MarkLodato (changes requested):

docs/spec/v1.0/principles.md
olivekl: I'll take out the numbers if they're not immutable/complete. I was actually trying to draw attention to the "finalness" by numbering them that way.

docs/spec/v1.0/principles.md
olivekl: Done

docs/spec/v1.0/principles.md
olivekl: done

docs/spec/v1.0/principles.md
olivekl: added back

docs/spec/v1.0/principles.md
olivekl: added back

docs/spec/v1.0/principles.md
olivekl: Your explanation is clearer -- I added text along these lines.

docs/spec/v1.0/principles.md
```diff
@@ -2,24 +2,29 @@
...
+produced by those systems.
 
-Reason: Trusted computing bases are unavoidable, but analysis to establish that trust is
-difficult and expensive. It's infeasible to harden and audit thousands of
-bespoke build systems.  
+**Reasoning**: Trusted computing bases are unavoidable---there's no choice but to
+trust some systems. It's infeasible to harden and audit thousands of bespoke build systems.
+It's difficult and expensive manual work to harden and verify each individual system,
+and each system increases the attack surface. Verifying that an artifact is produced by a
+trusted system, though, is easy to automate.
```
MarkLodato, line 20: Second and third sentence repeat each other. Here's a suggestion to clean up wording (and wrap to 80 columns while we're at it):

```suggestion
**Reasoning**: Trusted computing bases are unavoidable---there's no choice but
to trust some systems. Hardening and verifying systems is difficult and
expensive manual work, and each trusted system expands the attack surface of the
supply chain. Verifying that an artifact is produced by a trusted system,
though, is easy to automate.
```

docs/spec/v1.0/principles.md
```diff
@@ -2,24 +2,29 @@
...
 
-We can, however, do this for a small number of general purpose systems. **Once
-those systems are trusted, we can verify the supply chain of _artifacts_ by
-ensuring that they were produced by trusted systems**.
+To simultaniously scale and reduce attack surfaces, it is most efficient to trust a limited
+numbers of systems and then automate verification of the artifacts produced by those systems.
+The attack surfaces and work to estalish trust does not scale with the number of artifacts produced,
```
MarkLodato, line 24:
```suggestion
The attack surface and work to establish trust does not scale with the number of artifacts produced,
```

docs/spec/v1.0/principles.md
```diff
@@ -43,22 +48,22 @@ manual security analysis. Where possible:
...
-Securely trace all software back to source code.  
+Securely trace all software back to source code rather than trust individual maintainers who have write access to package registries.
 
-Reason: Code is static and analyzable. People, on the other hand, are prone to mistakes,
+**Reasoning**: Code is static and analyzable. People, on the other hand, are prone to mistakes,
 credential compromise, and sometimes malicious action.
 
-Benefits: Removes the possibility for an individual maintainer to tamper with source code
+**Benefits**: Removes the possibility for an individual maintainer to tamper with source code
 after it has been committed.
```
MarkLodato, line 59: suggestion: call out compromised maintainer creds since this is a major risk and also more palatable to people on a team. No one wants to think their colleague is malicious, while credential compromise is widely recognized as a problem.

```suggestion
**Benefits**: Removes the possibility for an individual maintainer---or an
attacker abusing a compromised maintainer's credentials---to tamper with source
code after it has been committed.
```

MarkLodato (changes requested):
Looks good, just a few more suggestions

docs/spec/v1.0/principles.md
olivekl: Done, thanks!

MarkLodato (approved):

Issue #235, opened 2021-12-01
Project maintainer package vs. Distribution package


mlieberman85:
SLSA currently doesn't provide guidance or elaborate on the distinction between the things that are being built and packaged and the how the packaging itself is being maintained. It might

For example: A python package that is included in a linux distribution will have two sets of "source" to deal with. First is the actual python code, and second is the packaging code as maintained by the linux distribution itself.

This differs from how a package built and distributed by the package maintainers themselves might work since that will have just a single set of maintainers and the code is for all intents and purposes coming from one place. When you have a distribution package there are often patches and other things that modify the upstream code.

How do we want to manage this sort of thing?

david-a-wheeler:
This was discussed at the 2021-12-01 SLSA meeting. Thanks for raising this! This is a challenging issue.

Here are a few notes from that meeting:

* Distributors often take a source tree, make additions/changes, build something, & release that
* Mark Lodato: The same issue happens when a company creates a local distribution
* David A. Wheeler: At the least (“floor”), if a distribution package claims SLSA compliance, the distribution package development/release process must implement then SLSA requirements
* Problem: If a distribution package has a high SLSA level but is creating a package of a source code from a no/low SLSA level, this could be really confusing to users.
    1. Idea: Maybe we require that distribution packages say their level AND “on” the SLSA level of their source, e.g., “SLSA level 3 on 1” means this distributor does SLSA 3, but they’re packaging source at SLSA 1. Could recurse if it’s re-repackaged “3 on 1 on 1”.
    2. Idea: Take the minimum, simpler to understand [but gives less insight]
    3. Idea: Consider the distribution packagers + upstream for that package as if they were a single organization, and measure SLSA that way
* Question: How transitive should this be?

Others, please add your thoughts.

david-a-wheeler:
I think the underlying challenge is to be clear and *NOT* mislead.

As was discussed on the call, many distribution packages take the source code, make some minor changes, & then release it. The distributors often don't review every line of code (though that does happen, especially when they're the same people). The distributors add code that the upstream didn't see. That makes describing this challenging.

TomHennen:
So it sounds like the first questions to answer is "how should artifacts be _labeled_?", once we have an idea of what we want there we can thinking about the technical approach?

mlieberman85:
Some additional thoughts. I think we should not fall down the transitive rabbit hole right now, just want to make sure that we're clear what folks are attesting to.

[...]

PR #673, opened 2023-03-08
Add page on distributing provenance

### LINKED ISSUE #269: Provenance publication & discovery
Like Policy & Verification in https://github.com/slsa-framework/slsa/issues/46, and related to, how/where to publish provenance and how users discover published provenance feels like an open gap in the current set of guidance around SLSA.

I think we should capture recommendations/thoughts around how provenance attestations could be published and discovered for different scenarios.

Today folks can:
* publish alongside their release artefacts, which isn't especially discoverable
* store attestations in sigstore's rekor, discovery becomes a rekor query

One possibility is package ecosystems publishing provenance attestations alongside the packages and package managers growing features to fetch the associated attestations.

Has anyone thought about this in detail?

di:
Fixes #269.

### DIFF
--- a/docs/spec/v1.0/distributing-artifacts.md
+++ b/docs/spec/v1.0/distributing-artifacts.md
@@ -1,0 +1,157 @@
+# Distributing Artifacts
+
+SLSA requires the distribution and verification of provenance metadata in the
+form of SLSA attestations, in order to make provenance for related artifacts
+available after generation for verification.
+
+This document specifies how these attestations should be distributed, or the
+relationship between build artifacts and attestations. It is primarily
+concerned with artifacts for ecosystems that distribute "built distributions",
+but some attention is also paid to ecosystems that distribute container images
+or only distribute source artifacts, as many of the same principles generally
+apply to any artifact or group of artifacts.
+
+In addition, this document is primarily for the benefit of artifact
+distributors, to understand how they should adopt the distribution of SLSA
+provenance. It is primarily concerned with the means of distributing
+attestations and the relationship of attestations to build artifacts, and not
+with the specific format of the attestation or attestation bundle itself.
+
+## Overview
+
+[Package ecosystem]: #package-ecosystem
+
+A <dfn>package ecosystem</dfn> is a set of conventions and tooling for package
+distribution. Every package has an ecosystem, whether it is formal or ad-hoc.
+Some ecosystems are formal, such as language distribution (e.g.
+[Python/PyPI](https://pypi.org)), operating system distribution (e.g.
+[Debian/Apt](https://wiki.debian.org/DebianRepository/Format)), or artifact
+distribution (e.g. [OCI](https://github.com/opencontainers/distribution-spec)).
+Other ecosystems are informal, such as a convention used within a company. Even
+ad-hoc distribution of software, such as through a link on a website, is
+considered an "ecosystem".
+
+The package ecosystem's maintainers are responsible for reliably redistributing
+artifacts and provenance, making the producers' expectations available to consumers,
+and providing tools to enable safe artifact consumption (e.g. whether an artifact
+meets its producer's expectations).
+
+## Relationship between releases and attestations
+
+For many ecosystems, a single "release" of a project, package, or library may
+include multiple artifacts, resulting from because builds that may happen over
+multiple differing platforms, architectures or environments, and may not happen
+at roughly the same point in time (and might even span multiple days).
+
+As a result, the complete set of all attestations for a given release may be
+incomplete until a build is finished, and in reality, a build may never be
+truly 'finished' as in many ecosystems, it is permissible for new artifacts
+that support new platforms or architectures to be introduced for old releases
+long after the initial release.
+
+As a result, using a single attestation bundle per release is unnecessarily
+restrictive and does not work with the way complex, modern software artifacts
+are built and distributed.
+
+Thus, it is recommended to support multiple individual attestations or
+attestation bundles per release. At the time of a given build, the relevant
+build attestations for that build can be added to the release, depending on the
+relationship to the given artifacts.
+
+## Relationship between artifacts and attestations
+
+Given the previous recommendation that a single release may have multiple
+attestations, this raises the question of the relationship between the
+one-or-more artifacts that comprise the output of a release build and the
+attestations that correlate to them.
+
+It is recommended to support a one-to-many relationship between build artifacts
+and attestations, to ensure that anyone is free to produce and publish any
+attestation they may need, but require a strict one-to-one relationship between
+the build artifact and the build provenance attestation for that artifact. This
+may require a custom filename schema that uniquely identifies the build
+attestation over other attestation types, or it may happen as a de-facto
+standard based on where the attestation is published.
+
+For example, the build attestation should have a filename that is directly
+related to the build artifact filename. For example, for an artifact
+`<filename>.<extension>`, the attestation is `<filename>.attestation` (or some
+similar extension, for example [in-toto](https://in-toto.io/) recommends
+`<filename>.intoto.jsonl`), or for an artifact with a SHA-256 hash of
+`abc123...`, the attestation incorporates that hash into the filename or
+metadata.
+
+## Where attestations should be published
+
+There are a number of opportunities and venues to publish attestations during
+and after the build process. It is recommended to publish attestations in a
+number of places:
+
+* **Publish attestations in a transparency log**: Once a build has been made,
+  the resulting attestation should be published to a third-party transparency
+  log that exists outside the source repository and artifact repository.
+* **Publish attestations alongside source releases in the source repository**:
+  For ecosystems where source releases are commonly published to source
+  repositories as well as to artifact repositories (such as GitHub releases),
+  if the source repository supports attaching additional artifacts on a
+  release, it is recommended to additionally include provenance attestations as
+  part of these releases. This option requires no changes to the artifact
+  repository to support build attestation formats, but means that the source
+  repository in use is now in the dependency chain for installers that want to
+  verify policy at build-time.
+* **Publish attestations alongside the artifact in the artifact repository**:
+  Many software repositories already support some variety of publishing 1:1
+  related files alongside an artifact, sometimes known as “sidecar files”. For
+  example, PyPI supports publishing `.asc` files representing the PGP signature
+  for an artifact with the same filename (but different extension). This
+  requires the mapping between artifact and attestation (or attestation vessel)
+  to be 1:1.
+
+A combination of these options allows for a bootstrap process where projects
+that want to generate build provenance attestations may begin publishing
+attestations in a way that is directly tied to a source release or built
+artifact without needing to wait, by publishing it as part of the source
+repository release.
+
+Long-term, in order to maintain a single dependency on the artifact repository
+already in use, repositories should gain support for both uploading the build
+attestation format, and distributing it 1:1 alongside the artifact.
+
+Short term consumers of build artifacts can bootstrap a manual policy by using
+the source repository only for projects that publish all artifacts and
+attestations to the source repository, and later extend this to all artifacts
+published to the artifact repository via the canonical installation tools once
+a given ecosystem supports them.
+
+## Immutability of attestations
+
+Attestations should be immutable. Once a build attestation is published as it
+corresponds to a given artifact, that attestation is immutable and cannot be
+overwritten later with a different attestation that refers to the same
+artifact. Instead, a new release (and new artifacts) should be created instead.
+
+## Format of the attestation
+
+The provenance is available to the consumer in a format that the consumer
+accepts. The format SHOULD be in-toto [SLSA Provenance](/provenance),
+but another format MAY be used if both producer and consumer agree and it meets
+all the other requirements.
+
+## Considerations for source-based ecosystems
+
+Some ecosystems have support for installing directly from source repositories
+(optional for Python/`pip`, Go, etc). For these examples in these ecosystems,
+as there is no "build" step that translates between a source repository and an
+artifact that is being installed, there is no need to publish or verify build
+provenance.
+
+However, for ecosystems that install from source repositories _via_ some
+intermediary (e.g. [Homebrew installing from GitHub release artifacts generated
+from the repository or GitHub Packages](https://docs.brew.sh/Bottles), [Go
+installing through the Go module proxy](https://proxy.golang.org/), these
+intermediaries as transforming the original source repository in some way and
+as a result should be providing build provenance, and the recommendations
+outlined here apply.
+
+
+
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -150,7 +150,7 @@ The build process MUST generate provenance that unambiguously identifies the
 output package and describes how that package was produced.
 
 The format MUST be acceptable to the
-[package ecosystem](verifying-artifacts.md#package-ecosystem) and/or
+[package ecosystem](distributing-artifacts.md#package-ecosystem) and/or
 [consumer](verifying-artifacts.md#consumer). It
 is RECOMMENDED to use the [SLSA Provenance] format and [associated suite]
 because it is designed to be interoperable, universal, and unambiguous when
--- a/docs/spec/v1.0/verifying-artifacts.md
+++ b/docs/spec/v1.0/verifying-artifacts.md
@@ -20,27 +20,6 @@ verification.
 Consumers can trust their package ecosystem or some other third party to verify
 artifacts, or they can verify artifacts against their own set of expectations.
 
-## Package ecosystem
-
-[Package ecosystem]: #package-ecosystem
-
-> ⚠ **RFC:** Is there a better term that is more obvious to most readers?
-
-A <dfn>package ecosystem</dfn> is a set of conventions and tooling for package
-distribution. Every package has an ecosystem, whether it is formal or ad-hoc.
-Some ecosystems are formal, such as language distribution (e.g.
-[Python/PyPA](https://www.pypa.io)), operating system distribution (e.g.
-[Debian/Apt](https://wiki.debian.org/DebianRepository/Format)), or artifact
-distribution (e.g. [OCI](https://github.com/opencontainers/distribution-spec)).
-Other ecosystems are informal, such as a convention used within a company. Even
-ad-hoc distribution of software, such as through a link on a website, is
-considered an "ecosystem".
-
-The package ecosystem's maintainers are responsible for reliably redistributing
-artifacts and provenance, making the producers' expectations available to consumers,
-and providing tools to enable safe artifact consumption (e.g. whether an artifact
-meets its producer's expectations).
-
 ### Setting Expectations
 
 <dfn>Expectations</dfn> are known provenance values that indicate the corresponding


docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+attestation bundles per release. At the time of a given build, the relevant
+build attestations for that build can be added to the release, depending on the
+relationship to the given artifacts.
+
+## Relationship between artifacts and attestations
+
+Given the previous recommendation that a single release may have multiple
+attestations, this raises the question of the relationship between the
+one-or-more artifacts that comprise the output of a release build and the
+attestations that correlate to them.
```
kpk47, line 66: Maybe nix this sentence.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+
+It is recommended to support a one-to-many relationship between build artifacts
+and attestations, to ensure that anyone is free to produce and publish any
+attestation they may need, but require a strict one-to-one relationship between
+the build artifact and the build provenance attestation for that artifact. This
```
kpk47, line 71:
```suggestion
Package ecosystems SHOULD support a one-to-many relationship between build artifacts
and attestations to ensure that anyone is free to produce and publish any
attestation they may need. Package ecosystems MUST require a strict one-to-one relationship between
the build artifact and the build provenance attestation for that artifact. 
```

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+may require a custom filename schema that uniquely identifies the build
+attestation over other attestation types, or it may happen as a de-facto
+standard based on where the attestation is published.
```
kpk47, line 74:
```suggestion
The mappings can be either implicit (e.g. require a custom filename schema that uniquely identifies the build
attestation over other attestation types) or explicit (e.g. it may happen as a de-facto
standard based on where the attestation is published).
```

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+
+For example, the build attestation should have a filename that is directly
```
kpk47, line 76:
```suggestion
The build attestation SHOULD have a filename that is directly
```

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+`<filename>.<extension>`, the attestation is `<filename>.attestation` (or some
+similar extension, for example [in-toto](https://in-toto.io/) recommends
+`<filename>.intoto.jsonl`), or for an artifact with a SHA-256 hash of
+`abc123...`, the attestation incorporates that hash into the filename or
+metadata.
+
+## Where attestations should be published
+
+There are a number of opportunities and venues to publish attestations during
+and after the build process. It is recommended to publish attestations in a
```
kpk47, line 87:
```suggestion
and after the build process. Producers SHOULD publish attestations in a
```

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+number of places:
+
+-   **Publish attestations in a transparency log**: Once a build has been made,
```
kpk47, line 90: IIUC, builds are processes and artifacts/packages are their outputs. Do you mean the publishing should happen at the end of the build process, once the artifact is build (but not necessarily at the end of the full build process), or something else?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+    the resulting attestation should be published to a third-party transparency
+    log that exists outside the source repository and artifact repository.
+-   **Publish attestations alongside source releases in the source
+    repository**: For ecosystems where source releases are commonly published
+    to source repositories as well as to artifact repositories (such as GitHub
+    releases), if the source repository supports attaching additional artifacts
+    on a release, it is recommended to additionally include provenance
+    attestations as part of these releases. This option requires no changes to
```
kpk47, line 98:
```suggestion
    releases), producers SHOULD include provenance
    attestations as part of these releases if the source repository supports attaching additional artifacts
    to a release. This option requires no changes to
```

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+    the artifact repository to support build attestation formats, but means
+    that the source repository in use is now in the dependency chain for
+    installers that want to verify policy at build-time.
+-   **Publish attestations alongside the artifact in the artifact repository**:
+    Many software repositories already support some variety of publishing 1:1
+    related files alongside an artifact, sometimes known as “sidecar files”.
+    For example, PyPI supports publishing `.asc` files representing the PGP
+    signature for an artifact with the same filename (but different extension).
+    This requires the mapping between artifact and attestation (or attestation
```
kpk47, line 107:
```suggestion
    This option requires the mapping between artifact and attestation (or attestation
```

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+    vessel) to be 1:1.
+
+A combination of these options allows for a bootstrap process where projects
+that want to generate build provenance attestations may begin publishing
+attestations in a way that is directly tied to a source release or built
+artifact without needing to wait, by publishing it as part of the source
```
kpk47, line 113: What would producers be waiting for?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+repository release.
+
+Long-term, in order to maintain a single dependency on the artifact repository
+already in use, repositories should gain support for both uploading the build
+attestation format, and distributing it 1:1 alongside the artifact.
```
kpk47, line 118:
```suggestion
Long-term, in order to maintain a single dependency on the artifact repository
already in use, repositories should gain support for both uploading and distributing the build
attestation alongside the artifact.
```

kpk47: We might also want to rework the sentence to avoid a lower case 'should'.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+
+Short term consumers of build artifacts can bootstrap a manual policy by using
```
kpk47, line 120: Should I read this as "consumers who will use a package for a short while ..." or "while waiting for canonical tooling ..."?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+the source repository only for projects that publish all artifacts and
+attestations to the source repository, and later extend this to all artifacts
+published to the artifact repository via the canonical installation tools once
+a given ecosystem supports them.
+
+## Immutability of attestations
+
+Attestations should be immutable. Once a build attestation is published as it
```
kpk47, line 128: Maybe this should be MUST? Or are there ecosystems that can't support immutability?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+corresponds to a given artifact, that attestation is immutable and cannot be
+overwritten later with a different attestation that refers to the same
+artifact. Instead, a new release (and new artifacts) should be created instead.
```
kpk47, line 131:
```suggestion
artifact. Instead, a new release (and new artifacts) should be created.
```

kpk47: When would someone need a new attestation to refer to an identical artifact?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+
+As a result, the complete set of all attestations for a given release may be
+incomplete until a build is finished, and in reality, a build may never be
+truly 'finished' as in many ecosystems, it is permissible for new artifacts
+that support new platforms or architectures to be introduced for old releases
+long after the initial release.
+
+As a result, using a single attestation bundle per release is unnecessarily
+restrictive and does not work with the way complex, modern software artifacts
+are built and distributed.
```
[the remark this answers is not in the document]
MarkLodato: Can we simplified this as, "Attestations should be bound to artifacts, not releases"?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+already in use, repositories should gain support for both uploading the build
+attestation format, and distributing it 1:1 alongside the artifact.
+
+Short term consumers of build artifacts can bootstrap a manual policy by using
+the source repository only for projects that publish all artifacts and
+attestations to the source repository, and later extend this to all artifacts
+published to the artifact repository via the canonical installation tools once
+a given ecosystem supports them.
+
+## Immutability of attestations
```
MarkLodato, line 126: nit: attestation **bundle** throughout, right? In other words, a single artifact SHOULD be immutable, and once you attach an attestation bundle, that bundle SHOULD also be immutable.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+Some ecosystems have support for installing directly from source repositories
+(optional for Python/`pip`, Go, etc). For these examples in these ecosystems,
+as there is no "build" step that translates between a source repository and an
+artifact that is being installed, there is no need to publish or verify build
+provenance.
+
+However, for ecosystems that install from source repositories _via_ some
+intermediary (e.g. [Homebrew installing from GitHub release artifacts generated
+from the repository or GitHub Packages](https://docs.brew.sh/Bottles), [Go
+installing through the Go module proxy](https://proxy.golang.org/), these
```
MarkLodato, line 151:
```suggestion
installing through the Go module proxy](https://proxy.golang.org/)), these
```

MarkLodato (changes requested):
Thanks, @di! This is great. I added a bunch of nitpicks but overall I like where this is going.

MarkLodato:
Friendly ping.

docs/spec/v1.0/distributing-artifacts.md
di: Possibly if: someone migrates to a new build system, does a reproducible build of the artifact, and has a new attestation for the same artifact. The original attestation should not be overwritable in this case.

docs/spec/v1.0/distributing-artifacts.md
di: I don't think it matters, I think the recommendation is just "once you generate attestations, you should publish them". 

Should this be rephrased to "Once an attestation has been generated"?

docs/spec/v1.0/distributing-artifacts.md
di: 169ddf9

docs/spec/v1.0/distributing-artifacts.md
di: a798577

docs/spec/v1.0/distributing-artifacts.md
di: ee9c949

docs/spec/v1.0/distributing-artifacts.md
di: 72352ed

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+similar extension, for example [in-toto](https://in-toto.io/) recommends
+`<filename>.intoto.jsonl`), or for an artifact with a SHA-256 hash of
+`abc123...`, the attestation incorporates that hash into the filename or
+metadata.
+
+## Where attestations should be published
+
+There are a number of opportunities and venues to publish attestations during
+and after the build process. It is recommended to publish attestations in a
+number of places:
```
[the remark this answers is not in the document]
di: I think there isn't an order of preference: the recommendation is that they're published in all these places, if possible.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+
+-   **Publish attestations in a transparency log**: Once a build has been made,
+    the resulting attestation should be published to a third-party transparency
+    log that exists outside the source repository and artifact repository.
```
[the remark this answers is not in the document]
di: I think 142eb4d addresses this -- it could be either or.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+For many ecosystems, a single "release" of a project, package, or library may
+include multiple artifacts, resulting from because builds that may happen over
+multiple differing platforms, architectures or environments, and may not happen
+at roughly the same point in time (and might even span multiple days).
+
+As a result, the complete set of all attestations for a given release may be
+incomplete until a build is finished, and in reality, a build may never be
+truly 'finished' as in many ecosystems, it is permissible for new artifacts
+that support new platforms or architectures to be introduced for old releases
+long after the initial release.
```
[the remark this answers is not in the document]
di: fa67fd4

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+or only distribute source artifacts, as many of the same principles generally
+apply to any artifact or group of artifacts.
+
+In addition, this document is primarily for the benefit of artifact
+distributors, to understand how they should adopt the distribution of SLSA
+provenance. It is primarily concerned with the means of distributing
+attestations and the relationship of attestations to build artifacts, and not
+with the specific format of the attestation or attestation bundle itself.
+
+## Overview
```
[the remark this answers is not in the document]
di: 89ff846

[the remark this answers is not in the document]
di: https://github.com/slsa-framework/slsa/commit/89ff846e7d0e98cd397e56b7c03b588dcd9bd643

di:
Thanks for all your comments @kpk47 and @MarkLodato! I think I've addressed all your concerns and this is ready for a re-review.

This PR is also being shared at the OpenSSF Securing Software Repos WG later today, so we should hold off on merging until later this week to give folks there one last chance to review.

MarkLodato (changes requested):
Overall I think this is good. Thank you! I have a few suggestions on specific parts.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,149 @@
...
+A combination of these options allows for a bootstrap process where projects
+that want to generate build provenance attestations may begin publishing
+attestations in a way that is directly tied to a source release or built
+artifact without needing to wait for the upstream package repository to support
+publishing attestations, by publishing it as part of the source repository
+release.
+
+Long-term, in order to maintain a single dependency on the artifact repository
+already in use, repositories should gain support for both uploading and
+distributing the build attestation alongside the artifact.
```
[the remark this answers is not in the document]
di: Hmm, the other aspect is trust -- consumers already trust one service, and don't need to extend trust into some other additional service.

I think I'd change "failure domain" to "point of failure" here as well, so as not to be confused with literal domain names, WDYT?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,149 @@
...
+build can be added to the release, depending on the relationship to the given
+artifacts.
+
+## Relationship between artifacts and attestations
+
+Package ecosystems SHOULD support a one-to-many relationship between build
+artifacts and attestations to ensure that anyone is free to produce and publish
+any attestation they may need. Package ecosystems MUST require a strict
+one-to-one relationship between the build artifact and the build provenance
+attestation for that artifact.  The mappings can be either implicit (e.g.
```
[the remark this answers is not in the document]
di: This is meant to say that there is a one-to-one inverse relationship: an attestation should only have a relationship to a single artifact.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,149 @@
...
+-   **Publish attestations alongside source releases in the source
+    repository**: For ecosystems where source releases are commonly published
+    to source repositories as well as to artifact repositories (such as GitHub
+    releases), producers SHOULD include provenance attestations as part of
+    these releases if the source repository supports attaching additional
+    artifacts to a release. This option requires no changes to the artifact
+    repository to support build attestation formats, but means that the source
+    repository in use is now in the dependency chain for installers that want
+    to verify policy at build-time.
+-   **Publish attestations alongside the artifact in the artifact repository**:
```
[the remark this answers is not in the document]
di: 9755259

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,149 @@
...
+the attestation is `<filename>.attestation` (or some similar extension, for
+example [in-toto](https://in-toto.io/) recommends `<filename>.intoto.jsonl`),
+or for an artifact with a SHA-256 hash of `abc123...`, the attestation
+incorporates that hash into the filename or metadata.
+
+## Where attestations should be published
+
+There are a number of opportunities and venues to publish attestations during
+and after the build process. Producers SHOULD publish attestations in a number
+of places:
```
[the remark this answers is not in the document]
di: I don't think there's an order of preference here, and these aren't mutually exclusive. I think there's always a benefit to publishing these in multiple places, e.g. once a package registry supports publishing, that doesn't mean one should stop publishing in source repository releases.

> Publishing to a transparency log is also a good idea. (Why?)

Added an answer to this in 21e8091.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,150 @@
...
+    monitored transparency log during verification helps ensure the attestation
+    is trustworthy.
+-   **Publish attestations alongside source releases in the source
+    repository**: For ecosystems where source releases are commonly published
+    to source repositories as well as to package registries (such as GitHub
+    releases), producers SHOULD include provenance attestations as part of
+    these releases if the source repository supports attaching additional
+    artifacts to a release. This option requires no changes to the artifact
+    repository to support build attestation formats, but means that the source
+    repository in use is now in the dependency chain for installers that want
```
MarkLodato, line 94: nit: "dependency chain" might be confused with build-time "dependencies" that we talk about elsewhere in SLSA. Maybe "critical path"?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,150 @@
...
+    to verify policy at build-time.
+-   **Publish attestations alongside the artifact in the package registry**:
```
trishankatdatadog, line 96: So is this a SHOULD or MAY or something else? Not clear right now. Are all three possibilities kept to a SHOULD right now, so that it is acceptable not to publish any build provenance attestations at all?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,150 @@
...
+
+Short term, consumers of build artifacts can bootstrap a manual policy by using
+the source repository only for projects that publish all artifacts and
+attestations to the source repository, and later extend this to all artifacts
+published to the package registry via the canonical installation tools once
+a given ecosystem supports them.
+
+## Immutability of attestations
+
+Attestations MUST be immutable. Once a build attestation is published as it
```
trishankatdatadog, line 123: How does one enforce this? Does that mean that package repositories MUST use a transparent log or at least a [WORM](https://aws.amazon.com/blogs/storage/protecting-data-with-amazon-s3-object-lock/) storage bucket?

trishankatdatadog (commented):
Thanks for your contribution! I think one of the biggest missing pieces right now is precisely how consumers or SLSA tooling would verify various provenance attestations signed by different build systems using different keys. There are also related tricky cases such as: is a source repository allowed to use a different key than a package repository? Is the idea to push these Qs to a future MAJOR version of SLSA?

trishankatdatadog:
> Thanks for your contribution! I think one of the biggest missing pieces right now is precisely how consumers or SLSA tooling would verify various provenance attestations signed by different build systems using different keys. There are also related tricky cases such as: is a source repository allowed to use a different key than a package repository? Is the idea to push these Qs to a future MAJOR version of SLSA?

We [discussed](https://docs.google.com/document/d/1-f6m442MHg9hktrbcp-4sM9GbZC3HLTpZPpxMXjMCp4/edit#bookmark=id.iwo5seoi4lyf) this earlier at the OpenSSF Securing Software Repos WG meeting today, and the current thinking seems to be that we will slowly but surely iterate by figuring out exactly what sorts of standardised in-toto attestations we will need. For example, right now, this means build provenance attestations for SLSA 1.0, but maybe in the future we will add new types of attestations (e.g., maybe a specific one for the [Source](https://github.com/in-toto/attestation/issues/124) track) as we see fit. Then, we would need in-toto [policies](https://github.com/in-toto/ITE/pull/38) (aka layouts) to link all the different attestations together. Until policies arrive, we might need to special-case existing SLSA tooling/verifiers with known good public keys of public build systems. When they do arrive, we can then discuss how to securely associate different artefacts with different policies. But all of this will obviously take some time to organically develop. Did I capture the thinking correctly @MarkLodato and @znewman01? If so, I wonder if it might be worth noting somewhere in SLSA 1.0.

docs/spec/v1.0/distributing-artifacts.md
di: The line that introduces this list says "Producers SHOULD publish attestations in a number of places:", which I would interpret to mean that these all are SHOULDs, but if that's unclear we could repeat that for each item in the list.

docs/spec/v1.0/distributing-artifacts.md
di: I think this is just one of many ways in which consumers need to trust artifact repositories, i.e. that their behavior ensures that attestations are immutable in the same way consumers assume artifacts are immutable.

di:
> Thanks for your contribution! I think one of the biggest missing pieces right now is precisely how consumers or SLSA tooling would verify various provenance attestations signed by different build systems using different keys. There are also related tricky cases such as: is a source repository allowed to use a different key than a package repository? Is the idea to push these Qs to a future MAJOR version of SLSA?

Great questions -- I think these are probably better discussed in the context of [the "Verifying artifacts" topic](https://slsa.dev/spec/v1.0/verifying-artifacts), since this is just focused on distributing provenance.

> We [discussed](https://docs.google.com/document/d/1-f6m442MHg9hktrbcp-4sM9GbZC3HLTpZPpxMXjMCp4/edit#bookmark=id.iwo5seoi4lyf) this earlier at the OpenSSF Securing Software Repos WG meeting today, and the current thinking seems to be that we will slowly but surely iterate by figuring out exactly what sorts of standardised in-toto attestations we will need.

The background for this page explicitly says:

> It is primarily concerned with the means of distributing attestations and the relationship of attestations to build artifacts, and not with the specific format of the attestation itself.

GIven that, I think the contents of the attestations are definitely out of scope here 🙂

docs/spec/v1.0/distributing-artifacts.md
di: c1e5e56

docs/spec/v1.0/distributing-artifacts.md
di: 2bef566

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
di: a6bdf5d

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
di: b06879c

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -80,18 +78,22 @@ of places:
...
 -   **Publish attestations in a transparency log**: Once a build has been made,
     the resulting attestation (or a hash of the attestation and a pointer to
-    where it is indexed) should be published to a third-party transparency log
-    that exists outside the source repository and artifact repository.
+    where it is indexed) SHOULD be published to a third-party transparency log
+    that exists outside the source repository and package registry. Not only
+    are transparency logs are guaranteed to be immutable, but they also more
+    easily enable monitoring.  Requiring the presence of the attestation in a
+    monitored transparency log during verification helps ensure the attestation
+    is trustworthy.
```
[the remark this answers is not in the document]
di: I'm weighing how technical we want to get about explaining transparency logs here vs. just linking to something like https://docs.sigstore.dev/security/#rekor-security-model or https://transparency.dev/. What would you suggest?

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
MarkLodato: Sorry, I still don't understand. What is the reasoning behind the requirement? What would happen if this requirement is violated?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+release. At the time of a given build, the relevant build attestations for that
+build can be added to the release, depending on the relationship to the given
+artifacts.
+
+## Relationship between artifacts and attestations
+
+Package ecosystems SHOULD support a one-to-many relationship from build
+artifacts to attestations to ensure that anyone is free to produce and publish
+any attestation they may need. Package ecosystems MUST require a strict
+one-to-one relationship between from a build provenance attestation to a given
```
jchestershopify, line 60: Nit: `between from a` should probably be `from a`.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+but some attention is also paid to ecosystems that distribute container images
+or only distribute source artifacts, as many of the same principles generally
+apply to any artifact or group of artifacts.
+
+In addition, this document is primarily for the benefit of artifact
+distributors, to understand how they should adopt the distribution of SLSA
+provenance. It is primarily concerned with the means of distributing
+attestations and the relationship of attestations to build artifacts, and not
+with the specific format of the attestation itself.
+
```
jchestershopify, line 26: Should the usual RFC 2119 boilerplate be included?

jchestershopify (approved):
Some small nits, but otherwise LGTM.

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
di: The reasoning is to enforce the recommendation on attestation filenames and their relationship to artifacts. E.g., if we have `artifact_a.<extension>` and `artifact_b.<extension>`, an attestation named `artifact_a.attestation` shouldn't have any relationship to `artifact_b.<extension>` (especially given the fact that attestation formats like `.intoto.jsonl` might contain multiple attestations in a single file).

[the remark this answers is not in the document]
MarkLodato: But why is that a problem? Suppose you create a build that generates both `artifact_a` and `artifact_b` in the same build, with a single attestation listing both artifacts in the subject. You publish the same attestation as both `artifact_a.attestation` and `artifact_b.attestation`. That seems to violate your requirement here, but seems perfectly valid to me. Or am I misunderstanding?

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
di: I think there's not much distinction here between distributing a provenance file, or distributing a reference to that provenance file -- either way, we're including it in the transparency log.

That said, if the attestation is signed and the signature is added to the transparency log instead, I don't see that this would provide any added benefit -- and since that is more about the provenance format, I think I agree that we can remove this here.

If we do that, do you think that it would make sense for https://slsa.dev/spec/v1.0/verifying-artifacts  to declare a preference for signing methodologies that rely on transparency logs?

(Also, just found some prior discussion about this here: https://github.com/slsa-framework/slsa/issues/235#issuecomment-1013800183)

docs/spec/v1.0/distributing-artifacts.md
di:
```suggestion
one-to-one relationship from a build provenance attestation to a given
```

docs/spec/v1.0/distributing-artifacts.md
di: 6d931c0

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
di: I agree that would be fine, what wouldn't be ok is having to know something about `artifact_b` (like the filename) to get the attestation for `artifact_a`.

So maybe this should be rephrased as "ecosystems MUST have a one-to-one mapping between an artifact and an attestation such that the attestation can be discovered from the artifact alone" or something to that effect.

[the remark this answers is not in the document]
MarkLodato: But is this a real concern? Can we just drop this requirement? I'm having a hard time thinking of a situation where we drop this requirement, someone implements it anyway, it works, and yet that's a bad thing. Either it won't work, but it's likely fine?

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
MarkLodato: > If we do that, do you think that it would make sense for https://slsa.dev/spec/v1.0/verifying-artifacts to declare a preference for signing methodologies that rely on transparency logs?

Yeah, that sounds like a good idea.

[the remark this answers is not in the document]
di: https://github.com/slsa-framework/slsa/issues/727

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
+---
+title: Distributing artifacts
+description: This page covers the detailed technical requirements for distributing artifacts at each SLSA level. The intended audience is system implementers and software distributors.
+prev_page:
+  url: requirements
+next_page:
+  url: verifying-artifacts
```
MarkLodato, line 7: These can now be removed as of a few minutes ago (#724). Same for the other files. (If you sync you'll get merge conflicts. The right resolution is to remove.)

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
MarkLodato: LGTM.

MarkLodato (approved):
Thanks, Dustin!

MarkLodato:
Hey all, I'd like to get this submitted today or Monday so that it's locked in. If there are any small last-minute comments, please review ASAP. Otherwise I suggest that we defer any other major discussions to follow-up issues or PRs.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
...
+## Relationship between releases and attestations
+
+Attestations SHOULD be bound to artifacts, not releases.
+
+A single "release" of a project, package, or library might include multiple
+artifacts. These artifacts result from builds on different platforms,
+architectures or environments. The builds need not happen at roughly the same
+point in time and might even span multiple days.
+
+The complete set of attestations for a given release is incomplete until all
```
kpk47, line 48:
```suggestion
The set of attestations for a given release is incomplete until all
```

docs/spec/v1.0/distributing-artifacts.md
kpk47: Perhaps we should make it clear that they MUST publish provenance in at least one place and SHOULD publish it in many.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
...
+    to verify policy at build-time.
+-   **Publish attestations alongside the artifact in the package registry**:
+    Many software repositories already support some variety of publishing 1:1
+    related files alongside an artifact, sometimes known as “sidecar files”.
+    For example, PyPI supports publishing `.asc` files representing the PGP
+    signature for an artifact with the same filename (but different extension).
+    This option requires the mapping between artifact and attestation (or
+    attestation vessel) to be 1:1.
+
+A combination of these options allows for a bootstrap process where projects
```
kpk47, line 105:
```suggestion
A combination of these options allows for a bootstraping process where projects
```

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
...
+that want to generate build provenance attestations can begin publishing
+attestations in a way that is directly tied to a source release or built
+artifact without needing to wait for the upstream package registry to support
+publishing attestations, by publishing it as part of the source repository
+release.
```
kpk47, line 110: I'm having trouble parsing this sentence. 

Are you saying that publishing provenance to the source repo is a stepping stone to publishing provenance to the pkg repo? How does publishing build provenance to a source repo help in cases where the build isn't reproducible?

kpk47 (commented):
Mostly LGTM. I'm just a bit confused about publishing provenance to source repos.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
...
+release. At the time of a given build, the relevant build attestations for that
+build can be added to the release, depending on the relationship to the given
+artifacts.
+
+## Relationship between artifacts and attestations
+
+Package ecosystems SHOULD support a one-to-many relationship from build
+artifacts to attestations to ensure that anyone is free to produce and publish
+any attestation they might need. Package ecosystems MUST provide a way to map a
+given artifact to a build attestation. The mappings can be either implicit
```
znewman01, line 64: Nit: "a build attestation" -> "corresponding build attestations" or something?

I can imagine requiring >1 canonical build attestation (e.g., if builds are reproducible).

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
...
+point in time and might even span multiple days.
+
+The complete set of attestations for a given release is incomplete until all
+builds for a given release are finished. However, it is difficult or impossible
+to determine when a release is 'finished' because many ecosystems allow adding
+new artifacts to old releases when adding support for new platforms or
+architectures.
+
+Thus, package ecosystems SHOULD support multiple individual attestations per
+release. At the time of a given build, the relevant build attestations for that
```
znewman01, line 55: Build attestation is never defined. I think that's worth a quick sentence or two (or a link to the terminology page)

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
...
+
+Package ecosystems SHOULD support a one-to-many relationship from build
+artifacts to attestations to ensure that anyone is free to produce and publish
+any attestation they might need. Package ecosystems MUST provide a way to map a
+given artifact to a build attestation. The mappings can be either implicit
+(e.g. require a custom filename schema that uniquely identifies the build
+attestation over other attestation types) or explicit (e.g. it could happen as
+a de-facto standard based on where the attestation is published).
+
+The build attestation SHOULD have a filename that is directly related to the
```
znewman01, line 69: What if there's more than one? Do we smush all attestations into the same file?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
...
+
+Attestations SHOULD be bound to artifacts, not releases.
+
+A single "release" of a project, package, or library might include multiple
+artifacts. These artifacts result from builds on different platforms,
+architectures or environments. The builds need not happen at roughly the same
+point in time and might even span multiple days.
+
+The complete set of attestations for a given release is incomplete until all
+builds for a given release are finished. However, it is difficult or impossible
```
znewman01, line 49: This seems to imply that a given *build's* attestation is finished as soon as the build completes. I think that's insufficiently flexible: if we have reproducible builds and want attestations from two different builders, those would happen at different times. For dynamic information (like known vulnerabilities in an old package), you might even want new attestations years in the future, from a recurring scan.

Maybe the key thing here (see also the section on immutability) is that new attestations (and builds) can always be added to a release, but never removed.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
...
+the source repository only for projects that publish all artifacts and
+attestations to the source repository, and later extend this to all artifacts
+published to the package registry via the canonical installation tools once
+a given ecosystem supports them.
+
+## Immutability of attestations
+
+Attestations MUST be immutable. Once a build attestation is published as it
+corresponds to a given artifact, that attestation is immutable and cannot be
+overwritten later with a different attestation that refers to the same
```
znewman01, line 132: We can add new attestations, though–right?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,157 @@
...
+The provenance is available to the consumer in a format that the consumer
+accepts. The format SHOULD be in-toto [SLSA Provenance](/provenance), but
+another format MAY be used if both producer and consumer agree and it meets all
+the other requirements.
+
+## Considerations for source-based ecosystems
+
+Some ecosystems have support for installing directly from source repositories
+(an option for Python/`pip`, Go, etc). In these cases, there is no need to
+publish or verify provenance because there is no "build" step that translates
```
znewman01, line 146: SLSA almost always uses the term provenance to mean "build provenance" but I think that's in conflict with the common usage.

The source still has some notion of "provenance" (where'd it come from? is a given tag supposed to indicate a release? was this commit merged by a real maintainer). I can imagine wanting to check this (though it's explicitly out-of-scope for SLSA). This information could come from VCS/forge mechanisms for signing/auth, but I could imagine a proper provenance attestation being useful too.

I'd suggest "In these cases, there is no need to publish or verify *build* provenance." You could (but don't need to) say something about source provenance, noting that it's out-of-scope, as well.

znewman01 (commented):
So sorry for the last minute drive-by! I just became aware of this via the OpenSSF Securing Software Repos group a few days back.

Overall LGTM though I think my comments might be worth addressing to prevent ruling out useful verification policies later on.

docs/spec/v1.0/distributing-artifacts.md
znewman01: If I'm reading correctly, the point here is that you could start building a totally third-party system for verifying build provenance without needing the repository itself to implement anything.

I think the confusion is that a "release" on the source repository can actually be a *built* release. So you have GitHub Actions do the build and sign an attestation when it creates a GitHub release. This is exactly the same as the attestation that *would* be published to the package repository, once the package repository supports it.

docs/spec/v1.0/distributing-artifacts.md
behnazh-w: [Here](https://github.com/micronaut-projects/micronaut-core/releases/tag/v4.0.0-M1) is one example project that produces 30+ artifacts. We had to create an archive file and use `multiple.intoto.jsonl` to attest to the artifacts.

docs/spec/v1.0/distributing-artifacts.md
di: Yes, what @znewman01 said, and additionally:

- starting to publish attestations to a source repository gets the project 90% of the way there to publishing them in a package repository once it is supported
- a package repository that doesn't support publishing attestations might still support some verifiable link between the package and the source repository, which allows the attestations to be discovered there.

docs/spec/v1.0/distributing-artifacts.md
di: a78555c

docs/spec/v1.0/distributing-artifacts.md
di: I think "provenance" is a better term for this (that's conveniently already defined): 559e22d

docs/spec/v1.0/distributing-artifacts.md
di: > What if there's more than one? Do we smush all attestations into the same file?

* If there's more than one artifact (`a.zip`, `b.zip`) there should be more than one attestation (`a.intoto.jsonl`, `b.intoto.jsonl`)
* If theres one artifact (`a.zip`) but more than one provenance attestation (e.g. multiple reproducible builds with the same result) these should be in the same file (`a.intoto.jsonl`).

If that makes sense, I can make an update here to make that more clear.

> [Here](https://github.com/micronaut-projects/micronaut-core/releases/tag/v4.0.0-M1) is one example project that produces 30+ artifacts. We had to create an archive file and use multiple.intoto.jsonl to attest to the artifacts.

I would argue that in this case, `artifacts.zip` is the actual artifact, and `multiple.intoto.jsonl` is the attestation for that artifact, not the files within that artifact (which seems to be the case, as it contains a single attestation). The only difference is that `multiple` seems arbitrary, and that the filename should be something like `artifacts.intoto.jsonl`, or some other convention that the ecosystem has determined.

docs/spec/v1.0/distributing-artifacts.md
di: > This seems to imply that a given build's attestation is finished as soon as the build completes. I think that's insufficiently flexible: if we have reproducible builds and want attestations from two different builders, those would happen at different times. For dynamic information (like known vulnerabilities in an old package), you might even want new attestations years in the future, from a recurring scan.

I think the provenance for a given artifact (the "build attestation") shouldn't be published until the reproducible builders have finished and provided individual attestations which should be combined.

> Maybe the key thing here (see also the section on immutability) is that new attestations (and builds) can always be added to a release, but never removed.

Yes, exactly. Do you think this needs additional clarification, or does the immutability section cover it?

docs/spec/v1.0/distributing-artifacts.md
di: Yes.

docs/spec/v1.0/distributing-artifacts.md
di: I think that given that we explicitly define "provenance" as "attestations about how build artifacts were created", I'll stick with that interpretation here as implied, and will save discussions about source for the future "source" track.

znewman01: Fair enough!

docs/spec/v1.0/distributing-artifacts.md
znewman01: > I think the provenance for a given artifact (the "build attestation") shouldn't be published until the reproducible builders have finished and provided individual attestations which should be combined.

Who's responsible for combining them? I generally prefer a model where they remain separate and then the verification policy requires >1 individual attestation, retrieved from whatever source. We don't have to go into that detail in this doc but I'd prefer to leave that model possible based on what's here.
 
My vote would be to make a change like:

> The set of attestations for a given release is incomplete until all builds for a given release are finished.

to

> The set of attestations for a given release may grow (but never shrink) over time as additional builds and attestations are created.

If you have a strong preference for something closer to what exists, I can live with that, though 🙂 

> Yes, exactly. Do you think this needs additional clarification, or does the immutability section cover it?

I wouldn't mind some foreshadowing here, though it may not be strictly necessary. Your call.

znewman01 (approved):
Thanks for the updates! Feel free to resolve my prior comments. LGTM, though I just want to give one more (optional) piece of feedback.

The change to make this doc about "provenance" does seem to address my main concerns, at least from a "everything is now technically correct level."

My concern: the set of all attestations about an artifact may in fact grow over time. So once you expand from "provenance" to "attestations more generally," the recommendations in this doc no matter apply. That's fine, because this doc is just about provenance, but it seems like a waste because with only tiny modifications this would be a perfect doc about distributing attestations more generally (with a couple callouts about specific requirements for provenance)

Probably don't need to do that now, especially since I know everybody's eager to get this in, but wanted to explain the reason for my pedantry 😄

di:
Zach and I had a synchronous chat about his feedback. We determined there isn't anything here necessarily blocking future work, but there's a few changes we can make to ensure that's clear:

- clarify that there are lots of possible attestations, but in this context SLSA only cares about provenance (build attestations)
- it should be more clear that this is _only_ considering build provenance, produced by the same maintainers as the artifacts themselves (but possibly across multiple reproducers prior to publication)
- by uploading provenance alongside a release in the manner specified, publishers are 'elevating' build attestations (provenance) above all other possible attestations
- the goal is to provide all the attestations necessary for a repository to be able to verify some potential policy that requires a certain SLSA level for publication, not support the publication of arbitrary attestations
- therefore, all attestations necessary to publish the artifact should accompany the artifact at publish time

@znewman01, does that fully capture our discussion? If so, I'll move forward w/ the necessary updates.

znewman01:
SGTM!

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,155 @@
...
+between a source repository and an artifact that is being installed.
+
+However, for ecosystems that install from source repositories _via_ some
+intermediary (e.g. [Homebrew installing from GitHub release artifacts generated
+from the repository or GitHub Packages](https://docs.brew.sh/Bottles), [Go
+installing through the Go module proxy](https://proxy.golang.org/)), these
+ecosystems distribute "source archives" that are not the bit-for-bit identical
+form from version control.  These intermediaries as transforming the original
```
joshuagl, line 152:
```suggestion
form from version control.  These intermediaries are transforming the original
```
?

docs/_data/nav/v1.0.yml
```diff
@@ -54,14 +54,18 @@
     url: /spec/v1.0/requirements
     description: Detailed technical requirements for producing software artifacts, intended for system implementers
 
-  - title: Verifying build systems
-    url: /spec/v1.0/verifying-systems
-    description: Guidelines for securing SLSA Build L3+ builders, intended for system implementers
+  - title: Distributing artifacts
+    url: /spec/v1.0/distributing-artifacts
+    description: Detailed technical requirements for distributing software artifacts, intended for system implementers and software distributors
```
joshuagl, line 59: Should this just be named "Distributing provenance"?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,155 @@
+---
+title: Distributing artifacts
+description: This page covers the detailed technical requirements for distributing artifacts at each SLSA level. The intended audience is system implementers and software distributors.
```
joshuagl, line 3: Should this page be called "Distributing provenance"? We don't offer requirements for other types of artifact.

joshuagl (approved):
This LGTM. I left some comments, most notably on the page title and the terms "built distribution" and "source repository".

docs/spec/v1.0/distributing-artifacts.md
behnazh-w: > I would argue that in this case, artifacts.zip is the actual artifact, and multiple.intoto.jsonl is the attestation for that artifact, not the files within that artifact (which seems to be the case, as it contains a single attestation)

The `artifacts.zip` was created to avoid cluttering the release assets. The `multiple.intoto.jsonl` still attests to every subject included in  `artifacts.zip`  individually. We still want to keep a separate subject for every JAR artifact in the provenance, hence using `multiple.intoto.jsonl`.

It's common for a Java project to have multiple sub-projects and produce several JAR files as release artifacts, so we need a practical solution here. 

@laurentsimon @ianlewis any thoughts?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,155 @@
...
+    attestation in a monitored transparency log during verification helps
+    ensure the attestation is trustworthy.
+-   **Publish attestations alongside source releases in the source
+    repository**: For ecosystems where source releases are commonly published
+    to source repositories as well as to package registries (such as GitHub
+    releases), producers SHOULD include provenance as part of these releases if
+    the source repository supports attaching additional artifacts to a release.
+    This option requires no changes to the package registry to support
+    provenance formats, but means that the source repository in use is now in
+    the critical path for installers that want to verify policy at build-time.
```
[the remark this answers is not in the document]
di: Yes. Per https://slsa.dev/spec/v1.0/terminology this could just be "repository" but I think "source repository" makes it more clear. Maybe: "source platform"? I'm open to ideas.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,155 @@
...
+  url: verifying-artifacts
+---
+
+SLSA requires the distribution and verification of provenance metadata in the
+form of SLSA attestations, in order to make provenance for related artifacts
+available after generation for verification.
+
+This document provides specifications for distributing attestations, and the
+relationship between build artifacts and attestations. It is primarily
+concerned with artifacts for ecosystems that distribute "built distributions",
```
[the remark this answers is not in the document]
di: Yes, this is my Python slipping through. This should probably become "package" or "package artifact" per https://slsa.dev/spec/v1.0/terminology#package-model, thoughts?

docs/_data/nav/v1.0.yml
di: I think so, however our other complimentary pages are:

- ["Producing Artifacts"](https://slsa.dev/spec/v1.0/requirements)
- ["Verifying Artifacts"](https://slsa.dev/spec/v1.0/verifying-artifacts)

So I would be hesitant to change this unless we decided to change those as well. @MarkLodato, @kpk47, thoughts?

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,155 @@
...
+description: This page covers the detailed technical requirements for distributing artifacts at each SLSA level. The intended audience is system implementers and software distributors.
+prev_page:
+  url: requirements
+next_page:
+  url: verifying-artifacts
+---
+
+SLSA requires the distribution and verification of provenance metadata in the
+form of SLSA attestations, in order to make provenance for related artifacts
+available after generation for verification.
```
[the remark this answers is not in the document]
di: 9e8bb46

docs/spec/v1.0/distributing-artifacts.md
MarkLodato: As I understand it, the main requirement is that there is a well-defined API for retrieving the attestations for a particular artifact. Maybe that's not coming across clearly in the text, leading to confusion?

For example, suppose the ecosystem has an API where you fetch `<package>/<version>/<filename>`, e.g. `micronaut-aop/4.0.0-M1/micronaut-aop-4.0.0-M1-javadoc.jar`. There SHOULD (MUST?) be a well-known transform from that path to a corresponding `.intoto.jsonl` file, say `micronaut-aop/4.0.0-M1/micronaut-aop-4.0.0-M1-javadoc.jar.intoto.jsonl`. If a particular release has a one-to-may relationship between attestation bundles and artifacts (as in the link from @behnazh-w), then the recommendation is to make copies or links so that the API continues to work. The alternative of requiring multiple artifacts (`<package>/<version>/*`) to _always_ have a single bundle (`<package>/<version>/multiple.intoto.jsonl`) would likely no work since it would preclude cases where each artifact has a separate bundle.

Does that help?

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
MarkLodato: I was also confused by this. Defining a new term might be a good idea. In our model GitHub is actually serving as a package registry where the package name _is_ the source repository name.

docs/_data/nav/v1.0.yml
MarkLodato: No opinion - I'm OK with either. @olivekl in case you have thoughts

olivekl: I'd be inclined to go with "Distributing provenance" (or if you want to be more general, "Distributing attestations"). I think it's okay to leave the two pages names as is, since they have technical guidance related to artifacts apart from provenance.

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
MarkLodato: Maybe just avoid the confusion by saying "alongside the source repository releases"? Maybe:

> **Publish attestations alongside the source repository releases**: If the source repository hosting provider offers an artifact "release" feature, such as [GitHub releases](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases) or [GitLab releases](https://docs.gitlab.com/ee/user/project/releases/), producers SHOULD include provenance as part of such releases. This option has the benefit of requiring no changes to the package registry to support provenance formats, but has the disadvantage of putting the source repository hosting providing in the critical path for installers that want to verify policy at build-time.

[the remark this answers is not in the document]
joshuagl: That avoids the confusion and avoids defining a new term. WFM!

docs/spec/v1.0/distributing-artifacts.md
kpk47: Got it. So if I understand correctly, then the bootstrapping process is
1. generate provenance with your CI/CD system
2. publish provenance to the source repo
3. when possible, publish provenance to the pkg repo.

Can we make these steps more explicit? Maybe something like
> Combining these options gives us a process for bootstrapping SLSA adoption within an ecosystem, even if
> the package registry doesn't support publishing attestations. First, interested projects modify their release 
> process to produce SLSA provenance. Then, they publish that provenance to their source repository.
> Finally, they publish the provenance to the package registry, if and when the registry supports it.

kpk47 (changes requested):

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,156 @@
+---
+title: Distributing artifacts
+description: This page covers the detailed technical requirements for distributing artifacts at each SLSA level. The intended audience is system implementers and software distributors.
+prev_page:
+  url: requirements
+next_page:
+  url: verifying-artifacts
+---
+
+In order to make provenance for related artifacts available after generation
```
ianlewis, line 10: nit: 'related' feels redundant.
```suggestion
In order to make provenance for artifacts available after generation
```

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,156 @@
...
+release. At the time of a given build, the relevant provenance for that build
+can be added to the release, depending on the relationship to the given
+artifacts.
+
+## Relationship between artifacts and attestations
+
+Package ecosystems SHOULD support a one-to-many relationship from build
+artifacts to attestations to ensure that anyone is free to produce and publish
+any attestation they might need. Package ecosystems MUST provide a way to map a
+given artifact to corresponding provenance. The mappings can be either implicit
```
ianlewis, line 63: nit: trying to use consistent language.
```suggestion
given artifact to its corresponding attestation. The mappings can be either implicit
```

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,156 @@
...
+attestation is `<filename>.attestation` or some similar extension (for example
+[in-toto](https://in-toto.io/) recommends `<filename>.intoto.jsonl`.)
+
+## Where attestations are published
+
+There are a number of opportunities and venues to publish attestations during
+and after the build process. Producers MUST publish attestations in at least
+one place, and SHOULD publish attestations in more than one place:
+
+-   **Publish attestations in a transparency log**: Once an attestation has
```
laurentsimon, line 79: I don't think Rekor aims at being a distribution service per se. It's intended to be an append-only log, but there's no intention for Rekor to be used for discovery. That's why SIgstore bundles were created, ie to avoid the need to hit the log to discover the attestation.

I agree it's good to have attestations recorded in an immutable log. What's unclear is that we say `Producers MUST publish attestations in at least one place`, but if this only place is Rekor then verifiers will hit the log for discovery and use a hash of the artifact as index? This leads to attacks where someone DoSes the verifier by uploading many entries for a good artifact.

It's also a little unclear what "attestation" means. The signed attestation (payload + DSSE) or just the raw intoto provenance attestation?

/cc @asraa @bobcallaway @haydentherapper

ianlewis: It does seem that sigstore-js already does include a full sigstore bundle in the tlog entry which would include the dsse envelope and the full attestation payload. So I wonder if there was some discussion and consensus that was built prior to this.

https://github.com/sigstore/sigstore-js/blob/2484d689d9c6f54c0a65c80a9acac771ae5d7ed9/src/tlog/format.ts#L74-L118

/cc @bdehamer

docs/spec/v1.0/distributing-artifacts.md
ianlewis: I agree that the goal is for there to be a well known way to get the attestation given the location of the artifact. Ideally this is consistent throughout the ecosystem.

Ideally also it has a similar naming scheme so it's obvious to humans what attestations go with what artifacts.

behnazh-w: Thanks @MarkLodato . I agree that we need a very clear API. 

>The alternative of requiring multiple artifacts (<package>/<version>/*) to always have a single bundle (<package>/<version>/multiple.intoto.jsonl) would likely no work since it would preclude cases where each artifact has a separate bundle.

We don't need to recommend it in this way only. We can recommend two ways at least for the Java ecosystem:

1. allow multiple subjects being attested in a single attestation bundle by using `<group>/<version>.intoto.jsonl` (better than `multiple.intoto.jsonl` for sure). 
2. allow `<group>/<version>/<filename>.intoto.jsonl` for single artifacts in case that's preferred for a project. 

The recommendation for GitHub releases can be: artifacts can be published as individual files, or in an archive file that can be unpacked by a verifier. That greatly helps projects that have many artifacts. 

I predict that this is very soon going to be even more important as we start seeing SBOMs and other byproducts of a pipeline being attested by a SLSA provenance. Considering the way [slsa-github-generator](https://github.com/slsa-framework/slsa-github-generator) works, if we want to mandate "attestation SHOULD have a filename that is directly related to", we would need to call the reusable workflow each time, which will make release pipelines very slow. So an attatation bundle with multiple subjects should be definitely allowed.

>I'm also curious what happens with artifacts.zip and multiple.intoto.jsonl when it's published to the package registry: is it published as-is? Or is it broken apart into separate artifacts?

@di Right now we are using GitHub releases and creating `artifacts.zip` to publish the attestations next to the artifacts. But at the same time each of the artifacts included in `artifacts.zip` are separately published to Maven central, e.g., see [here](https://repo1.maven.org/maven2/io/micronaut/micronaut-aop/4.0.0-M1/). Once Maven and Gradle support publishing SLSA provenances, i.e.,  there will be a `multiple.intoto.jsonl` (or `<group>/<version>.intoto.jsonl`) to query from [Maven central](https://repo1.maven.org/maven2/io/micronaut/micronaut-aop/4.0.0-M1/) as well. Here the `multiple.intoto.jsonl` would have a one-to-many relationship but it could actually be beneficial to know what other artifacts were built by the same pipeline.

Note that right now we can manually verify the individual artifacts like below, but that's because I know the query URL in Maven central for the subjects in the generated provenance. Otherwise tools like [macaron](https://github.com/oracle-samples/macaron) can't establish this connection, hence the need for `artifacts.zip` in the GitHub release.
```bash
$curl --output micronaut-aop-4.0.0-M1-javadoc.jar -L https://repo1.maven.org/maven2/io/micronaut/micronaut-aop/4.0.0-M1/micronaut-aop-4.0.0-M1-javadoc.jar

$slsa-verifier verify-artifact micronaut-aop-4.0.0-M1-javadoc.jar --provenance-path multiple.intoto.jsonl --source-uri github.com/micronaut-projects/micronaut-core
```

docs/spec/v1.0/distributing-artifacts.md
haydentherapper: Agreed @laurentsimon, Rekor is meant to only record what it has been presented and not be storage. Logs ideally should be treated as untrusted, and everything is verifiable - inclusion of an entry, consistency of the log, etc. Trusting Rekor to provide the attestation is antithetical to the purpose of the log.

asraa: @ianlewis to clarify what's going on there

> It does seem that sigstore-js already does include a full sigstore bundle in the tlog entry which would include the dsse envelope and the full attestation payload. So I wonder if there was some discussion and consensus that was built prior to this.

When you send an intoto type to rekor, you add the dsse envelope including the full attestation payload. The intoto type recorded on the log does NOT include the full attestation payload, but rather a hash commit to it. A separate GCS bucket on Rekor currently stores a attestation payloads up to a configured side, but they are not on the immutable log (they are returned if available when querying the log). Check out the raw body of this [attestation](https://search.sigstore.dev/?logIndex=16138354)

It's not obvious but there is a difference in the proposed entry you send to Rekor and the committed entry to the log.

docs/spec/v1.0/distributing-artifacts.md
jchestershopify: Speaking about attestations retrieved from the package repository (as opposed to the source repository), I think it's important to keep in mind that a goal is to _not_ require an API to find that file's URL. It should be derivable from the name of the package in a predictable way.

Doing so makes it CDN-friendly, which is a huge deal for package repos. Without Fastly, services like RubyGems.org and PyPI would probably buckle in seconds.

mlieberman85: It's not clear to me the details need to be hashed out in the specification vs. it being more of a tooling discussion?

In some of the other working groups, like the securing software repos wg, there appeared to be some reluctance on being too prescriptive as far as an api goes.

mlieberman85: Yeah I think API is the wrong term here. Also, there appeared to be some reluctance from other working groups, like the securing software repos wg, to be too prescriptive on how folks should distribute and consume SLSA attestations. Looking at the wording in the distributing-artifacts.md though it does seem pretty flexible. Is there any concern it's not clear enough?

docs/spec/v1.0/distributing-artifacts.md
di: c731b0e

docs/spec/v1.0/distributing-artifacts.md
di: 7ad4d3a, which also includes changes from the rest of our discussion.

I dropped "but never shrink" because I think deleting artifacts (for ecosystems that support it) could/should result in attestations being deleted as well.

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
di: f8eddce

docs/_data/nav/v1.0.yml
di: 4cb2965

docs/spec/v1.0/distributing-artifacts.md
di: 4cb2965

docs/spec/v1.0/distributing-artifacts.md
di: What I'm taking from this discussion is that this should only recommend the "hash of an attestation and a pointer to where it is indexed" option. How about:

> -   **Publish attestations in a transparency log**: Once an attestation has
>     been generated for a build, a hash of the attestation and a pointer to
>     where it is indexed SHOULD be published to a third-party transparency log
>     that exists outside the source repository and package registry. Not only
>     are transparency logs guaranteed to be immutable, but they also more easily
>     enable monitoring.  Requiring the presence of the attestation in a
>     monitored transparency log during verification helps ensure the attestation
>     is trustworthy.

di:
Hi folks, thanks for all your feedback. I'd like this to be merged before the 2023-03-31 deadline to be included in RC2, and am working to close out the current discussion threads here.

I'd like to ask that unless you have very small and easily-changed fixes like typos, please hold adding any new large discussions that could delay merging by that deadline, and raise them as issues during the RC2 comment period instead.

Thanks!

docs/spec/v1.0/distributing-artifacts.md
di: > or in an archive file that can be unpacked by a verifier

I'd be concerned that this pattern would be adding an extra transformative step that lacks verification -- I think verification should be just "artifact" + "attestations" with no transformation necessary.

docs/spec/v1.0/distributing-artifacts.md
```diff
@@ -0,0 +1,154 @@
...
+long after the initial release.
+
+As a result, using a single attestation bundle per release is unnecessarily
+restrictive and does not work with the way complex, modern software artifacts
+are built and distributed.
+
+Thus, it is recommended to support multiple individual attestations or
+attestation bundles per release. At the time of a given build, the relevant
+build attestations for that build can be added to the release, depending on the
+relationship to the given artifacts.
```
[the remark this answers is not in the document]
di: @kpk47, this is still unresolved, any thoughts about whether we need changes here?

kpk47 (approved):

docs/spec/v1.0/distributing-artifacts.md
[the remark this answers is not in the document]
kpk47: Maybe turn around the first sentence to use SHOULD?

> Package ecosystems SHOULD support multiple individual attestations or attestation bundles per release.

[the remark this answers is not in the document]
di: Already done in d1c2621b15a2f6f108ab67e52c53b803ab63e1ac!

kpk47 (approved):

docs/spec/v1.0/distributing-artifacts.md
ianlewis: I think it should also be clear when we say:

> Producers MUST publish attestations in at least one place, and SHOULD publish attestations in more than one place:

that publishing a hash, or even the full attestation to the tlog only would not satisfy the MUST part of this statement. Right now it seems like a tlog is just one of the possible places to publish.

docs/spec/v1.0/distributing-artifacts.md
behnazh-w: > I'd be concerned that this pattern would be adding an extra transformative step that lacks verification -- I think verification should be just "artifact" + "attestations" with no transformation necessary.

Fair enough. It might be a good idea to keep the recommendation at "artifact" + "attestations" level. 

>Also, there appeared to be some reluctance from other working groups, like the securing software repos wg, to be too prescriptive on how folks should distribute and consume SLSA attestations.

Agreed. Because attestations allow multiple subjects (one-to-many relationship to artifacts), in practice we can't be prescriptive and require "The build attestation SHOULD have a filename that is directly related to the build artifact filename". Otherwise, we should just allow a one-to-one relationship between an attestation bundle and artifact to begin with.

lehors (changes requested):
A couple of minor points. LGTM otherwise. Could you please add the new file to onepage.md as well?
Thanks.

docs/spec/v1.0/distributing-artifacts.md
di: Thanks y'all, makes sense. Addressed in edb21fc.

lehors (approved):
I pushed a couple of small changes. LGTM now.

di:
@lehors Thanks. In the future, can we wait to merge until all discussions have been resolved? We might even want to turn on the "Require conversation resolution before merging" option for branch protection for this repo.

lehors:
@di, sorry I don't see any that isn't!? I scrolled through and didn't see any. I didn't see any request for changes either so I thought it was good to merge.

di:
@lehors, no worries! There's still an unresolved discussion here: https://github.com/slsa-framework/slsa/pull/673#discussion_r1148404932

lehors:
Ok, I'll blame github for hiding an unresolved discussion!! Sorry about that.

docs/spec/v1.0/distributing-artifacts.md
behnazh-w: @di Just to answer the question:
>I'm not sure if I follow: there's nothing here preventing an ecosystem from providing the exact same attestation as different filenames (even if it contains multiple subjects). Why do attestations with multiple subjects prevent us from being prescriptive here?

Thanks for the clarification. So you're saying we should copy the same attestation bundle under different names for every subject included. That would be doable when package repositories like Maven central start hosting the attestations, but until then it won't be practical for GitHub releases when a project has 30+ artifacts.

Let's continue the discussion in a follow-up PR :+1:

Issue #650, opened 2023-02-28
SLSA v1: unify "Verifying Artifacts" across provenance and spec


MarkLodato:
Right now the steps to verify an artifact is split between two pages:

- https://slsa.dev/provenance/v1#verification --- Detailed steps how to verify
- https://slsa.dev/spec/v1.0/verifying-artifacts --- Broader picture with fewer details, plus an assignment between Package Ecosystem vs Consumer that may not be accurate (e.g. what if the Consumer does the verification)

@joshuagl and I discussed discussed a possible reorganization:

- Move all of the content to the spec, with the provenance page just linking there. This way there's no jumping back and forth.
- In the spec, say that we assume the reader is using slsa.dev/provenance/v1, but explain that if they are using something else they need to do the equivalent. This would also force us to explain each step to say why we're doing it.
- Organize the page as follows (not sure of section titles):
  - Overview / what the purpose of verification is
  - Talk about architecture options:
    - package ecosystem vs consumer --- we recommend moving the work to the ecosystem but it's not mandatory.
    - upload time vs install time vs monitoring (first half of [Verifying Expectations](https://slsa.dev/spec/v1.0/verifying-artifacts#verifying-expectations-for-artifacts))
  - How to verify = merge content from [provenance/v1#verification](https://slsa.dev/provenance/v1#verification) plus second half of [Verifying Expectations](https://slsa.dev/spec/v1.0/verifying-artifacts#verifying-expectations-for-artifacts)
- Remove [Provenance Distribution](https://slsa.dev/spec/v1.0/verifying-artifacts#provenance-distribution) since that is out of place (already found on Producing Artifacts)

PR #675, opened 2023-03-09
Unify "Verifying Artifacts" across provenance and spec

kpk47:
Addresses #650 .


docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +168,204 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+
+[verify-step-3]: #step-3-check-dependencies-recursively
+
+Finally, recursively check the `resolvedDependencies` as available and to the
+extent desired. This mitigates [threat "E"]. While SLSA v1.0 does not have any
+requirements on the completeness or verification of `resolvedDependencies`, one
+might wish to go beyond SLSA's minimum requirements in order to protect against
+threats further up the supply chain.
+
+One possible approach is to recursively verify each entry in
```
olivekl, line 361: Just to clarify, is this one possible approach to checking dependencies recursively, or to going beyond SLSA's requirements to protect against threats further up the supply chain? Or are they the same?

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +168,204 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+
+[Step 1]: #step-1-check-slsa-build-level
+
+First, check the SLSA Build level by comparing the artifact to its provenance
+and the provenance to a preconfigured root of trust. The goal is to ensure that
+the provenance actually applies to the artifact in question and to assess the
+trustworthiness of the provenance. This mitigates some or all of [threats] "D",
+"F", "G", and "H", depending on SLSA Build level and where verification happens.
+
+Up front:
```
olivekl, line 200: Nitpicky, but does "up front" mean you do it once and not again? Is there a more precise phrase you can use here?

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +168,204 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+referenced here.
+
+### Step 1: Check SLSA Build level
+
+[Step 1]: #step-1-check-slsa-build-level
+
+First, check the SLSA Build level by comparing the artifact to its provenance
+and the provenance to a preconfigured root of trust. The goal is to ensure that
+the provenance actually applies to the artifact in question and to assess the
+trustworthiness of the provenance. This mitigates some or all of [threats] "D",
```
MarkLodato, line 197: broken link: `[threats]` is not defined (shows up with brackets in rendered page instead of link)

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +168,204 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+Note: This section assumes that the provenance is in the recommended
+[provenance format](/provenance/v1). If it is not, then the verifier must
+perform equivalent checks on provenance fields that correspond to the ones
+referenced here.
+
+### Step 1: Check SLSA Build level
+
+[Step 1]: #step-1-check-slsa-build-level
+
+First, check the SLSA Build level by comparing the artifact to its provenance
```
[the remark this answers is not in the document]
MarkLodato: You mean what we currently call "Step 1"? The thinking was that "determine the SLSA level" is an atomic step, and it's broken down into the four sub-steps below.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +168,204 @@ under the malicious actor's control. Some ways this could be achieved include:
...
-artifacts, they may still gain some benefit by relying on third-party verification
-(e.g. a monitoring service that publicizes artifacts that violate their expectations).
-
-The consumer may have to opt-in to enable SLSA verification, depending on the
-package ecosystem.
-
-Consumers may either audit the build systems
-themselves using the prompts in [verifying systems](verifying-systems.md) or
-rely on the [SLSA certification program](certification.md) (coming soon).
+![Threats covered by each step](/images/supply-chain-threats-build-verification.svg)
```
MarkLodato, line 183: @olivekl @kpk47 For here and below, it might be valuable to have all this copied verbatim in this PR so that we don't need to review that part, then have a separate PR to edit it. Otherwise it's hard to follow the change history.

docs/spec/v1.0/verifying-artifacts.md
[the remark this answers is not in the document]
olivekl: ok, that makes sense!

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +168,204 @@ under the malicious actor's control. Some ways this could be achieved include:
...
-> **TODO:** expand on this map model. Provide examples for ecosystems to follow,
-perhaps in the use-cases, and link to certification.
+## How to verify
 
 Verification MUST include the following steps:
 
 -   Ensuring that the builder identity is one of those in the map of trusted
     builder id's to SLSA level.
--   [Verification of the provenance](/provenance/v1#verification) metadata.
+-   Verification of the provenance metadata.
```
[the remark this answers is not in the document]
kpk47: Done.

docs/spec/v1.0/verifying-artifacts.md
kpk47: Done

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +168,204 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+In our threat model, the adversary has ability to invoke a build and to publish
+to the registry but not to write to the source repository, nor do they have
+insider access to any trusted systems. Expectations MUST be sufficient to detect
+or prevent this adversary from injecting unofficial behavior into the package.
+Example threats in this category include building from an unofficial fork or
+abusing a build parameter to modify the build. Usually expectations identify the
+canonical source repository (which is the entry in `externalParameters`) and any
+other security-relevant external parameters.
+
+The expectations SHOULD cover the following:
```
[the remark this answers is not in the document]
kpk47: The verifier sets expectations, though they may take input from the producer. For example, the verifier could ask the producer for the canonical source repo (producer input), or they could treat the source repo in the first provenance uploaded as canonical (no producer input).

Is the text confusing if you read it without the assumption that the producer sets expectations?

docs/spec/v1.0/verifying-artifacts.md
kpk47: Leaving this comment open for a follow-up PR.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +168,204 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+First, check the SLSA Build level by comparing the artifact to its provenance
+and the provenance to a preconfigured root of trust. The goal is to ensure that
+the provenance actually applies to the artifact in question and to assess the
+trustworthiness of the provenance. This mitigates some or all of [threats] "D",
+"F", "G", and "H", depending on SLSA Build level and where verification happens.
+
+Up front:
+
+-   Configure the verifier's roots of trust, meaning the recognized builder
+    identities and the maximum SLSA Build level each builder is trusted up to.
```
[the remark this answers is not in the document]
kpk47: We've got guidance in "Verifying Build Systems", and we're working on a conformance program (#515).

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -1,30 +1,39 @@
...
 SLSA uses provenance to indicate whether an artifact is authentic or not, but
-provenance doesn't do anything unless somebody inspects it. SLSA calls that inspection
-verification, and this page describes how to verify artifacts and their SLSA
-provenance. The intended audience is system implementers, security engineers,
-and software consumers.
+provenance doesn't do anything unless somebody inspects it. SLSA calls that
+inspection verification, and this page describes how to verify artifacts and
+their SLSA provenance.
 
 ## Overview
```
MarkLodato, line 8: nit: I think you can drop this heading. Also if you wrap the above paragraph in `<div class="subtitle">` it will show up bigger, like the other pages

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -1,30 +1,39 @@
...
+implementers must make regarding verifying provenance. The second describes
+for tool implementers the procedure for verifying an artifact and its
+provenance against a set of expectations.
 
-Package ecosystems verify an artifact's provenance against a set of expectations.
-Those expectations are set by the package ecosystem, with optional input from the
-artifact producer. Package ecosystems may choose to distribute only artifacts that
-pass verification, or they may choose to require that consumers opt-in to
-verification.
+## Architecture options
```
MarkLodato, line 15: note for the future: it would be nice to have a diagram here

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -1,30 +1,39 @@
...
-and software consumers.
+provenance doesn't do anything unless somebody inspects it. SLSA calls that
+inspection verification, and this page describes how to verify artifacts and
+their SLSA provenance.
 
 ## Overview
 
-Artifact verification is a process that consists of several stages that involves
-the artifact distribution system ("package ecosystem") and artifact consumer.
+This page is divided into two sections. The first discusses choices system
```
MarkLodato, line 10: nit: this is a bit confusing since there are three following sections (`##` headings)

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -1,30 +1,39 @@
...
+### Where to verify
+
+There are three options for where in a system to verify provenance: at the
+[package ecosystem], at the [consumer], or at an [external monitor].
+
+System implementers SHOULD prefer to perform verification at the package
+ecosystem because doing so provides security benefits to all of the
+package ecosystem's clients.
+
+#### Package ecosystem
```
MarkLodato, line 32: (Feel free to defer to a future PR)

nit: I'm having trouble following this many levels of heading.  Can we make this inner one a bulleted list (here and for "when to verify")? Something like:

```markdown
*   **Package ecosystem:** The package registry verifies ...

File: docs/spec/v1.0/verifying-artifacts.md [hunk 13]
```diff
@@ -1,30 +1,39 @@
...
-Those expectations are set by the package ecosystem, with optional input from the
-artifact producer. Package ecosystems may choose to distribute only artifacts that
-pass verification, or they may choose to require that consumers opt-in to
-verification.
+## Architecture options
 
-Consumers can trust their package ecosystem or some other third party to verify
-artifacts, or they can verify artifacts against their own set of expectations.
+System implementers must make two architectural decisions about verifying
+provenance: where in the system to verify and when to do it. Each option
```
MarkLodato, line 18: note for the future: I find this hard to follow since the "where" and "when" are usually tied in practice:

- "ecosystem" usually means "upload" (or at least that's what I think)
- "consumers" usually means "download"
- "monitor" means "continuously"

Are they actually independent? Can we simplify?

olivekl: +1, I had also wondered if "where" and "when" could be clearer.

docs/spec/v1.0/verifying-artifacts.md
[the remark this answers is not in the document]
olivekl: Hm, maybe it was my own confusion on that point muddying things. It makes sense if the verifier is setting expectations, yes. I will look double check in Terminology to see whether that definition ("setting expectations") is clear enough.

docs/spec/v1.0/verifying-artifacts.md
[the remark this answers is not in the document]
olivekl: If there's something short and simple to link to in the Verifying Build Systems page, let's add the link.  This sentence is almost a step in itself (i.e., "figure out what SLSA build level the builder is trusted up to"). Or am I overthinking this and the information won't be hard for users to get?

docs/spec/v1.0/verifying-artifacts.md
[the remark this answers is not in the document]
olivekl: Reopening this: I think the source of my confusion comes from the definition in [package terminology](https://slsa.dev/spec/v1.0/terminology): 

"Expectations: A set of constraints on the package’s provenance metadata. The package producer sets expectations for a package, whether explicitly or implicitly."

We should find alignment between that definition and what's being described here. Do both parties set expectations in different ways? (Producer sets expectations based on how they produced the package; consumer sets them based on what's expected from... what the producer set?)

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +148,206 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+    other expectations are defined in the source repository. This is how go
+    works, for example, since the package name *is* the source repository
+    location.
+
+TIP: Difficulty in setting meaningful expectations for `externalParameters` can
+be a sign that the `buildType`'s level of abstraction is too low. For example,
+`externalParameters` that record a list of commands to run is likely impractical
+to verify because the commands change on every build. Instead, consider a
+`buildType` that defines the list of commands in a configuration file in a
+source repository, then make put only the source repository in
```
joshuagl, line 326:
```suggestion
source repository, then put only the source repository in
```

joshuagl (commented):
thanks, this is a solid improvement. I added a few minor comments.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +148,206 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+| What | Why |
+| ---- | --- |
+| Builder identity from [Step 1] | To prevent an adversary from building the correct code on an unintended system |
+| `buildType` | To ensure that `externalParameters` are interpreted as intended |
+| `externalParameters` | To prevent an adversary from injecting unofficial behavior |
+
+Verifiers SHOULD reject unrecognized fields in `externalParameters` to err on
+the side of caution. It is acceptable to allow a parameter to have a range of
+values (possibly any value) if it is known that any value in the range is safe.
+Implementations need not special-case the `buildType` if JSON comparisons are
```
lehors, line 295: I'm not sure what "special-case" mean here.

docs/spec/v1.0/verifying-artifacts.md
lehors: This raises the question about how far one should go.

lehors (commented):
LGTM overall but I'm concerned by the significant change brought in with regard to who sets the expectations.

docs/spec/v1.0/terminology.md
```diff
@@ -96,7 +96,7 @@ build system the package was built.
 
 | Term         | Description |
 |--------------|---- |
-| Expectations | A set of constraints on the package's provenance metadata. The package producer sets expectations for a package, whether explicitly or implicitly. |
+| Expectations | A set of constraints on the package's provenance metadata. The provenance verifier sets expectations for a package, whether explicitly or implicitly. |
```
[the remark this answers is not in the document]
kpk47: It is, but I think it's a worthwhile change. The current division of labor makes the producer responsible for the verification tool's configuration, even if the producer and verifier are separate parties with no relationship. Shifting the responsibility to the verifier doesn't prevent the producer from influencing the expectations, but it does ensure that they aren't responsible for infrastructure they might not control.

There is precedent for this model with SSH keys. The user is ultimately responsible for the contents of their known_hosts file. There are better and worse ways to populate that file, but the remote isn't to blame if you trust an invalid key. The analogy isn't perfect, but I think the overall idea applies. 

I'm open to changing the definition, but I want to be sure that the spec doesn't give any party responsibility for things they can't control and that it supports trust-on-first-use as a model for setting expectations.

docs/spec/v1.0/verifying-artifacts.md

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -1,62 +1,104 @@
...
-and software consumers.
+provenance doesn't do anything unless somebody inspects it. SLSA calls that
+inspection verification, and this page describes how to verify artifacts and
+their SLSA provenance.
 
 ## Overview
 
-Artifact verification is a process that consists of several stages that involves
-the artifact distribution system ("package ecosystem") and artifact consumer.
+This page is divided into the sections. The first discusses choices system
```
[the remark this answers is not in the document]
kpk47: I was intentionally being a bit vague because verification can happen in the package ecosystem or on the consumer side, and those are built by different parties. I clarified that this section is targeting distribution and deployment systems.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -1,62 +1,104 @@
...
-Example threats in this category include building from an unofficial fork or
-abusing a build parameter to modify the build. Usually expectations identify
-the canonical source repository (which is the main external parameter) and
-any other security-relevant external parameters.
+A package's <dfn>consumer</dfn> is the organization or individual that uses the
+package.
+
+Consumers can set their own expectations for artifacts and use client-side
+verification tooling to ensure that the artifact's provenance matches the
+consumer's expectations for that package name before use (e.g. during
```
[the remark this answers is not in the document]

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -106,87 +148,206 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+
+Once, when bootstrapping the verifier:
+
+-   Configure the verifier's roots of trust, meaning the recognized builder
+    identities and the maximum SLSA Build level each builder is trusted up to.
+    Different verifiers might use different roots of trust, but usually a
+    verifier uses the same roots of trust for all packages. This configuration
+    is likely in the form of a map from (builder public key identity,
+    `builder.id`) to (SLSA Build level) drawn from the [SLSA Conformance
+    Program](certification.md) (coming soon).
```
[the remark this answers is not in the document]
kpk47: Fixed above

docs/spec/v1.0/terminology.md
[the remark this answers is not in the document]
olivekl: I think I understand now, and I believe what's happening here is confusion over a term meaning two things:

- The producer SETS expectations, meaning that they communicate authenticity features to help others verify their artifact;
- The consumer INPUTS (not sets) expectations, meaning that they take the expectations from the producer and feed them into a tool to verify the artifact.

Would splitting the idea of "set expectations" into two terms based on those roles help disambiguate what's going on here? Line 99 would revert to the original ("The package producer sets expectations for a package...") and we could add a sentence explaining that the consumer inputs these expectations into tooling.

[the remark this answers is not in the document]
lehors: I like @joshuagl's point of view better: the package producer and/or ecosystem sets _default_ expectations that the consumer may use but can override at verification time.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -9,63 +9,106 @@ next_page:
...
+
+[Monitor]: #monitor
+
+A <dfn>monitor</dfn> is a service that verifies provenance for a set
+of packages and publishes the result of that verification. The set of
+packages verified by a monitor is arbitrary, though it MAY mimic the set
+of packages published through one or more package ecosystems. The monitor
+MUST publish its expectations for all the packages it verifies.
+
+Verifiers can continuously poll a monitor to detect artifacts that
```
MarkLodato, line 91: nit: "Verifier" is not defined here, so I found it confusing who this meant. Can we replace with "Consumers"?

docs/spec/v1.0/verifying-artifacts.md
MarkLodato: The intention was to say this this is something that one probably wants to do, without calling it out as a requirement because, as @lehors said, there is a question of how far one should go.

docs/spec/v1.0/verifying-artifacts.md
MarkLodato: The intention was to proactively address a question that came up: does a verification tool have to have custom logic for each `buildTypes` it supports, e.g. to understand the schema of `externalParameters`? The answer is "probably not - regular JSON comparisons are likely fine". Any thoughts on how to word better?

MarkLodato (approved):
It would be good to get this submitted sooner rather than later, so maybe we can submit something that is good enough and then iterate on wording?

docs/spec/v1.0/terminology.md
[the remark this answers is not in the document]
kpk47: I reverted this change and am double-checking that this PR is consistent throughout wrt who is responsible for setting expectations.

I'll open a separate PR separating the two meanings of "expectations" so that we can discuss that change in isolation.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -9,63 +9,106 @@ next_page:
...
-Consumers can trust their package ecosystem or some other third party to verify
-artifacts, or they can verify artifacts against their own set of expectations.
+More than one component can verify provenance. For example, if a package
+ecosystem verifies provenance, then consumers who get artifacts from that
+package ecosystem do not have to verify provenance. Consumers can do so with
+client-side verification tooling or by polling a monitor, but there is no
+requirement that they do so.
 
-## Package ecosystem
+> **TODO** Add a diagram.
```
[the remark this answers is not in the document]

docs/spec/v1.0/verifying-artifacts.md

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -9,63 +9,106 @@ next_page:
...
+ecosystem tooling tests those expectations during upload to ensure all packages
+in the ecosystem are built from their canonical source repo, which
+indicates their authenticity.
+
+Expectations MUST be sufficient to detect or prevent an adversary from injecting
+unofficial behavior into the package. Example threats in this category include
+building from an unofficial fork or abusing a build parameter to modify the
+build. Usually expectations identify the canonical source repository (which is
+the main external parameter) and any other security-relevant external
+parameters.
```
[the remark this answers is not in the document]

docs/spec/v1.0/verifying-artifacts.md
kpk47: I reordered the text a bit to emphasize that this step is optional and best-effort.

docs/spec/v1.0/verifying-artifacts.md
kpk47: I tried turning it around. Instead of saying you don't need special-casing, I said that JSON comparisons are sufficient. WDYT?

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -339,24 +338,22 @@ repository is constant across builds.
...
-this process more efficient by recording the result of prior verifications. A
-trimming heuristic or exception mechanism will almost always be necessary
-because there will always be some transitive dependencies that are SLSA Build
+extent desired. Note that SLSA v1.0 does not have any requirements on the
+completeness or verification of `resolvedDependencies`. However, one might wish
+to verify dependencies in order to mitigate [threat "E"] and protect against
+threats further up the supply chain. If `resolvedDependencies` is incomplete,
+these checks can be done on a best-effort basis.
+
+A [Verification Summary Attestation (VSA)][VSA] can make dependency verificaion
```
MarkLodato, line 352:
```suggestion
A [Verification Summary Attestation (VSA)][VSA] can make dependency verification
```

MarkLodato (approved):

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -115,87 +158,203 @@ under the malicious actor's control. Some ways this could be achieved include:
...
+Finally, recursively check the `resolvedDependencies` as available and to the
+extent desired. Note that SLSA v1.0 does not have any requirements on the
+completeness or verification of `resolvedDependencies`. However, one might wish
+to verify dependencies in order to mitigate [threat "E"] and protect against
+threats further up the supply chain. If `resolvedDependencies` is incomplete,
+these checks can be done on a best-effort basis.
+
+A [Verification Summary Attestation (VSA)][VSA] can make dependency verificaion
+more efficient by recording the result of prior verifications. A trimming
+heuristic or exception mechanism is almost always necessary while verifying
```
lehors, line 354: nit: I'm not a native English speaker so feel free to ignore me on that one but the following sounds better to me:
```suggestion
heuristic or exception mechanism is almost always necessary when verifying
```

PR #676, opened 2023-03-09
Add spec level previous and next links (all versions)

mparsonspsycle:
- Added `prev_page` and `next_page` config to each spec page `.md` file (across all versions)
- Added logic to output links in `standard`, `default` and `specifications` layout templates

### Please Note...
- This has been done in a more manual/hardcoded way to ensure the flexibility needed to name/link to all spec pages. It could technically be done using Jeykell "Collections" and a more dynamic implimentation, but doing so would mean a larger rework of the folder structure and would be impacted by the current method of versioning. This hardcoded method seemed the safer option in the short term
- We have taken the order of pages in versions `1.0` and `1.0-rc1` from that shown in their respective `onepage` templates
- Both `1.0` and `1.0-rc1` specs have pages in their directories which don't appear in the `onepage` versions, so we have left them out for now (please feel free to add them in in the correct order if needed):
 - `1.0`: `verifying-artifacts`
 - `1.0-rc1`: `verifying-artifacts`, `whats-new`

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -16,7 +16,19 @@
           {% include toc.html html=content %}
         </aside>
         <div class="content">
+
           {{ content }}
+
+          {% if page.prev_page or page.next_page %}
+            <div class="mt-10 pt-10 border-t flex flex-col sm:flex-row space-between gap-5">
+              {% if page.prev_page %}
+                <a href="{{ page.prev_page.url }}" class="border rounded px-4 py-2 text-left">&lsaquo; {{ page.prev_page.title }}</a>
+              {% endif %}
+              {% if page.next_page %}
+                <a href="{{ page.next_page.url }}" class="sm:ml-auto border rounded px-4 py-2 text-right">{{ page.next_page.title }} &rsaquo;</a>
+              {% endif %}
+            </div>
+          {% endif %}
         </div>
       </main>
       {%- include footer.html -%}
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -16,19 +16,8 @@
           {% include toc.html html=content %}
         </aside>
         <div class="content">
-
           {{ content }}
-
-          {% if page.prev_page or page.next_page %}
-            <div class="mt-10 pt-10 border-t flex flex-col sm:flex-row space-between gap-5">
-              {% if page.prev_page %}
-                <a href="{{ page.prev_page.url }}" class="border rounded px-4 py-2 text-left">&lsaquo; {{ page.prev_page.title }}</a>
-              {% endif %}
-              {% if page.next_page %}
-                <a href="{{ page.next_page.url }}" class="sm:ml-auto border rounded px-4 py-2 text-right">{{ page.next_page.title }} &rsaquo;</a>
-              {% endif %}
-            </div>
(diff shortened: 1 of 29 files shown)


docs/_layouts/default.html
```diff
@@ -16,7 +16,19 @@
...
+          {% if page.prev_page or page.next_page %}
+            <div class="mt-10 pt-10 border-t flex flex-col sm:flex-row space-between gap-5">
+              {% if page.prev_page %}
+                <a href="{{ page.prev_page.url }}" class="border rounded px-4 py-2 text-left">&lsaquo; {{ page.prev_page.title }}</a>
+              {% endif %}
+              {% if page.next_page %}
+                <a href="{{ page.next_page.url }}" class="sm:ml-auto border rounded px-4 py-2 text-right">{{ page.next_page.title }} &rsaquo;</a>
+              {% endif %}
+            </div>
+          {% endif %}
```
MarkLodato, line 31: Q: Since this is the same on every page, is it worth creating a template for it?

docs/_layouts/default.html
```diff
@@ -16,7 +16,19 @@
...
         <div class="content">
+
           {{ content }}
+
+          {% if page.prev_page or page.next_page %}
+            <div class="mt-10 pt-10 border-t flex flex-col sm:flex-row space-between gap-5">
+              {% if page.prev_page %}
+                <a href="{{ page.prev_page.url }}" class="border rounded px-4 py-2 text-left">&lsaquo; {{ page.prev_page.title }}</a>
+              {% endif %}
+              {% if page.next_page %}
```
MarkLodato, line 27: Would it be difficult to add an "Index" link in the middle that goes to `./`? That could be useful.

MarkLodato (approved):
This is perfect. Thank you! I tested it on mobile and desktop and it works great.

I agree with the decision to have it be manual for now. This gives us the UX we want - we can always clean up internals later.

docs/_layouts/default.html
mparsonspsycle: @MarkLodato Yep, we can add this. Do you want it as a separate PR (along with the change above), or hold on merging this PR until we can get it updated?

MarkLodato: Either way is fine. Whatever's easiest for you.

scott-psycle: @MarkLodato Before we do make this change I'm not sure it really helps in this situation.  The links are navigation between pages of the spec, so dropping an `index` link in between previous and next may be confusing, rather than helpful.

MarkLodato: OK, so let's not do it in this PR at least.

The reason I asked was because when I was on the page, I wanted to to back to the index page (I do this a lot!) but there's no link. Adding a link between the two was one idea, but it could also be breadcrumbs or something else.

scott-psycle: Thanks @MarkLodato .  Lets pick it up separately.

docs/_layouts/default.html
mparsonspsycle: I've set this PR as draft for now until we can get this change in place 👌

MarkLodato (approved):
Amazing, thank you!

Issue #649, opened 2023-02-27
provenance/v1-rc1: Ability to store inline byproducts


lcarva:
In version 0.2 of SLSA Provenance, the unstructured nature of `buildConfig` allowed build systems to store additional information about the build that didn't necessarily fit neatly under other attributes. [Tekton Chains](https://github.com/tektoncd/chains/) leverages this when generating SLSA Provenance v0.2 for builds generated by a [Tekton Pipeline](https://tekton.dev/docs/pipelines/). 

Today Chains includes Task results in the SLSA Provenance v0.2 attestation. (A Pipeline is composed of Tasks, and Tasks can generate a result.) These results are small. Having them in the attestation means a verifier can leverage their value during a policy evaluation. 

For example, I have a Pipeline that includes a Task that performs certain checks against the checked out code, and produces a mini report, e.g. `{"x": 5, "y": 0, "z": 99}`. Once the build is complete, I evaluate its provenance attestation against a policy that looks at those values to decide whether or not the underlying build artifact should be promoted to the next environment.

Having the ability to specify an Artifact, not just an ArtifactReference in byproducts would allow me to continue to leverage this workflow.

The workaround is to push those results to an external location, and simply use the existing ArtifactReference. However, this has a few downsides. First, infrastructure to store such results is required. Second, verification becomes more complex.

MarkLodato:
I'm not opposed to this for cases where you really do want to add just a tiny bit of extra information. Maybe a `value` field that is an arbitrary JSON object, probably mutually exclusive with `digest`?

@marcelamelara FYI since you're probably interested and might want to weigh in

MarkLodato:
> A URI wouldn't be needed either since the artifact info is in-line, meaning that it's no longer a reference, I think. Could we retrofit the `mediaType` field to indicate that the object should be parsed as an in-line object by the consumer?

I think these are separate concepts:

- What the artifact represents: `uri`
- How it was obtained: `downloadLocation` (or `uri`)
- Its value: `digest`, `value` (if accepted)
- How the value was interpreted: `mediaType`, `localName`

marcelamelara:
[...]

Issue #614, opened 2023-02-08
Simplify ArtifactReference fields?


marcelamelara:
Do we need both `uri` and `downloadLocation` fields when artifacts are always matched on `digest`? 

This question came out of a discussion happening over at https://github.com/in-toto/attestation/pull/133, so I wanted to bring this topic into the SLSA community. 

Currently the ArtifactReference description in the Provenance v1.0 draft says that the `uri` must be set when the `digest` is not, implying that matching can in fact occur on the `uri`. This seems to be in conflict with our current definition of "artifact", which is any object that is hashable.

My take is that, we might not need the `uri` field in an ArtifactReference. For one, it makes the model for artifact matching less ambiguous. It seems like the `uri` is being used as both a type and (in some cases) location hint of the artifact, but IMO these two aspects can be clearly indicated via the `mediaType` and `downloadLocation` in an ArtifactReference.

Would love to hear folks' thoughts on this. cc/ @MarkLodato @TomHennen

di:
My two cents: an "artifact" implies something to be hashable.

In https://github.com/in-toto/attestation/pull/133#issuecomment-1422692712, an example of a URI for a builder is given as being something that could be a non-hashable `ArtifactReference`, but I think that would not be a common assumption for most people, and I agree with @marcelamelara's suggestion https://github.com/in-toto/attestation/pull/133#issuecomment-1425044269 that we should handle artifact and service references separately.

Given that, I think the concrete things we could do here in Provenance v1.0 is:

- make `digest` required for `ArtifactReference`;
- create a separate `ServiceReference` for un-hashable references;
- expand `builderDependencies` to permit both `ArtifactReference` and `ServiceReference` types.

That said, I think retaining the `uri` field for artifacts is worthwhile, as it provides additional metadata that might not otherwise be available via the `mediaType` or `downloadLocation`, like ecosystem, version, etc.

marcelamelara:
Thanks for your input @di!

There's actually a third use case for references that aren't services, but more broadly non-hashable resources/entities, such as GitHub pull requests (see [ITE-4](https://github.com/in-toto/ITE/tree/master/ITE/4#use-case-2-github-entities)). In that case, maybe defining something like a `ResourceReference` would encapsulate all of these non-artifact use cases?

[...]

PR #684, opened 2023-03-14
Replace ArtifactReference with new in-toto ResourceDescriptor type

marcelamelara:
This PR replaces the `ArtifactReference` type with the slightly more general [ResourceDescriptor](https://github.com/in-toto/attestation/blob/main/spec/v1.0-draft/resource_descriptor.md) type from the in-toto Attestation Framework, ahead of its v1.0 release.

Note: in-toto doesn't have the protobuf for `ResourceDescriptor` yet, so I have left those in here.

Addresses #614 and #649

### DIFF
--- a/docs/provenance/schema/v1-rc1/provenance.cue
+++ b/docs/provenance/schema/v1-rc1/provenance.cue
@@ -1,6 +1,6 @@
 {
     // Standard attestation fields:
-    "_type": "https://in-toto.io/Statement/v0.1",
+    "_type": "https://in-toto.io/Statement/v1",
     "subject": [...],
 
     // Predicate:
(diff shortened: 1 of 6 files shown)


docs/provenance/schema/v1-rc1/provenance.cue
```diff
@@ -1,6 +1,6 @@
 {
     // Standard attestation fields:
-    "_type": "https://in-toto.io/Statement/v0.1",
+    "_type": "https://in-toto.io/Statement/v1",
```
MarkLodato, line 3: nit: let's leave v1-rc1 as-is (snapshot in time) and just update v1. Thanks.

docs/provenance/schema/v1/provenance.cue
```diff
@@ -37,9 +37,11 @@
         // TODO: list the other standard algorithms
         [string]: string,
     },
-    "localName": string,
+    "name": string,
     "downloadLocation": string,
     "mediaType": string,
+    "content": string, // base64-encoded bytes
```
MarkLodato, line 43:
```suggestion
    "content": bytes, // base64-encoded
```

docs/provenance/schema/v1/provenance.proto
```diff
@@ -18,27 +18,29 @@ message BuildDefinition {
...
 
-message ArtifactReference {
+message ResourceDescriptor {
   string uri = 1;
   map<string, string> digest = 2;
-  string local_name = 3;
+  string name = 3;
   string download_location = 4;
   string media_type = 5;
+  string content = 6; // base64-encoded bytes
```
MarkLodato, line 30:
```suggestion
  bytes content = 6;
```

MarkLodato (changes requested):
Thanks!

docs/provenance/schema/v1-rc1/provenance.cue
marcelamelara: Makes perfect sense! I wasn't sure if we wanted to update the rc1 spec as well, so just went ahead, but I'll revert the changes.

docs/provenance/schema/v1/provenance.cue
marcelamelara: FWIW, I noticed that digests are typed as strings right now (in the `map<string, string>` type)... should we revisit that since the digest values are also really encoded bytes?

MarkLodato: The `bytes` field in CUE and protobuf means binary data that is encoded as base64 in JSON. For the digests, it's not necessarily base64 so we need to use `string`.

Issue #20, opened 2021-04-14
Add rationale for all of the requirements


MarkLodato:
It helps to explain not just *what* a system must do to meet SLSA N, but also *why*.

Thanks @david-a-wheeler for the suggestion.

MarkLodato:
Prior art: https://bestpractices.coreinfrastructure.org/en/criteria

david-a-wheeler:
An easy way to do this would be to use HTML `<details>` ...rationale... `</details>`

david-a-wheeler:
E.g., Requirement "Scripted build" says:

> All build steps were fully defined in some sort of “build script”. The only manual command, if any, was to invoke the build script.
> 
> Examples:
>
> - Build script is Makefile, invoked via make all.
> - Build script is .github/workflows/build.yaml, invoked by GitHub Actions.

I think we need to document the rationale, something like this:

Rationale: If the only way to rebuild the software is to manually enter special commands, over time rebuilders are likely to forget/omit steps, leading to subtly wrong builds. In addition, an automated script is necessary to implement automated pipelines to rebuild and test built artifacts. Note that it's fine for rebuilding scripts to support options (e.g., via parameters and environment variables) so users can instruct an automated build for a specific purpose while still having an automated script.

david-a-wheeler:
Here's a tweaked rationale for "Scripted build" - I think every rationale needs to explain *why* it's relevant to security/integrity. I also add a reference to the best practices badge requirement for the same thing if there's a build step available. I chose "scripted build" because I figured it shouldn't be hard to write a rationale for it :-).

Rationale: If rebuilding the software requires manual steps, rebuilders are likely to forget/omit steps or execute them incorrectly in some cases, leading to subtly wrong builds that can, for example, have latent vulnerabilities. In addition, an automated script is necessary to implement automated pipelines to rebuild and test built artifacts, and automated testing is vital for early detection of vulnerabilities (preferably before the software is released for use). Note that it's fine for rebuilding scripts to support options (e.g., via parameters and environment variables) so users can instruct an automated build for a specific purpose while still having an automated script. This requirement is not applicable if there are no build steps, however, note that building a package for distribution is still a build step, even if it's for an interpreted library. This is also required by the OpenSSF Best Practices badge ["build"](https://bestpractices.coreinfrastructure.org/en/criteria/0#0.build) requirement.

FYI, here's the full text of the OpenSSF Best Practices badge requirement "build":

[...]

PR #686, opened 2023-03-15
Align spec stages with Community Specification License framework

lehors:
This addresses [Issue #20 opened in the governance repo](https://github.com/slsa-framework/governance/issues/20)

### DIFF
--- a/docs/_data/versions.yml
+++ b/docs/_data/versions.yml
@@ -4,7 +4,7 @@
 #       versions:
 #         <version>:
 #           name: <string> # display name
-#           status: <string> # one of: Working Draft, In Review, Stable, Retired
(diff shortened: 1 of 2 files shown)


docs/_data/versions.yml
```diff
@@ -17,13 +17,13 @@ spec:
   versions:
     v0.1:
       name: Version 0.1
-      status: Stable
+      status: Approved
```
MarkLodato, line 20: Need to update line 7 as well

MarkLodato (approved):

docs/_data/versions.yml
lehors: Indeed. Good catch! Thanks.

joshuagl (approved):
LGTM, thanks Arnaud!

Issue #689, opened 2023-03-15
Preview image in link unfolding is too large


arewm:
When I post a link to slsa.dev (i.e. in Slack), the preview image that unfolds is massive. Can we associate a smaller image to make image linking work better?

The referenced URL of the image (by Slack again) is:
https://slack-imgs.com/?c=1&o1=ro&url=http%3A%2F%2Fslsa.dev%2Fimages%2Ficons%2Fandroid-chrome-512x512.png

MarkLodato:
Looks like it fixes the issue with it being huge. Probably still more to do to improve the social media preview, but I'll call this fixed for now.

PR #690, opened 2023-03-15
Change social image to be slightly smaller

arewm:
This may not completely remedy the situation presented in #689, but it should at least be a little bit better.


MarkLodato (approved):
Thanks! Confirmed this worked in Slack.

PR #694, opened 2023-03-16
Simplify implementation of nav template/css

MarkLodato:
There is no user-visible difference.

- Remove unnecessary HTML classes.
- Remove unused liquid variables.
- Remove duplication between nav.html and nav-level.html.
- Combine relevant SCSS bits into one file.


joshuagl (approved):
LGTM and seems much more maintainable for folks, like me, who aren't front-end developers. Thanks Mark!

Issue #693, opened 2023-03-16
Preview text in unfolded links displays static content


arewm:
When pasting links to slsa.dev into common platforms that perform link unfurling, it would be great if a scoped and accurate description were displayed instead of every page having a description of 

> SLSA: Security framework to ensure software supply chain integrity 

I believe that this arises because the [default description](https://github.com/slsa-framework/slsa/blob/1fd20a9a9b76b4eba3e9565edcd2d94eccfd5ba1/docs/_config.yml#L2) is always used since it is never specified in the [YAML frontmatter](https://github.com/slsa-framework/slsa/blob/1fd20a9a9b76b4eba3e9565edcd2d94eccfd5ba1/docs/_includes/specifications-list.html#L7-L15). 

```yaml
    ---
    subpages:
      - title: Some page
        description: Text to display below the link
        relative_url: foo
      - title: Another page
        description: More text
        relative_url: ../bar
    ---
```

I believe that if we add a `description` to all pages that the preview text should be customized in link unfurling. 

This is further supported by inspecting the HTML on pages and finding
```html
<meta property="og:description" content="SLSA: Security framework to ensure software supply chain integrity">
```
Additionally, if you _wanted_ to specify a different image for the page's link unfurling, I assume that that would be accomplished via the HTML tag

```html
<meta property="og:image" content="http://slsa.dev/images/icons/android-chrome-192x192.png">
```

The related markup is `og:*`: https://ogp.me/

arewm:
The Jekyll plugin that is actually configuring this seems to be [jekyll-seo-tag](https://github.com/jekyll/jekyll-seo-tag/blob/master/docs/usage.md).

> The SEO tag will respect the following YAML front matter if included in a post, page, or document:
>
> * `title` - The title of the post, page, or document
> * `description` - A short description of the page's content
> * `image` - URL to an image associated with the post, page, or document (e.g., /assets/page-pic.jpg)
> * `author` - Page-, post-, or document-specific author information (see [Advanced usage](https://github.com/jekyll/jekyll-seo-tag/blob/master/docs/advanced-usage.md#author-information))
> * `locale` - Page-, post-, or document-specific locale information. Takes priority over existing front matter attribute lang.

PR #695, opened 2023-03-16
Add description across SLSA pages.

arewm:
Link unfurling and SEO currently only uses the default description for the entire webpage. In order to provide more clarity to this, I unified pages' frontmatter to specify the title and provide a description.

Resolves: #693

Blog posts don't need to have descriptions added as they seem to be handled differently. Their description is populated by the initial content in the post.

I added basic one-liner content to each of the pages, but the HTML tag (and unfurling) can handle longer descriptions. Feel free to offer suggestions about what the description should be.

I also modified the earlier versions (i.e. 1.0-rc draft) as I figured that this was beneficial and does not directly affect the content displayed in the specifications.


arewm:
A preview of the difference:
Screenshot 2023-03-16 at 11 03 57 AM

lehors (approved):
LGTM.

arewm:
@lehors, I rebased on a recent commit to deconflict, but I also inflated the descriptions in the v1.0 directory as I since realized that it is okay for the snippets to be more than just a simple short line.

@olivekl, FYI as well as I moved the first headings into the frontmatter as a title to ensure that it is picked up by the SEO plugin.

docs/attestation-model.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Software attestations
+description: Overview of software attestations and is specifics related to SLSA.
```
joshuagl, line 3:
```suggestion
description: Overview of software attestations and its specifics related to SLSA.
```
?

joshuagl (changes requested):
This is a great set of changes, thank you! Two high-level questions/suggestions:
1. can we finish each description with a period (.) for consistency?
2. should we include the version in descriptions, so that it's easy to tell, i.e., v0.1 requirements form v1.0 requirements in a preview?

joshuagl (approved):
Thanks for the fixes Andrew. Your rationale for not including the versions seems sane to me, let's not do anything with that suggestion for now.

lehors:
We're better off not adding the version to this because it would be one more thing we'd have to maintain. We need to try and limit the number of things that need updating every time we produce a new release.

trishankatdatadog (approved):

PR #696, opened 2023-03-17
Add a format selector to switch between onepage and multipage

lehors:
This change adds a format selector when a onepage exists so that the user can switch between formats.

### DIFF
--- a/docs/_includes/format-selector.html
+++ b/docs/_includes/format-selector.html
@@ -1,0 +1,39 @@
+{%- comment %}
+If a onepage exists adds a selector to choose between the multiple
+pages and one page formats.
+{%- endcomment %}
+
+{%- if page.name == "index.md" %}
+{%- assign onepageurl = page.url | append: "onepage" %}
+{%- else %}
+{%- assign url_parts = page.url | split: '/' %}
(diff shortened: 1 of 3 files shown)


MarkLodato:
Thanks! This is helpful, but I'm concerned that that area is getting too crowded. See screenshot - almost the entire screen is filled with selectors rather than content.

Can we move it to the nav bar instead? That seems like a simpler solution and also avoids clutter in the main content area. FYI I have an upcoming PR to redo the nav bar on the left; I'm waiting for #694 to merge before sending.

image

MarkLodato:
Here is my new PR, which includes the onepage in the nav bar: https://github.com/slsa-framework/slsa/pull/698

lehors:
On what device to you get that view??
Of course there is always a trade-off with these things. Adding it to the navbar makes the navbar even longer... If it were up to me I would still go with having it the way this PR presents it but this isn't a hill I'm willing to die on either. :-)

lehors:
I agree this a nice layout! (python.org)

MarkLodato:
Shall we close this since we now have links in the left-hand bar?

lehors:
I still like it better but ok.

PR #683, opened 2023-03-14
Update docs structure 1

olivekl:
First in a series of PRs to update docs; see [Issue 672](https://github.com/slsa-framework/slsa/issues/672)

- Move content and create pages according to [Issue 672](https://github.com/slsa-framework/slsa/issues/672)
- No changes to the moved content (content updates are forthcoming in another PR)
- Add descriptive material to pages/sections for clarity about what’s moved
- Add or update Previous/Next nav buttons
Note: some nav buttons are missing because not all files are in the Spec directory yet (will move and add in later PR)

Issue #662, opened 2023-03-06
Separate hosting for community-maintained buildTypes?


MarkLodato:
In https://github.com/slsa-framework/slsa/pull/623#issuecomment-1456240866, @lehors brings up the point that slsa.dev should not be the home for buildType definitions for proprietary or commercial build systems. I still think it's valuable to have a central repository for community-developed definitions, even for commercial build systems, since no build system maintains an official one yet. My thinking:

1. For community-maintained definitions, there is no natural place to put it.
2. Even for official definitions, hosting can be difficult, especially coming up with a good short URL.
3. Having multiple definitions adjacent to one another brings consistency and ease of comparison.

What do you think about creating a new domain for this, e.g. `https://buildtype.slsa.dev`? This could use a different theme and clearly state that these are community maintained and not owned by the OpenSSF. The top-level page (`/`) could serve as the index, even for ones that are not hosted there.

lehors:
To summarize my point: I think there is value in having an index. I don't think we should host the definitions.
I would also add that there are examples of such registries we could look at. One I have in mind is the W3C DID registry, although I would keep each entry even simpler: https://github.com/w3c/did-spec-registries

MarkLodato:
But we are writing some of these build types and we need a place to host them. What should we use for those?

You suggested slsa-github-generator but that's not right either - it's not specific to that project.

TomHennen:
FWIW for in-toto attestations we provide [a common location](https://github.com/in-toto/attestation/tree/main/spec/predicates) for folks to define these definitions.  When reviewing PRs we provide some feedback to the specifiers to see if what they're proposing makes sense.

A similar structure for buildType definitions would seem to make sense?

MarkLodato:
Discussed at today's [spec meeting](https://slsa.dev/notes/spec) but didn't come to any decision.

Nikokrock:
Maybe an idea would be that SLSA maintains an index of index instead of directly the proper index. This way the only burden would be to register **only once** a community (a default one could be created) or a "company/provider".

Additional checks could be provided (such as relation between a given index url and the built type urls, ...)

joshuagl:
I strongly agree with the sentiment that we should have an index. 

[...]

PR #698, opened 2023-03-17
Customize nav bar per version, and populate spec index automatically

MarkLodato:
Now the nav bar (almost) always shows the page the user is visiting, rather than only linking to the latest stable version. Users can select version by using the existing version selector, though that is not particularly discoverable. Might be good to move the version selector to the nav bar in a future PR.

**The nav layout has changed for v1.0.** Now the pages are all at the top level, rather than being nested under "specifications". I believe this is in line with #683 but it warrants specific consideration from reviewers.

Also auto-populate the spec/v1.0/index page from the same data file so that we don't have to repeat all the links and descriptions. (A good future PR would be to figure out how to do this for the next/previous links, but I haven't yet figured out a way to do this that isn't slow and clunky.)

Known issues:

- When viewing /spec/v0.1/, the attestation links are for /provenance/v0.1 and /verification_summary/v0.1 instead of v0.2. Ideas for fixes are welcome.

Intended omissions:

- /example is still not shown in the nav bar (probably OK?)
- /github-actions-workflow/... not shown in the nav bar (intended - will be moved as per https://github.com/slsa-framework/slsa/issues/662)

### Previews

v0.1

v1.0-rc1

v1.0


lehors (approved):
Updating the navbar according to the version of the spec being looked at is very much needed. So thank you for this!

I also support the automatic rendering of the table in the index page. We really have to reduce the number of files we have to update every time we make changes to the structure of the spec.

I don't know what performance problem you faced but for that matter I think we ought to have one config file of sort that defines what is to be considered part of the spec and use this as the only source of truth for the pagination links and the onepage format. The pagination links are already badly broken as a result of #683 and it's going to be an ongoing battle to keep these straight and files are being added without onepage.md being updated.
Anyway, I digress, this doesn't need to be part of this PR.

Besides, I'm not really fond of the "single-page view" entry which seems a bit lost in all this but I recognize that this is a matter of taste.

MarkLodato:
Sounds good! Yes, I'm considering this an iteration towards something better.

I absolutely would like to get the next/previous automatic sooner rather than later. I do think it's possible with some effort. Jekyll/Liquid is so painful to work with that it takes hours to figure out tricks to work around lack of basic features.

olivekl:
> **The nav layout has changed for v1.0.** Now the pages are all at the top level, rather than being nested under "specifications". I believe this is in line with #683 but it warrants specific consideration from reviewers.

Very glad for the updated nav bar and auto-populated index! I do think, though, that having Understanding SLSA, etc at the same level as Specification Index could be confusing. Is three layers of nesting too much visually for the nav bar to handle?

lehors:
I know what you mean about Jekyll/Liquid being so painful to work with!!! I always have to go through a bunch of trials with printouts to figure out the right incantation!
I have a patch to automatically get the page title for the navigation prev/next links. It's not a complete solution but that's one fewer piece of info we can avoid having to maintain and get wrong. With that change we only need to specify the page.prev_url and page.next_url front matter.
I've seen a case where we had the title right and the url wrong or vice versa so that simplification seems worthwhile already.

MarkLodato:
I just pushed a version that simplified a bit, based on feedback from @olivekl.

kpk47 (approved):
LGTM

MarkLodato:
Thanks, all. I'll submit this now and then iterate on the design. For example, we probably want to make each category collapsible so that the list isn't so long by default, and also to render the non-clickable titles differently. But I think this is still an improvement from the status quo.

PR #699, opened 2023-03-17
Review requirements related to provenance completeness for L2 and L3.

Nikokrock:
Describe which part of the provenance MUST be complete and which part is best effort for each level.


MarkLodato (approved):
Thanks!

Issue #657, opened 2023-03-03
SLSA v1.0 RC feedback - Ephemeral and isolated


arewm:
> The build SHOULD NOT call out to remote execution unless it’s considered part of the “builder” within the trust boundary [[ref](https://slsa.dev/spec/v1.0-rc1/requirements#:~:text=The%20build%20SHOULD%20NOT%20call%20out%20to%20remote%20execution%20unless%20it%E2%80%99s%20considered%20part%20of%20the%20%E2%80%9Cbuilder%E2%80%9D%20within%20the%20trust%20boundary)]

Can this statement be clarified as it seems impossible to verify in its current form. Some possible points of confusion that result from this statement include:
* Would this prevent the vendoring of binary content as some process of the build was performed outside the trust boundary?
* Would this prevent `curl`ing pre-built artifacts to be used in a resulting artifact?
* Would leveraging something like NTP within the build be remote execution?
* Should this prevent pulling of untrusted container artifacts as well since this can also be a "building" process (i.e. https://github.com/imjasonh/kontain.me/tree/main/cmd/buildpack)?
* Since any remote access can potentially result in remote execution, should this be interpreted as a requirement for hermetic builds in complete network isolation?

Additionally, one component of isolation from other processes (and therefore reproducibility) is to _not_ use tag-based container pullspecs and instead use digest-pinned pullspecs. Since tags can change between build runs, it is possible that two builds can violate isolation if any tag-based references are used.

arewm:
This comment seems related:

> The way we defined the core elements of a "secure" build service in the Secure Software Factory Ref Arch is pretty much as follows:
> * Secure orchestration of the build pipeline. i.e. only approved pipelines that follow established policy and rules should be allowed to run.
> * Secure orchestration of the build workload. i.e. only approved build workloads are allowed to be run, and if someone tampers with the orchestration of a workload it should be detected or prevented. An example here is using SPIFFE or other workload identities.
> * Secure the actions in the build workload itself. i.e. a build that tries to perform malicious actions should be prevented and the build workload should be isolated enough to limit blast radius. [[ref](https://github.com/slsa-framework/slsa/issues/599#issuecomment-1412319904:~:text=The%20way%20we,limit%20blast%20radius)]

[...]

Issue #423, opened 2022-07-13
Consider removing "optional" requirements


MarkLodato:
Feedback received from a customer: the "optional" requirements add confusion because it's not clear how to use them. If they are optional, why are they listed as requirements at all? And why are they listed as optional for some levels but not all?

For v1.0, I think we should consider whether they continue to make sense in the spec. If the intent is to nudge or recommend things, perhaps there's a more clear way to do it?

marcelamelara:
I assume by optional requirements, you mean the ones labeled with "O" on this page? https://slsa.dev/spec/v0.1/requirements

Source requirements aside, there are two optional requirements "Reproducible" (Build reqs), which only applies to L4, and "Includes Metadata" (Provenance reqs), which is indeed rather vague. 

I'm inclined to recommend removing the Includes Metadata requirement from 1.0 entirely. The stated intent currently is "The provenance includes metadata to aid debugging and investigations." which seems orthogonal to SLSA's main goal of enabling trust decisions. And, any effort to define requirements around including additional metadata would likely end up replacing this general "Includes Metadata" requirement in any case.

shaunmlowry:
+1 on resolving for 1.0.  If they're optional at a given level, we should just push them up a level and make them mandatory, even if that means having a level 5 :)

marcelamelara:
@shaunmlowry Good point. For optional requirements that apply to a specific level, I think it makes sense to bump them up to the next level. As an aside, I can think of other build requirements in addition to "Reproducible" (e.g., run your builder on attested hardware) that would fit into a L5.

The "Includes Metadata" common requirement, specifically, is currently optional in all L1-4, so if we want to pin it to a specific level for 1.0, the meaning of this requirement needs to be made more precise.

MarkLodato:
Let's decide this for v1.0. Suggestion for the two optional build requirements:

- Reproducible: Move this to the list of "possible future requirements" that will house the existing SLSA 4 requirements (since v1.0 will not define a Build L4.) The purpose of this "optional" requirement was more to guide readers that that is an ideal future state, and I think that will serve this purpose.
- Includes Metadata: Remove it from the requirements and instead:
  - In the provenance spec, have more clear guidance on how to progressively implement, which fields are more important and map to what level, etc. (Related: #425). Here we can recommend but not require this extra metadata.
  - In the Build requirements, suggest using SLSA provenance unless there is a reason to do so otherwise, and recommend using the same fields.

joshuagl:
[...]

PR #700, opened 2023-03-17
Combine ephemeral and isolated to isolated

### LINKED ISSUE #291: Isolated builds and build caches
Part of the [Isolated](https://slsa.dev/spec/v0.1/requirements#isolated) build requirement is that:

> Build caches, if used, MUST be purely content-addressable to prevent tampering.

At SLSA level 4, where hermetic is also a requirement, this would be scoped to only include caches which are provided by the build service (as any external dependencies must be pre-fetched and networking disabled).

However, this requirement can be interpreted in such a way that it makes SLSA level 3 extremely difficult to obtain for open source projects on GitHub (ref #185) because many "caches" of artifacts (i.e. dependencies in ecosystem package managers) are file addressed.

I think this item of the isolated requirement is focused solely on outputs of the build process which are cached between build steps (i.e. [actions/cache](https://github.com/actions/cache)), but wanted to seek clarity from others. Any thoughts/opinions?

### LINKED ISSUE #321: How strong of a guarantee is required for Isolated and Ephemeral Environment at SLSA 3?

arewm:
In order to help reduce confusion around ephemeral and isolated properties, these have been merged into a single isolated property. Additional clarity is added to the isolated build requirement, relating it to the previous hermetic requirement.

Relates to #657

Fixes #291 by being more specific around build caches.

Fixes #321 and addresses comments about self-hosted GitHub runners

Some content taken from https://github.com/slsa-framework/slsa/pull/685#discussion_r1140522178 by @MarkLodato.

### DIFF
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -60,7 +60,7 @@ Responsibility to implement SLSA is spread across the following parties.
   <td><a href="#build-service">Build service</a>
   <td> <td>✓<td>✓
 <tr>
-  <td><a href="#ephemeral-isolated">Ephemeral and isolated</a>
+  <td><a href="#isolated">Isolated</a>
   <td> <td> <td>✓
 </table>
 
@@ -262,23 +262,48 @@ workstation.
 Examples: GitHub Actions, Google Cloud Build, Travis CI.
 
 <td> <td>✓<td>✓
-<tr id="ephemeral-isolated">
-<td>Ephemeral and isolated
+<tr id="isolated">
+<td>Isolated
 <td>
 
-The build service ensured that the build steps ran in an ephemeral and isolated
-environment provisioned solely for this build, free of influence from other
-build instances, whether prior or concurrent.
+The build service ensured that the build steps ran in an isolated environment,
+free of unintended external influence. In other words, any external influence on
+the build was specifically requested by the build itself. This MUST hold true
+even between builds within the same tenant project.
 
--   It MUST NOT be possible for a build to access any secrets of the build service, such as the provenance signing key.
--   It MUST NOT be possible for two builds that overlap in time to influence one another.
--   It MUST NOT be possible for one build to persist or influence the build environment of a subsequent build.
--   Build caches, if used, MUST be purely content-addressable to prevent tampering.
--   The build SHOULD NOT call out to remote execution unless it's considered part of the "builder" within the trust boundary.
--   The build SHOULD NOT open services that allow for remote influence.
+The build system MUST guarantee the following:
 
-Note: This requirement was split into "Isolated" and "Ephemeral Environment"
+-   It MUST NOT be possible for a build to access any secrets of the build
+    service, such as the provenance signing key, because doing so would
+    compromise the authenticity of the provenance.
+-   It MUST NOT be possible for two builds that overlap in time to influence one
+    another, such as by altering the memory of a different build process running
+    on the same machine.
+-   It MUST NOT be possible for one build to persist or influence the build
+    environment of a subsequent build. In other words, an ephemeral build
+    environment MUST be provisioned for each build.
+-   It MUST NOT be possible for one build to inject false entries into a build
+    cache used by another build, also known as "cache poisoning". In other
+    words, the output of the build MUST be identical whether or not the cache is
+    used.
+-   The build system MUST NOT open services that allow for remote influence
+    unless all such interactions are captured as `externalParameters` in the
+    provenance.
+
+There are no sub-requirements on the build itself. Build L3 is limited to
+ensuring that a well-intentioned build runs securely. It does not require that
+build systems prevent a producer from performing a risky or insecure build. In
+particular, the "Isolated" requirement does not prohibit a build from calling
+out to a remote execution service or a "self-hosted runner" that is outside the
+trust boundary of the build platform.
+
+NOTE: This requirement was split into "Isolated" and "Ephemeral Environment"
 in the initial [draft version (v0.1)](../v0.1/requirements.md).
+
+NOTE: This requirement is not to be confused with "Hermetic", which roughly
+means that the build ran with no network access. Such a requirement requires
+substantial changes to both the build system and each individual build, and is
+considered in the [future directions](future-directions.md).
 
 <td> <td> <td>✓
 </table>
--- a/docs/spec/v1.0/threats.md
+++ b/docs/spec/v1.0/threats.md
@@ -386,33 +386,25 @@ provenance.
 
 These threats are directly addressed by the SLSA Build track.
 
-<details><summary>Compromise build environment of subsequent build <span>(Build L3)</span></summary>
-
-*Threat:* Perform a "bad" build that persists a change in the build environment,
-then run a subsequent "good" build using that environment.
-
-*Mitigation:* Builder ensures that each build environment is [ephemeral], with
-no way to persist changes between subsequent builds.
-
-*Example:* Build service uses the same machine for subsequent builds. Adversary
-first runs a build that replaces the `make` binary with a malicious version,
-then runs a subsequent build that otherwise would pass expectations. Solution:
-Builder changes architecture to start each build with a clean machine image.
-
-</details>
-<details><summary>Compromise parallel build <span>(Build L3)</span></summary>
+<details><summary>Compromise other build <span>(Build L3)</span></summary>
 
 *Threat:* Perform a "bad" build that alters the behavior of another "good" build
-running in parallel.
+running in parallel or subsequent environments.
 
 *Mitigation:* Builds are [isolated] from one another, with no way for one to
-affect the other.
+affect the other or persist changes.
 
-*Example:* Build service runs all builds for project MyPackage on the same
-machine as the same Linux user. Adversary starts a "bad" build that listens for
-the "good" build and swaps out source files, then starts a "good" build that
-would otherwise pass expectations. Solution: Builder changes architecture to
-isolate each build in a separate VM or similar.
+*Example 1:* Build service runs all builds for project MyPackage on
+the same machine as the same Linux user. Adversary starts a "bad" build that
+listens for the "good" build and swaps out source files, then starts a "good"
+build that would otherwise pass expectations. Solution: Builder changes
+architecture to isolate each build in a separate VM or similar.
+
+*Example 2:* Build service uses the same machine for subsequent
+builds. Adversary first runs a build that replaces the `make` binary with a
+malicious version, then runs a subsequent build that otherwise would pass
+expectations. Solution: Builder changes architecture to start each build with a
+clean machine image.
 
 </details>
 <details><summary>Steal cryptographic secrets <span>(Build L3)</span></summary>
@@ -663,9 +655,8 @@ accept cryptographic hashes with strong collision resistance.
 <!-- Links -->
 
 [authentic]: requirements.md#provenance-authentic
-[ephemeral]: requirements.md#ephemeral-isolated
 [exists]: requirements.md#provenance-exists
-[isolated]: requirements.md#ephemeral-isolated
+[isolated]: requirements.md#isolated
 [non-forgeable]: requirements.md#provenance-non-forgeable
 [service]: requirements.md#build-service
 [supply-chain threats]: threats-overview


docs/spec/v1.0/threats.md
```diff
@@ -386,33 +386,25 @@ provenance.
...
 
 *Threat:* Perform a "bad" build that alters the behavior of another "good" build
-running in parallel.
+running in parallel or subsequent environments.
 
 *Mitigation:* Builds are [isolated] from one another, with no way for one to
-affect the other.
+affect the other or persist changes.
+
+*Example 1 (Build L3):* Build service runs all builds for project MyPackage on
```
MarkLodato, line 397: nit: do we need "(Build L3)"? That was only added to "Set values of the provenance" below because there are two different levels. None of the other examples have this.

docs/spec/v1.0/requirements.md
```diff
@@ -262,23 +266,54 @@ workstation.
...
+    environment MUST be provisioned for each build.
+-   It MUST NOT be possible for one build to inject false entries into a build
+    cache used by another build, also known as "cache poisoning". In other
+    words, the output of the build MUST be identical whether or not the cache is
+    used.
+-   The build system MUST NOT open services that allow for remote influence
+    because doing so would introduce additional attack vectors to compromise
+    the build system.
+-   The build system SHOULD NOT call out to remote execution outside of the
+    trust boundary. In other words, if any remote execution, like `ntp`, is
```
MarkLodato, line 297: What is ntp? I always thought it was the Network Time Protocol, so I don't see how that's remote execution. Would something like [Bazel's Remote Build Execution](https://bazel.build/remote/rbe) be a better example?

docs/spec/v1.0/requirements.md
```diff
@@ -262,23 +266,54 @@ workstation.
...
+    weighed and considered.
+
+There are no sub-requirements on the build itself, but the following are
+RECOMMENDED in order to honor the spirit of "Isolated":
+
+-   The build SHOULD NOT call out to remote execution unless it's considered
+    part of the build system within the trust boundary because doing so could
+    compromise the authenticity of the provenance.
+-   The build SHOULD NOT open services that allow for remote influence because
+    doing so could compromise the authenticity of the provenance.
```
MarkLodato, line 308: nit: I think the "doing so could compromise the authenticity of the provenance" doesn't add much here. I had suggested it for the first bullet to explain why access to a key violates this principle. Here I think it's more clear?

```suggestion
-   The build SHOULD NOT call out to remote execution unless it's considered
    part of the build system within the trust boundary.
-   The build SHOULD NOT open services that allow for remote influence.
```

docs/spec/v1.0/requirements.md
```diff
@@ -249,6 +249,10 @@ The SLSA Build level describes the minimum bar for isolation strength. For more
 information on assessing a build system's isolation strength, see
 [Verifying build systems](verifying-systems.md).
 
+Note: The isolation strength only applies to the build system and not to the
+build itself. A similar hermetic property was associated to the build in the
+initial [draft version (v0.1)](../v0.1/requirements.md#hermetic).
+
```
MarkLodato, line 255: Is this still needed? I would imagine that the updated text addresses this possible source of confusion. Also, I think this statement only applies to the "Isolated" row and not the whole section?

MarkLodato (changes requested):
Thanks!

docs/spec/v1.0/requirements.md
arewm: Yes, a better example can be used. I used `ntp` because I had it in the original issue and I was commenting about the lack of clarity around what is remote execution as we have no _guarantee_ that ntp servers are not doing some kind of execution.

docs/spec/v1.0/requirements.md
arewm: I liked the pattern that you introduced with your original section -- providing explanation behind the suggestions. These _recommendations_ are here to ensure that there is proper provenance generation. Going back to my original questions, I think that this is something that I didn't understand before.

> * Would this prevent the vendoring of binary content as some process of the build was performed outside the trust boundary?

No, this would be fine since the dependency would be present in the provenance. We are not trying to make suggestions about the _content_ of the built artifact or the recursive provenance.

> * Would this prevent curling pre-built artifacts to be used in a resulting artifact?

This would be an anti-pattern assuming that the artifacts are not included in the provenance.

> * Would leveraging something like NTP within the build be remote execution?

It doesn't affect provenance.

> * Should this prevent pulling of untrusted container artifacts as well since this can also be a "building" process (i.e. https://github.com/imjasonh/kontain.me/tree/main/cmd/buildpack)?

That would be fine from a provenance perspective since we are not trying to create recursive provenance, but this might not fly for the build hardening aspect.

> * Since any remote access can potentially result in remote execution, should this be interpreted as a requirement for hermetic builds in complete network isolation?

This has been clarified to some extent at least.

docs/spec/v1.0/requirements.md
arewm: It isn't needed. The hermetic link is also made in the `Isolated` row as well.

docs/spec/v1.0/requirements.md
MarkLodato: I think this section is trying to say that the "build system MUST" rules above are only about what is enabled by default. When a tenant properly configures the build system in the normal way, they MUST be secure. However, it is **not** a requirement to **prevent** tenants from making themselves insecure - that is a much more difficult task and might be a future level. (Relevant issue: #321)

The list here contains the most common situations that we would consider "less secure". We list them here for two reasons:

1. To call out things the build system need not prevent to give more clarify on where the line is.
2. To provide guidance to tenants on how to secure their own builds, even if not strictly required.

In particular, I think this needs to definitely answer the question of whether [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) allowed at Build L3 and if they need a separate `builder.id`. I believe where we landed on #321 was to say that self-hosted runners are OK at Build L3.

I'm kind of doubting the wisdom of reason 2 above (guidance to tenants). We got feedback in #423 that it's confusing to have "optional" requirements. SHOULD here is another form of optional requirement. Might be less confusing to just say that the build system need not prevent it.

docs/spec/v1.0/requirements.md
```diff
@@ -262,23 +262,55 @@ workstation.
...
+-   It MUST NOT be possible for one build to persist or influence the build
+    environment of a subsequent build. In other words, an ephemeral build
+    environment MUST be provisioned for each build.
+-   It MUST NOT be possible for one build to inject false entries into a build
+    cache used by another build, also known as "cache poisoning". In other
+    words, the output of the build MUST be identical whether or not the cache is
+    used.
+-   The build system MUST NOT open services that allow for remote influence
+    because doing so would introduce additional attack vectors to compromise
+    the build system.
```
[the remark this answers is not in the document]
arewm: I see a difference between these two bullets, at least in my intention of wording it like I did

### Second bullet
> It MUST NOT be possible for two builds that overlap in time to influence one
    another, such as by altering the memory of a different build process running
    on the same machine.

* This is about the builder enforcing isolation between each running build
* Any two running builds must be isolated from each other, one producer-triggered event cannot affect another producer-triggered event
* The build system should ensure that builds running cannot influence/affect one another

### Fifth bullet
> The build system MUST NOT open services that allow for remote influence
    because doing so would introduce additional attack vectors to compromise
    the build system.

* This is about the builder ensuring that builds run as _configured_
* Once the builder is configured, it cannot be changed until another deployment occurs (i.e. updating Tekton Pipeline definitions or GitHub actions updates)
* An example of this would be the ability to [debug Travis CI runs](https://docs.travis-ci.com/user/running-build-in-debug-mode/) -- if debug is enabled and remote influence is potentially introduced, then the resulting artifact would not be Build L3
* The support at this line might be misleading as it isn't necessarily around compromising the build system, but more about compromising the ability for the build system to know the steps called to build the artifact.

docs/spec/v1.0/requirements.md
```diff
@@ -262,23 +262,55 @@ workstation.
...
+-   The build system SHOULD NOT call out to remote execution outside of the
+    trust boundary. In other words, if any remote execution, like Bazel
+    [Remote Execution](https://bazel.build/remote/rbe), is used by the build
+    system, the full implications of such decision SHOULD be weighed and
+    considered.
```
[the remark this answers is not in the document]
arewm: I created this bullet initially because of the confusion around mentioning `build`s within the context of `build system`. Thinking about it more, this really just comes down to ensuring that appropriate build system threat modeling has been done to ensure that the build process is not affected by any process outside of the trust boundary. 

It makes sense that any build actions within the build system MUST be within the trust boundary. Would that mean that it MUST be considered part of the builder identified by the `builder.id`? If this is the place, what would be the best place to describe this detail -- within `build model` in the Terminology?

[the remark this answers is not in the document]
MarkLodato: Maybe on the provenance page? I filed #706 to track.

docs/spec/v1.0/requirements.md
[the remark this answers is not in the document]
MarkLodato: Oh, I see now. Thanks for explaining. I don't think the current wording quite gets that through because it's still a bit abstract. Maybe something like the following would be more clear?

> The build system MUST NOT open services on the build work that allow for external parties to run commands or otherwise alter the behavior of the running build process, such as [Debug Mode in Travis CI](https://docs.travis-ci.com/user/running-build-in-debug-mode/). Doing so would constitute an external parameter that is not captured in the provenance.

docs/spec/v1.0/requirements.md
MarkLodato: I think I follow. Could you propose some updated wording?

docs/spec/v1.0/requirements.md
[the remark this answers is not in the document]
arewm: Ah, interesting wording. I can see how these would be considered external parameters. Based on the completeness wording for provenance L3

> -   External parameters MUST be fully enumerated.

we can actually change this wording since parameterless wasn't considered previously until L4, this _may_ still be fine.

> Any open services on the build system which allow for external parties to alter the behavior of the running build process, such as [Debug Mode in Travis CI](https://docs.travis-ci.com/user/running-build-in-debug-mode/), are considered external parameters and MUST be fully enumerated if allowed.

It isn't until the external parameters are not enumerable that the relevant services MUST NOT be opened.

docs/spec/v1.0/requirements.md
arewm: I won't push content since it would remove the content that is in here right now. I am also not going to further address the build system isolation (i.e. L4) here as that just needs to go in a new issue. The best way I see to make it explicit (using a pattern I don't see elsewhere) would be to:

1. Add a row to the `Producer` table around isolation without any check-marks
2. Add a section under Producer for isolation similar to this:

>   There are no requirements set forth to producers related to isolation in order to achieve build L1-L3. If the build system allows package builds to occur with access to the network or the internet, producers SHOULD NOT intentionally subvert the build system's best effort for provenance completion.

3. Removing the section on the build system starting with:

> There are no sub-requirements on the build itself, but the following are
RECOMMENDED in order to honor the spirit of "Isolated":

docs/spec/v1.0/requirements.md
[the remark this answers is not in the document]
MarkLodato: > > Any open services on the build system which allow for external parties to alter the behavior of the running build process, such as [Debug Mode in Travis CI](https://docs.travis-ci.com/user/running-build-in-debug-mode/), are considered external parameters and MUST be fully enumerated if allowed.
> 
> It isn't until the external parameters are not enumerable that the relevant services MUST NOT be opened.

While this is technically correct, I suspect in practice that this might be more confusing than just disallowing such services. I imagine that almost all cases allow arbitrary commands or behavior to be injected, so they likely are never enumerable. Can you think of an example where such a thing would be enumerated as an external parameter?

docs/spec/v1.0/requirements.md
MarkLodato: Hmm. I'm hesitant to add a row with no check marks (might be confusing). It would be good to have other feedback on possible wording, if anyone else is reading this.

docs/spec/v1.0/requirements.md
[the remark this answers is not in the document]
MarkLodato: Actual example. 😁  Maybe we can just say it's generally disallowed, with a footnote or comment explaining that if you record all of the inputs in `externalParameters` then it's OK?

Regarding Travis, I'd say a `builder.id` should indicate that Travis was used in a configuration that prohibits Debug Mode (at least before provenance is fully generated). What do you think?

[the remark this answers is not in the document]
arewm: As long as a service like Travis who has this functionality has the ability to clearly indicate the L2/L3 differentiation that would happen, it comes down to an implementation detail. It doesn't even matter that it is used in a configuration that prohibits the debug mode -- even if it is enabled, what matters is whether an artifact was generated with the mode enabled.

IMO, adding a footnote will just introduce more confusion than addressing it _well_ in a bullet. Even just changing `open services on the build system` to `build system services` simplifies comprehension.

MarkLodato:
FYI I added "Fixes #291 by being more specific around build caches." to the description.

docs/spec/v1.0/requirements.md
abacchi: I think its fair to show that the build system holds the burden for the L3 requirements- and eventually at L4 the producer will need to make some effort to meet the future requirement. 
I hesitate to show a blank row as well because it sort of makes the 1.0 release look unfinished - but I see the benefit of having it there as a placeholder for the future.
@arewm I like the 2 & 3 suggestions

arewm: I can also do 2&3 without 1, I just thought 1 would be needed for completeness/adhering to the pattern.

MarkLodato: I'm not sure #2 is needed - it seems weird to talk about Isolation in the Producer section.

Why not just replace the "There are no sub-requirements..." with a more direct explanation that there are no requirements to block the build process from connecting to external resources, and that things like self-hosted runners are OK?

arewm: I only updated the content in the isolation table. I also modified the PR description to add `Fixes #321`.

docs/spec/v1.0/requirements.md
```diff
@@ -262,23 +262,47 @@ workstation.
...
+    words, the output of the build MUST be identical whether or not the cache is
+    used.
+-   The build system MUST NOT open services that allow for remote influence
+    because doing so would introduce additional attack vectors to compromise
+    the build system.
+
+There are no sub-requirements on the build itself. As provenance generation up
+to Level 3 is *best effort*, any reasonable functionality for the builds are
+acceptable considering non-malicious producer intentions. This includes builds
+running with access to the internet as well as leveraging [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners)
```
MarkLodato, line 296: I like where this is going, but provenance generation is not "best effort" at L3 - only the resolvedDependencies bit is. What do you think about the following alternative? I tried to more directly state what we discussed in the thread. (Also avoid linking to a specific vendor, though I'm open to re-adding the link.)

```suggestion
There are no sub-requirements on the build itself. Build L3 is limited to
ensuring that the a well-intentioned build runs securely. It does not require
preventing a producer from performing a risky or insecure build. In particular,
the "Isolated" requirement does not prohibit a build from calling out to a
remote execution service or a "self-hosted runner" that is outside the trust
boundary of the build platform.
```

docs/spec/v1.0/requirements.md
```diff
@@ -262,23 +262,47 @@ workstation.
...
+as part of a build system.
```
MarkLodato, line 297: I like where this is going, but provenance generation is not "best effort" at L3 - only the resolvedDependencies bit is. What do you think about the following alternative? I tried to more directly state what we discussed in the thread. (Also avoid linking to a specific vendor, though I'm open to re-adding the link.)

```suggestion
There are no sub-requirements on the build itself. Build L3 is limited to
ensuring that the a well-intentioned build runs securely. It does not require
preventing a producer from performing a risky or insecure build. In particular,
the "Isolated" requirement does not prohibit a build from calling out to a
remote execution service or a "self-hosted runner" that is outside the trust
boundary of the build platform.
```

arewm: I meant to indicate that provenance completeness is best effort, not that provenance generation is best effort as this is what is mentioned earlier in the requirements (https://github.com/slsa-framework/slsa/pull/700/files#diff-3a161534a67508acd917c97c9e5d57ed648c50c3a7da23acd58aede57fafbdb2R230-R231).

I like the wording in general. Do we want to tie back to provenance completeness as a best effort?

MarkLodato: Sure, maybe a few words either here or in the "hermetic" note?

docs/spec/v1.0/requirements.md
```diff
@@ -262,23 +262,49 @@ workstation.
...
+    cache used by another build, also known as "cache poisoning". In other
+    words, the output of the build MUST be identical whether or not the cache is
+    used.
+-   The build system MUST NOT open services that allow for remote influence
+    because doing so would introduce additional attack vectors to compromise
+    the build system.
+
+There are no sub-requirements on the build itself. Build L3 is limited to
+ensuring that a well-intentioned build runs securely and that provenence
+generated completeness is best effort. It does not require that build systems
```
MarkLodato, line 295: I can't find the original thread, but "provenence generated completeness is best effort" doesn't make sense to me. What is the intention?

arewm: I can remove this line if the PR is ready to be merged without it and open another issue to discuss the confusion around this phrase in the provenance requirements and how I end up interpreting that across the board.

MarkLodato: Yeah, let's just remove that line for now to get this merged, and then have a separate PR to discuss that particular piece. I think what you have so far (minus this) is a strict improvement so let's lock that in!

arewm: I opened #717 to continue the conversation.

docs/spec/v1.0/requirements.md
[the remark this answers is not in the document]
arewm: @MarkLodato, this is the only remaining thread. We never settled on whether or how to include external parameters in isolation. To some extent, this comes down to implementation details. One possible change could be from 

> The build system MUST NOT open services that allow for remote influence
    because doing so would introduce additional attack vectors to compromise
    the build system.

to 

> The build system MUST NOT open services that allow for remote influence
    unless all such interactions are captured as `externalParameters` in the provenance.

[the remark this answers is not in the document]
MarkLodato: I'm OK with that change!

joshuagl (approved):
Thank you for this change. I really appreciate all of the detailed discussion in this PR review, I'm sorry I missed it! 😁

docs/spec/v1.0/requirements.md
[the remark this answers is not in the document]
arewm: Change pushed, thanks.

MarkLodato (approved):
Thanks! I know this back-and-forth wasn't easy but I appreciate your hard work in resolving these issues!

arewm:
No problem. I enjoyed the back and forth I just wish that there was a better way to converse about it than in PR comment threads that get easily outdated. 

This PR should be ready to merge.

PR #710, opened 2023-03-20
Provenance v1: builder.id should explain itself

### LINKED ISSUE #594: SLSA v1: how to represent what capabilities a builder has w.r.t. provenance values
_Originally posted by @laurentsimon in https://github.com/slsa-framework/slsa/pull/582#discussion_r1087222031_

> A bit of a tangent question: is it correct that the builderType will indicate whether the builder is capable (or not) of reporting these values? (similar question for `env` and `vars`). 
>
> Example: the default npm CLI cannot capture this in an unforgeable way (since it's under the control of the project maintainer), unless the registry is able to verify it by querying the runner at package upload time.
>
> Hence, then do we need to define different builder types: one for the "runner" builder and one for the "reusable workflow" builder... and maybe a hybrid one if only certain fields can be reported?
>
> /cc @asraa @feelepxyz

I suspect we want to say that the `builder.id` indicates how complete and accurate the provenance is, independent of `buildType`, but I'm forking this conversation to get #582 submitted.

MarkLodato:
Add a recommendation for the `builder.id` URI to resolve to documentation that explains the scope, level, and accuracy of the builder.

Fixes #594.

### DIFF
--- a/docs/provenance/v1.md
+++ b/docs/provenance/v1.md
@@ -339,6 +339,12 @@ security attributes or SLSA Build levels, each mode MUST have a different
 `builder.id` and SHOULD have a different signer identity. This is to minimize
 the risk that a less secure mode compromises a more secure one.
 
+The `builder.id` URI SHOULD resolve to documentation explaining:
+
+-   The scope of what this ID represents.
+-   The claimed SLSA Build level.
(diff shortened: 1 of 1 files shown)


TomHennen (approved):

lehors (approved):
LGTM

Issue #367, opened 2022-04-15
Define "trusted control plane"


bobcatfish:
Another phrase which occurs a few times [in the SLSA requirements](https://slsa.dev/spec/v0.1/requirements) but is not clearly defined (or I haven't found the definition!) is **trusted control plane**.

It would be very useful to have this clearly defined in [the SLSA terminology](https://slsa.dev/spec/v0.1/terminology), especially how it could apply to different systems, for example this phrase would likely mean something different for GitHub Actions which is hosted and has complete control over the control plane vs. for something like Tekton where the trusted control plane depends on how Tekton is configured and also may include components such as [Tasks](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md)

(More context available in [SLSA + Tekton: Case Study](https://docs.google.com/document/d/12yYP0M_zz5Tc9ftGnLwhDZ_tdZBq4mEod70ngWtHmLs/edit#) - particularly [the section around what trusted control plane could mean for Tekton](https://docs.google.com/document/d/12yYP0M_zz5Tc9ftGnLwhDZ_tdZBq4mEod70ngWtHmLs/edit#bookmark=id.jiiws37hhbri) - the doc is visible to anyone in mailing list slsa-discussion@googlegroups.com ).

marcelamelara:
+1 to resolving this for v1.0.

kpk47:
This may have been addressed by https://github.com/slsa-framework/slsa/pull/568.

MarkLodato:
This can be further clarified because requirements.md says "trusted control plane" without linking to a definition. Simple solution is to link to the corresponding page in verifying build systems.

PR #712, opened 2023-03-21
Provenance v1: clarify "builder" and "platform"

### LINKED ISSUE #593: Provenance v1: Add `builder` to the SLSA Provenance model
In the [provenance v1 spec](https://slsa.dev/provenance/v1), it would be nice to clarify how the diagram and particularly the "Build Process" relate to the `builder`. The description talks about `builder` but is not shown in the diagram. *(Originally posted by @rbehjati in https://github.com/slsa-framework/slsa/issues/582#issuecomment-1406680513.)*

I'm not sure how best to do this so I'm creating this issue to track ideas.

First, what exactly is the builder in the diagram? Is it a collection of multiple boxes (e.g. platform, build type, build cache, build process)? Is it just the platform? Does it not really map to the diagram?

Once we decide that, we can figure out how to represent it. For example, I tried adding dashed lines around a collection of boxes but that added too much clutter.

MarkLodato:
Previously we used "builder" and "build platform" interchangably, but this was never stated explicitly anywhere, leading to confusion for several readers. Also we never defined "control plane", and the diagram used "platform" inconsistently.

Changes:

-   Update diagram to add "Build Platform (builder)" to the diagram, showing a box over the pieces that are trusted, and rename the old "Platform" to "Control Plane". This is now consistent with the model in Verifying Build Systems.
-   In the text, prefer "build platform" over "builder" unless it is very clear that the latter is short for the former. Now the only places that "builder" is used are:
    -   As a field name (`builder`)
    -   In the "Builder" section of the doc, where it is explained that
        the two terms mean the same thing.
    -   In the Verification section, where "builder" is unambiguous.
-   Explain that "build platform" includes not just software but also hardware, people, and organizations, and link from the Provenance page to the terminology's Build model.

Fixes #593.
Partially addresses #367 by showing the control plane in the diagram.

/cc @asraa @bobcatfish @rbehjati @olivekl

Preview: https://deploy-preview-712--slsa.netlify.app/provenance/v1


docs/provenance/v1.md
TomHennen, line 1: This seems fine to me but maybe one thing that is unaddressed is the when code can be re-used and run in multiple places (e.g. Tekton).  Would each instance be considered a separate build platform and have its own builder ID?  Is there a term for this type of software?

MarkLodato: Good point, thanks for reminding me. I just added d15b272 to hopefully address. PTAL

TomHennen: Great, thanks!

TomHennen (approved):

kpk47 (approved):

docs/provenance/v1.md
```diff
@@ -354,8 +354,13 @@ or security guarantees.
 
 </table>
 
-The builder represents the transitive closure of all the entities that are, by
-necessity, [trusted] to faithfully run the build and record the provenance.
+The build platform, or <dfn>builder</dfn> for short, represents the transitive
+closure of all the entities that are, by necessity, [trusted] to faithfully run
```
rbehjati, line 358: Why is [trusted] in brackets?

rbehjati (approved):
Thanks :)

docs/provenance/v1.md
rbehjati: Oh! Thanks. I did not know it works like that.

MarkLodato:
I'm going to merge this now just to avoid long-standing PRs that lead to merge conflicts with other PRs, but I'm certainly open to feedback on this PR, including "I don't like this". :-)

PR #719, opened 2023-03-22
Make the draft banner more useful.

MarkLodato:
Most readers probably don't want to be taken to the latest approved version, but they might want to select a release candidate (or even know that a release candidate exists). So refer them to the version selector rather than adding a link.

Also display what version it is a draft of, to make it more clear to readers.

I'm thinking that when we send out a call to review v1.0, the current banner would be confusing to people.


lehors (approved):
I agree. LGTM.

PR #720, opened 2023-03-23
Version /docs/images/

kpk47:


MarkLodato:
This breaks several images: https://deploy-preview-720--slsa.netlify.app/

Do we need this? Could we just version images as we go?

kpk47:
Versioning as we go SGTM, or maybe we start snapshotting with v1.0-rc2. I'll close this PR.

PR #722, opened 2023-03-23
Describe SLSA's versioning conventions.

### LINKED ISSUE #570: Add section about how to refer to SLSA versions
In the [SLSA 1.0 proposal](https://github.com/slsa-framework/slsa-proposals/tree/main/0003), we decided the following (bullets added for clarity):

> - The specification is versioned but we agree to never change the gist of each level ("hybrid" below). Effectively, the "SLSA 1/2/3/4" namespace is taken. It would cause too much confusion to change the meanings significantly, and requiring everyone to always specify a revision number is impractical. We can still make minor revisions to the levels, so long as we do not believe it would cause too much confusion.
> - When precision is not critical, unversioned references are OK, e.g. "SLSA 3."
> - When precision is required, references MUST include a revision number, preferably using the syntax "SLSA 3 (v1.0)," and ideally backed by [evidence of security claims].

We need to incorporate that into the v1.0 spec somewhere.

kpk47:
This change clarifies when/why SLSA will increment its version number and how users should refer to SLSA levels in context where the version number matters.

Fixes #570

### DIFF
--- a/docs/spec-stages.md
+++ b/docs/spec-stages.md
@@ -1,8 +1,10 @@
 ---
-title: Specification Stages
+title: Specification Stages and Versioning
 description: SLSA specifications go through various stages of development from which you should have different expectations. This document defines the different stages the SLSA project uses and their meaning for readers and contributors.
 layout: specifications
 ---
+
+## Specification Stages
 
 Specifications go through various stages of development from which you
 should have different expectations. This document defines the different
@@ -13,7 +15,7 @@ Every specification page should prominently display a *Status* section
 stating which stage the specification is in with a link to its
 definition.
 
-## Draft
+### Draft
 
 This is the first stage of development a specification goes
 through. At this point, not much should be expected of it. The
@@ -31,7 +33,7 @@ See the
 for other considerations related to a
 [Draft Specification](https://github.com/slsa-framework/governance/blob/main/1._Community_Specification_License-v1.md).
 
-## Candidate
+### Candidate
 
 At this stage the document is considered to be feature complete and is
 published as a way to invite final reviews. Editorial changes may
@@ -44,7 +46,7 @@ See the
 for other considerations related to a
 [Candidate for Approved Specification](https://github.com/slsa-framework/governance/blob/main/1._Community_Specification_License-v1.md).
 
-## Approved
+### Approved
 
 At this stage the document is considered stable. No changes that would
 constitute a significant departure from the existing specification are
@@ -56,10 +58,28 @@ See the
 for other considerations related to an
 [Approved Specification](https://github.com/slsa-framework/governance/blob/main/1._Community_Specification_License-v1.md).
 
-## Retired
+### Retired
 
 This stage indicates that the specification is no longer maintained.
 It has been either rendered obsolete by a newer version or
 abandoned for some reason. The status of the document may provide
 additional information and point to the new document to use intead if
 there is one.
+
+## Versioning
+
+SLSA uses a variant of [Semantic Versioning](https://www.semvar.org) to assign
+version numbers to the Core Specification and Attestation Formats, collectivly
+known as the SLSA Specification.
+
+Given a version MAJOR.MINOR.PATCH, we will increment
+
+-   MAJOR version when making backwards incompatible changes to the Core
+  Specification.
+-   MINOR version when making backwards compatible changes to the Core
+  Specification.
+-   PATCH version when making changes to the Attestation Formats.
+
+We expect to release a MINOR version at least every three (3) months, and we
+expect to release a MAJOR version at most once per calendar year. We will
+release PTACH versions as often as necessary.
--- a/docs/spec-stages.md
+++ b/docs/spec-stages.md
@@ -68,18 +68,24 @@ there is one.
 
 ## Versioning
 
-SLSA uses a variant of [Semantic Versioning](https://www.semvar.org) to assign
+SLSA needs revision from time to time, so we version the specification to
+facilitate conformance efforts and prevent confusion. We assign
 version numbers to the Core Specification and Attestation Formats, collectivly
 known as the SLSA Specification.
 
-Given a version MAJOR.MINOR.PATCH, we will increment
-
+Given a version MAJOR.MINOR, we will increment
 -   MAJOR version when making backwards incompatible changes to the Core
-  Specification.
+  Specification. 
 -   MINOR version when making backwards compatible changes to the Core
-  Specification.
--   PATCH version when making changes to the Attestation Formats.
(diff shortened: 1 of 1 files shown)


docs/spec-stages.md
```diff
@@ -56,10 +58,28 @@ See the
...
 additional information and point to the new document to use intead if
 there is one.
+
+## Versioning
+
+SLSA uses a variant of [Semantic Versioning](https://www.semvar.org) to assign
+version numbers to the Core Specification and Attestation Formats, collectivly
+known as the SLSA Specification.
+
+Given a version MAJOR.MINOR.PATCH, we will increment
```
arewm, line 75: Version numbers have only been `MAJOR.MINOR` so far. Should we either note that there is an assumed `.0` in the absence of it or should we change the current version to `v1.0.0`?

docs/spec-stages.md
```diff
@@ -56,10 +58,28 @@ See the
...
+
+-   MAJOR version when making backwards incompatible changes to the Core
+  Specification.
+-   MINOR version when making backwards compatible changes to the Core
+  Specification.
+-   PATCH version when making changes to the Attestation Formats.
+
+We expect to release a MINOR version at least every three (3) months, and we
+expect to release a MAJOR version at most once per calendar year. We will
+release PTACH versions as often as necessary.
```
arewm, line 85: Is `expect` too strong of a word here? There is also some time-based component here. Short term, I would expect that the release cadence would be higher/more significant changes. Longer term, there might not be as much need for as many large changes.

docs/spec-stages.md
arewm, line 75: Version numbers have only been MAJOR.MINOR so far. Should we either note that there is an assumed .0 in the absence of it or should we change the current version to v1.0.0?

MarkLodato: Do we really need patches? Right now we don't have a good release process, so actually calling something a patch will be painful. Can we just have major and minor but say that we follow the semantic versioning otherwise? Is anyone confused by the lack of ".PATCH"?

docs/spec-stages.md
```diff
@@ -56,10 +58,28 @@ See the
...
+Given a version MAJOR.MINOR.PATCH, we will increment
+
+-   MAJOR version when making backwards incompatible changes to the Core
+  Specification.
+-   MINOR version when making backwards compatible changes to the Core
+  Specification.
+-   PATCH version when making changes to the Attestation Formats.
+
+We expect to release a MINOR version at least every three (3) months, and we
+expect to release a MAJOR version at most once per calendar year. We will
```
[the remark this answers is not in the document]
MarkLodato: Right. I think we would ideally _never_ have a MAJOR version, and I agree that we should limit it to at most once per year. I also think we shouldn't set any expectations on the cadence of MINOR versions for now.

docs/spec-stages.md
arewm: SemVer is confused: https://semver.org/#spec-item-2 :)

docs/spec-stages.md
```diff
@@ -56,10 +58,28 @@ See the
...
+version numbers to the Core Specification and Attestation Formats, collectivly
+known as the SLSA Specification.
+
+Given a version MAJOR.MINOR.PATCH, we will increment
+
+-   MAJOR version when making backwards incompatible changes to the Core
+  Specification.
+-   MINOR version when making backwards compatible changes to the Core
+  Specification.
+-   PATCH version when making changes to the Attestation Formats.
```
MarkLodato, line 81: Can we be more precise? I think the rules vary by attestation formats vs the slsa spec.

- Attestation formats: Either link to or copy https://github.com/in-toto/attestation/blob/main/spec/versioning.md#versioning-rules-1
- SLSA specification: What is a breaking change? We need to spell out criteria for determining the minimum type of version bump. I'm not sure precisely what this should be.
  - Clarifying text in a way that doesn't change meaning: PATCH (even if we don't report the PATCH level)
  - Adding new tracks or levels: MINOR
  - Adding more explanation to an existing requirement, in a way that doesn't fundamentally change the requirement but provides better guidance: PATCH?
  - Altering an existing requirement to be more or less stringent, but that doesn't fundamentally change the meaning: MINOR?
  - Adding a completely new requirement to an existing level: MAJOR

Furthermore, we promise to never change the high-level meaning of a SLSA level once it has been released. For example, "SLSA Build L3" SHOULD NOT mean something fundamentally different even in a new major version number. If we want to do that, we'll create a new name, e.g. "SLSA Compile L3". https://github.com/slsa-framework/slsa-proposals/tree/main/0003#versioning

docs/spec-stages.md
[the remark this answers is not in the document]
kpk47: I'm happy enough not to schedule MAJOR versions, but I think there are benefits to scheduling MINOR releases. It sets expectations for infrastructure providers, and gives us a basis for instituting some basic project management (e.g. bugs can be classified as "do immediately for a PATCH" vs "hold for the quarterly MINOR release"). It would also be useful for writing the roadmap to be able to say something like "<feature x> is scheduled for <MINOR version>", though I suppose that's not too different to scheduling for a month or quarter.

I am, however, open to the idea that it's simply too early for this sort of scheduling.

docs/spec-stages.md
```diff
@@ -56,10 +58,36 @@ See the
...
+
+Given a version MAJOR.MINOR, we will increment
+
+-   MAJOR version when making backwards incompatible changes to an
+    Attestation Format or adding new requirements to an existing level
+    in the Core Specification.
+-   MINOR version when adding new tracks or levels to the Core Specification,
+    modifying an existing level without fundamentally changing its meaning, or
+    adding new fields to an Attestation Format in a backwards-compatible way.
+    For more details on Attestation Format versioning, see the [intoto
```
MarkLodato, line 84:
```suggestion
    For more details on Attestation Format versioning, see the [in-toto
```

docs/spec-stages.md
```diff
@@ -56,10 +58,36 @@ See the
...
+    Versioning Rules](https://github.com/in-toto/attestation/blob/main/spec/versioning.md).
+
+Although we can revise the contents of level, we will never change a level's
+high-level meaning after publication (e.g. SLSA Build Level 2 will retain its
+general meaning between major versions). If you require precision when referring
+to SLSA levels, then include their version number using the syntax `SLSA
+[Track] [Level] ([Version]` (e.g. `SLSA Build L3 (v1.0)`). For more
```
MarkLodato, line 91:
```suggestion
[Track] [Level] ([Version])` (e.g. `SLSA Build L3 (v1.0)`). For more
```

MarkLodato (approved):

PR #723, opened 2023-03-24
Move version selector to header or sidebar

lehors:
This uses the space we have in the header and reduces the amount of stuff we have on the right side.
This seems to scale down well enough to accommodate for a mobile phone.
FYI, I tried to move the status info as well but it doesn't fit on mobile phones.


MarkLodato (changes requested):
## Problem 1: not visible on desktop

The header doesn't show up when at full width, in which case there is no version selector.

image

## Problem 2: everything is shrunken on mobile

On mobile, now the entire site is shrunken with white space on the right.

Before:

after

After:

after

lehors:
Ok, I wasn't aware that the header didn't show in full page (I *never* do that! :-)
Forget it then.

lehors:
By the way, I don't see that problem with the white on mobile...

MarkLodato:
I do agree that putting the version selector in the header is a better UI than the status quo, it's just that our current template makes it tricky to do. If we could get the header to always show, we should revive this PR.

> By the way, I don't see that problem with the white on mobile...

Oh, it must be my machine. I am using Chrome's mobile preview (ctrl-shift-i, then the little mobile phone icon on the upper left of that widget.) I had the problem on your deploy preview but not slsa.dev so I assumed it was a problem with this PR, but now that I test it on other sites, something weird is going on on my machine.

lehors:
Ok, maybe I can have another look...

lehors:
Ok, @MarkLodato I figured out how the topbar+sidebar stuff works and added the selector to the sidebar so that when the topbar is hidden it shows up there.
Let me know what you think.

MarkLodato (commented):
Can we always show the version selector, maybe using the same logic that the nav bar uses? Right now it's a bit jarring that as you navigate pages, it appears and disappears.

It would also be nice to style it a bit nicer, but I realize that's lower priority.

lehors:
> Can we always show the version selector, maybe using the same logic that the nav bar uses? Right now it's a bit jarring that as you navigate pages, it appears and disappears.

I can do something like that:

Screenshot 2023-03-25 at 10 00 41 AM

But it looks a bit weird on the home page, community page, blog, etc:

Screenshot 2023-03-25 at 10 04 32 AM

And if I remove it from those pages then we are back to having it coming on and off as you move between pages.

> It would also be nice to style it a bit nicer, but I realize that's lower priority.

I'm not fond of the style either. I reduced the font size a bit and I think it looks a bit better but I'm sure someone more graphic design oriented could do better!

lehors:
Ok, I found a way to keep the layout constant without clutter. Let me know what you think.

MarkLodato (approved):
Thanks. It would be nice for a follow-up to style it a bit nicer (e.g. nicer colors, full width, maybe say "Version:") but this is an improvement from having it be mostly undiscoverable on the right.

lehors:
I guess the color & style is a matter of taste. I quite like it that way but I understand that's entirely personal...
I did put the "Version " piece back though. I wanted it to have outside the selector but there is no room for it (on mobile) so it's better this way.

MarkLodato:
Let's merge?

> I guess the color & style is a matter of taste. I quite like it that way but I understand that's entirely personal...

From your [screenshot](https://user-images.githubusercontent.com/6464618/227707826-534c3be5-788d-4993-b5bd-83b0011fbad2.png) it looks like it appears green in your browser, which does look good. But in [mine](https://user-images.githubusercontent.com/58860/228536518-a93d708d-01cc-43b7-aab9-386a19f96b21.png) it appears white, which is more jarring.

MarkLodato:
Looks like the latest change broke it:

image

lehors:
That was merely while the new content was being deployed. Just refresh it should be fine now.

PR #724, opened 2023-03-24
Automatically generate next/previous page links

MarkLodato:
Now the "next page" and "previous page" links automatically generate from the same data source as the nav bar, so that we have a single source of truth. This avoids the error-prone process of manually changing the `next_page`/`prev_page` linked lists.

NOTE: The "use cases" link is currently broken in the nav. I'll fix that via a separate PR.

### DIFF
--- a/docs/_includes/pagination-parse.liquid
+++ b/docs/_includes/pagination-parse.liquid
@@ -1,0 +1,15 @@
+{% comment -%}
+  For each page in `include.collection` (recursively), print:
+    [URL]<ENDURL>[TITLE]<ENDTITLE>
+{% endcomment -%}
+{% for entry in include.collection -%}
+  {% if entry.url and entry.skip_next_prev != true -%}
+    {% assign parts = entry.url | split: '/' -%}
+    {% if parts[1] == include.first_url_part -%}
+      {{entry.url}}<ENDURL>{{entry.title}}<ENDTITLE>
+    {%- endif -%}
+  {% endif -%}
+  {% if entry.children -%}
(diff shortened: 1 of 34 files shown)


docs/_includes/pagination-parse.liquid
```diff
@@ -0,0 +1,15 @@
+{% comment -%}
+  For each page in `include.collection` (recursively), print:
+    [URL]<ENDURL>[TITLE]<ENDTITLE>
+{% endcomment -%}
```
lehors, line 4: Nice trick!

MarkLodato:
Yeah, "fun" indeed 😝

di:
Not sure if this is intentional but this seems to break at https://deploy-preview-724--slsa.netlify.app/spec/v1.0/threats which does not link to https://deploy-preview-724--slsa.netlify.app/attestation-model which is next in the nav tree.

di:
(probably due to `skip_next_prev` being set on the single-page view)

MarkLodato:
That is intentional and should have been the case prior to this PR. We cannot link from a versioned page to an unversioned page because then we'd have multiple inbound links but we can only have one outbound link. For example, if both "/spec/v1.0-rc1/threats" and "/spec/v1.0/threats" say that the next page is "/attestation-model", then there is no single "previous" page that we can list in the latter. So right now we're skipping over those pages.

lehors:
That's by design. It's not supposed to link every page on the website. Only pages within a given spec.
If we go ahead and merge the specs together that will be different.

lehors:
well, Mark beat me to it. :-)

lehors:
By the way, this isn't new. It was already like that before this PR got merged.

PR #733, opened 2023-03-24
Color nav bar section titles differently from links

MarkLodato:
Previously all nav bar text entries were rendered identically, whether or not it was clickable. (Section titles are not clickable.) This made it a little hard to read.

Now (non-clickable) section titles are rendered differently from (clickable) links. The new color is light green, the same color as the prior hover text. This seems like a decent compromise to me. I experimented with bold, small caps, and others, but a slight change in color seemed the most subtle and useful.

To make this distinct from hover, hover/focus is now rendered the same as on the main page: with an underline rather than a change of color.

**For reviewers:** Do you think this is better?

In the screenshots below, "About SLSA" is the current page and the mouse is over "Verifying build systems" (i.e. hover/focus).

Before | After
--- | ---
before | after

### DIFF
--- a/docs/_includes/nav-level.html
+++ b/docs/_includes/nav-level.html
@@ -11,7 +11,7 @@
           {{ subitem.title | escape }}
         </a>
       {%- elsif subitem.children.size > 0 %}
-        {{ subitem.title | escape }}
+        <span class="section-title">{{ subitem.title | escape }}</span>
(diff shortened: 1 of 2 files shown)


joshuagl:
Nice change. I really like using a different style for the titles! Could we further distinguish it from the active/hovered over entry?

joshuagl (approved):
Thank you

PR #737, opened 2023-03-24
Split 'expectations' into two terms

### LINKED ISSUE #736: Who sets expectations?
Following up on the discussion from #675 , I think we can still improve the language around setting expectations. We currently have

>During package upload, a package ecosystem can ensure that the artifact’s provenance matches the known expectations for that package name before accepting it into the registry

and

> Consumers can set their own expectations for artifacts or use default expectations provided by the package producer and/or package ecosystem.

@olivekl pointed out that these sentences imply that "expectations" can have two meanings: the expected values publicized by the producer and the expected values the consumer uses for verification. https://github.com/slsa-framework/slsa/pull/675#discussion_r1140712387

I propose that we use "expected values" or "expected <provenance field>" for the former (publicized by producer) and "verification policy" for the latter (enforced by party performing verification).

kpk47:
Following up on the discussion from https://github.com/slsa-framework/slsa/pull/675 , I think we can still improve the language around setting expectations. We currently have

> During package upload, a package ecosystem can ensure that the artifact’s provenance matches the known expectations for that package name before accepting it into the registry

and

> Consumers can set their own expectations for artifacts or use default expectations provided by the package producer and/or package ecosystem.

@olivekl pointed out that these sentences imply that "expectations" can have two meanings: the expected values publicized by the producer and the expected values the consumer uses for verification. https://github.com/slsa-framework/slsa/pull/675#discussion_r1140712387

I propose that we use "expected values" or "expected " for the former (publicized by producer) and "verification policy" for the latter (enforced by party performing verification).

Fixes #736


docs/spec/v1.0/terminology.md
```diff
@@ -281,6 +283,7 @@ implemented for different, broadly defined, package ecosystems.
 | Term | Example |
 | ---- | ------- |
 | Expectations | Defined separately for each package and stored in the package registry. |
+| Verification policy | An exact copy of the expectations. |
```
olivekl, line 286: Is "an exact copy of the expectation" in this example meant to show that in some cases you could choose to set exactly the expectations as the verification policy? I think that makes sense, but is there any chance this could be misunderstood as meaning the verification policy should always match expectations in an open source context? (Assuming that's not what you want to imply?)

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -81,35 +81,35 @@ A <dfn>monitor</dfn> is a service that verifies provenance for a set
...
+ecosystem may maintain a mapping between package names and their canonical
+source repositories. That mapping constitutes a verification policy. The package
+ecosystem tooling compares artifacts' provenance against its verification policy
+during upload to ensure all packages in the ecosystem are built from their
+canonical source repo, which indicates their authenticity.
+
+A verification policy MUST be sufficient to detect or prevent an adversary from
+injecting unofficial behavior into the package. Example [threats](threats.md) in
+this category include building from an unofficial fork or abusing a build
+parameter to modify the build. Usually verification policy identifies the
```
olivekl, line 104:
```suggestion
parameter to modify the build. Usually the verification policy identifies the
```

docs/spec/v1.0/requirements.md
```diff
@@ -102,10 +102,10 @@ choose a builder capable of producing Build Level 3 provenance.
 ### Follow a consistent build process
 
 The producer MUST build their artifact in a consistent
-manner such that verifiers can form expectations about the build process. In
-some implemenatations, the producer MAY provide explicit metadata to a verifier
-about their build process. In others, the verifier will form their expectations
-implicitly (e.g. trust on first use).
+manner such that verifiers can form a verification policy for the build process.
```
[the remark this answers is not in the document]
kpk47: Done

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -120,28 +120,30 @@ artifact are defined by the package ecosystem.
 <table>
 <tr><th>Requirement<th>Description<th>L1<th>L2<th>L3
 
-<tr id="expectations-known">
-<td>Expectations known
+<tr id="verification-policy-known">
+<td>Verification policy known
```
olivekl, line 124:
```suggestion
<td>Verification policy
```

docs/spec/v1.0/terminology.md
kpk47: This may be a lack of imagination on my part, but I can't think of any other way an OSS ecosystem would set their verification policy.  

I tried softening the language a bit. WDYT?

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -308,27 +310,27 @@ exhaustive):
...
-
-TIP: Difficulty in setting meaningful expectations for `externalParameters` can
-be a sign that the `buildType`'s level of abstraction is too low. For example,
-`externalParameters` that record a list of commands to run is likely impractical
-to verify because the commands change on every build. Instead, consider a
-`buildType` that defines the list of commands in a configuration file in a
-source repository, then put only the source repository in
-`externalParameters`. Such a design is easier to verify because the source
+-   **Explicit policy:** Package producer defines expectations for the
+    package and distributes them to the verifier; the verifier uses these
```
olivekl, line 314: verifier --> consumer?

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -308,27 +310,27 @@ exhaustive):
...
+    expectations as a verification policy after verifying their authenticity.
+    In this model, there MUST be some protection against an adversary
+    unilaterally modifying the verification policy. For example, this might
+    involve two-party control over policy modifications, or having consumers
+    accept each policy change (another form of trust on first use).
+
+-   **Immutable policy:** The verification policy for a package cannot change.
+    In this model, the verification policy consists of an immutable mapping from
+    package name to source repository, and the build process must be defined
+    in the source repository. This is how go works, for example, since the
```
olivekl, line 324: nit, maybe "the go ecosystem"? just for clarity. Or do we capitalize Go?

olivekl (approved):
A few final comments, but LGTM. I think this division makes this document much clearer -- thank you!

docs/spec/v1.0/verifying-artifacts.md

docs/spec/v1.0/verifying-artifacts.md

joshuagl (commented):
Thank you for this, I like the intent and think this will help improve clarity.

We had explicitly moved away from using the term "policy" as it has a lot of baggage. I think the more precise term "verification policy" helps a lot here, but I'd appreciate some broader feedback.

The proposed changes to the "requirements" and "terminology" pages draw a distinction between "expectations" and "verification policy" which is not continued in the "verifying artifacts" page. That makes sense for the "How to verify" section, but I think we want to mention expectations in the "Architecture options" and "Setting verification policy" sections --- WDYT?

docs/spec/v1.0/terminology.md
```diff
@@ -259,7 +259,8 @@ build system the package was built.
 
 | Term         | Description |
 |--------------|---- |
-| Expectations | A set of constraints on the package's provenance metadata. The package producer sets expectations for a package, whether explicitly or implicitly. |
+| Expectations | A set of expected values for the purpose of verifying the package's provenance metadata. The package producer sets expectations for a package, whether explicitly or implicitly. |
+| Verification policy | A set of constraints on the package's provenance metadata. The provenance consumer sets the verification policy based on the expectations. |
```
MarkLodato, line 263: Use "verifier" for consistency with #782?

```suggestion
| Verification policy | A set of constraints on the package's provenance metadata. The provenance verifier sets the verification policy based on the expectations. |
```

MarkLodato, line 263: I'm having a hard time understanding the difference between these two concepts. They sound very similar, and I can't really follow why some places use "verification policy" while others use "expectations". I also find it confusing that *producers* set expectations.

Do we need both concepts? Can we get away with one?

I suspect that the original problem arose from two underlying issues:

1.  Trying to use "expectations" as a technical term with special meaning (jargon) rather than using natural English phrases whose meaning is obvious. Compare the following examples from [threats](https://slsa.dev/spec/v1.0/threats):
    -   OK: "Verifier requires the provenance’s source location to match an expected value."
        - This reads fine to me. I don't need a definition of "expected value" (though I might wonder where those expectations come from.)
    -   Confusing: "Modify the expectations to accept something that would not otherwise be accepted."
        - How can someone "modify expectations"? What does that mean? (Here it's referring to a particular record of expectations.)
2.   Trying to be too generic about how verification works, leading to confusion about whether the producer or consumer sets expectations. If instead we talked specifically about the different models (upload time, download time) that might avoid the need to define a new term.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -81,35 +81,35 @@ A <dfn>monitor</dfn> is a service that verifies provenance for a set
...
-ecosystem tooling tests those expectations during upload to ensure all packages
-in the ecosystem are built from their canonical source repo, which
-indicates their authenticity.
-
-Expectations MUST be sufficient to detect or prevent an adversary from injecting
-unofficial behavior into the package. Example [threats](threats.md) in this
-category include building from an unofficial fork or abusing a build parameter
-to modify the build. Usually expectations identify the canonical source
-repository (which is the main external parameter) and any other
+## Setting verification policy
```
MarkLodato, line 91: This section repeats a lot of what is already in Step 2 below. Should we just eliminate it? I wonder if it raises more questions than it answers.

MarkLodato (changes requested):

kpk47:
I think this change is too big to make it into RC2. I've instead mailed #795, which is a more modest change that may also clarify what we mean by setting and testing expectations.

Issue #672, opened 2023-03-08
Restructure documentation and improve content


olivekl:
I’m planning to restructure and rewrite the docs on slsa.dev over the course several PRs before the v1.0 launch.

The goal end state is:

- Overview (slsa.dev landing page)
- Understanding SLSA
  - What's new in v1.0
  - About SLSA
  - Threats overview
  - Use cases
  - Guiding principles
  - Future directions
- Core specification
  - Security levels
  - Producing
  - Verifying
  - Threats in detail
- Attestations
  - General model
  - Provenance
  - VSA
- How to SLSA
  - For developers
  - For organizations
  - For implementers
- Specification stages
- Community
- Blog

I’m planning these PRs to keep reviews manageable:

PR 1: [DONE](https://github.com/slsa-framework/slsa/pull/683) Move / divide existing content, create new structure, but no content changes

- Change Spec index page to have multiple tables reflecting new structure
- Create Understanding–Concepts page (rename Guiding Principles to Concepts, move What is SLSA, etc into that page)
- Create Understanding–Threats Overview page (split and move 1st half of current threats info there)
- Move Use Cases under Understanding heading

PR 2:[DONE](https://github.com/slsa-framework/slsa/pull/705): Update content, focusing on Understanding SLSA section

- Edit, rewrite, and and add new content to make SLSA more understandable to newcomers

PR 3: [DONE](https://github.com/slsa-framework/slsa/pull/739) Standardize and reduce terms across all pages 

- I.e., developer vs. creator vs. maintainer; build vs. release vs. publish)

PR 4: Add new pages How to SLSA (for implementers and organizations)

- Basic instructions for three main user journeys (developers, organizations, implementers)

PR 5: [DONE](https://github.com/slsa-framework/slsa/pull/743)Add General Model under Attestations 
- Anything else lost in the shuffle

PR 6: Final nav bar reorganization, banner link to v1.0

Happy for feedback and discussion!  @di @MarkLodato @joshuagl @lehors @kpk47 @mlieberman85

MarkLodato:
To clarify thinking here: It seems valuable to have a "SLSA v1.0" header in the menu even if technically attestation formats are versioned independently. The goal is to make it easy for a reader to find the information they are looking for. The organization of the menu might not reflect the URL scheme in all cases.

arewm:
[...]

PR #743, opened 2023-03-27
Explain attestation formats

olivekl:
PR to explain attestations in a series of documentation updates outlined in Issue #672.

Adds a section telling users how to choose among attestations formats based on whether they are using SLSA for first party, open source, or third party close source code.

@MarkLodato @kpk47 @arewm @lehors

### DIFF
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -38,13 +38,13 @@ dependencies (as in the case of SLSA [Provenance]).
 
 This section explains how to choose the attestation format best for your
 situation by considering factors such as use case and who will be consuming
-the attestation. 
+the attestation.
 
 ### First party
 
 Producers of first party code might consider the following questions:
 -   Will SLSA be used only within our organization?
--   Is the primary use case to manage insider risk?
+-   Is SLSA's primary use case to manage insider risk?
 -   Are we developing entirely in a closed source environment?
 
 If these are the main considerations, the organization can choose any format
@@ -55,10 +55,10 @@ attestations and the easiest to verify using the [Generic SLSA Verifier].
 
 ### Open source
 
-Producers of open source code might consider these questions: 
--   Is the primary use case to convey trust in how your code was developed?
+Producers of open source code might consider these questions:
+-   Is SLSA's primary use case to convey trust in how your code was developed?
 -   Do you develope software with standard open source licenses?
--   Will the code be consumed by others? 
+-   Will the code be consumed by others?
 
 In these situations, it is recommended you use the [SLSA Provenance format]. The SLSA
 Provenance format offers a path towards interoperability and cohesion across the open
@@ -69,8 +69,8 @@ using the [Generic SLSA Verifier].
 
 Producers of closed source code that is consumed by others might consider
 the following questions:
--   Is my code produced for third party consumers, backed with contracts?
--   Is the primary use case to create trust in our organization or to comply with
+-   Is my code produced for the sole purpose of specific third party consumers?
+-   Is SLSA's primary use case to create trust in our organization or to comply with
 audits and legal requirements?
 
 In these situation, you may not want to make all the details of your
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -11,7 +11,7 @@ The primary intended use case is to feed into automated policy engines, such as
 
 This page provides a high-level overview of the attestation model, including
 standardized terminology, data model, layers, conventions for software
-attestations, and recommended formats for different use cases.
+attestations, and suggested formats for different use cases.
 
 ## Overview
 
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -60,7 +60,7 @@ Producers of open source code might consider these questions:
 -   Do you develope software with standard open source licenses?
 -   Will the code be consumed by others? 
 
-In these situations, you should use the [SLSA Provenance format]. The SLSA
+In these situations, it is recommended you use the [SLSA Provenance format]. The SLSA
 Provenance format offers interoperability and cohesion across the open
 source ecosystem. Users can verify any provenance statement in this format
 using the [Generic SLSA Verifier].
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -11,7 +11,7 @@ The primary intended use case is to feed into automated policy engines, such as
 
 This page provides a high-level overview of the attestation model, including
 standardized terminology, data model, layers, conventions for software
-attestations, and suggested formats for different use cases.
+attestations, and formats for different use cases.
 
 ## Overview
 
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -51,8 +51,8 @@ Producers of first party code might consider the following questions:
 If these are the main considerations, the organization can choose any format
 for internal use. To make an external claim of meeting a SLSA level, however,
 there needs to be a way for external users to consume and verify your provenance.
-Currently, [SLSA Provenance format] is the suggested format for SLSA
-attestations and the easiest to verify using the [Generic SLSA Verifier].
+Currently, SLSA recommends using the [SLSA Provenance format] for SLSA
+attestations since it is easy to verify using the [Generic SLSA Verifier].
 
 ### Open source
 
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -42,7 +42,7 @@ choose the format best for your situation.
 ### First party
 
 As a first party producer, you use SLSA only within your organization,
-primarily manage insider risk. If you are developering entirely in
+primarily to manage insider risk. If you are developing entirely in
 a closed source environment, no particular format is required for internal use.
 
 If you want to make an external claim of meeting a SLSA level, however, there
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -36,9 +36,9 @@ dependencies (as in the case of SLSA [Provenance]).
 
 ## Formats
 
-This section explains how to choose the attestation format best for your
-situation by considering factors such as use case and who will be consuming
-the attestation.
+This section explains how to choose the attestation format that's best suited
+for your situation by considering factors such as intended use and who will be
+consuming the attestation.
 
 ### First party
 
@@ -62,7 +62,7 @@ Producers of open source code might consider these questions:
 -   Do you develop software with standard open source licenses?
 -   Will the code be consumed by others?
 
-In these situations, it is suggested you use the [SLSA Provenance format]. The SLSA
+In these situations, we encourage you to use the [SLSA Provenance format]. The SLSA
 Provenance format offers a path towards interoperability and cohesion across the open
 source ecosystem. Users can verify any provenance statement in this format
 using the [Generic SLSA Verifier].
@@ -159,4 +159,4 @@ recognize that other choices MAY be necessary in various cases.
 [SLSA Provenance format]: /provenance/v1.md
 [sigstore/cosign]: https://github.com/sigstore/cosign
 [SPDX]: https://github.com/in-toto/attestation/blob/main/spec/predicates/spdx.md
-[Verification Summary Attestation]: /attestation-model.md
+[Verification Summary Attestation]: /verification_summary/v1.md
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -59,7 +59,7 @@ attestations since it is easy to verify using the [Generic SLSA Verifier].
 Producers of open source code might consider these questions:
 
 -   Is SLSA's primary use case to convey trust in how your code was developed?
--   Do you develope software with standard open source licenses?
+-   Do you develop software with standard open source licenses?
 -   Will the code be consumed by others?
 
 In these situations, it is suggested you use the [SLSA Provenance format]. The SLSA
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -50,7 +50,7 @@ Producers of first party code might consider the following questions:
 If these are the main considerations, the organization can choose any format
 for internal use. To make an external claim of meeting a SLSA level, however,
 there must be a way for external users to consume and verify your provenance.
-Currently, [SLSA Provenance format] is the most widely used format for SLSA
+Currently, [SLSA Provenance format] is the recommended format for SLSA
 attestations and the easiest to verify using the [Generic SLSA Verifier].
 
 ### Open source
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -61,7 +61,7 @@ Producers of open source code might consider these questions:
 -   Will the code be consumed by others? 
 
 In these situations, it is recommended you use the [SLSA Provenance format]. The SLSA
-Provenance format offers interoperability and cohesion across the open
+Provenance format offers a path towards interoperability and cohesion across the open
 source ecosystem. Users can verify any provenance statement in this format
 using the [Generic SLSA Verifier].
 
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -76,7 +76,7 @@ the following questions:
 -   Is SLSA's primary use case to create trust in our organization or to comply with
 audits and legal requirements?
 
-In these situation, you may not want to make all the details of your
+In these situations, you may not want to make all the details of your
 provenance available externally. Consider using Verification Summary
 Attestations (VSAs) to summarize provenance information in a sanitized way
 that's safe for external consumption. For more about VSAs, see the [Verification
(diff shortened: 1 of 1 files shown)


docs/attestation-model.md
```diff
@@ -34,6 +34,45 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+must be a way for external users to consume and verify your provenance.
+Currently, [SLSA Provenance format] is the most widely used format for SLSA
+attestations and the easiest to verify using the [Generic SLSA Verifier].
+
+### Open source
+
+As an open source producer, you produce software with standard open source
+licenses (no contracts or warranties). You probably use SLSA so that others can
+trust how your code was developed. If you develop open source projects that
+will be consumed by others, you should use the [SLSA Provenance format].
```
arewm, line 58: > You probably use SLSA so that others can
trust how your code was developed.

This phrasing seems a little odd/presumptuous. Do we need it in there?

Maybe instead, we can word it in a way that the SLSA provenance format is recommended to use in the absence of any other provenance format requirements.

olivekl: Yeah, I can see that. I was trying to guide and disambiguate based on the likely use case. I'm okay with just removing it. The other use cases also make presumptions too, though; what do you think of those?

arewm: Apparently I didn't notice anything on a first pass, but I see what you are saying now. Instead of chunking up like you have it, maybe we can just reinforce in the parent section (`Formats`) that there are no requirements but that there may be certain criteria to include in making a decision. Then in each section, just highlight the potential questions/concerns that might be relevant.

We can then put a recommendation at the end with a brief defense similar to what you did..

docs/attestation-model.md
```diff
@@ -34,6 +34,45 @@ raw signing. For example, an attestation might state exactly how an artifact
...
 
+## Formats
+
+There is no single required attestation format. This section explains how to
+choose the format best for your situation.
+
+### First party
+
+As a first party producer, you use SLSA only within your organization,
+primarily manage insider risk. If you are developering entirely in
```
joshuagl, line 45:
```suggestion
primarily to manage insider risk. If you are developing entirely in
```

docs/attestation-model.md
joshuagl: I agree with the comments on presumptuousness and like the idea of communicating the criteria but not presuming why those criteria are relevant.

joshuagl (changes requested):

docs/attestation-model.md
olivekl: Updated! Does this work better? @arewm @joshuagl

arewm: Generally better, yes. I added more comments.

docs/attestation-model.md
```diff
@@ -34,6 +34,52 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+In these situations, you should use the [SLSA Provenance format]. The SLSA
+Provenance format offers interoperability and cohesion across the open
+source ecosystem. Users can verify any provenance statement in this format
+using the [Generic SLSA Verifier].
+
+### Closed source, third party
+
+Producers of closed source code that is consumed by others might consider
+the following questions:
+-   Is my code produced for third party consumers, back with contracts?
```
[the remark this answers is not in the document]
olivekl: Mistake: "backed with contracts." To differentiate from open source licenses.

docs/attestation-model.md
```diff
@@ -34,6 +34,52 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+the attestation. 
+
+### First party
+
+Producers of first party code might consider the following questions:
+-   Will SLSA be used only within our organization?
+-   Is the primary use case to manage insider risk?
+-   Are we developing entirely in a closed source environment?
+
+In these situations, no particular format is required for internal use.
```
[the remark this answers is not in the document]
arewm: I see what you are saying. How about if we change the first sentence slightly to something like

> If these are the main considerations, the organization can choose any format for internal use.

I am not attached to it, but I wanted to still impress that there may be some entity that has a format preference. It may be too specific.

docs/attestation-model.md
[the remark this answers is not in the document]
arewm: I think we can still make that clearer. I feel like the fact that a contract exists doesn't matter as much as the fact that there is a specific delivery mechanism of the code. I think that I understand what you are trying to say but I don't have the words for it. I can think about it some more.

docs/attestation-model.md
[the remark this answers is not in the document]
olivekl: Sounds good to me. Updated (and combined with next paragraph)

docs/attestation-model.md
```diff
@@ -34,6 +34,52 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+attestations and the easiest to verify using the [Generic SLSA Verifier].
+
+### Open source
+
+Producers of open source code might consider these questions: 
+-   Is the primary use case to convey trust in how your code was developed?
+-   Do you develope software with standard open source licenses?
+-   Will the code be consumed by others? 
+
+In these situations, you should use the [SLSA Provenance format]. The SLSA
```
lehors, line 63: I suggest sticking with the same form:
```suggestion
In these situations, it is recommended you use the [SLSA Provenance format]. The SLSA
```

docs/attestation-model.md
```diff
@@ -34,6 +34,52 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+Provenance format offers interoperability and cohesion across the open
+source ecosystem. Users can verify any provenance statement in this format
```
lehors, line 65: This sounds like wishful thinking to me at this point. I'd rather we use some more nuance form such as:
```suggestion
Provenance format offers a path towards interoperability and cohesion across the open
source ecosystem. Users can verify any provenance statement in this format
```

lehors (commented):
I think we need to be explicit what we wish and avoid making presumptuous claims.

lehors (approved):
LGTM. Thanks!

MarkLodato (approved):

docs/attestation-model.md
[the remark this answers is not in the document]
arewm: How about if we just change it to some wording like

> Is my code produced for the sole purpose of specific third party consumers?

I feel like this more clearly represents the specific targeted relationships between parties.

arewm (approved):

docs/attestation-model.md
```diff
@@ -34,6 +34,54 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+### Closed source, third party
+
+Producers of closed source code that is consumed by others might consider
+the following questions:
+
+-   Is my code produced for the sole purpose of specific third party consumers?
+-   Is SLSA's primary use case to create trust in our organization or to comply with
+audits and legal requirements?
+
+In these situation, you may not want to make all the details of your
```
lehors, line 79: nit:
```suggestion
In these situations, you may not want to make all the details of your
```

olivekl:
> I think RFC2119 isn't a very good fit for this actually but in my opinion there is no real harm in capitalizing it to make the linter happy. Otherwise you need to figure another way to say the same thing without using may! Which I know is something @MarkLodato has done in the past...

I agree, I'd rather this section be more about helpful guidance than MUST, MAY, etc. Maybe we can do a dance around all the linter words (though I think "recommend" is one of them and that could be hard).

docs/attestation-model.md
```diff
@@ -34,6 +34,54 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+
+Producers of first party code might consider the following questions:
+
+-   Will SLSA be used only within our organization?
+-   Is SLSA's primary use case to manage insider risk?
+-   Are we developing entirely in a closed source environment?
+
+If these are the main considerations, the organization can choose any format
+for internal use. To make an external claim of meeting a SLSA level, however,
+there must be a way for external users to consume and verify your provenance.
```
kpk47, line 53:
```suggestion
there needs to be a way for external users to consume and verify your provenance.
```

docs/attestation-model.md
```diff
@@ -34,6 +34,54 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+### Closed source, third party
+
+Producers of closed source code that is consumed by others might consider
+the following questions:
+
+-   Is my code produced for the sole purpose of specific third party consumers?
+-   Is SLSA's primary use case to create trust in our organization or to comply with
+audits and legal requirements?
+
+In these situations, you may not want to make all the details of your
```
kpk47, line 79:
```suggestion
In these situations, you might not want to make all the details of your
```

docs/attestation-model.md
```diff
@@ -10,8 +10,8 @@ The primary intended use case is to feed into automated policy engines, such as
 [in-toto] and [Binary Authorization].
 
 This page provides a high-level overview of the attestation model, including
-standardized terminology, data model, layers, and conventions for software
-attestations.
+standardized terminology, data model, layers, conventions for software
+attestations, and recommended formats for different use cases.
```
kpk47, line 14:
```suggestion
attestations, and suggested formats for different use cases.
```

docs/attestation-model.md
```diff
@@ -34,6 +34,54 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+Producers of first party code might consider the following questions:
+
+-   Will SLSA be used only within our organization?
+-   Is SLSA's primary use case to manage insider risk?
+-   Are we developing entirely in a closed source environment?
+
+If these are the main considerations, the organization can choose any format
+for internal use. To make an external claim of meeting a SLSA level, however,
+there must be a way for external users to consume and verify your provenance.
+Currently, [SLSA Provenance format] is the recommended format for SLSA
```
kpk47, line 54:
```suggestion
Currently, [SLSA Provenance format] is the suggested format for SLSA
```

docs/attestation-model.md
```diff
@@ -34,6 +34,54 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+
+### Open source
+
+Producers of open source code might consider these questions:
+
+-   Is SLSA's primary use case to convey trust in how your code was developed?
+-   Do you develope software with standard open source licenses?
+-   Will the code be consumed by others?
+
+In these situations, it is recommended you use the [SLSA Provenance format]. The SLSA
```
kpk47, line 65:
```suggestion
In these situations, it is suggested you use the [SLSA Provenance format]. The SLSA
```

kpk47 (approved):
I agree with @lehors that RFC2119 don't apply to the sections you're writing, but it's (unfortunately) easier to dance around the keywords than fix the linter. I left a few suggested dance moves, but otherwise LGTM.

olivekl:
> I agree with @lehors that RFC2119 don't apply to the sections you're writing, but it's (unfortunately) easier to dance around the keywords than fix the linter. I left a few suggested dance moves, but otherwise LGTM.

Thank you! I took all the suggestions. (Danced around? SLSAed around, surely.)

lehors:
I appreciate @kpk47's effort to workaround the linter but I think the change from recommended formats to suggested is wrong. Indeed, [the spec states that one SHOULD use SLSA Provenance](https://slsa.dev/spec/v1.0/distributing-provenance#format-of-the-attestation) so, it doesn't seem right to say here that this is merely a suggestion.

docs/attestation-model.md
```diff
@@ -34,6 +34,54 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+
+-   Will SLSA be used only within our organization?
+-   Is SLSA's primary use case to manage insider risk?
+-   Are we developing entirely in a closed source environment?
+
+If these are the main considerations, the organization can choose any format
+for internal use. To make an external claim of meeting a SLSA level, however,
+there needs to be a way for external users to consume and verify your provenance.
+Currently, [SLSA Provenance format] is the suggested format for SLSA
+attestations and the easiest to verify using the [Generic SLSA Verifier].
```
lehors, line 55:
```suggestion
Currently, SLSA recommends using the [SLSA Provenance format] for SLSA
attestations since it is easy to verify using the [Generic SLSA Verifier].
```

docs/attestation-model.md
```diff
@@ -10,8 +10,8 @@ The primary intended use case is to feed into automated policy engines, such as
 [in-toto] and [Binary Authorization].
 
 This page provides a high-level overview of the attestation model, including
-standardized terminology, data model, layers, and conventions for software
-attestations.
+standardized terminology, data model, layers, conventions for software
+attestations, and suggested formats for different use cases.
```
lehors, line 14: I think we can simply skip the word in this instance:
```suggestion
attestations, and formats for different use cases.
```

docs/attestation-model.md
```diff
@@ -107,5 +156,7 @@ recognize that other choices MAY be necessary in various cases.
 [Provenance]: /provenance
 [remote attestation]: https://en.wikipedia.org/wiki/Trusted_Computing#Remote_attestation
 [RFC 2119]: https://tools.ietf.org/html/rfc2119
+[SLSA Provenance format]: /provenance/v1.md
```
lehors, line 159:
```suggestion
[Provenance format]: /provenance/v1.md
```

lehors (commented):
Suggested a couple of changes to avoid triggering the linter but keeping in line with the recommendation we set in the spec.

olivekl:
> Suggested a couple of changes to avoid triggering the linter but keeping in line with the recommendation we set in the spec.

Thank you! I've incorporated them.

I think that addresses everything. @joshuagl do you have any requests or is it okay to merge?

docs/attestation-model.md
```diff
@@ -34,6 +34,54 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+there needs to be a way for external users to consume and verify your provenance.
+Currently, SLSA recommends using the [SLSA Provenance format] for SLSA
+attestations since it is easy to verify using the [Generic SLSA Verifier].
+
+### Open source
+
+Producers of open source code might consider these questions:
+
+-   Is SLSA's primary use case to convey trust in how your code was developed?
+-   Do you develope software with standard open source licenses?
```
marcelamelara, line 62: Nit: 
```suggestion
-   Do you develop software with standard open source licenses?
```

docs/attestation-model.md
```diff
@@ -107,5 +156,7 @@ recognize that other choices MAY be necessary in various cases.
 [Provenance]: /provenance
 [remote attestation]: https://en.wikipedia.org/wiki/Trusted_Computing#Remote_attestation
 [RFC 2119]: https://tools.ietf.org/html/rfc2119
+[SLSA Provenance format]: /provenance/v1.md
 [sigstore/cosign]: https://github.com/sigstore/cosign
 [SPDX]: https://github.com/in-toto/attestation/blob/main/spec/predicates/spdx.md
+[Verification Summary Attestation]: /attestation-model.md
```
marcelamelara, line 162: This link seems to point to this same document. Should this link be pointing to: verification_summary/v1.md?

marcelamelara (changes requested):
Thanks! I only a couple minor changes.

docs/attestation-model.md
```diff
@@ -34,6 +34,54 @@ raw signing. For example, an attestation might state exactly how an artifact
 was produced, including the build command that was run and all of its
 dependencies (as in the case of SLSA [Provenance]).
 
+## Formats
+
+This section explains how to choose the attestation format best for your
+situation by considering factors such as use case and who will be consuming
+the attestation.
```
joshuagl, line 41: This doesn't read quite right, is it missing a word? Perhaps: 
```suggestion
This section explains how to choose the attestation format best suited for your
situation by considering factors such as use case and who will be consuming
the attestation.
```

docs/attestation-model.md
```diff
@@ -34,6 +34,54 @@ raw signing. For example, an attestation might state exactly how an artifact
...
+
+### Open source
+
+Producers of open source code might consider these questions:
+
+-   Is SLSA's primary use case to convey trust in how your code was developed?
+-   Do you develope software with standard open source licenses?
+-   Will the code be consumed by others?
+
+In these situations, it is suggested you use the [SLSA Provenance format]. The SLSA
```
joshuagl, line 65: I'd be tempted to go so far as to say we encourage use of the SLSA Provenance format.

joshuagl (commented):
Apart from the fixes @marcelamelara suggested this LGTM. Thanks.

docs/attestation-model.md
olivekl: Thanks! I took your suggestions and added even a bit more clarification:
"This section explains how to choose the attestation format that's best suited
for your situation by considering factors such as intended use and who will be
consuming the attestation."

docs/attestation-model.md
olivekl: done, thank you!

docs/attestation-model.md
joshuagl: nice, thanks

joshuagl (approved):

marcelamelara (approved):
Thanks @olivekl ! LGTM

PR #745, opened 2023-03-27
Update parsing rules to match latest in-toto extension field rules

### LINKED ISSUE #670: Extension field format and periods
I am trying to make use of extension fields and the SLSA specification defers to the in-toto specification:

> Extension fields: Producers MAY add extension fields to any JSON object by using a property name that is a [TypeURI](https://github.com/in-toto/attestation/blob/main/spec/v1.0-draft/scalar_field_types.md#TypeURI). The use of URI is to protect against name collisions. Consumers MAY parse and use these extensions if desired. The presence or absence of the extension field MUST NOT influence the meaning of any other field, and the field MUST follow the monotonic princple.

This would mean that field names (not field values) would look like URLs.   I am storing my Provenances in MongoDB, but MongoDB does not allow periods in field names.  This is so that you can reference subdocuments with a period.    This means that I would have to restore to either mangling the extension field names before storing in the database.    I suspect this will be a problem with other databases and JSON tools.  The TypeURI seems like it was really meant for unique field values and not intended to create unique field names.   Please consider some other approach where the names do not have to include so many special characters (colon, period, and slash right now).

marcelamelara:
Updates the SLSA Provenance parsing rules for extension fields to match the latest parsing guidelines from the in-toto Attestation Framework. In particular, the new rules remove the recommendation to use TypeURIs and add an explanation to avoid certain special characters in extension field names.

Fixes #670

### DIFF
--- a/docs/provenance/v1.md
+++ b/docs/provenance/v1.md
@@ -93,9 +93,12 @@ This predicate follows the in-toto attestation [parsing rules]. Summary:
 -   Consumers MUST ignore unrecognized fields unless otherwise noted.
 -   The `predicateType` URI includes the major version number and will always
     change whenever there is a backwards incompatible change.
--   Minor version changes are always backwards compatible and "monotonic." Such
-    changes do not update the `predicateType`.
--   Producers MAY add extension fields using field names that are URIs.
+-   Minor version changes are always backwards compatible and "monotonic."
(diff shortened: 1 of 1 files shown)


MarkLodato (approved):
Thanks!

PR #748, opened 2023-03-27
Rename "non-forgeable" to "unforgeable"

### LINKED ISSUE #721: SLSA v1: rename "non-forgeable" to "unforgeable"?
We currently say [non-forgeable](https://slsa.dev/spec/v1.0/requirements#provenance-non-forgeable) but "unforgeable" is a much more natural English word.

MarkLodato:
The latter is a more common and natural English word.

Also simplify the language of the requirement slightly without changing meaning.

Fixes #721.

**Question for reviewers:** Does this term sound as good as the old one? Is there a different term that is even better?


joshuagl (approved):
This reads much more naturally, thank you!

PR #749, opened 2023-03-27
VSA v1: convert to lowerCamelCase

### LINKED ISSUE #741: v1 VSA: use lowerCamelCase?
The Provenance format uses lowerCamelCase naming convention for fields. Should we do the same for VSA for consistency?

MarkLodato:
For consistency with Provenance.

Fixes #741.


joshuagl (approved):

MarkLodato:
Since we have several changes queued up for VSA, I'm going to submit this now to avoid merge conflicts. I believe there is community consensus on the rename so I don't think this is controversial.

PR #750, opened 2023-03-27
SLSA v1: Remove references to expectations from levels.md

### LINKED ISSUE #729: SLSA V1.0 Spec wording - "Higher SLSA build levels provide increased confidence that a package truly came from the correct sources", is 'correct' accurate?
A nit about the wording on https://slsa.dev/spec/v1.0-rc1/levels#build-track

"Higher SLSA build levels provide increased confidence that a package truly came from the **correct** sources, without unauthorized modification or influence."

My question is around the accuracy of the word 'correct', I don't think this makes any claims or improvements as to the 'correctness' of the sources, rather just that the provenance is an accurate representation of the sources. I.e. you can still be importing the wrong thing, SLSA won't help with that but it will accurately describe the wrong thing you're importing

### LINKED ISSUE #735: SLSA V1.0 spec - "Downstream tooling" requirement
In the SLSA V1.0 spec, in L1 of the Build track there is the requirement 

"Downstream tooling automatically verifies that the artifact’s provenance exists and matches the expectation."

((https://slsa.dev/spec/v1.0-rc1/levels#build-l1-provenance))

If this is a requirement for L1 and therefore any subsequent levels, how can we verify that this is occurring and/or do we have to implement it ourselves as a SLSA builder?

MarkLodato:
de5117ad added expectations and verification to the build track in levels.md, but since then we have decided to split that into a separate concept ("Verifying Artifacts"). However, the old language remained in levels.md.

This change addresses those remnants by explaining that the build level is purely about the provenance, but mentioning that the primary purpose is to enable verification and matching against expectations. Hopefully this is the right balance to strike.

Also removes a now-irrelevant TODO.

Fixes #729.
Fixes #735.


MarkLodato:
@khalkie does this address your concerns?

docs/spec/v1.0/levels.md
```diff
@@ -23,7 +23,7 @@ tracks without invalidating previous levels.
 | Track/Level | Requirements | Benefits |
 | ----------- | ------------ | -------- |
 | [Build L0]  | (none)       | (n/a)    |
-| [Build L1]  | Attestation showing that the package was built as expected | Documentation, mistake prevention, inventorying |
+| [Build L1]  | Attestation showing how the package was built | Documentation, mistake prevention, inventorying |
```
khalkie, line 26: Does it still cover mistake prevention?

khalkie:
> @khalkie does this address your concerns?

Yes this makes sense, thanks!

joshuagl (approved):

docs/spec/v1.0/levels.md
```diff
@@ -101,25 +94,25 @@ SLSA---other than tamper protection---without changing their build workflows.
...
-    built, including the canonical source repository and build command.
+-   Software producer follows a consistent build process so that others can form
+    expectations about what a "correct" build looks like.
 
--   On each build, the release process automatically generates and distributes a
-    [provenance attestation] describing how the artifact was *actually* built,
-    including: who built the package (person or system), what process/command
-    was used, and what the input artifacts were.
+-   Build system automatically generates [provenance attestation] describing how
+    the artifact was built, including: who built the package (person or system),
```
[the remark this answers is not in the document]
arewm: The `builder.id` is required by L1, so that should theoretically describe the system which built the artifact.

docs/spec/v1.0/levels.md
```diff
@@ -90,9 +84,8 @@ n/a
 <dl class="as-table">
```
arewm, line 84: I cannot comment higher at the section heading, but I see the provenance heading as confusing.

I understand that `Build L1` only requires provenance to be generated, but the initial reading of the sections feels like it is only a requirement for L1 and not that there are additional provenance requirements for L2/L3.

The summary of each section makes it clearer. Would it be reasonable to remove the content following the `:`, instead adding some subheading with more information?

docs/spec/v1.0/levels.md
```diff
@@ -23,7 +23,7 @@ tracks without invalidating previous levels.
 | Track/Level | Requirements | Benefits |
 | ----------- | ------------ | -------- |
 | [Build L0]  | (none)       | (n/a)    |
-| [Build L1]  | Attestation showing that the package was built as expected | Documentation, mistake prevention, inventorying |
+| [Build L1]  | Attestation showing how the package was built | Documentation, mistake prevention, inventorying |
 | [Build L2]  | Signed attestation, generated by a hosted build service | Reduced attack surface, weak tamper protection |
 | [Build L3]  | Hardened build service | Strong tamper protection |
```
arewm, line 28: Can we qualify tamper protection a little more here?

docs/spec/v1.0/levels.md
```diff
@@ -23,7 +23,7 @@ tracks without invalidating previous levels.
 | Track/Level | Requirements | Benefits |
 | ----------- | ------------ | -------- |
 | [Build L0]  | (none)       | (n/a)    |
-| [Build L1]  | Attestation showing that the package was built as expected | Documentation, mistake prevention, inventorying |
+| [Build L1]  | Attestation showing how the package was built | Documentation, mistake prevention, inventorying |
 | [Build L2]  | Signed attestation, generated by a hosted build service | Reduced attack surface, weak tamper protection |
```
arewm, line 27: Attestation is used in the summary table, but it is `provenance attestation` elsewhere. Which should be the shortened version that is preferred if only using one word? I feel like we just use `provenance` elsewhere instead of just `attestation`.

MarkLodato: Fixed. I filed #758 to do a full pass throughout the spec.

docs/spec/v1.0/levels.md
MarkLodato: Do you mind filing a separate issue for this? It seems unrelated to this PR. I think it's useful to have a short description of each level to stick in the reader's minds, and I had been interpreting this as the difference from the prior level.

docs/spec/v1.0/levels.md
MarkLodato: Let's merge with the comment thread above since you're both getting at the same thing.

docs/spec/v1.0/levels.md
MarkLodato: [joining with the other thread]

> (@khalkie) Does it still cover mistake prevention?
> (@arewm) Can we qualify tamper protection a little more here?

IMO it does *when verified against expectations*. My intention was to present the levels as:
L1 = mistake protections (stop unintentional badness)
L2 = weak tamper protection (stop unmotivated / risk-averse adversaries)
L3 = strong tamper protection (stop motivated adversaries)

Do you think that framing is useful? Is it confusing that it's "when verified against expectations"? Any other ideas on how to frame it that avoids those issues?

docs/spec/v1.0/levels.md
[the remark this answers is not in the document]
MarkLodato: Yeah, implicit in that statement is that at L1, the `builder.id` could be a person since it was built on their workstation. But I think that's more confusing than helpful. Let me think of alternatives...

docs/spec/v1.0/levels.md
arewm: I just created a PR for it. I think that this might be sufficient: https://github.com/slsa-framework/slsa/pull/761

docs/spec/v1.0/levels.md
MarkLodato: What about replacing the current "Benefits" column with the following?

Level | Focus
--- | ---
Build L1 | Mistakes, documentation
Build L2 | Unsophisticated adversaries
Build L3 | Sophisticated adversaries

(I'm not 100% sold, but maybe this spurs further ideas?)

MarkLodato: Good point. What about the following?

| Level | Focus
| --- | ---
| Build L1 | Mistakes, documentation
| Build L2 | Tampering after build
| Build L3 | Tampering during build
| Build L4 | Completeness of provenance

arewm: I think that makes more sense. L4, when it comes, _can_ include reproducibility ... or it can also be included in a L5 as it also has a clear separation.

I also like how it clearly represents what _is not_ in focus in levels based on what _is_ in focus in future levels.

MarkLodato: [cc812c3](https://github.com/slsa-framework/slsa/pull/750/commits/cc812c33a03ec98f8bdb9b22304a38b130e9f23e)

OK to resolve?

docs/spec/v1.0/levels.md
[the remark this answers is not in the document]
MarkLodato: I replaced with "what entity built the package". How does that sound?

docs/spec/v1.0/levels.md
arewm: I am okay with it, yes. Thanks!

docs/spec/v1.0/levels.md
```diff
@@ -90,9 +87,8 @@ n/a
 <dl class="as-table">
 <dt>Summary<dd>
 
-Package has a provenance attestation showing how it was built, and a downstream
-system automatically verifies that packages were built as expected. Prevents
-mistakes but is trivial to bypass or forge.
+Package has a provenance attestation showing how it was built. Can be used to
```
arewm, line 90: L1 just needs provenance, not a provenance attestation.

docs/spec/v1.0/levels.md
```diff
@@ -20,37 +20,34 @@ to recognize progress made in one aspect of security without blocking on an
 unrelated aspect. Tracks also allow the SLSA spec to evolve: we can add more
 tracks without invalidating previous levels.
 
-| Track/Level | Requirements | Benefits |
-| ----------- | ------------ | -------- |
+| Track/Level | Requirements | Focus |
```
arewm, line 23: I am okay with the change up here, but it is inconsistent with the rest of the page as we talk about benefits lower down. If you want to be consistent, we should change it there too.

arewm (changes requested):

docs/spec/v1.0/levels.md
MarkLodato: Done

docs/spec/v1.0/levels.md
MarkLodato: It reads OK to me as-is? I interpret "focus" as a brief summary, while the "benefits" is a full list. The two contents are consistent; just the "focus" is a subset of "benefits.

@olivekl any opinions?

arewm (approved):

docs/spec/v1.0/levels.md
MarkLodato: Got it, thanks! I'll leave it as-is then.

PR #761, opened 2023-03-28
Clarify level differentiation

arewm:
I commented that differentiation between the levels wasn't intuitive in https://github.com/slsa-framework/slsa/pull/750/files#r1150745321. I think that if I just modify it to explicitly state that this is the basic provenance that it becomes clearer.

### DIFF
--- a/docs/spec/v1.0/levels.md
+++ b/docs/spec/v1.0/levels.md
@@ -85,7 +85,7 @@ n/a
 </section>
 <section id="build-l1">
 
-### Build L1: Provenance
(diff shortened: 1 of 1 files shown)


MarkLodato (approved):
:+1: I had the same thought. Let's do it!

MarkLodato:
Oops, I didn't mean to click merge so quickly without waiting for other opinions. Sorry about that. I can't easily roll back without a separate PR, so just comment on here if you disagree and we can easily revert.

PR #763, opened 2023-03-29
Remove obsolete reference to version selector.

MarkLodato:
Also the spec now encompasses both the levels and the formats, so remove that part as well.


lehors (approved):
Good catch.

PR #764, opened 2023-03-29
Provenance v1: raise prominence of extensions + other cleanups

MarkLodato:
- Fix type of `builder.version`
- Sort fields of Builder alphabetically
- Better document extension fields so that it is more prominent
- Remove stale RFC tags
- Update changelog to reflect latest changes.

The main change is to raise the prominence of extension fields so that implementers know that it's an option.

### DIFF
--- a/docs/provenance/v1.md
+++ b/docs/provenance/v1.md
@@ -279,11 +279,6 @@ Guidelines:
     to the build server, record the location in `externalParameters` as a URI
     and record the `uri` and `digest` in `resolvedDependencies`.
 
-> ⚠ **RFC:** We are particularly looking for feedback on this schema from
-> potential implementers. Does this model map cleanly to existing build systems?
-> Is it natural to identify and express the external parameters? Is anything
-> confusing or ambiguous?
-
 ### RunDetails
 
 [RunDetails]: #rundetails
@@ -385,8 +380,6 @@ the case where one signer generates attestations for more than one builder, as
 in the GitHub Actions example above. The field is REQUIRED, even if it is
 implicit from the signer, to aid readability and debugging. It is an object to
 allow additional fields in the future, in case one URI is not sufficient.
-
-> ⚠ **RFC:** Do we need more explicit guidance on how to choose a URI?
 
 > ⚠ **RFC:** Do we want/need to identify the tenant of the build system,
 > separately from the build system itself? If so, a single `id`
--- a/docs/provenance/v1.md
+++ b/docs/provenance/v1.md
@@ -526,10 +526,13 @@ Major refactor to reduce misinterpretation, including a minor change in model.
 -   Altered the model slightly to better align with real-world build systems,
     align with reproducible builds, and make verification easier.
 -   Grouped fields into `buildDefinition` vs `runDetails`.
--   Renamed, with slight changes in semantics:
-    -   `parameters` -> `externalParameters`
-    -   `environment` -> `systemParameters`
-    -   `materials` -> `resolvedDependencies`
+-   Renamed:
+    -   `parameters` -> `externalParameters` (slight change in semantics)
+    -   `environment` -> `systemParameters` (slight change in semantics)
+    -   `materials` -> `resolvedDependencies` (slight change in semantics)
+    -   `buildInvocationId` -> `invocationId`
+    -   `buildStartedOn` -> `startedOn`
+    -   `buildFinishedOn` -> `finishedOn`
 -   Removed:
     -   `configSource`: No longer special-cased. Now represented as
         `externalParameters`  + `resolvedDependencies`.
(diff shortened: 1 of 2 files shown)


docs/provenance/v1.md
```diff
@@ -361,6 +348,11 @@ Dependencies used by the orchestrator that are not run within the workload
 and that do not affect the build, but might affect the provenance generation
 or security guarantees.
 
+<tr id="builder.version"><td><code>version</code>
+<td>map (string→string)<td>
+
+Version numbers of components of the build platform.
```
laurentsimon, line 354: Thoughts on re-using `ArtifactDescriptor` which can record richer information? Or is `version` the only relevant information for the platform components / dependencies? 
Would it a bad idea to add a version field to `ArtifactDescriptor`?

TomHennen: FWIW I'm also curious about this field, it's not clear what the key is supposed to be?

Replacing `version` with some sort of ResourceDescriptor list (as Laurent is suggesting) might be more interesting?  But also would result in less information actually in the provenance or be pretty awkward to express inline.

I could also see how any change like that would be better off in a different PR (since this is just alphabetizing)?

TomHennen (approved):

docs/provenance/v1.md
MarkLodato: The intention is that `version` is just for string values. To record the actual artifacts used by the builder, there is `builderDependencies`.

I think such a change is outside the scope of this PR.

docs/provenance/v1.md
```diff
@@ -433,6 +418,26 @@ The timestamp of when the build completed.
 
 </table>
 
+### Extension fields
+
+[Extension fields]: #extension-fields
```
laurentsimon, line 423: so these fields are not in an "extension" namespace?

docs/provenance/v1.md
```diff
@@ -433,6 +418,26 @@ The timestamp of when the build completed.
...
+
+Implementations MAY add extension fields to any JSON object to describe
```
laurentsimon, line 425: What happens if the next version uses a field that is already used by someone else's extension?

MarkLodato: It will be ambiguous what the author meant. Are you requesting a change in the doc?

laurentsimon: That could help, but I suspect extension are buildType dependent so it should be OK?

docs/provenance/v1.md
laurentsimon: Yes, I find it hard to avoid collisions, in particular if more future versions of SLSA specs may add new fields. That makes it impossible to define extension names that don't collide with future versions. If we at least say that the extension should start with `ext_`, then this would be future-proof. The namespace is another way to do it.

docs/provenance/v1.md
laurentsimon: But also, telling people how to name extensions to avoid collision could be useful. Ideally they would be unique to builderID+buildType, e.g. `builderID_BuildType_NAME` although that would make the fields rather long :(

TomHennen: FWIW I'm imagining doing something like `<product|company name>_fieldName`.  E.g. `gerrit_codeReview`

lumjjb: In the same vein, i also do like what OCI does with using the domains like "com.google.xyz" that way

MarkLodato: PTAL at 3527436, which uses Tom's suggestion of `<vendor>_<name>`. I think that's a good compromise between full collision avoidance and having a concise, easy-to-read name.

I made it a MUST but I'm not confident about that.

Note: We originally suggested using full URLs as the field name to avoid collisions, but that's very verbose and also hard to parse due to the `.`s. A reverse domain name seems unnecessarily complicated to me.

docs/provenance/v1.md
joshuagl, line 354: This is a map of component->version, right? Can we somehow clarify that in the description?
```suggestion
Map of names of components of the build platform to their version.
```

joshuagl (approved):
I like the change to prefix extensions. LGTM, thanks!

docs/provenance/v1.md
laurentsimon: LGTM.

docs/provenance/v1.md
```diff
@@ -435,6 +420,25 @@ The timestamp of when the build completed.
...
+information that is not captured in a standard field. Guidelines:
+
+-   Extension fields MUST use names of the form `<vendor>_<fieldname>`, e.g.
+    `examplebuilder_isCodeReviewed`. This practice avoids field name collisions
+    by namespacing each vendor. Non-extension field names never contain an
+    underscore.
```
lehors, line 433: While this is a reasonable way to provide for what we cannot anticipate it should be noted that this kind of extension mechanism has some inherent flaws. Namely, experience shows that it becomes very difficult to ever change those names. For an example, consider all the x-* HTTP headers. See:

> Custom proprietary headers have historically been used with an X- prefix, but this convention was deprecated in June 2012 because of the inconveniences it caused when nonstandard fields became standard

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

MarkLodato: I downgraded this to a SHOULD to give us flexibility in the future, but I do think the convention is useful. Consensus seems to be that we'd prefer to avoid naming collisions, even at the expense of uglier names. In the future if we find this to be a problem, we can always come up with a better system. But the bigger problem seems like naming collisions now than ugly names in the future.

lehors (approved):

PR #770, opened 2023-03-30
Clarify there might be more than the Source track

JanZerebecki:
missing. The future versions page also lists other things, so clarify the wording to also encompass e.g. the Build track.

### DIFF
--- a/docs/spec/v1.0/levels.md
+++ b/docs/spec/v1.0/levels.md
@@ -32,7 +32,7 @@ hermeticity or completness of provenance -->
 
 > Note: The [previous version] of the specification used a single unnamed track,
 > SLSA 1–4. For version 1.0 the Source aspects were removed to focus on the
-> Build track. A Source track may be added in [future versions].
+> Build track. It is currently incomplete, see [future versions].
 
 ## Build track
 


docs/spec/v1.0/levels.md
```diff
@@ -32,7 +32,7 @@ hermeticity or completness of provenance -->
 
 > Note: The [previous version] of the specification used a single unnamed track,
 > SLSA 1–4. For version 1.0 the Source aspects were removed to focus on the
-> Build track. A Source track may be added in [future versions].
+> Build track. It is currently incomplete, see [future versions].
```
kpk47, line 35: I disagree with this change. While we expect SLSA to grow in the future, I don't think it's fair to call the spec incomplete. 

(Also nit: the "it" is ambiguous and could refer to either "Source aspects" or "Build track".)

JanZerebecki: The ambiguity is intentional, see description of the change. However if being less ambiguous is preferable that would work for me, too.

olivekl: Are you trying to say that Source is incomplete, or Build? Grammatically, this sentence means Build is incomplete, not Source. 

If you want to say both, then it should be "they." But I disagree with that statement: the Build track is complete for v1.0. It may be expanded later, but that doesn't mean it's incomplete. I.e., a house that gets a third story added on wasn't incomplete when it was a smaller, finished building.

MarkLodato:
Since there doesn't seem to be consensus on this change, and we have not gotten any other feedback about the confusion, I'll close this PR to keep our list of open PRs clean.

PR #775, opened 2023-03-30
Add recommended crypto algorithms for DSSE.

kpk47:
This closes out the TODO on the Attestation Model page.

I chose these algorithms because we recommend SHA-256 elsewhere in the spec and ECDSA over P-256 has the same security strength. I'm open to other suggestions.

### DIFF
--- a/docs/attestation-model.md
+++ b/docs/attestation-model.md
@@ -90,7 +90,7 @@ recognize that other choices MAY be necessary in various cases.
 
 | Component | Recommendation |
 | --- | --- |
-| Envelope | **[DSSE]** (ECDSA over NIST P-256 and SHA-256) |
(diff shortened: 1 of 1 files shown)


laurentsimon (approved):
You may also say users may use stronger curves with the corresponding SHA

asraa (approved):
Yes, recommending this one is probably better than others because its more widely supported and prevents people from potentially implementing something themselves (something like ed25519 may not be officially supported or integrated in base crypto libs). +1 for stronger if chosen (so at least that many bits of security)

ECDSA sigs are malleable (https://www.derpturkey.com/inherent-malleability-of-ecdsa-signatures/) but I'd figure it would be somewhat difficult to construct a validly typed message because we're working with PAE

Issue #717, opened 2023-03-22
SLSA v1: Define tenant, admin, control plane, and untrusted build as it relates to a build model


arewm:
Can we provide additional clarity around what it means for provenance completeness to be best effort? Is this just best effort because we are assuming that up to L3 there are just well-intentioned users? Is this best-effort of completeness the driving force behind other requirements as well (i.e. isolation) and the specific absence of additional requirements to around checking/enforcing against tampering in the build system?

I started to include statement around best effort completeness in #700 but I ended up pulling them out as the conversation was distracting from the specific improvement made around isolation. 

@Nikokrock also touched on this bit in #699.

MarkLodato:
I think "well-intentioned" is not the right term. It's misleading.

Rather, I wanted to say that the any sort of external influence on the build that violates the Isolated principles MUST NOT be the fault of the build system. In other words, a build that does not do anything "wrong" MUST be isolated, but that a "wrong thing" MAY exist and a producer need not protect against it.

arewm:
That makes sense, we can iterate on wordings around violations in the isolation principles MUST NOT be the fault of the build system.

What does this require of the build system? Would it require that the build system to perform some form of self-validation of its integrity before being able to claim a build L3? For example, in a build system leveraging Tekton, would the build system need to ensure that producers do not modify the Pipeline/task definitions?

MarkLodato:
> What does this require of the build system? Would it require that the build system to perform some form of self-validation of its integrity before being able to claim a build L3? For example, in a build system leveraging Tekton, would the build system need to ensure that producers do not modify the Pipeline/task definitions?

Sorry, I don't understand the question. Is this with respect to how much effort the build system must go through to prevent a user from violating the spirit of the requirement? I don't know what you mean by "self-validation of its integrity". As for Tekton, I'm not very familiar with it, but if they are akin to GitHub Actions workflows or actions, it seems like we'd need to support custom ones?

MarkLodato:
> Are there other requirements imposed on the build system (other than isolation) that are "okay" to be violated as long as they MUST NOT be the fault of the build system?

Good question. No. The only other requirement is that the [provenance is non-forgeable](https://slsa.dev/spec/v1.0/requirements#provenance-non-forgeable), and that MUST NOT be violated by builds.

> If a build system can be customized by users, would it be the responsibility of the build system to only allow certain modifications?

[...]

PR #778, opened 2023-03-31
Add 'control plane' to the terminology page

kpk47:


docs/spec/v1.0/terminology.md
```diff
@@ -60,6 +60,7 @@ describing this whole process.
 | Platform | System that allows tenants to run builds. Technically, it is the transitive closure of software and services that must be trusted to faithfully execute the build. It includes software, hardware, people, and organizations.
 | Build | Process that converts input sources and dependencies into output artifacts, defined by the tenant and executed within a single environment on a platform.
 | Steps | The set of actions that comprise a build, defined by the tenant.
+| Control plane | Build system component that orchestrates each independent build execution and produces provenance.
```
MarkLodato, line 63:
```suggestion
| Control plane | Build system component that orchestrates each independent build execution and produces provenance. It is trusted an outside the control of the tenant.
```

MarkLodato (approved):
I'm OK submitting this since it's strictly better than before, but the model build model needs further revision as per #717. For example, the text and diagram should probably say "control plane" where it says "platform" now, we should define "tenant", etc.

PR #789, opened 2023-04-03
Avoid MUST in "verifying artifacts" and "distributing provenance"

### LINKED ISSUE #786: "Verifying artifacts" is unclear how it relates to levels
It is currently unclear how [verifying artifacts](https://slsa.dev/spec/v1.0/verifying-artifacts) relates to the levels and requirements.

- It uses the term "MUST" without indicating what it is required for. Is this for the build track? For anything SLSA? Or is this just guidelines
- It uses "L1", "L2", and "L3" without any reference to tracks. If this is the build track, why is it not listed in the requirements? (Also I thought we decided verification is *not* a requirement for the build track?) If it is not, what is it?

I suggest removing the term "MUST" and references to levels from the doc, and turning it into guidelines or recommendations instead.

MarkLodato:
As discussed at the 2023-04-03 Specification Meeting, we agreed to downgrade the requirements (MUST) in "verifying artifacts" and "distributing provenance" to be recommendations (SHOULD) since we have not yet sufficiently defined what a "SLSA-conformant system" means. Hopefully we can address this in a future revision of the specification.

This leaves one use of "MUST", namely "Producers MUST publish attestations in at least one place", because that is a requirement from "producing artifacts".

Fixes #786.

### DIFF
--- a/docs/spec/v1.0/verifying-artifacts.md
+++ b/docs/spec/v1.0/verifying-artifacts.md
@@ -5,14 +5,13 @@ description: SLSA uses provenance to indicate whether an artifact is authentic o
 
 SLSA uses provenance to indicate whether an artifact is authentic or not, but
 provenance doesn't do anything unless somebody inspects it. SLSA calls that
-inspection **verification**, and this page describes how to verify artifacts and
-their SLSA provenance.
+inspection **verification**, and this page describes recommendations for how to
+verify artifacts and their SLSA provenance.
 
-This page is divided into several sections. The first discusses choices
-software distribution and/or deployment system implementers must make regarding
-verifying provenance. The second describes how to set the expectations used to
-verify provenance. The third describes the procedure for verifying an artifact
-and its provenance against a set of expectations.
+This page is divided into several sections. The first discusses choices for
+where provenance verification can happen. The second describes how to set the
+expectations used to verify provenance. The third describes the procedure for
+verifying an artifact and its provenance against a set of expectations.
 
 ## Architecture options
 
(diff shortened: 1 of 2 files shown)


kpk47 (approved):

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -277,7 +274,7 @@ order to mitigate [threat "C"].
 
 In our threat model, the adversary has ability to invoke a build and to publish
 to the registry but not to write to the source repository, nor do they have
-insider access to any trusted systems. Expectations MUST be sufficient to detect
+insider access to any trusted systems. Expectations SHOULD be sufficient to detect
```
lehors, line 277: Is that a requirement on the build system in disguise? Like the build should provide enough information to detect...

lehors: Either it is and it doesn't belong here or it's not and the use of an RFC 2119 SHOULD is odd.

MarkLodato: Do you mean just this sentence (and line 98 above) or all uses of SHOULD in this doc?

My interpretation is that this is already covered by the requirements:

- [Producer](https://slsa.dev/spec/v1.0/requirements#follow-a-consistent-build-process): "MUST build their artifact in a consistent manner such that verifiers can form expectations about the build process."
- [Build system](https://slsa.dev/spec/v1.0/requirements#provenance-unforgeable): "External parameters MUST be fully enumerated [in the provenance]."
- Verifiers: SHOULD [compare the provenance to expectations that are] sufficient to detect ... (this sentence)

Should we be more explicit to tie "consistent manner", "external parameters", "expectations", and "sufficient to detect..." together? If so, do you have recommendations on how?

lehors: Ok, I have now realized that this is about expectations set by the verified not the producer. The confusion still lingers but this isn't what this PR is about so forget my suggestion.

lehors (approved):

PR #715, opened 2023-03-22
SLSA v1: Update supply chain threats diagram

kpk47:
- Move threat C to the Source Track.
- Swap threats D and E. Now the division is ABC == Source, D == Dependencies, EFGH == Build.

Addresses #708 and #647.

PR #790, opened 2023-04-03
Fix diagrams to use consistent style

MarkLodato:
This fixes the change from PR #715, which had started from an older source figma file. The new version fixes a typo ("Depedencies"), uses the same style as the other supply chain diagrams, and fixes the aspect ratio of the threats diagram.

Preview links:

- https://deploy-preview-790--slsa.netlify.app/spec/v1.0/verifying-artifacts#how-to-verify
- https://deploy-preview-790--slsa.netlify.app/spec/v1.0/threats-overview


lehors (approved):
LGTM

PR #796, opened 2023-04-04
Fix references to threats on verifying-systems.md

MarkLodato:
- Fix references to threats D and E, whose labels have swapped.
- Use relative links instead of absolute.


lehors (approved):
LGTM

PR #800, opened 2023-04-04
SLSA v1 RC2

MarkLodato:
This PR creates SLSA v1.0 RC2. It is best viewed as two separate commits:

1.  94cdd12: copy files verbatim from v1.0
    -   Can be verified with `git show 94cdd12 --find-copies-harder`.
2.  960ef54: update version references within the files


docs/_data/versions.yml
```diff
@@ -21,6 +21,9 @@ spec:
     v1.0-rc1:
       name: Version 1.0 RC1
       status: Candidate
```
kpk47, line 23: Should RC1 move to "Retired" status now?

lehors (approved):
LGTM

joshuagl (approved):
Looks great, thanks.

`git diff --no-index --find-copies-harder docs/spec/v1.0 docs/spec/v1.0-rc2` is the way to review locally

di (approved):

TomHennen (approved):

PR #807, opened 2023-04-06
Clarify insider threats on "About SLSA"

### LINKED ISSUE #805: Who is SLSA - Software Producers needs a bit more clarity.
I hope this isn't pedantic but when reading https://slsa.dev/spec/v1.0-rc2/about#who-is-slsa-for:

> Software producers, such as a development team or open source maintainers. SLSA gives you protection against insider risk and tampering along the supply chain to your consumers, increasing confidence that the software you produce reaches your consumers as you intended.

I feel like this should be made clearer that it give you protection against tampering, including insider risk. I think here it's a bit unclear. I'm trying to think through when you would have a supply chain compromise caused by insider risk that wasn't tampering along the supply chain.

This came up in the 1.0 release announcement draft. I think it could lead to confusion due to 1.0 being purely about the build, and so the majority of risk that SLSA would be protecting against would be tampering with the build. If that's the case, what's the insider risk if the compromise is coming from an external build system?

 I think we should either have something like:

Remove it completely:
> SLSA gives you protection against tampering along the supply chain

Reword it a bit:
> SLSA gives you protection against tampering along the supply chain to your consumers, including insider risk, thereby increasing confidence...

MarkLodato:
Previously the "Who is SLSA for? / Software producers" bullet described producers as individuals rather than organizations, and said "insider risk and tampering" as though they were distinct threats. Similarly, the "What SLSA doesn't cover / Developer trust" bullet talked about the threat of individuals writing malicious code, which seems to be exactly the kind of "insider risk and tampering" threat described above.

This change fixes that by distinguishing between individuals and organizations. SLSA addresses the threat of individuals subverting the organization's goals, but does not address the organization having bad goals.

Addresses most of #805. There are still pending discussion around "insider" and "producer trust."

### DIFF
--- a/docs/spec/v1.0/about.md
+++ b/docs/spec/v1.0/about.md
@@ -131,7 +131,7 @@ important to consider together with SLSA such as:
     source code followed secure coding practices.
 -   Producer trust: SLSA does not address organizations that intentionally
     produce malicious software, but it can reduce insider risks within an
-    organization you trust. SLSA's Build track protects against tampering during
+    organization you trust. SLSA's Build Track protects against tampering during
     or after the build, and [future SLSA tracks](future-directions.md) intend to
     protect against unauthorized modifications of source code and dependencies.
 -   Transitive trust for dependencies: the SLSA level of an artifact is
--- a/docs/spec/v1.0/about.md
+++ b/docs/spec/v1.0/about.md
@@ -85,7 +85,7 @@ infrastructure for software.
 
 **Software producers**, such as an open source project, a software vendor, or a
 team writing first-party code for use within the same company. SLSA gives you
-protection against tampering along the supply chain to your consumers, thereby
+protection against tampering along the supply chain to your consumers, both
 reducing insider risk and increasing confidence that the software you produce
 reaches your consumers as you intended.
 
@@ -133,7 +133,7 @@ important to consider together with SLSA such as:
     produce malicious software, but it can reduce insider risks within an
     organization you trust. SLSA's Build track protects against tampering during
     or after the build, and [future SLSA tracks](future-directions.md) intend to
-    protect against unauthorized modifications of source code and dependenies.
+    protect against unauthorized modifications of source code and dependencies.
 -   Transitive trust for dependencies: the SLSA level of an artifact is
     independent of the level of its dependencies. You can use SLSA recursively to
     also judge an artifact's dependencies on their own, but there is
--- a/docs/spec/v1.0/about.md
+++ b/docs/spec/v1.0/about.md
@@ -131,11 +131,9 @@ important to consider together with SLSA such as:
     source code followed secure coding practices.
 -   Producer trust: SLSA does not address organizations that intentionally
     produce malicious software, but it can reduce insider risks within an
-    organization you trust. Currently SLSA's build track protects against
-    tampering during or after the build; a
-    [future SLSA Source track](future-directions#source-track) will also protect
-    against tampering before the build, particularly modifications to source
-    code.
+    organization you trust. SLSA's Build track protects against tampering during
+    or after the build, and [future SLSA tracks](future-directions.md) intend to
+    protect against unauthorized modifications of source code and dependenies.
 -   Transitive trust for dependencies: the SLSA level of an artifact is
     independent of the level of its dependencies. You can use SLSA recursively to
     also judge an artifact's dependencies on their own, but there is
--- a/docs/spec/v1.0/about.md
+++ b/docs/spec/v1.0/about.md
@@ -83,10 +83,11 @@ For more exploration of this analogy, see the blog post
 In short: everyone involved in producing and consuming software, or providing
 infrastructure for software.
 
-**Software producers**, such as a development team or open
-source maintainers. SLSA gives you protection against insider risk and tampering
-along the supply chain to your consumers, increasing confidence that the
-software you produce reaches your consumers as you intended.
+**Software producers**, such as an open source project, a software vendor, or a
+team writing first-party code for use within the same company. SLSA gives you
+protection against tampering along the supply chain to your consumers, thereby
+reducing insider risk and increasing confidence that the software you produce
+reaches your consumers as you intended.
 
 **Software consumers**, such as a development team using open source packages, a
 government agency using vendored software, or a CISO judging organizational
@@ -128,10 +129,13 @@ important to consider together with SLSA such as:
 
 -   Code quality: SLSA does not tell you whether the developers writing the
     source code followed secure coding practices.
--   Developer trust: SLSA cannot tell you whether a developer is
-    intentionally writing malicious code. If you trust a developer to write
-    code you want to consume, though, SLSA can guarantee that the code will
-    reach you without another party tampering with it.
+-   Producer trust: SLSA does not address organizations that intentionally
+    produce malicious software, but it can reduce insider risks within an
+    organization you trust. Currently SLSA's build track protects against
+    tampering during or after the build; a
+    [future SLSA Source track](future-directions#source-track) will also protect
+    against tampering before the build, particularly modifications to source
+    code.
 -   Transitive trust for dependencies: the SLSA level of an artifact is
     independent of the level of its dependencies. You can use SLSA recursively to
     also judge an artifact's dependencies on their own, but there is


docs/spec/v1.0/about.md
```diff
@@ -128,10 +129,13 @@ important to consider together with SLSA such as:
...
-    intentionally writing malicious code. If you trust a developer to write
-    code you want to consume, though, SLSA can guarantee that the code will
-    reach you without another party tampering with it.
+-   Producer trust: SLSA does not address organizations that intentionally
+    produce malicious software, but it can reduce insider risks within an
+    organization you trust. Currently SLSA's build track protects against
+    tampering during or after the build; a
+    [future SLSA Source track](future-directions#source-track) will also protect
+    against tampering before the build, particularly modifications to source
+    code.
```
joshuagl, line 138: Nit: it may be clearer if we say "unauthorised modification"?
```suggestion
    against tampering before the build, particularly unauthorized modification
    of source code.
```
Also, should we mention dependencies as a potential part of the source track? They seem to be top-of-mind as a risk for many.

joshuagl (commented):
Thanks for addressing this so quickly.

docs/spec/v1.0/about.md
```diff
@@ -128,10 +129,13 @@ important to consider together with SLSA such as:
 
 -   Code quality: SLSA does not tell you whether the developers writing the
     source code followed secure coding practices.
--   Developer trust: SLSA cannot tell you whether a developer is
-    intentionally writing malicious code. If you trust a developer to write
-    code you want to consume, though, SLSA can guarantee that the code will
-    reach you without another party tampering with it.
+-   Producer trust: SLSA does not address organizations that intentionally
+    produce malicious software, but it can reduce insider risks within an
+    organization you trust. Currently SLSA's build track protects against
```
arewm, line 134: Nit: Production of malicious software isn't necessarily intentional.
```suggestion
-   Producer trust: While SLSA does not address trust that organizations
    won't produce malicious software, it can reduce insider risks within an
    organization you trust. Currently SLSA's build track protects against
```

arewm (changes requested):

docs/spec/v1.0/about.md
```diff
@@ -83,10 +83,11 @@ For more exploration of this analogy, see the blog post
...
 infrastructure for software.
 
-**Software producers**, such as a development team or open
-source maintainers. SLSA gives you protection against insider risk and tampering
-along the supply chain to your consumers, increasing confidence that the
-software you produce reaches your consumers as you intended.
+**Software producers**, such as an open source project, a software vendor, or a
+team writing first-party code for use within the same company. SLSA gives you
+protection against tampering along the supply chain to your consumers, thereby
+reducing insider risk and increasing confidence that the software you produce
```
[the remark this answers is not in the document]
arewm: I see it as a fuzzy/fine line. It reduces risk and increase confidence that the software that you produce reaches your consumers as intended. It is "internal risk" whether by internal compromise or insider threat as I suspect the supply chain attack vectors are harder to exploit by someone who is not internal.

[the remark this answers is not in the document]
lehors: I think @mlieberman85 has a point. It does read as if this is specifically against insider risk. Should this read something like "reducing risk, including insider's, and increasing..."?

docs/spec/v1.0/about.md
MarkLodato: Will apply the suggestion. Isn't dependencies covered by the next bullet?

olivekl: The next bullet is more about transitive dependencies, right? Not about risk to your immediate dependencies, but that a SLSA level doesn't take into account your dependency tree?

MarkLodato: Went with a variant of @olivekl's suggestion. PTAL

docs/spec/v1.0/about.md
```diff
@@ -131,11 +131,9 @@ important to consider together with SLSA such as:
...
 -   Producer trust: SLSA does not address organizations that intentionally
     produce malicious software, but it can reduce insider risks within an
-    organization you trust. Currently SLSA's build track protects against
-    tampering during or after the build; a
-    [future SLSA Source track](future-directions#source-track) will also protect
-    against tampering before the build, particularly modifications to source
-    code.
+    organization you trust. SLSA's Build track protects against tampering during
+    or after the build, and [future SLSA tracks](future-directions.md) intend to
+    protect against unauthorized modifications of source code and dependenies.
```
joshuagl, line 136:
```suggestion
    protect against unauthorized modifications of source code and dependencies.
```

docs/spec/v1.0/about.md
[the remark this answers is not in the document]
MarkLodato: Went with "both". Does that address your concern?

docs/spec/v1.0/about.md
MarkLodato: Sorry, I'm not sure I follow. What do you mean by unintentional malicious software? Is that not covered by "Code quality" above? My thinking is that it's valuable to call out both unintentional vulnerabilities (currently called "code quality") and intentional badness (currently called "producer trust").

docs/spec/v1.0/about.md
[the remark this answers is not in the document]
mlieberman85: Yes! Sorry if it came off as pedantic. I think within context of the page it makes sense but I found that in the announcement it felt a bit unclear.

mlieberman85 (approved):
LGTM

docs/spec/v1.0/about.md
```diff
@@ -128,10 +129,11 @@ important to consider together with SLSA such as:
 
 -   Code quality: SLSA does not tell you whether the developers writing the
     source code followed secure coding practices.
--   Developer trust: SLSA cannot tell you whether a developer is
-    intentionally writing malicious code. If you trust a developer to write
-    code you want to consume, though, SLSA can guarantee that the code will
-    reach you without another party tampering with it.
+-   Producer trust: SLSA does not address organizations that intentionally
+    produce malicious software, but it can reduce insider risks within an
+    organization you trust. SLSA's Build track protects against tampering during
```
olivekl, line 134:
```suggestion
    organization you trust. SLSA's Build Track protects against tampering during
```

olivekl (approved):
tiny formatting nit, LGTM!

lehors (approved):
LGTM

docs/spec/v1.0/about.md
lehors: I like @olivekl's attention to these details!

docs/spec/v1.0/about.md
arewm: I think that there are lots of nuances in the comment which is why I phrased it as a nit.

I would consider code quality to catch common mistakes like bugs and vulnerabilities in code more than malicious code. 

I think that the nit is that while _someone_ would ultimately be responsible for intentionally introducing malicious code, that individual/entity might not always be the producer/organization itself.

If the organization is a private company, then there could be potential risk of a compromised account being used to introduce malicious code (intentional from an adversary, unintentional from a company). If the organization is some open source entity, it may be possible for a "random" contributor to intentionally add malicious code that isn't identified (yes, I admit, this would likely fall at least partially within code quality and/or future directions).

I feel like we are trying to establish that SLSA cannot _create_ confidence in a producer/artifact in isolation. The SLSA specification is a tool that producers can use to help strengthen consumers' trust in an already trusted producer (as it supports the producer themselves to adhering to their best practices, guidelines, etc. by mitigating supply chain vulnerabilities).

lehors: What @arewm is saying sounds right to me and I'm supportive of his proposed change. I agree with him on the origin of the possible threat which is why I also responded to the point about "insider risk" @mlieberman85 raised. This said, this is merely the "about" section of the spec and although we should make sure it is not misrepresenting anything we have to accept that there is only so much we can address in a concise way. That's why I approved the PR as it stands anyway.

MarkLodato: Since there are enough approvals and I think everyone agrees that this is a strict improvement than before, I'm going to submit this as-is but not mark the underlying issue fixed. I do want to get agreement on this point because it gets at the heart of the scope of SLSA, so it's an important thing for us as an organization to agree upon.

I'll continue the discussion in the issue.

Issue #787, opened 2023-04-03
v1: Clarify differentiation between build system and build service


arewm:
In https://github.com/slsa-framework/slsa/pull/779#discussion_r1154764715, it was raised that both build system and build service are used without a clear distinction. This distinction should be clarified and the specification should be reviewed to ensure that the distinction is consistently used.

My comments on the distinction from the thread:

> There isn't clear consistency across the requirements page. I see the build system as a generic term for the trusted boundary producing an artifact. It may be a service (GitHub Actions, etc.) or it may be a non-service (individual's machine).
> 
> Since the isolation requirements for L2 are a build service, it is consistent to start talking about a build service at L2 requirements in provenance.

MarkLodato:
Let's consider this as part of #714.

Issue #714, opened 2023-03-21
Review use of terminology associated with the build word


Nikokrock:
I have re-read a bit the spec looking specially to terminology we are using. Here are my notes. English is not my native language so some of my comments may not be relevant.

- **Build process**: in terminology this is the definition of **build**
- **Build**: most of the time refers to the instance of a build process
- **Build service**: _not defined_. Sometime refers to build platform, sometimes to “control plane” of the build platform
- **Builder**: _not defined_. Equivalent to build tools ?
- **Build system**: __not defined__. Same as build platform ?
- **Build environment**: same as **Build Executor** (though less tainted towards an implementation from my point of view) 
- **Build tools**: _not defined_. Equivalent to builder ?
- **Build model**: _not defined_ but probably not needed
- **Build platform**: most of the time **Build service** is used but platform is used in terminology
- **Build recipe**: Equivalent to build steps ?
- **Build infrastructure**: Equivalent to control plane ?
- **Build configuration source**: Equivalent to build external parameters ?
- **Build worker**: Should we use **Build environment** ?
- **Build cache**: Do we need to define cache in the spec. Looks like it’s an implementation “detail”
- **Builder model**: Same as build model ?
- **Build executor**: Equivalent to build environment ?
- **Build definition**: this is just a build external parameter (most of time as a source of the the project to build)
- **Control Plane**: __not defined__. Probably need definition in terminology ?

olivekl:
Thank you, @Nikokrock, this is really useful!

MarkLodato:
From #787 by @arewm:

> In [#779 (comment)](https://github.com/slsa-framework/slsa/pull/779#discussion_r1154764715), it was raised that both build system and build service are used without a clear distinction. This distinction should be clarified and the specification should be reviewed to ensure that the distinction is consistently used.
> 
> My comments on the distinction from the thread:
> 
> > There isn't clear consistency across the requirements page. I see the build system as a generic term for the trusted boundary producing an artifact. It may be a service (GitHub Actions, etc.) or it may be a non-service (individual's machine).
> > Since the isolation requirements for L2 are a build service, it is consistent to start talking about a build service at L2 requirements in provenance.

chtiangg:
[...]

Issue #706, opened 2023-03-20
SLSA v1: Explain that you must define a security model for your build system 


MarkLodato:
The spec should (somewhere) explain that the build system implementer must define a security model for the build system, with clear boundaries, actors, and interfaces. This will then inform how to fix the system and implement provenance. Maybe on the provenance page?

Related discussion: https://github.com/slsa-framework/slsa/pull/700#discussion_r1142073167

/cc @arewm

arewm:
In order to have a reasonable guarantee about the authenticity of the provenance (i.e. L3), all build activity needs to happen within the trust boundary and the threat model needs to ensure that appropriate mitigations have been made with all entities outside the trust boundary (as needed).

arewm:
I started to fix this in #810 but then the PR exploded to unify on `platform`. In the PR, however, I did call out whether the wording that I added should also be added to system verification. 
> Does this MUST mean that we should add this to verifying-systems? [[ref](https://github.com/slsa-framework/slsa/pull/810#discussion_r1160201391)]

MarkLodato:
> Does this MUST mean that we should add this to verifying-systems? [[ref](https://github.com/slsa-framework/slsa/pull/810#discussion_r1160201391)]

Yeah, I think so, now that you say it. This is similar https://github.com/slsa-framework/slsa/pull/789#discussion_r1156334736 by @lehors. I think his argument was that all requirements on the build platform should be in requirements.md.

arewm:
Should the change then be to include the `MUST` wording and to include it in verifying systems or to change it to `SHOULD` (and still include it in verifying systems)?

MarkLodato:
I lean towards SHOULD.

MUST means that a build platform failing to define such a security model _cannot_ be considered SLSA conformant. But this seems more like guidelines than a strict requirement. Could there be a situation where a build platform does not do this and yet one would still consider it SLSA conformant? I imagine so. Also if we make it a MUST, we'd probably need guidelines on how to verify it.

arewm:
SHOULD + noting in verifying systems or without the note?

I support having it as a SHOULD as it is a best practice and can be used to inform other aspects of the specification. The ideas that I have about when this might not be needed seem to revolve around simpler systems that can make determinations like this without any formal modeling.

MarkLodato:
I'd say in two or three places:

[...]

Issue #802, opened 2023-04-05
Future Suggestion: Build Systems Track


mattfarina:
The "build track" appears to look at things from the standpoint of a user of a build system. In the [future directions](https://slsa.dev/spec/v1.0-rc2/future-directions) sections it covers expanding the "build track" and a "source track".

While these are great, there is a track that's missing. A build systems track that covers the security of a build system. The material on [verifying build systems](https://slsa.dev/spec/v1.0-rc2/verifying-systems) is a nice start. But, there are some limitations with it:

1. In many cases people aren't empowered to do a self assessment of a build system.
2. Folks can't see, at a glance, how a provider does. We are accustomed to being able to view things at a glance.
3. Without something more formal you can't have a compliance program. Given the attack vectors that can come through a build service, it would be great to have a 3rd party compliance program for them. To do that you need something to comply against. #515 showed me that folks were thinking of compliance.

A build system could be a complete mess. Very much insecure. Or, it could be really good. At present, there is no way to know and nothing in the future plans to solve for it.

I came to this when I noticed the Common Requirements have been removed. This is already captured in #674.

TheFoxAtWork:
This is a good call out @mattfarina. Given that a significant portion of open source projects rely heavily on managed build systems, conveying the security of build systems to users is important, akin to how 3P Cloud Providers convey the security of their services to their users.
When we (CNCF TAG Security) were developing the [Supply Chain Security Whitepaper](https://github.com/cncf/tag-security/blob/main/supply-chain-security/supply-chain-security-paper/sscsp.md), this was a point of long discussion that is often overlooked when we talk about the supply chain - the security controls, constraints, and considerations in the design and architecture of a given build system. This is why there is an entire section of the [security of a build infrastructure](https://github.com/cncf/tag-security/blob/main/supply-chain-security/supply-chain-security-paper/sscsp.md#build-infrastructure) within the paper.  Its not perfect, but by approaching it from the perspective owning the breadth of the build system, we were able to identify areas where weakness in the build infrastructure itself could interfere with the integrity of the artifacts being built.

CC: @jonmuk who was instrumental to our development of the paper.

kpk47:
Thank you for bringing this up! @JoshuaMulliken and I are still working on #515 but got a bit sidetracked working on the v1.0-RC2 spec release. Now that the release is ready, we'll be focusing on the conformance program more.

[...]

PR #810, opened 2023-04-06
Unify on build platform

arewm:
Quote taken from hidden note in `terminology.md`.
> Using terms such as "platform," "system," and "service" interchangeably can cause
> confusion.
> 
> For consistency: Whenever possible, default to "platform." "Service" is reserved
> for use within the requirements and "system" is reserved for use to describe both
> software and tools internal to a platform. External self-sescribed services and
> systems can continue to be called by these terms.

Addresses #787 and by extension, some of #714
(No longer addresses #706)


arewm:
My only focus on clarifications were the terminology, provenance, and requirements pages. I haven't yet done a review of the other instances of the terms to identify further locations for clarification.

docs/spec/v1.0/terminology.md
```diff
@@ -72,7 +72,9 @@ describing this whole process.
 
 | Primary Term | Description
 | --- | ---
-| Platform | System that allows tenants to run builds. Technically, it is the transitive closure of software and services that must be trusted to faithfully execute the build. It includes software, hardware, people, and organizations.
+| Platform | The transitive closure of software and services that must be trusted to faithfully execute builds in the build system. It includes software, hardware, people, and organizations.
+| System | Collection of processes and tools which allows tenants to run builds. This includes both the trusted control plane as well as the processes that execute untrusted tenant build steps.
```
arewm, line 76: I think that this should be included on the image, but I don't have a way to do that off hand. My thoughts were to take the current platform as the system and make the platform larger, including a "people" icon labeled as "people and organizations".

docs/provenance/v1.md
```diff
@@ -50,6 +50,12 @@ The model is as follows:
     record the provenance. (Note: The same model can be used for platform-less
     or single-tenant build systems.)
 
+    -   The build platform implementer MUST define a security model for the build
```
arewm, line 53: Does this MUST mean that we should add this to verifying-systems?

docs/spec/v1.0/terminology.md
MarkLodato, line 76: I don't understand the difference, and I suspect other readers will not either. My strong preference is to pick one or the other and use that consistently. In general we should strive to minimize the number of special terms that the reader needs to understand.

Is there a need to differentiate between these concepts? If so, could you point to an example? In that case, can we just spell out the alternate concept, similar to how we approached #739?

arewm: The difference between the two is that the platform contains the individuals/entities while the system does not. I changed references in the requirements document from `system` to `platform`, but I defined `system` as it is a step towards describing the difference between a system and a service.

We could remove system and go from platform to service if we still want to define service here. I guess a service would still be a transitive closure of all individuals and organizations (even more so since there would be SREs involved too).

Is your strong preference to remove the added system _and_ service definitions or just the system? If the only place that we use `build service` is in the requirements, I can just try to only clarify the definition there.

arewm: One example of a location where the differentiation is needed is in this table: https://github.com/slsa-framework/slsa/blob/main/docs/get-started.md?plain=1#L114

If we don't define a build service in the terminology, then this would only be defined in the requirements or we would have to change to something like  `No hosted build platform`.

MarkLodato: Sorry for the ambiguity. I meant just `platform` vs `system`. It seems like we previously were using these two terms interchangeably, whereas this PR makes a distinction that I think is quite subtle. Would it work to use a single term for the combined concept, and then spell out the (hopefully rare!) cases where we do need to make a distinction? This is how #739 handled `producer`, which I think improved readability.

I'm fine keeping `service` for now, and adding it to the Terminology page seems like a good idea since it's used in multiple places. Perhaps we could eliminate it, too, saying "hosted build platform/service" as you mentioned, but I'm not confident about that. Would help to see it, but leaving that to a future PR is fine IMO.

arewm: I just pushed two changes to unify from service and system to platform. Please re-review.

docs/_data/nav/v1.0.yml
```diff
@@ -56,19 +56,19 @@
...
+    description: Detailed technical requirements for distributing provenance, intended for platform implementers and software distributors
 
   - title: Verifying artifacts
     url: /spec/v1.0/verifying-artifacts
-    description: Guidance for verifying software artifacts and their SLSA provenance, intended for system implementers and software consumers
+    description: Guidance for verifying software artifacts and their SLSA provenance, intended for platform implementers and software consumers
 
-  - title: Verifying build systems
+  - title: Verifying build platforms
     url: /spec/v1.0/verifying-systems
```
MarkLodato, line 70: Q: Should we rename this page (with a redirect from the old name) in a future PR?

docs/provenance/v1.md
```diff
@@ -213,7 +219,7 @@ values to aid verification. The same field name SHOULD NOT be used for both
 `externalParameters` and `systemParameters`.
```
MarkLodato, line 219: Not for this PR, but I do wonder if we should rename this to `internalParameters` or `trustedParameters`? It's unfortunate that we now have the word "system" as part of the schema 🫤.

docs/community.md
```diff
@@ -27,7 +27,7 @@ practical standards, tools, and documentation to reduce software supply chain
 risk in the real world. To succeed, we rely on contributors from a variety
 organizations to help us improve. Whether that's reporting successes or
 challenges, contributing changes to the specification or documentation, or
-developing tooling and services, we welcome your contributions.
+developing tooling and platforms, we welcome your contributions.
```
MarkLodato, line 30: I think in some cases like this, the word "services" actually reads better.

docs/spec/v1.0/future-directions.md
```diff
@@ -25,9 +25,9 @@ following requirements, which **may or may not** be part of a future Build L4:
 -   Pinned dependencies, which guarantee that each build runs on exactly the
     same set of inputs.
 -   Hermetic builds, which guarantee that no extraneous dependencies are used.
--   All dependencies listed in the provenance, which enables downstream systems
+-   All dependencies listed in the provenance, which enables downstream platforms
```
MarkLodato, line 28: nit: I don't think this was intended as "build system". Maybe "verifiers"?

```suggestion
-   All dependencies listed in the provenance, which enables downstream verifiers
```

docs/spec/v1.0/future-directions.md
```diff
@@ -25,9 +25,9 @@ following requirements, which **may or may not** be part of a future Build L4:
...
     to recursively apply SLSA to dependencies.
--   Reproducible builds, which enable other systems to corroborate the
+-   Reproducible builds, which enable other platforms to corroborate the
```
MarkLodato, line 30: nit: Spelling out "build platforms" I think makes it a bit more clear what we mean here.

```suggestion
-   Reproducible builds, which enable other build platforms to corroborate the
```

docs/spec/v1.0/levels.md
```diff
@@ -129,7 +129,7 @@ SLSA---other than tamper protection---without changing their build workflows.
 </section>
 <section id="build-l2">
 
-### Build L2: Build service
+### Build L2: Build platform
```
MarkLodato, line 132: I think a bare "build platform" is too ambiguous here.

```suggestion
### Build L2: Hosted build platform
```

docs/provenance/v1.md
```diff
@@ -524,7 +530,7 @@ The following fields from v0.2 are no longer present in v1.0:
 Major refactor to reduce misinterpretation, including a minor change in model.
 
 -   Significantly expanded all documentation.
--   Altered the model slightly to better align with real-world build systems,
+-   Altered the model slightly to better align with real-world build platforms,
     align with reproducible builds, and make verification easier.
 -   Grouped fields into `buildDefinition` vs `runDetails`.
 -   Renamed:
```
MarkLodato, line 536: The link on line 585 needs to be updated since the heading changed. (I can't wait to switch off Jekyll so that we can create explicit link anchors!)

docs/provenance/v1.md
MarkLodato: Can we move this to a separate PR to make it easier to review?

MarkLodato (commented):
Sorry, I only had time to review half of the files, but here's my first round of comments. 

I did the following search for existing terms.

```
git grep -i '\bsystems\?\b' spec/v1.0 provenance/v1.md verification_summary/v1.md
```

docs/_data/nav/v1.0.yml
arewm: I would support that in a future PR.

docs/provenance/v1.md
arewm: I think the parallelism with `internalParameters` sounds better for a future PR.

docs/community.md
arewm: What are the services that we are trying to describe here? Are we trying to describe hosted platforms? Some generic set of hosted tools?

I agree that platforms doesn't fit in here well, but would `hosted platforms` be better? We use `systems` more than `services`, so would that be a better term?

docs/provenance/v1.md
arewm: Yeah, I will do that. This is why I started the PR and then it exploded on me.

docs/provenance/v1.md
arewm: Thanks. I knew that there should have been at least one link to update (that is the only one I see).

docs/community.md
MarkLodato: 🤷‍♂️ Good question! I don't know what was originally intended, but developing a service seems like a pretty big request! What about either deleting "and platforms" or replacing with "and implementations"?

```suggestion
developing tooling, we welcome your contributions.
```

docs/spec/v1.0/requirements.md
```diff
@@ -276,8 +276,8 @@ information on assessing a build system's isolation strength, see
 <td>Build service
 <td>
 
-All build steps ran using some build service, not on an individual's
-workstation.
+All build steps ran using a hosted build platform on shared or dedicated
+infrastructure, not on an individual's workstation.
```
MarkLodato, line 280: Much better wording :)

(No action needed)

docs/spec/v1.0/requirements.md
```diff
@@ -261,13 +261,13 @@ Note: This requirement was called "non-falsifiable" in the initial
 [Isolation strength]: #isolation-strength
 [Isolation between builds]: #isolation-strength
 
-The build system is responsible for isolating between builds, even within the
+The control plane is responsible for isolating between builds, even within the
```
MarkLodato, line 264: nit: I don't understand this change. Could you explain? I would have thought "build platform" fit better here, but I might be missing something.

docs/spec/v1.0/terminology.md
```diff
@@ -2,16 +2,28 @@
...
+
 For consistency: Whenever possible, default to "producer," in line with the
-model of producer--consumer--infrastructure provider.
-"Maintainer" is reserved for sections specifying the act of continuing to
-maintain a project after its creation, or when used in a less technical context
-where it is unlikely to cause confusion. Author is reserved for the act of
-making source code commits or reviews. Individual is used when the context's
-focus is specifying a single person (i.e., "an individual's workstation" or
-"compromised individual").--->
+  model of producer--consumer--infrastructure provider. "Maintainer" is reserved
```
MarkLodato, line 11: nit: why are the other lines indented? If these other parts aren't changed, I suggest reverting them.

docs/spec/v1.0/terminology.md
```diff
@@ -2,16 +2,28 @@
...
+  confusion. Author is reserved for the act of making source code commits or
+  reviews. Individual is used when the context's focus is specifying a single
+  person (i.e., "an individual's workstation" or "compromised individual").
+
+Using terms such as "platform," "system," and "service" interchangeably can cause
+confusion.
+
+For consistency: Whenever possible, default to "platform." "Service" is reserved
+  for use within the requirements and "system" is reserved for use to describe both
+  software and tools internal to a platform. External self-sescribed services and
```
MarkLodato, line 23: nit: sounds like "system" is no longer used?

docs/spec/v1.0/requirements.md
```diff
@@ -192,27 +192,27 @@ provenance attestation in order to:
 -   *Ensure integrity:* Verify that the digital signature of the provenance
     attestation is valid and the provenance was not tampered with after the
     build.
--   *Define trust:* Identify the build system and other entities that are
+-   *Define trust:* Identify the build platform and other entities that are
     necessary to trust in order to trust the artifact they produced.
 
 This SHOULD be through a digital signature from a private key accessible only to
 the service that generated the provenance attestation.
```
MarkLodato, line 199: Should this change too? Maybe:

```suggestion
the build platform component that generated the provenance attestation.
```

Or alternatively use "control plane" somehow.

docs/spec/v1.0/requirements.md
```diff
@@ -192,27 +192,27 @@ provenance attestation in order to:
...
-*Accuracy:* The provenance MUST be generated by the build system (i.e. within
-the trust boundary identified in the provenance) and not by a tenant of the
-build system (i.e. outside the trust boundary), except as noted below.
+*Accuracy:* The provenance MUST be generated by the control plane (i.e. within
+the trust boundary [identified in the provenance]) and not by a tenant of the
+build platform (i.e. outside the trust boundary), except as noted below.
 
 -   The data in the provenance MUST be obtained from the build service, either
     because the generator *is* the build service or because the provenance
     generator reads the data directly from the build service.
```
MarkLodato, line 210: Same - should this change?

```suggestion
-   The data in the provenance MUST be obtained from the build platform, either
    because the generator *is* the build platform or because the provenance
    generator reads the data directly from the build platform.
```

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -110,7 +110,7 @@ Resulting threat mitigation:
         platform itself, such as by a malicious insider. Instead, verifiers
         SHOULD carefully consider which build platforms are added to the roots
         of trust. For advice on establishing trust in build platforms, see
-        [Verifying build systems](verifying-systems).
+        [Verifying build platforms](verifying-systems).
 -   [Threat "F"]: SLSA Build L2 covers tampering of the artifact or provenance
     after the build. This is accomplished by verifying the `subject` and
     signature in the steps above.
```
MarkLodato, line 116: Line 152 below: `system` should be `platform`

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -228,11 +228,11 @@ provenance file.
 
 ## Architecture options
 
-System implementers decide which part(s) of the system will verify provenance:
+System implementers decide which part(s) of the platform will verify provenance:
 the package ecosystem at upload time, the consumers at download time, or via a
 continuous monitoring system. Each option comes with its own set of
 considerations, but all are valid. The options are not mutually exclusive, but
-at least one part of a SLSA-conformant system SHOULD verify provenance.
+at least one part of a SLSA-conformant platform SHOULD verify provenance.
```
MarkLodato, line 235: "Platform" doesn't really work here. The intention was really the overall producer-to-consumer chain of events. Here's a possible rewording that avoids the problem:

> There are several options for where provenance verification can happen, and these options are not mutually exclusive:
> [...]
> considerations, but all are valid and at least one SHOULD be used.

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -259,7 +259,7 @@ considered an "ecosystem". For more background, see
 
 During package upload, a package ecosystem can ensure that the artifact's
 provenance matches the expected values for that package name's provenance before
-accepting it into the package registry.  If possible, system implementers SHOULD
+accepting it into the package registry.  If possible, platform implementers SHOULD
 prefer this option because doing so benefits all of the package ecosystem's
 clients.
```
MarkLodato, line 264: similarly:

```suggestion
accepting it into the package registry.  This option is RECOMMENDED whenever
possible because doing so benefits all of the package ecosystem's clients.
```

docs/spec/v1.0/verifying-artifacts.md
```diff
@@ -296,5 +296,5 @@ SHOULD publish its expectations for all the packages it verifies.
 
 Consumers can continuously poll a monitor to detect artifacts that
 do not meet the monitor's expectations. Detecting artifacts that fail
-verification is of limited benefit unless a human or another part of the system
+verification is of limited benefit unless a human or another part of the platform
 responds to the failed verification.
```
MarkLodato, line 300: similarly:

```suggestion
verification is of limited benefit unless a human or automated system takes
action in response to the failed verification.
```

docs/spec/v1.0/requirements.md
MarkLodato, line 1: Should we rename the "Build service" requirement to "Hosted"? That seems like the last remaining use of "service".

MarkLodato (commented):
Thank you so much for this. It's a monster of a PR, but I think it significantly improves readability of the spec.

I have a bunch of minor comments but overall I like this change.

docs/community.md
```diff
@@ -81,7 +81,7 @@ developing tooling and services, we welcome your contributions.
             <div class="w-full md:w-1/2">
                 <div class="rounded-lg text-green p-5 border border-green-400 inline-block mb-8 h4">SLSA v1.0 is coming soon!</div>
                 <p>The SLSA v1.0 release candidate specification is out and <a href="/blog/2023/02/slsa-v1-rc">available for community review</a>. We anticipate a stable release soon. <br><br>
-We’ve released a set of tools and services to generate up to SLSA 3 provenance, and plan to develop further tooling soon.<br><br>
+We’ve released a set of tools and platforms to generate up to SLSA 3 provenance, and plan to develop further tooling soon.<br><br>
```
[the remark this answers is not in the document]
arewm: This is within the community document. While I think it makes sense to not mention the tools/platforms/services, people from the SLSA community are also involved in SLSA-adjacent communities and interested parties/organizations ... often helping to develop these tools/platforms/services.

docs/community.md
arewm: This relates to the other comment in this file. The wording depends on what "community involvement" means. I imagine that it is more likely that tooling can happen within the purview of the community while hosted platforms/implementations are more likely in community-adjacent spaces. _Notifications_ of compliant systems happens within the community since these are recorded in the repository in a self-reported manner.

I feel like removing the clause is likely going to be better but I will wait before doing it.

docs/index.md
```diff
@@ -244,7 +244,7 @@ It’s adaptable, and it’s been designed with the wider security ecosystem in
             <div class="w-full md:w-1/2">
                 <div class="rounded-lg text-green p-5 border border-green-400 inline-block mt-8 md:mt-0 mb-8 h4 font-semibold">SLSA v1.0 is coming soon!</div>
                 <p>The SLSA v1.0 release candidate specification is out and <a href="/blog/2023/02/slsa-v1-rc">available for community review</a>. We anticipate a stable release soon. <br><br>
-We’ve released a set of tools and services to generate up to SLSA 3 provenance, and plan to develop further tooling soon.<br><br>
+We’ve released a set of tools and platforms to generate up to SLSA 3 provenance, and plan to develop further tooling soon.<br><br>
```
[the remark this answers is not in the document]
arewm: What are the tools that we are referring to? Are they specifically used to generate SLSA 3 provenance? Should this whole sentence be removed?

docs/spec/v1.0/terminology.md
arewm: I was trying to make the massive block of notes more legible now that there are multiple different cases to watch out for. That resulted in adding newlines as well as indenting to separate the two sections.

docs/spec/v1.0/requirements.md
arewm: I was trying to be more precise here baed on the terminology definitions

> | Platform | The transitive closure of software and services (systems) that must be trusted to faithfully execute builds. The systems include both the trusted control plane as well as the processes that execute untrusted tenant build steps. The platform includes software, hardware, people, and organizations.
> | Control plane | Build platform component that orchestrates each independent build execution and produces provenance. The control plane is managed by an admin and trusted to be outside the tenant's control.

Re-reading these definitions, the platform mentions the scheduling of untrusted tenant build steps so this is likely what we are actually referring to. The precision of referencing the control plane specifically will likely be most important when talking about the provenance generation.

We can make the separation between the two clearer in the terminology section. Part of this was added by me in this PR.

docs/spec/v1.0/verifying-artifacts.md
arewm: Thanks. Changed locally to be included in the next push.

docs/spec/v1.0/verifying-artifacts.md
arewm: Thanks. Changed locally to be included in the next push.

docs/spec/v1.0/requirements.md
arewm: Thanks. I want to make it control plane, but that would only come in as a requirement at L3. Changed locally to be included in the next push.

docs/spec/v1.0/requirements.md
arewm: I like that change as we are effectively deprecating the use of `service`. We still use the word on like 130 within `Build Platform`:

>  A package's <dfn>build platform</dfn> is the infrastructure used to transform the
> software from source to package. This includes the transitive closure of all
> hardware, software, persons, and organizations that can influence the build. A
> build platform is often a hosted, multi-tenant build service, but it could be a
> system of multiple independent rebuilders, a special-purpose build platform used
> by a single software project, or even an individual's workstation. Ideally, one
> build platform is used by many different software packages so that consumers can
> [minimize the number of trusted platforms](principles.md). For more background,
> see [Build Model](terminology.md#build-model).

It isn't as clear to change that service to `platform` as we are using platform to define platform.

MarkLodato: Change LGTM. I think that use of "service" should be kept.

arewm: ack. Resolving thread.

trishankatdatadog (approved):

docs/community.md
abacchi: I like removing the "and platforms" like Mark had suggested to keep it simple.

docs/community.md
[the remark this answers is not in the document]
abacchi: I'd say remove the sentence or just make it " a set of tools to generate..."

docs/index.md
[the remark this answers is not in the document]
arewm: I will remove for clarity. We can always re-add something later as needed.

docs/spec/v1.0/terminology.md
arewm: Changed to bullets for more directed clarity..

docs/community.md
arewm: suggestion accepted!

docs/spec/v1.0/requirements.md
arewm: @MarkLodato, I think that this is the last unresolved thread. I will go ahead and change this back to `build platform` here as I think it reads just fine. I suspect that no further action will be needed to resolve this thread.

MarkLodato: 👍

MarkLodato (approved):
LGTM!

docs/spec/v1.0/faq.md
```diff
@@ -84,3 +84,15 @@ in-toto's official implementations written in
...
 Provenance such as Sigstore's cosign, the SLSA GitHub Generator, and the in-toto
 Jenkins plugin.
+
+## Q. What is the difference between a build platform, system, and service?
+
+Build platform and build system have been used interchangably in the past. With
+the v1.0 specification, however, there has been a unification around the term
+platform as indicated in the [Terminology](terminology.md). The use of the word
+`system` still exists related to software and services within the build platform
+and to systems outside of a build platform like change managent systems.
```
joshuagl, line 94:
```suggestion
and to systems outside of a build platform like change management systems.
```

docs/spec/v1.0/terminology.md
```diff
@@ -72,10 +80,10 @@ describing this whole process.
 
 | Primary Term | Description
 | --- | ---
-| Platform | System that allows tenants to run builds. Technically, it is the transitive closure of software and services that must be trusted to faithfully execute the build. It includes software, hardware, people, and organizations.
+| Platform | The transitive closure of software and services (components) that must be trusted to faithfully execute builds. The systems include both the trusted control plane as well as the processes that execute untrusted tenant build steps. The platform includes software, hardware, people, and organizations.
```
lehors, line 83: Introducing the notion of systems in the middle of the definition of the platform is confusing to me. I suspect the goal is to differentiate the two but it doesn't provide the common point of reference that would make it clear what the relationship/difference between the system and the platform is.

I would suggest deleting the sentence in the middle and if we want to define systems to do this in a different entry.

I also wonder whether this definition of (build) systems aligns with what @mattfarina is talking about in issue #802.

docs/spec/v1.0/terminology.md
```diff
@@ -2,16 +2,24 @@
...
+     confusion. Author is reserved for the act of making source code commits or
+     reviews. Individual is used when the context's focus is specifying a single
+     person (i.e., "an individual's workstation" or "compromised individual").
+* Using terms such as "platform," "system," and "service" interchangeably can cause
+  confusion.
+  * For consistency: Whenever possible, default to "platform." Instead of using "service,"
+    a reference to a "hosted platform" should be used. A reference to some specific
+    software or tools internal to a platform can be made with "platform component" unless
+    there is a more appropriate definition to use directly like "control plane." External
+    self-sescribed services and systems can continue to be called by these terms.
```
joshuagl, line 21:
```suggestion
    self-described services and systems can continue to be called by these terms.
```

joshuagl (approved):
Thank you for working on this. I spotted a couple of minor typos, but otherwise this looks great.

docs/spec/v1.0/terminology.md
arewm: Thanks for the catch. On a previous version I had `(systems)` in the parenthesis. Would it make more sense if I changed systems in the second sentence to components? I added that line because I wanted to make the association of different trust levels in the platform as a whole.

I don't see this as being the same as #802 as I view that as trying to add a track for the operation of a build platform -- enabling the gamification of platform (system) operation hardening.

arewm: I just pushed the update to change `systems` -> `components` but will wait further review before taking further action.

arewm: Are there two definitions within `terminology.md` or two including the one in `requirements.md`? If you are just commenting about multiple definitions of the platform in the one table row, I was trying to add clarity and not add multiple definitions. If I failed on that, do you have a way to make it better?

We define build platform in requirements.md, but it is a consistent definition which just goes through the hops of expanding the recursive definition. I think it improves readability of the requirements to have it duplicated there as opposed to making a reader jump to the terminology page. In addition to just providing the definition in the requirements, it also relates to the ideals of the build platform (hosted, shared by multiple software packages, etc). We do direct back to the build model terminology at the end of the paragraph to link the two.

arewm: That previous response was just to @lehors 

re: @MarkLodato 's comment -- yes, I was trying to make the inclusion of trusted and untrusted aspects explicit for readers that might not understand the association/need for differentiation.

lehors: Sorry, I really wasn't clear. What I meant by multiple definitions is in that text alone. It says that the platform is:
1. The transitive closure of software and services (components) that must be trusted to faithfully execute builds (in which the components include both the trusted control plane as well as the processes that execute untrusted tenant build steps).
2. The platform includes software, hardware, people, and organizations.

I don't think this reads as well as the original text did, nor the way it is defined in the requirements.md file which is why I was pointing out that we also had it defined there.

arewm: I will back out this one line change. Clarifying this in the future will be easy to do, so we can clarify as needed if we get further feedback on it.

MarkLodato (approved):

PR #811, opened 2023-04-07
css: improve look of :target element

MarkLodato:
When loading a link to a section within the page:

- Add enough scroll margin at the top to ensure that the header does not hide the target of the link.
- Briefly highlight the target of the link with a subtle background color to make it more clear what was actually clicked on.

Known limitations:

- Visually the highlight would look nicer the `line-height` of `h2` and family were about 1.6 (right now `h2` is 1 and `h3` and up are 1.3) but changing that would require some adjusting of other margins which I don't feel like doing right now.
- This doesn't fix the "draft" banner from being drawn over the link target. Fixing that would require either always setting a larger scroll margin (which would be too big once the banner is dismissed) or some more complicated solution. What is in this commit seemed like a good first step, at least.

To preview:

1.  Open a spec page, e.g. /spec/v1.0/requirements.
2.  Resize the window to less than 1200 px so that the header appears on scroll.
3.  Click on a link in the TOC to jump around within the page.

### DIFF
--- a/docs/_sass/minima.scss
+++ b/docs/_sass/minima.scss
@@ -18,6 +18,7 @@ $grey-border: #dedede;
 
 $green-dark: #155757;
 $green-light: #58DA95;
+$green-lighter: #B4F3C9;
 $green-darkest: #102B2D;
 $green-pastel: #ECF8F2;
 $green-button: #23907E;
--- a/docs/_sass/minima/_custom.scss
+++ b/docs/_sass/minima/_custom.scss
@@ -556,6 +556,22 @@ body {
     }
 }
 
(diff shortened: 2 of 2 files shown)


MarkLodato:
If you don't like this change, that feedback would also be valuable.

di (approved):
Looks really nice to me!

lehors:
Thank you for doing this! I have to say that I find the general behavior regarding the header to be fairly unstable in that when I click around sometimes the header appears and sometimes it doesn't but at least with this change I always get to see the anchor no matter what. Not always being able to see the anchor has been annoying me for a long time so I'm really grateful for this fix! :-)

PR #812, opened 2023-04-07
Fix color of threat C in "how to verify" diagram

MarkLodato:
Previously threat C was erroneously greyed out in the legend. Now it has full color to match the image above it. Also un-grey the "source threats" category name.


joshuagl (approved):
Thank you

PR #813, opened 2023-04-08
Remove final "RFC" comment in provenance.md

MarkLodato:
The last TODO or RFC comment was asking about adding a "tenant" field. I had anticipated sending out a PR to add such a field before RC2 but I didn't get it out in time. Let's remove the RFC since we're no longer asking for comments on this. Users who want this can file an issue.


abacchi (approved):
Sounds good - Approved

lehors (approved):
Yes.

PR #814, opened 2023-04-08
[DRAFT] Provenance v1: add `tenant` field

MarkLodato:
There have been some requests to denote the tenant of build platform, distinct from the builder ID. It seems like valuable information that is common to most platforms, even if the tenant is not part of the trust base.

**This is a draft PR waiting for next release.** I created this before RC2 but forgot to send it out, and now it's too late to get in for v1.0. We should defer adding this until the next version. I'm just creating this PR now in case anyone is looking for this. It's also valuable to get feedback against this, if you think it's a bad idea.


MarkLodato:
Since no one has brought this up since then, I'm going to close this PR. We can always revive it if there is interest.

PR #817, opened 2023-04-10
Provenance v1: rename `systemParameters` to `internalParameters`

### LINKED ISSUE #815: Provenance: Should we rename `systemParameters` to `internalParameters`?
In #810, we are converging on "build platform" instead of "build system " or "build service". But we currently have `systemParameters` in the provenance schema, and we cannot rename that without a breaking change.

Should we rename `systemParameters` to `internalParameters` (to mirror `externalParameters`) or perhaps `trustedParameters` (to denote trust)? Either of those two alternates are probably better names than the current one.

Is this high enough priority to warrant a change before v1.0? After v1.0, I'm guessing it's not worth breaking backwards compatibility.

abacchi:
Action item from community meeting 4/10/2023 - fixes #815
I specifically left the RC1 and RC2 files alone and only updated the "v1" files- please let me know if you think that is wrong and we can correct that.

### DIFF
--- a/docs/provenance/schema/v1/provenance.cue
+++ b/docs/provenance/schema/v1/provenance.cue
@@ -9,7 +9,7 @@
         "buildDefinition": {
             "buildType": string,
             "externalParameters": object,
-            "systemParameters": object,
+            "internalParameters": object,
             "resolvedDependencies": [ ...#ResourceDescriptor ],
         },
         "runDetails": {
--- a/docs/provenance/schema/v1/provenance.proto
+++ b/docs/provenance/schema/v1/provenance.proto
@@ -17,7 +17,7 @@ message Provenance {
 message BuildDefinition {
   string build_type = 1;
   google.protobuf.Struct external_parameters = 2;
-  google.protobuf.Struct system_parameters = 3;
+  google.protobuf.Struct internal_parameters = 3;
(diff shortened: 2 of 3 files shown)


docs/provenance/schema/v1/provenance.cue
MarkLodato, line 1: Also need to edit provenance.proto in the same directory.

MarkLodato (commented):
Looks great, thanks! Just one more file to edit.

MarkLodato (approved):
Thanks, @abacchi!

trishankatdatadog (approved):

joshuagl (approved):

MarkLodato:
> LGTM. @MarkLodato do you know how best to add some annotation to RC2 to point out the upcoming change here? This can be done in a separate PR though.

Separate PR sounds good to me since this one has already been approved.

In v1-rc1.md and v1-rc2.md, I suggest adding a note to the documentation for `systemParameters`, saying that it was renamed to `internalParameters` in the final v1.0 version.

I'll also update my What's New PR (#820) calling this out.

MarkLodato:
> I'll also update my What's New PR (#820) calling this out.

Actually I'll do that in a separate PR since it's somewhat significant.

PR #822, opened 2023-04-11
css: prevent "draft" banner from occluding :target

MarkLodato:
This is a follow-on to 9492b35 (#811).

When loading a link to a section within the page, add enough scroll padding to prevent the header from overlapping with the target.

While 9492b35 set the `scroll-padding-top` of `<html>` statically, this change sets the `scroll-margin-top` of all clickable elements dynamically based on whether or not the banner is visible. It's not possible to dynamically set the style of `<html>` based on the state of a checkbox (inside the `<html>`) since there is no CSS parent selector.

Preview: https://deploy-preview-822--slsa.netlify.app/spec/v1.0/requirements#provenance-unforgeable

In the following screenshot, the "Provenance is Unforgeable" link was clicked:

| Description | Screenshot |
| --- | --- |
| Before | before |
| After | after |

### DIFF
--- a/docs/_includes/header.html
+++ b/docs/_includes/header.html
@@ -54,8 +54,8 @@
 {%- assign spec_name = url_parts[1] %}
 {%- assign spec_version = url_parts[2] %}
 {%- if site.data.versions[spec_name].versions[spec_version].draft %}
+<input id="important-messages-checkbox" type="checkbox" hidden>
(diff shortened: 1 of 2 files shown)


joshuagl (approved):
Appreciate you doing all the CSS shovel work, thanks!

Issue #691, opened 2023-03-15
Recommend SPDX and CycloneDX for BOM links in `byproducts`


MarkLodato:
Feedback from @mrutkows: it would be valuable to call out explicitly putting a link to a BOM in `byproducts`. This is similar to #669 from @chitrangpatel that it's not clear where to put the build spec.

I'm thinking of guidelines that, if you want to describe more details about the runtime environment, we recommend using a well established format like SPDX or CycloneDX to represent these details.

MarkLodato:
No, at least not for this issue. `resolvedDependencies` (and `buildDefinition` in general) is about the *inputs* to the build. That does overlap with **Software** Bill of Materials, but is not the intention here. If you ran this build on another service, the `buildDefinition` (and most of the SBOM) would remain the same.

Rather, this issue is about describing the `runDetails`, i.e. the specific invocation of this build. The BOM here would be a "Service" Bill of Materials. For example, what version and configuration of Tekton was running on the build cluster.

laurentsimon:
Got it, thanks for the clarification

MarkLodato:
SBOM is intentionally not part of the Build track, though perhaps might be part of a future Vulnerabilities track. We really need a FAQ on this.

My quick take:

- Provenance and the Build track are focused on trustworthiness of the build process. Provenance is necessarily generated in the build platform's trusted control plane, which in practice results in it being coarse grained.
- SBOM is focused on vulnerability management and licensing. In practice, this requires a signal-to-noise ratio, which in turn usually means doing it in the tooling.

In other words, Provenance and SBOM are operating at different levels of abstraction. Furthermore, Provenance can cover the trustworthiness of the SBOM, i.e. how the SBOM was generated.

joshuagl:
> SBOM is intentionally not part of the Build track, though perhaps might be part of a future Vulnerabilities track. We really need a FAQ on this.

I've attempted to add a FAQ entry in https://github.com/slsa-framework/slsa/pull/827 based on Mark's quick take in https://github.com/slsa-framework/slsa/issues/691#issuecomment-1504083988

MarkLodato:
Looking at this again after half a year of experience, what is the recommendation for where to put SBOMs, `byproducts` or `subject`? Does it matter how the SBOM was produced? Should the SBOM have its own, separate provenance?

- If the SBOM was generated by the user-defined build process, then I suppose `subject` is likely more appropriate since it was indeed an output.
[...]

PR #827, opened 2023-04-12
Add FAQ item on SBOMs

joshuagl:
An attempt at answering this frequent source of confusion for users, as suggested in https://github.com/slsa-framework/slsa/issues/691#issuecomment-1504083988

See the new FAQ item at https://deploy-preview-827--slsa.netlify.app/spec/v1.0/faq#q-how-does-slsa-and-slsa-provenance-relate-to-sbom

### DIFF
--- a/docs/spec/v1.0/faq.md
+++ b/docs/spec/v1.0/faq.md
@@ -96,3 +96,22 @@ and to systems outside of a build platform like change management systems.
 A build service is a hosted build platform that is often run on shared infrastructure
 instead of individuals' machines and workstations. Its use has also been replaced outside
 of the requirements as it relates to the build platform.
+
+## Q: How does SLSA and SLSA Provenance related to SBOM
+
+Software Bill of Materials (SBOM) are a frequently recommended tool for
+increased software supply chain rigor. An SBOM is typically focused on
+understanding software in order to evaluate risk through known vulnerabilities
+and license compliance. These use-cases require fine-grained and timely data
+which can be refined to improve signal-to-noise ratio.
+
+SLSA Provenance and the Build track are focused on trustworthiness of the build
+process. To improve trustworthiness, Provenance is generated in the build
+platform's trusted control plane, which in practice results in it being coarse
+grained.
+
+While they likely include similar data, SBOMs and SLSA Provenance operate at
+different levels of abstraction. While an SBOM is not well suited as a
+requirement for the Build track, SBOMs are good practice which may form part
+of a future SLSA Vulnerabilities track. Further, SLSA Provenance can increase the
+trustworthiness of an SBOM by describing how the SBOM was created.
--- a/docs/spec/v1.0/faq.md
+++ b/docs/spec/v1.0/faq.md
@@ -97,16 +97,16 @@ A build service is a hosted build platform that is often run on shared infrastru
 instead of individuals' machines and workstations. Its use has also been replaced outside
 of the requirements as it relates to the build platform.
 
-## Q: How does SLSA and SLSA Provenance related to SBOM
+## Q: How does SLSA and SLSA Provenance related to SBOM?
 
-Software Bill of Materials (SBOM) are a frequently recommended tool for
+[Software Bill of Materials (SBOM)] are a frequently recommended tool for
 increased software supply chain rigor. An SBOM is typically focused on
 understanding software in order to evaluate risk through known vulnerabilities
 and license compliance. These use-cases require fine-grained and timely data
 which can be refined to improve signal-to-noise ratio.
 
-SLSA Provenance and the Build track are focused on trustworthiness of the build
-process. To improve trustworthiness, Provenance is generated in the build
+[SLSA Provenance] and the [Build track] are focused on trustworthiness of the
+build process. To improve trustworthiness, Provenance is generated in the build
 platform's trusted control plane, which in practice results in it being coarse
 grained.
 
@@ -115,3 +115,7 @@ different levels of abstraction. While an SBOM is not well suited as a
 requirement for the Build track, SBOMs are good practice which may form part
 of a future SLSA Vulnerabilities track. Further, SLSA Provenance can increase the
 trustworthiness of an SBOM by describing how the SBOM was created.
+
+[Software Bill of Materials (SBOM)]: https://ntia.gov/sbom
+[SLSA Provenance]: ../provenance/v1.md
+[Build track]: levels.md#build-track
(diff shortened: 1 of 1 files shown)


docs/spec/v1.0/faq.md
```diff
@@ -96,3 +96,22 @@ and to systems outside of a build platform like change management systems.
 A build service is a hosted build platform that is often run on shared infrastructure
 instead of individuals' machines and workstations. Its use has also been replaced outside
 of the requirements as it relates to the build platform.
+
+## Q: How does SLSA and SLSA Provenance related to SBOM
+
+Software Bill of Materials (SBOM) are a frequently recommended tool for
```
MarkLodato, line 102: Link to https://ntia.gov/sbom?

MarkLodato (approved):
Thanks! This is great.

To make sure everyone is on board with this, I suggest we get a few more approvals, particularly the Positioning SIG. @melba-lopez @mlieberman85

docs/spec/v1.0/faq.md
melba-lopez, line 1: Here are several definitions to provenance --- which folks are going to question versus slsa's definition in the faq vs slsa.dev.  SLSA.DEV seems to match NIST, but this post does not.  


**ON SLSA.DEV  IT STATES:** 
`To trace software back to the source and define the moving parts in a complex supply chain, provenance needs to be there from the very beginning. It’s the verifiable information about software artifacts describing where, when and how something was produced. For higher SLSA levels and more resilient integrity guarantees, provenance requirements are stricter and need a deeper, more technical understanding of the predicate.`


**NIST DEF**

```Definition(s):
  The chronology of the origin, development, ownership, location, and changes to a system or system component and associated data. It may also include personnel and processes used to interact with or make modifications to the system, component, or associated data.
Source(s):
[NIST SP 800-161r1](https://doi.org/10.6028/NIST.SP.800-161r1)
[NIST SP 800-37 Rev. 2](https://doi.org/10.6028/NIST.SP.800-37r2)

  The chronology of the origin, development, ownership, location, and changes to a system or system component and associated data. It may also include the personnel and processes used to interact with or make modifications to the system, component, or associated data.
Source(s):
[NIST SP 800-53 Rev. 5](https://doi.org/10.6028/NIST.SP.800-53r5)

  The records describing the possession of, and changes to, components, component processes, information, systems, organization, and organizational processes. Provenance enables all changes to the baselines of components, component processes, information, systems, organizations, and organizational processes, to be reported to specific actors, functions, locales, or activities.
Source(s):
[NISTIR 7622](https://doi.org/10.6028/NIST.IR.7622) under Provenance

  Metadata pertaining to the origination or source of specified data.
Source(s):
[NIST Privacy Framework Version 1.0](https://doi.org/10.6028/NIST.CSWP.01162020) under Provenance from [NISTIR 8112](https://doi.org/10.6028/NIST.IR.8112) - Adapted```

Comment (melba-lopez) [reply docs/spec/v1.0/faq.md:1]: Another item is around this particular verbiage:

```
"While an SBOM is not well suited as a requirement for the Build track, SBOMs are good practice which may form part of a future SLSA Vulnerabilities track. Further, SLSA Provenance can increase the trustworthiness of an SBOM by describing how the SBOM was created."
```

If we follow SLSA.dev definition = NIST definition, then Provenance can be handled by SBOM (also referenced in NIST) ... The SBOM for the BUILD system effectively.  

If you want to say the build actions/configurations is not included in an SBOM (per specification), then that would be more accurate description.  There are other specifications being developed that are more equivalent to that.

melba-lopez (changes requested):
Please see my concerns documented in review

docs/spec/v1.0/faq.md
MarkLodato: Regarding the comment that SBOM is focused on more than vulnerabilities, here's a quote from NTIA's [Minimum Elements For a Software Bill of Materials (SBOM)](https://www.ntia.doc.gov/files/ntia/publications/sbom_minimum_elements_report.pdf):

> The primary security use case for SBOM today is to identify known vulnerabilities and risks in the software supply chain

joshuagl:
I just pushed a couple of extra commits which try to address the concerns raised here:
1. 2b6d97b attempts to convey Arnaud's suggestion, that the intersecting slsa/attestations and SBOM spaces are rapidly evolving in 
2. 7891bc3 adds some more detail to the claims of abstraction and granularity differences

joshuagl:
> What do you think about the following, breaking up the question into two sub-questions?

This could work. I'm curious what others think?

The answer to the question "(2) Why does SLSA not recommend an existing SBOM format for provenance?" seems especially useful to include.

trishankatdatadog:
> This could work. I'm curious what others think?

This is better, but it still handwaves how in-toto provenance attestations could be integrated with SBOMs. I don't think it's a good idea to allow for that much confusion while people are busy producing both SBOMs and in-toto attestations. What do others think?

NicoleSchwartz:
I love the revisions thank you - as discussed in the meeting today I like that we are linking out to SBOM documents as they are iterating frequently, and also I like us not taking a stance for requiring them.

joshuagl:
@lehors I'd appreciate your review, if you have time? Do my suggested edits address your concerns?

lehors:
I like Mark's proposed update. It defines SBOMs very broadly and remains open with regard to the future. This protects us from criticism of trying to paint SBOMs into a too narrow space which may be seen as inaccurate by some, especially given the various developments underway, and it protects us from criticism about not being willing to leverage work happening elsewhere. What is said about the current state of SBOM formats is factual.

trishankatdatadog:
As I discussed in the Spec meeting earlier this week, I can see that it's important not for SLSA to require any SBOMs at least for v1.0. However, at the same time, I think it might clarify the matter to say that, if the user so wishes, they _could_ add any SBOMs with released artifacts in the `subject` of any provenance. This will give them the benefit of strongly (especially in >= L2) associating SBOMs with artifacts. Or is even this optional way of integrating SLSA with SBOMs too strong to say right now, and we should defer it to the future?

lehors:
I for one think that we will get to see both SBOMs and Provenance used together but I think we're better off waiting for the dust to settle on this before getting into any specifics.

trishankatdatadog:
SGTM, @mlieberman85 and @lehors, and concur, clarifying the ways one can integrate SLSA and SBOMs can wait.

trishankatdatadog (approved):

lehors:
Sorry @joshuagl I didn't mean to stall this. I like Mark's proposal better but I can live with the current proposal.

joshuagl:
Thanks Arnaud. I'll merge as-is and we can iterate based on feedback and how the space evolves in the coming months.

Issue #513, opened 2022-10-19
SLSA v1.0 final - release process


MarkLodato:
Once v1.0 has been reviewed and approved, we need to mark it as final and update the rest of slsa.dev.

- [x] Remove unused "current version" configuration
  - [x]  #828
  - [x] #830
  - [x] #831
- [x] #762
- [x] #829
- [x] #832

PR #828, opened 2023-04-12
Remove current_spec_version and unused code

MarkLodato:
This was intended to be an easy-to-use configuration parameter to point to the latest version of the spec, but in practice it was not actually helpful. In practice all of the pages have change between versions, so it's not helpful to have such a variable.

- It is only actually used in two places, neither of which are valuable:
  - The top-level index: this tried to point dynamically at the latest version but that won't work because all of the pages have changed between versions. We really do need to hard-code it here.
  - Provenance v0.1 and v0.2: we have since decided to tie the provenance and spec versions together, so these ought to be hard-coded to v0.1.
- All other uses are either comments or dead code. (The dead code is also removed in this commit.)

Part of #513.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -39,8 +39,6 @@ plugins:
   - jekyll-readme-index
   - jekyll-titles-from-headings
   - jekyll-relative-links
-# WARNING: Remember to update _data/nav.yml as well.
-current_spec_version: "v0.1"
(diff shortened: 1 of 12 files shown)


joshuagl (approved):

lehors (approved):
Always good to reduce technical debt!

PR #836, opened 2023-04-13
Update steering committee logos

joshuagl:
Remove VMware logo (I no longer work for VMware)
Add Verizon logo (I now work for Verizon)


kpk47 (approved):

MarkLodato (approved):
LGTM. (I filed #841 about changing the meaning of these logos, but in the meantime, this is a correct change.)

PR #839, opened 2023-04-14
Update supply chain diagram to say "Producer"

MarkLodato:
This brings the diagram in line with #739, which replace most uses of Developer with Producer.

This updates v1.0 and the front page completely, but leaves earlier version (v0.1, v1.0-rc1, v1.0-rc2) in an inconsistent state (the Terminology model now says "Producer" but the threat diagram still says "Developer"). If you think that's a problem, let me know and I can do more work to version the images.


david-a-wheeler:
I realize that I'm very late to the party, but do we *really* want to change "developer" to "producer"? To many, the organization running or controlling the *build* is the producer (as in, they produce the package). I think using "developer" everywhere would be easier to understand.

kpk47:
> I realize that I'm very late to the party, but do we _really_ want to change "developer" to "producer"? To many, the organization running or controlling the _build_ is the producer (as in, they produce the package). I think using "developer" everywhere would be easier to understand.

IIRC, we switch to "producer" because the "developer" isn't always the responsible party for a build. For example, an OSS project can have many contributing developers, but the project maintainers are responsible for the build. Similarly, an enterprise will have many developers, but the business as a whole is responsible for the build. "Developer" is really only accurate for solo projects.

kpk47 (approved):

MarkLodato:
> I realize that I'm very late to the party, but do we _really_ want to change "developer" to "producer"? To many, the organization running or controlling the _build_ is the producer (as in, they produce the package). I think using "developer" everywhere would be easier to understand.

@david-a-wheeler it sounds like you're commenting more generally throughout the site, rather than this particular diagram. Could you point out particular places in the text where you think "developer" makes more sense? #739 is probably a good starting point.

The approach we took was to collapse all roles into a single term, "producer", rather than use different specialized terms for each role, such as "developer",  "maintainer", "builder", or "packager". When we do need more nuance, we spell it out with plain English, such as the more specific ["insider threat"](https://github.com/slsa-framework/slsa/pull/739/files#diff-2597f711ef6f6b6366812af6b64dcae803202887aea265ded534127fac43b771R219) instead of a generic "maintainer".  This makes it easier for the reader by minimizing the number of jargon words they need to know and look up. It also forces the specification authors to state their intended meaning more clearly.

I do agree that "producer" is slightly more awkward in this diagram. When I hear "developer", I think of the individual engineer who is authoring code. That concept didn't really come up often in the v1.0 spec (and when it did, we just spelled it out, e.g. "individual"), presumably because we only have the Build track so far. But in this diagram, "developer" fits well because it is precisely the entity generating the source code.

That said, for consistency, I lean towards using "producer" to avoid confusion. "Producer" reads acceptably and has the benefit of being the same term we use everywhere else. This trains the reader to think "producer / consumer".

What do you think?

NicoleSchwartz:
Would it help if we linked under the image (and any image) to the terminology page - which has both explanations and examples? I am in favor of swapping to the more generic producer but get the point that it might be confusing for someone new to the spec.

lehors (approved):

trishankatdatadog (approved):
"Producer" does sound a lot more abstract than "developer", but it makes sense in the age of bots like Dependabot and code copilots!

melba-lopez (approved):
per comment in zoom 

If not done already…can we make sure we put the link to that definition close to the diagram 🙂 to help clear confusion

abacchi (approved):
I like the consistency with our verbiage

MarkLodato:
> Would it help if we linked under the image (and any image) to the terminology page - which has both explanations and examples?

Filed #844.

MarkLodato:
As discussed at the 2023-04-17 spec meeting, there was broad agreement that the consistency was a win. @david-a-wheeler if you'd like to reopen the terminology discussion, please file a separate issue. Thanks!

PR #840, opened 2023-04-14
css: distinguish multi-line entries in nav bar

MarkLodato:
Previously the nav bar did not have any visual distinction between two different entries vs one entry with multiple lines - they all ran together.

Now there is slightly more whitespace between entries and slightly less whitespace between lines within an entry, with all other spacing being *almost* pixel equivalent.

Note: 0fa8c9d did this for the TOC, without being as careful about pixel equivalence.

To preview, compare "Verifying build platforms" and "For infrastructure providers" in the screenshots below, or flip back and forth between the netlify [preview](https://deploy-preview-840--slsa.netlify.app/spec/v1.0/) and the [live site](https://slsa.dev/spec/v1.0/).

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/docs/_sass/minima/_custom.scss
+++ b/docs/_sass/minima/_custom.scss
@@ -656,27 +656,39 @@ html {
 }
 
 .site-nav {
+    // We want spacing between elements to *look* like line-height 1.7 but have
+    // reduced spacing within an element so it looks like a single element.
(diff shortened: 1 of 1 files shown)


joshuagl (approved):
I won't pretend to understand the CSS, but the results look good to me.

lehors (commented):
This is a nice trick!

PR #843, opened 2023-04-17
Provenance: raise visibility of changes since RCs

MarkLodato:
Previously it was not obvious what the changes were between RC1, RC2, and final. Now readers who are interested can see the specific section. Also call out the rename in v1-rc1 and v1-rc2 documentation, since some readers may get caught off-guard.

The current design shows all changes from the previous stable release (v0.2), then repeats the changes since RC1 and RC2. Most readers likely care about the former, so making the RC1/RC2 differences less prominent seems worthwhile. An alternative is to list RC1 and RC2 in the change history, but that is probably a loss of readability for most readers.


TomHennen (approved):

lehors (approved):
LGTM

PR #847, opened 2023-04-20
fix a broken link

hyangah:
"Verifying Artifacts"  is currently
  `/spec/v1.0/verifying-artifacts.md`
Change it to `/spec/v1.0/verifying-artifacts`.
The slsa.dev server doesn't recognize `.md` extension.

### DIFF
--- a/docs/provenance/v1.md
+++ b/docs/provenance/v1.md
@@ -439,7 +439,7 @@ information that is not captured in a standard field. Guidelines:
 
 ## Verification
 
-[Verification]: /spec/v1.0/verifying-artifacts.md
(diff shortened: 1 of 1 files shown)


joshuagl (approved):
Thank you for noticing the issue and submitting a PR to address this!

Jekyll _usually_ strips the extension, I'm not sure why that isn't happening in this case.

MarkLodato (approved):
The issue is that the `.md` is stripped only for *relative* links. This is an absolute link, which breaks. We should fix this throughout the site. It would also be nice to have some tests to check for broken links.

PR #852, opened 2023-04-27
Add lint check for links to slsa.dev and fix existing cases

MarkLodato:
The site should always use relative urls of the form `/absolute/path` or `path`, never `https://slsa.dev/absolute/path` because the latter break when when served on other domains (e.g. local development or deploy previews).

Fix existing uses and add a lint check to prevent this in future PRs.


joshuagl (approved):
Thanks especially for writing the lint check! LGTM.

PR #862, opened 2023-05-16
feat: Add notice if viewing outdated versions of the spec

ianlewis:
When viewing older versions of the spec, this PR adds a notice at the top of pages linking them to the corresponding page in the newest version of the spec (or the top page if the page doesn't exist).

Also removes older "Note about 0.x versions" which referenced v1.0 before it was released. Removed because notices at the top of the page should supersede them.

### DIFF
--- a/docs/_data/versions.yml
+++ b/docs/_data/versions.yml
@@ -1,18 +1,22 @@
 # This file controls the version selector
 # Schema:
 #     <spec>:
+#       name: <string>                # display name
+#       current: <version>            # the current version id.
 #       versions:
-#         <version>:
-#           name: <string> # display name
-#           status: <string> # one of: Draft, Candidate, Approved, Retired
-#           draft: <bool, optional> # if true, display a "draft" warning banner
-#           hidden: <bool, optional> # if true, do not show in dropdown
+#         <version>:                  # version id; used in url path.
+#           name: <string>            # display name
+#           status: <string>          # one of: Draft, Candidate, Approved, Retired
+#           draft: <bool, optional>   # if true, display a "draft" warning banner
+#           hidden: <bool, optional>  # if true, do not show in dropdown
 #
 # Where:
 #   <spec> = specification name, i.e. first component of the URL
 #   <version> = version name, i.e. second component of the URL, e.g. v1.0
 
 spec:
+  name: Core specification
+  current: v1.0
   versions:
     v0.1:
       name: Version 0.1
@@ -30,6 +34,8 @@ spec:
       status: Approved
 
 provenance:
+  name: Provenance attestation
+  current: v1
   versions:
     v0.1:
       name: Version 0.1
@@ -50,6 +56,8 @@ provenance:
       status: Approved
 
 verification_summary:
+  name: Verification summary (VSA)
+  current: v1
   versions:
     v0.1:
       name: Version 0.1
--- a/docs/_includes/header.html
+++ b/docs/_includes/header.html
@@ -48,16 +48,39 @@
 </header>
 
 {%- assign url_parts = page.url | split: '/' %}
-{%- assign spec_name = url_parts[1] %}
+{%- assign spec_id = url_parts[1] %}
 {%- assign spec_version = url_parts[2] %}
-{%- if site.data.versions[spec_name].versions[spec_version].draft %}
-<input id="important-messages-checkbox" type="checkbox" hidden>
-<section class="important-messages">
-  <div>
-    <label for="important-messages-checkbox" title="dismiss">✕</label>
-    <p>This is a working draft of {{ spec_version }}.</p>
-    <p style="font-size: smaller">For the latest release candidate or approved
-      version, please use the version selector.</p>
-  </div>
-</section>
+{%- assign spec = site.data.versions[spec_id] %}
+{%- assign version = spec.versions[spec_version] %}
+{%- if spec and version %}
+  {%- if spec_version != spec.current or version.draft %}
+    <input id="important-messages-checkbox" type="checkbox" hidden>
+    <section class="important-messages">
+      <div>
+        <label for="important-messages-checkbox" title="dismiss">✕</label>
+        {%- if version.draft %}
+        <p>This is a working draft of {{ spec_version }}.</p>
+        <p style="font-size: smaller">For the latest release candidate or approved
+          version, please use the version selector.</p>
+        {%- endif %}
+
+        {%- if spec_version != spec.current %}
+          {%- assign current_version = spec.versions[spec.current] %}
+          {%- if current_version %}
+            {%- comment %}
+            Check if there is a page for the newest version. If there is link there, otherwise link to the spec version top page.
+            {% endcomment %}
+            {%- assign other_version = page.url | replace: spec_version, spec.current | relative_url -%}
+            {%- assign other_page = site.pages | where_exp: "page", "page.url == other_version" | first -%}
+            {%- if other_page %}
+              {%- assign new_url = other_page.url %}
+            {%- else %}
+              {%- assign new_url = "/" + spec_id + "/" + spec.current %}
+            {%- endif %}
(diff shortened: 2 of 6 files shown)


docs/_data/versions.yml
```diff
@@ -13,6 +13,7 @@
 #   <version> = version name, i.e. second component of the URL, e.g. v1.0
 
 spec:
+  name: SLSA
```
joshuagl, line 16:
```suggestion
  name: Core specification
```

docs/_data/versions.yml
```diff
@@ -30,6 +31,7 @@ spec:
       status: Approved
 
 provenance:
+  name: Provenance
```
joshuagl, line 34:
```suggestion
  name: Provenance attestation
```

docs/_data/versions.yml
```diff
@@ -50,6 +52,7 @@ provenance:
       status: Approved
 
 verification_summary:
+  name: VSA
```
joshuagl, line 55:
```suggestion
  name: Verification summary (VSA)
```

joshuagl (changes requested):
Thanks for proposing this @ianlewis! I think this is a great addition.
I wonder if we would be better off using the styling, and perhaps mechanism, of the existing "working draft" banner? https://github.com/slsa-framework/slsa/blob/d799c7bf2f2f321e46956ee2bb2ad0bd85b31983/docs/_includes/header.html#L54

We could use the same styling and mechanism but with a different message if status is `retired`  vs. `draft`?

docs/_data/versions.yml
ianlewis: I think this describes the spec, so would it make sense to say something like "Provenance attestation version 1.0?".

docs/_includes/old-version-warning.html
```diff
@@ -0,0 +1,33 @@
+{%- assign url_parts = page.url | split: '/' %}
+{%- assign spec_id = url_parts[1] %}
+{%- assign spec_version = url_parts[2] %}
+{%- if spec_version and site.data.versions[spec_id].versions and site.data.versions[spec_id].name %}
+  {%- assign spec_name = site.data.versions[spec_id].name %}
+  {% comment %}
+  NOTE: this is a naive sort and may need to be adjusted to sort semver in the future.
```
MarkLodato: suggestion: instead of all this logic, what about just having a `current` field in _data/versions.yml? This was removed in d9308ee5dc02c6732c136b421aa69c72dc9c509d because it was unused, but if we now have a use case, let's just add it back. That seems easier to maintain.

docs/_includes/old-version-warning.html
```diff
@@ -0,0 +1,33 @@
...
+        Check if there is a page for the newest version. If there is link there, otherwise link to the spec version top page.
+        {% endcomment %}
+        {%- assign other_version = page.url | replace: spec_version, item[0] | relative_url %}
+        {%- assign other_page = site.pages | where_exp: "page", "page.url == other_version" | first %}
+        {%- if other_page %}
+          {%- assign new_url = other_page.url %}
+        {%- else -%}
+          {%- assign new_url = "/" + spec_id + "/" + item[0] %}
+        {%- endif %}
+        <div class="rounded-lg text-green p-5 border border-green-400 inline-block text-center w-full mt-6 h4">{{ spec_name }} {{ item[1].name | downcase }} is available now! See the newest documentation <a href="{{ new_url }}">here</a></div>
```
MarkLodato: nit: missing period at end

ianlewis:
> Thanks for proposing this @ianlewis! I think this is a great addition. I wonder if we would be better off using the styling, and perhaps mechanism, of the existing "working draft" banner?
> 
> https://github.com/slsa-framework/slsa/blob/d799c7bf2f2f321e46956ee2bb2ad0bd85b31983/docs/_includes/header.html#L54
> 
> We could use the same styling and mechanism but with a different message if status is `retired` vs. `draft`?

I redid the message to use this and allowed both the draft and new version message to be dismissed separately. I also updated the versions data to be a list rather than an object so it could be ordered.

docs/_data/versions.yml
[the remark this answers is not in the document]
ianlewis: On the other hand that might have implications regarding the Community Specification License that use the "Approved" language. IIUC an "Outdated" spec would still be "Approved" per the CSL, so changing it might be confusing.

https://github.com/slsa-framework/governance/blob/main/1._Community_Specification_License-v1.md

docs/_includes/old-version-warning.html
ianlewis: Done.

docs/_includes/old-version-warning.html
ianlewis: Done.

docs/_data/versions.yml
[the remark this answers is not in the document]
MarkLodato: Right. Or some other language that differentiates latest from older, e.g. "Approved (outdated)"

docs/_includes/header.html
```diff
@@ -48,16 +48,47 @@
...
 <section class="important-messages">
-  <div>
-    <label for="important-messages-checkbox" title="dismiss">✕</label>
-    <p>This is a working draft of {{ spec_version }}.</p>
-    <p style="font-size: smaller">For the latest release candidate or approved
-      version, please use the version selector.</p>
-  </div>
+  {%- assign version = spec.version[spec_version] %}
+
+  {%- if version.draft -%}
```
MarkLodato, line 58: nit: For consistency, all of the tags should be `{%- ... %}` to just consume leading whitespace.

docs/_includes/header.html
```diff
@@ -48,16 +48,47 @@
 </header>
 
 {%- assign url_parts = page.url | split: '/' %}
-{%- assign spec_name = url_parts[1] %}
+{%- assign spec_id = url_parts[1] %}
 {%- assign spec_version = url_parts[2] %}
-{%- if site.data.versions[spec_name].versions[spec_version].draft %}
-<input id="important-messages-checkbox" type="checkbox" hidden>
```
MarkLodato, line 54: This needs to stay outside the `<section>` so that it's a sibling of `<main>`, otherwise the CSS rule to fix the scroll margin doesn't work. Example: https://deploy-preview-862--slsa.netlify.app/provenance/v0.2#schema - notice how the header is covered by the banner.

I suggest keeping the HTML and CSS structure the same (don't change IDs) and instead just change the contents with an if/else. We won't ever need to display both draft AND outdated.

That is:

```liquid
{%- assign spec = site.data.versions[spec_name] %}
{%- if spec_version != spec.current or spec.versions[spec_version].draft %}
<input ...>
<section ...>
  <div>
    <label for=...>✕</label>
    {%- if spec.versions[spec_version].draft %}
      <p>This is a working draft...</p>
      <p>...</p>
    {%- else %}
      ... link to current version ...
    {%- endif %}
  </div>
</section>
```

ianlewis: Ok, I tried to make them separate messages but I'll just make it all one message and put them both in the div.

ianlewis: Done.

docs/_includes/header.html
ianlewis: Done.

docs/_data/versions.yml
```diff
@@ -30,13 +34,16 @@ spec:
...
+  name: Provenance attestation
+  current: v1
   versions:
     v0.1:
       name: Version 0.1
       status: Retired
     v0.2:
       name: Version 0.2
       status: Approved
+      draft: true
```
MarkLodato, line 46: remove (looks like it was added just for testing?)

MarkLodato (approved):
Thanks, Ian1

MarkLodato (approved):
Thanks, Ian!

docs/_data/versions.yml
ianlewis: ah, yeah... good catch. Done.

ianlewis:
@joshuagl PTAL. I think I need your approval to merge.

joshuagl (approved):
Looks great, thanks Ian!

PR #868, opened 2023-05-30
Fixes bad permalinks for building-secure-and-reliable-systems on the FAQ pages

chtiangg:
I think [this section](https://google.github.io/building-secure-and-reliable-systems/raw/ch14.html#hermeticcomma_reproduciblecomma_or_veri) is the right one. 

Both old [benefits](https://static.googleusercontent.com/media/sre.google/en//static/pdf/building_secure_and_reliable_systems.pdf#page=357) and [Hermetic, Reproducible, or Verifiable?](https://sre.google/static/pdf/building_secure_and_reliable_systems.pdf#page=357) don't work any more.


MarkLodato (approved):
Thanks!

For anyone wondering, the PDF was taken down and now the book is hosted as HTML. Hopefully these URLs will remain indefinitely.

PR #876, opened 2023-06-07
Adds clarity to build model terminology (issue #714)

chtiangg:
In the build model section: 
1. Define "Build Caches".
2. Adds a collapse card for "Ambiguous terms to avoid", and clarify "build recipe" and "builder".
3. Makes "Ambiguous terms to avoid" a collapse card in the Package model section.

After this change for the terms mentioned in [issue #714](https://github.com/slsa-framework/slsa/issues/714):

- Build process: is defined in "Build" before this PR, see [link](https://slsa.dev/spec/v1.0/terminology#:~:text=Build-,Process,-that%20converts%20input).
- Build: is defined before this PR.
- Build service: is defined in "Platform" before this PR, see [link](https://slsa.dev/spec/v1.0/terminology#:~:text=of%20software%20and-,services,-that%20must%20be) .
- Builder: is clarified in "Ambiguous terms to avoid".
- Build system: is defined in "Platform" before this PR, see [link](https://slsa.dev/spec/v1.0/terminology#:~:text=Platform-,System,-that%20allows%20tenants)
- Build environment: s defined before this PR.
- Build tools: is included in "Dependencies" before this PR, see [link](https://slsa.dev/spec/v1.0/terminology#:~:text=source%20artifacts%2C%20or-,build%20tools,-.)
- Build model: no need to define.
- Build platform: is defined before this PR.
- Build recipe: is clarified in "Ambiguous terms to avoid".
- Build infrastructure: **didn't change, but please let me know if it needs to be clarified.**
- Build configuration source: is clarified in "Ambiguous terms to avoid".
- Build worker: **didn't change, but not sure if this is really needed.**
- Build cache: is newly added to the term table. 
- Builder model: **didn't change, but not sure if this is really needed.**
- Build executor: is somewhat defined in "Build environment" before this PR, see [link](https://slsa.dev/spec/v1.0/terminology#:~:text=The%20independent-,execution,-context%20in%20which)
- Build definition: is clarified in "Ambiguous terms to avoid".
- Control Plane: is defined before this PR.


chtiangg:
Sorry, included a github action workflow script by mistake, and removing it now.

chtiangg:
Well, had a pretty bad experience with git rebase, but made it work.

docs/spec/v1.0/terminology.md
```diff
@@ -139,13 +153,15 @@ It is the primary identifier to which consumers attach expectations.
 | Package registry | An entity responsible for mapping package names to artifacts within a packaging ecosystem. Most ecosystems support multiple registries, usually a single global registry and multiple private registries. |
 | Publish [a package] | Make an artifact available for use by registering it with the package registry. In technical terms, this means associating an artifact to a package name. This does not necessarily mean making the artifact fully public; an artifact may be published for only a subset of users, such as internal testing or a closed beta. |
 
-Ambiguous terms to avoid:
+<details><summary>Ambiguous term to avoid</summary>
```
joshuagl, line 156:
```suggestion
<details><summary>Ambiguous terms to avoid</summary>
```

docs/spec/v1.0/terminology.md
```diff
@@ -101,6 +102,19 @@ dependency.
 For examples on how this model applies to real-world build platforms, see [index
 of build types](/provenance/v1#index-of-build-types).
 
+<details><summary>Ambiguous term to avoid</summary>
```
joshuagl, line 105:
```suggestion
<details><summary>Ambiguous terms to avoid</summary>
```

docs/spec/v1.0/terminology.md
```diff
@@ -101,6 +102,19 @@ dependency.
...
+
+-   *Build recipe:* Could mean `external parameters`, but may include concrete
```
joshuagl, line 107:
```suggestion
-   *Build recipe*: Could mean `external parameters`, but may include concrete
```

docs/spec/v1.0/terminology.md
```diff
@@ -67,30 +67,31 @@ chain.
...
-environment, and then starting execution. The build then performs arbitrary
-steps, possibly fetching additional dependencies, and outputs one or more
-artifacts. Finally, for SLSA Build L2+, the control plane outputs provenance
-describing this whole process.
+independent. (1) A tenant defines the build by specifying external parameters
+through some sort of interface, often including a reference to a configuration
+file in source control. The control plane then runs the build by (2)
+interpreting the parameters, (3) fetching some initial set of dependencies,
+initializing the build environment, and then starting execution. The build then
+performs arbitrary steps, (4) possibly fetching additional dependencies, (5) and
```
joshuagl, line 75: nit:
```suggestion
performs arbitrary steps, possibly (4) fetching additional dependencies, and (5)
```

I did initially wonder whether we should try and make these steps a list, but I worry that might reduce clarity by removing some of the context in the paragraph which doesn't map explicitly to the diagram. WDYT?

docs/spec/v1.0/terminology.md
```diff
@@ -67,30 +67,31 @@ chain.
...
+outputs one or more artifacts. Finally, for SLSA Build L2+, (6) the control
+plane outputs provenance describing this whole process.
 
 <p align="center"><img src="build-model.svg" alt="Model Build"></p>
 
 | Primary Term | Description
 | --- | ---
-| Platform | System that allows tenants to run builds. Technically, it is the transitive closure of software and services that must be trusted to faithfully execute the build. It includes software, hardware, people, and organizations.
```
joshuagl, line 83: Can you undo the changes to sort the terms alphabetically? We have a (perhaps undocumented?) convention to sort the terminology such that no definition uses a term which hasn't already been defined (see the other two terminology tables before this one).

docs/spec/v1.0/terminology.md
```diff
@@ -101,6 +102,19 @@ dependency.
...
+<details><summary>Ambiguous term to avoid</summary>
+
+-   *Build recipe:* Could mean `external parameters`, but may include concrete
+    steps of how to perform a build. To avoid implementation details, we don't
+    define this term, but always use `external parameters` which is the
+    interface to a `build platform`. Similar terms are *build configuration
+    source* and *build definition*.
+-   *Builder:* Is `build platform`, but sometimes is used as `build environment`
+    or a build tool from `dependencies`. To avoid confusion, we always use
+    `build platform` which allows tenants to run builds.
```
joshuagl, line 114:
```suggestion
    `build platform` to mean the system which allows tenants to run builds.
```
But, we define the term as 'platform' not 'build platform'

joshuagl (changes requested):
Thanks for working on these clarifications, these changes are a solid improvement.
I have a few minor requests and a couple of questions, LMK what you think.

docs/spec/v1.0/build-model.svg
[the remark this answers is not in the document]
lehors: Unfortunately this significantly reduces the size of all the labels. When they appear to be about the same size as the text on my browser they are now much smaller.
Old:
Screenshot 2023-06-09 at 1 57 33 PM
New:
Screenshot 2023-06-09 at 1 57 21 PM
Could you keep the same size? If that's the result of adding the "tenant" I don't think it's worth it to be honest.

docs/spec/v1.0/terminology.md
lehors: > But, we define the term as 'platform' not 'build platform'

Indeed and we can't claim that "we always use" either one because that's just not the case. We currently use a mix of these two forms through out.

docs/spec/v1.0/terminology.md
chtiangg: Let me describe my observation since I have a fresh mind and firstly read this doc recently. I'll then ask for suggestions at the end.

Diagram, text content, and aligning both are three parts. In my personal opinion:
1. A diagram could give a reader the most straightforward view without reading any text content, even if they have no technique context. Normally, a clear diagram can defeat thousands of words. When I first see the original diagram, I found:
   *  The diagram is too crowded: 3 different colors, more than 5 shapes with different sizes, 2 types of arrows, 3 different font sizes, text in vertical and horizontal direction, and 1 single annotation to 1 arrow dot-line. And some empty space on the right side of "interface", which looks strange. Because of these differences for each component, I cannot tell if they have different implicit meanings at first. The combination of these properties distract me from reading the following of this page, since my brain kept processing different things at that time.
   *  No steps exist. As I mentioned, if I have no context, I don't know which component is the start point, from left to right then top to bottom or the other way. Meanwhile, components don't always have arrows connected to each other.
   *  (I guess, these are the reason triggered me to make the diagram change in this PR, but please see my ask at the end.)

2. The text content is above the diagram. When reading this section the first time, I read the content, look at the diagram, then go back to try to find each diagram component in the text content (especially hard), and go back and forward several times to find the difference. However, I do think text content is more important than the diagram in the SLSA doc, since it is more precise and should have no ambiguity. The diagram may not be displayed in extreme conditions (e.g. super bad network).
   *  Not talking about this section, I personally prefer looking at a diagram first, then reading content to verify if the impression I got from then diagram matches what I read later, then I don't have to go back to verify.

3. Aligning both. I absolutely agree. I again personally think:
   * If the content doesn't describe a component in the diagram, maybe, we should add the content for it, e.g. Build Cache.
   * If the content cannot describe the step by step workflow as the diagram depicts, maybe it means either the diagram needs some improvement or the content needs to be changed to match the diagram.

After I wrote this, I found the change for the diagram and these steps are out of the scope for the issue itself. Drawing a nice diagram is always time costly, and needs more rounds of reviews. To follow the [small change list](https://google.github.io/eng-practices/review/developer/small-cls.html) practice, how about I revert all the changes in the diagram and the content. I can create a new issue to list my feedback for the diagram here. If there is an existing issue about the diagram, I can also add my feedback over there. However, please let me know if reverting the change for the diagram in this PR sounds good to you. Thank you!

docs/spec/v1.0/build-model.svg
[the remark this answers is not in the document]
chtiangg: Asked in the other comment. I will address this after the decision over there.

chtiangg (commented):
I will hold off on making any changes, before we resolve the comment thread for the "steps".

docs/spec/v1.0/terminology.md
lehors: I agree with @joshuagl You seem to have taken our comments as an opposition to this PR. It's not. I think your PR is moving us in the direction support it overall. In particular there is no doubt that adding labels to the diagram is a good idea. We just need to get it right. If you want to move that piece to a different PR that's fine but either way works for me.
Thanks.

docs/spec/v1.0/terminology.md
chtiangg: Done.

docs/spec/v1.0/terminology.md
chtiangg: Done.

chtiangg (commented):
Thanks for the review! I will update the description and remove the reverted changes.

docs/spec/v1.0/terminology.md
chtiangg: Done.

docs/spec/v1.0/build-model.svg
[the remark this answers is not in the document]
chtiangg: Revert the change.

docs/spec/v1.0/terminology.md
```diff
@@ -67,30 +67,31 @@ chain.
...
 | --- | ---
-| Platform | System that allows tenants to run builds. Technically, it is the transitive closure of software and services that must be trusted to faithfully execute the build. It includes software, hardware, people, and organizations.
 | Admin | A privileged user with administrative access to the platform, potentially allowing them to tamper with builds or the control plane.
-| Tenant | An untrusted user that builds an artifact on the platform. The tenant defines the build steps and external parameters.
-| Control plane | Build platform component that orchestrates each independent build execution and produces provenance. The control plane is managed by an admin and trusted to be outside the tenant's control.
 | Build | Process that converts input sources and dependencies into output artifacts, defined by the tenant and executed within a single build environment on a platform.
-| Steps | The set of actions that comprise a build, defined by the tenant.
+| Build caches | An intermediate artifact storage managed by the platform and keyed by the hash of input sources. A build may share build caches with another following build running on the platform.
```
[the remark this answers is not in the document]
lehors: But there is a big difference: In the text you are referring to it is used in an example. Here you are making it part of the definition of the term.

docs/spec/v1.0/terminology.md
```diff
@@ -87,6 +87,7 @@ describing this whole process.
 | Build | Process that converts input sources and dependencies into output artifacts, defined by the tenant and executed within a single build environment on a platform.
 | Steps | The set of actions that comprise a build, defined by the tenant.
 | Build environment | The independent execution context in which the build runs, initialized by the control plane. In the case of a distributed build, this is the collection of all such machines/containers/VMs that run steps.
+| Build caches | An intermediate artifact storage managed by the platform and keyed by the hash of input sources. A build may share build caches with any subsequent build running on the platform.
```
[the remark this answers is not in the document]
lehors: I'm really not sure why we can just leave it at "An intermediate artifact storage managed by the platform." but yes, that's already better. Thanks.

docs/spec/v1.0/terminology.md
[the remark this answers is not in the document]
chtiangg: Sounds good, I accepted the suggestion.

docs/spec/v1.0/terminology.md
```diff
@@ -101,6 +102,19 @@ dependency.
...
 of build types](/provenance/v1#index-of-build-types).
 
+<details><summary>Ambiguous term to avoid</summary>
+
+-   *Build recipe:* Could mean `external parameters`, but may include concrete
+    steps of how to perform a build. To avoid implementation details, we don't
+    define this term, but always use `external parameters` which is the
+    interface to a `build platform`. Similar terms are *build configuration
+    source* and *build definition*.
+-   *Builder:* Is `build platform`, but sometimes is used as `build environment`
```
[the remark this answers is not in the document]
MarkLodato: Actually the [general recommendation](https://english.stackexchange.com/a/9881) that "punctuation marks are printed in the same style of font of type as the word, letter, character, or symbol immediately preceding them." Otherwise it looks odd. See screenshot: 

image

I'll push a commit to fix.

MarkLodato (approved):
LGTM. I just pushed a few minor fixes, so another review by @joshuagl would be appreciated.

docs/spec/v1.0/terminology.md
```diff
@@ -104,21 +104,23 @@ of build types](/provenance/v1#index-of-build-types).
...
-    define this term, but always use `external parameters` which is the
-    interface to a `build platform`. Similar terms are *build configuration
+    define this term, but always use "external parameters" which is the
+    interface to a build platform. Similar terms are *build configuration
     source* and *build definition*.
--   *Builder*: Is `build platform`, but sometimes is used as `build environment`
-    or a build tool from `dependencies`. To avoid confusion, we always use
-    `build platform` to mean the system which allows tenants to run builds.
+-   *Builder:* Usually means *build platform*, but might be used for *build
+    environment*, the user who invoke the build, or a build tool from
```
joshuagl, line 113:
```suggestion
    environment*, the user who invoked the build, or a build tool from
```

joshuagl (approved):

lehors (approved):

Issue #801, opened 2023-04-05
Could npm use VSAs instead of a custom 'publish' attestation?


TomHennen:
npm has [a custom publish attestation](https://github.com/npm/attestation/tree/main/specs/publish/v0.1)  as a part of implementing [this RFC](https://github.com/npm/rfcs/blob/main/accepted/0049-link-packages-to-source-and-build.md).

I wonder if it would be possible to use [a VSA](https://slsa.dev/verification_summary/v1-rc2) instead?  Perhaps with some tweaks to the spec?

My understanding is that the publish attestation is meant to say "the npm registry allowed this package" which is very similar to what the VSA is trying to get across.  There may be some concern around npm not wanting to make an assertion about something's SLSA level, in which case I wonder if it would be helpful to add a `SLSA_LEVEL_UNEVALUATED` for the `verifiedLevels` field?

@feelepxyz, @laurentsimon, @joshuagl WDYT?

laurentsimon:
> Per the above point I think this should arguably just be PASSED as the registry did decide to admit the thing.

+ 1. I also think the verification status should be about the *entire* policy passing or failing, not just levels.

joshuagl:
> ```js
>      // TBD: what would an appropriate policy URI being if the artifact is not being verified?
>      // Could be a URI describing the authn checks performed by the registry?
> ```
> 
> I think that we should consider that the artifact itself _is_ verified, after all the registry did decide to admit it. I think it would be fine to reference some global policy though?

That seems sane and inline with what I was thinking, a global/uniform policy which describes the checks npm makes before accepting a package into the registry. 👍 

> 
> ```js
>   "verificationResult": "UNVERIFIED",
> ```
> 
> Per the above point I think this should arguably just be `PASSED` as the registry did decide to admit the thing.

Good call.

I've updated the JSON snippet in my comment above https://github.com/slsa-framework/slsa/issues/801#issuecomment-1497776568

TomHennen:
It's also conceivable that a registry would have custom policies for some packages and they could certainly indicate that in the policy field, but it's certainly not required.

MarkLodato:
In my mind these are two distinct concepts:

* VSA = attestation by **(someone)** that a particular artifact **met SLSA expectations** for a particular package name, _whether or not it was actually published_
* Publish = attestation by the **registry** that a particular artifact **was published** under a particular package name and version, _regardless of how it was authorized (SLSA or not)_

TomHennen:
[...]

PR #882, opened 2023-06-09
editorial: Clarify language in VSA spec

### LINKED ISSUE #878: Make more VSA fields optional, add verification section
Similar to #875, I would like to fix a few issues with VSA v1.0 as well for a v1.1 release.

A VSA is a claim by `verifier` that `subject` was passed the policy for `resourceUri`. That's the bare minimum, and those field MUST be specified or else it doesn't make sense. In particular:

- `verifier` is required for the same reason as `builder.id` (#622).
- `resourceUri` is required to ensure that an artifact that passed for one resource cannot be applied to a different resource (i.e. a confusion attack). For example, if a particular container image is allowed to be published as name "foo", you should not be able to take a VSA and claim that it can be published as "bar".

(We need a "how to verify" section to explain this.)

All other fields should be OPTIONAL. Specifically:

- `policy`: While recommended, it should not be required to indicate *which* policy was verified. You could imagine a situation where a closed source organization does not want to leak the identifier of its own internal policy.
- `policy.uri`: If a policy is specified, I don't see why a URI is required. What if you just want to include a digest?
- `verifiedLevels`: An organization might want to claim that a particular artifact passed its internal policies without making any public claim about what those policies were. The use case here would be something like CompanyX publishes a container image with a VSA saying the image is OK, and clients verify this. They don't need to know the level, just that CompanyX blessed it. (This is akin to a classic code signing case.)
- `timeVerified`: Nice to have, but I don't see why it's strictly required.

Actions:

1. Make the fields above optional.
2. Add a verification section.
3. Fix the intro to better explain this (and remove `bundle` which is not present?)

kpk47:
1. Explain the VSA's core assertion.
2. Update the introduction to be consistent with (1).
3. Update language to differentiate "users" between "VSA producers" and "VSA consumers".


docs/verification_summary/v1.md
```diff
@@ -240,6 +246,24 @@ WARNING: This is just for demonstration purposes.
...
+
+1.  Verify the signature on the VSA envelope using the preconfigured roots of trust.
+2.  Verify the statemet's `subject` matches the digest of the artifact in question.
+3.  Verify that the `predicateType` is `https://slsa.dev/verification_summary/v1`.
+4.  Verify  the `verifier` matches the public key (or equivalent) used to verify the signature in step 1.
+5.  Verify that the value for `resourceUri` in the VSA matches the expected value.
+
+Resulting threat mitigation: See
+[Verifying artifacts](/spec/v1.0/verifying-artifacts) for details about which
+threats are addressed by verifying each SLSA level.
```
joshuagl, line 261: This doesn't seem particularly useful, especially as due to a change elsewhere in this PR a VSA may not indicate the SLSA level that `subject` has been verified to.

joshuagl (commented):
This is a great start, thanks!

I compared the PR here to the suggested actions in #878 and noticed that the intro still includes `bundle`. I'm not entirely certain we need to _remove_ [`bundle`](https://github.com/in-toto/attestation/blob/main/spec/v1/bundle.md), perhaps we could spend a few more words describing that a VSA may be attesting to a single attestation OR a bundle of attestations?

docs/verification_summary/v1.md
```diff
@@ -34,11 +34,13 @@ Understanding of SLSA [Software Attestations](/attestation-model),
 
 A Verification Summary Attestation (VSA) is an attestation that some entity
 (`verifier`) verified one or more software artifacts (the `subject` of an
-in-toto attestation [Statement]) by evaluating the artifact and a `bundle`
-of attestations against some `policy`.  Users who trust the `verifier` may
-assume that the artifacts met the indicated SLSA level without themselves
-needing to evaluate the artifact or to have access to the attestations the
-`verifier` used to make its determination.
+in-toto attestation [Statement]) by evaluating the artifact and an
+attestation against some `policy`.  Specifically, a VSA is a claim by
```
AdamZWu, line 38: > the artifact and an attestation

Could this be "the artifact and its associated attestations"?
A policy verification may involve multiple attestations, for example attestations of vulnerability scanning, and/or SLSA provenances for transitive dependencies.

laurentsimon: re: multiple attestation types. Currently SLSA only supports build attestations `SLSA_BUILD_LEVEL_X`, so it may be better to defer mentioning different attestation types when the corresponding tracks are created?

AdamZWu: Even with just the SLSA provenance, there could still be the use case for transitive dependency verification. E.g. a policy specifies that "all transitive dependencies must be built by at least level X". The verification would then involve recursively checking build materials and verifying their respective SLSA provenances.

docs/verification_summary/v1.md
```diff
@@ -34,11 +34,13 @@ Understanding of SLSA [Software Attestations](/attestation-model),
...
+`verifier` that `subject` passed the policy for `resourceUri`. Users who trust
```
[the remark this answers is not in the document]
TomHennen: I don't think it should be.  The policies themselves aren't necessarily public, so consumers shouldn't have to read those policies.

docs/verification_summary/v1.md
mlieberman85: Though SLSA only includes build provenance today I would push for us to be open about allows VSA to support non-SLSA attestations.

TomHennen: 100%  We'd like to use VSAs as a generic vehicle to summarize all the policy decisions an org might make.  Some things that start as non-SLSA may also become SLSA in the future.  Allowing them in the VSA now allows the same infrastructure to be used for all the decisions a user of software might need to make.

docs/verification_summary/v1.md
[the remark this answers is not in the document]
kpk47: I agree with @TomHennen, though now that I think about it should we disallow a policy without verifiedLevels? I can't think of a use for that combination.

docs/verification_summary/v1.md
```diff
@@ -240,6 +246,24 @@ WARNING: This is just for demonstration purposes.
 
 <div id="slsaresult">
 
+## How to Verify
+
+Verification SHOULD include the following steps
+
+1.  Verify the signature on the VSA envelope using the preconfigured roots of trust.
+2.  Verify the statemet's `subject` matches the digest of the artifact in question.
```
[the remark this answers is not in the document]
kpk47: Done

docs/verification_summary/v1.md
```diff
@@ -240,6 +246,24 @@ WARNING: This is just for demonstration purposes.
...
+3.  Verify that the `predicateType` is `https://slsa.dev/verification_summary/v1`.
+4.  Verify  the `verifier` matches the public key (or equivalent) used to verify the signature in step 1.
```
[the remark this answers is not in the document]
kpk47: Done

docs/verification_summary/v1.md
kpk47: Good point. I'll remove this paragraph.

docs/verification_summary/v1.md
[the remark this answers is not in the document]
TomHennen: An verifier might have some policy but not have a canonical way to abbreviate that policy in verifiedLevels.

I've always thought of the policy field to be helpful to folks debugging issues, but not necessarily required for most users of the VSA.

Would there be any harm in allowing policy to be set if verifiedLevels is empty?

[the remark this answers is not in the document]
kpk47: No, I don't see any harm in leaving it. I was mainly throwing out the idea to see if anyone had a use in mind (which it seems you do).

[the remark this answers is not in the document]
laurentsimon: fine on my end as well. Clarifying in the doc if possible, in case others ask the same questions later

docs/verification_summary/v1.md
laurentsimon: I'm fine with being open. My main worry is around consistency with the rest of the specs and avoiding confusion for readers. If we say that VSAs can attest to other attestation types, I'd personally be confused what it means, given that other attestation types don't have a level associated with them (the current wording is around `SLSA level`). 
`by evaluating the artifact and its associated attestation(s) against some policy` LGTM!
/cc @olivekl

docs/verification_summary/v1.md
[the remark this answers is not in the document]
kpk47: Done

docs/verification_summary/v1.md
```diff
@@ -240,6 +248,24 @@ WARNING: This is just for demonstration purposes.
...
 <div id="slsaresult">
 
+## How to Verify
+
+Verification SHOULD include the following steps
+
+1.  Verify the signature on the VSA envelope using the preconfigured roots of trust.
+2.  Verify the statement's `subject` matches the digest of the artifact in question.
+3.  Verify that the `predicateType` is `https://slsa.dev/verification_summary/v1`.
+4.  Verify  that the `verifier` matches the public key (or equivalent) used to verify the signature in step 1.
```
TomHennen, line 258: nit: extra space between "Verify" and "that"

docs/verification_summary/v1.md
```diff
@@ -167,7 +173,7 @@ of the other top-level fields, such as `subject`, see [Statement]._
 > Either “PASSED” or “FAILED” to indicate if the artifact passed or failed the policy verification.
 
 <a id="verifiedLevels"></a>
-`verifiedLevels` _array ([SlsaResult]), required_
+`verifiedLevels` _array ([SlsaResult]), optional_
```
TomHennen, line 176: It's implicit (given the definition of SlsaResult) that users can put things besides SLSA levels here.  Would it make any sense to make that explicit in the documentation for the field too?

Something like "Users may add any additional, non-SLSA, properties they like to this list so long as they do not conflict with SlsaResult" ?

TomHennen (approved):
Thanks!

docs/verification_summary/v1.md
kpk47: Done

docs/verification_summary/v1.md
```diff
@@ -240,6 +251,24 @@ WARNING: This is just for demonstration purposes.
...
 
+## How to Verify
+
+Verification SHOULD include the following steps
+
+1.  Verify the signature on the VSA envelope using the preconfigured roots of trust.
+2.  Verify the statement's `subject` matches the digest of the artifact in question.
+3.  Verify that the `predicateType` is `https://slsa.dev/verification_summary/v1`.
+4.  Verify that the `verifier` matches the public key (or equivalent) used to verify the signature in step 1.
+5.  Verify that the value for `resourceUri` in the VSA matches the expected value.
```
AdamZWu, line 262: The "resourceUri", and the associate policy, hence the verification claim, are from the perspective of the verifier.
So should this really be a match on the tuple (verifier, resourceUri)? i.e. organization A's verification on resource `foo://bar` are not the same as organization B's verification on the same resource `foo://bar`.

TomHennen: Hmm, my hope was that in many cases folks verifying VSAs can blanket trust the verifier for any claims it makes.  I can see how that might not always work out and depends on how the verifier is configured.

E.g. Org A always trusts Org A's verifier for all decisions, but Org A trusts npmjs's verifier for decisions about npms, but not for decisions about packages from PyPI.

kpk47: I prefer to verify the two fields separately because I don't think we really want to tie `resourceUri` directly to `verifier` (the field). I think what we really want is a way to tie URI patterns to the verifier (the identity).

I prefer to leave that association to post-verification policy, but I'm open to changing my mind. What about adding a pre-configured map of verifiers to URI patterns alongside the pre-configured root of trust?

kpk47: I see a VSA as making the statement "`verifier` believes `subject` passes/fails <claim>". The question of whether you care about that statement is a policy decision, not a matter of whether the VSA is valid.

kpk47: I realize now that my mental model of VSA verification is inconsistent with SLSA provenance verification, where we discard provenance if the verifier doesn't trust the build platform to attest to the stated SLSA level. The two should be more consistent, so yes, we should add a preconfigured mapping from verifiers to what they can attest.

TomHennen: Oh but it does mean something.  It means it passed whatever policy `verifier` had for the resource.  If the verifier is a trusted authority that's all many users will actually care about.

docs/verification_summary/v1.md
```diff
@@ -240,6 +251,24 @@ WARNING: This is just for demonstration purposes.
 
 <div id="slsaresult">
 
+## How to Verify
+
+Verification SHOULD include the following steps
+
+1.  Verify the signature on the VSA envelope using the preconfigured roots of trust.
```
laurentsimon, line 258: verify `verificationResult` is missing?
verify level is missing?

docs/verification_summary/v1.md
MarkLodato, line 176: After some discussion and thinking, I'm now leaning towards keeping this field as required with the caveat below.

Previously, my thinking was that if party A wanted simply check that party B said the artifact was OK, then that should be possible. It shouldn't be required to invent a level if it doesn't actually care. In essence, it is just trusting that *any* VSA from B is OK, equivalent to traditional code signing.

However, similar to making `verifier.id` REQUIRED, it seems reasonable to keep this required and add guidance that, if you don't care, invent some fixed string here. The reason why it's required is for future flexibility. If you just rely on `verifier`, then if you want to change the implicit meaning of the VSA, you need to create a completely new verifier ID (i.e. a new trust root). But if you add a string here, then it's easier - you just create a new string. This doesn't seem like a very difficult requirement.

So I'm OK with either, but I'd lean towards keeping it to make verification have fewer special cases.

docs/verification_summary/v1.md
```diff
@@ -240,6 +251,24 @@ WARNING: This is just for demonstration purposes.
 
 <div id="slsaresult">
 
+## How to Verify
+
+Verification SHOULD include the following steps
```
MarkLodato, line 256: I think this could use more of an intro. This goes along with the comment about policy in the other thread. Perhaps something like:

"Suppose party X wants to delegate to party Y the decision for whether an artifact A is allowed to be published to resource R. To do so, party A checks the following..."

Then use X, Y, A, and R in the text below: signature and `verifier.id` match Y, `resourceUri` matches R, and so on.

docs/verification_summary/v1.md
kpk47: Done.

I avoided the word "policy" for the same reasons we avoided "policy" in the provenance docs. I went with "meeting the consumer's goal," but I'm open to suggestions for other wording.

docs/verification_summary/v1.md
kpk47: Done.

Rather than add `verifiedLevels` specifically, I called out that the verifier may need to check additional fields according to the consumer's goals.

docs/verification_summary/v1.md
kpk47: All good points. I could also go either way, so I decided to keep the field as OPTIONAL to prevent confusion in cases where the verifier is fully delegating the decision to the VSA producer (e.g. publish attestations).

docs/verification_summary/v1.md
kpk47: Done.

docs/verification_summary/v1.md
laurentsimon: It's unclear to me why we're trying to support non SLSA results. Is this in anticipation for future changes in the specs?

docs/verification_summary/v1.md
```diff
@@ -240,6 +255,77 @@ WARNING: This is just for demonstration purposes.
...
+    `https://slsa.dev/verification_summary/v1`.
+
+4.  Verify that the `verifier` matches the public key (or equivalent) used to
+    verify the signature in step 1.
+
+5.  Verify that the value for `resourceUri` in the VSA matches the expected
+    value.
+
+6.  Verify that the value for `slsaResult` is `PASSED`.
+
```
laurentsimon, line 285: note: also verify the expected level?

Unless this field is required, some end-users will forget to provide this level without realizing it's an important property.

docs/verification_summary/v1.md
laurentsimon: I'm worried of unsafe use cases: If we don't ask end-users to verify explicitly, will they simply forget to do so in practice?

TomHennen: The motivation for non-SLSA results was discussed some in #663, #801,  and #878.

docs/verification_summary/v1.md
TomHennen: Not all users will actually care about the SLSA build level and instead may just care that the verifier approved the thing.  See #801 for one example use case (though maybe they'll still use a publish attestation, unclear).

Another example use case is:

1. You have some script that downloads a tarball from cloud storage.
2. You want the script to verify the tarball before unpacking it.
3. The script can also fetch the attestation bundle next to the tarball <path>.intoto.jsonl
4. The script can verify the VSA in the bundle and ensure it passed according to the authority it trusts.
5. You don't want the script to care about the SLSA build level.  In this case the minimum build level you'd accept is set by the VSA issuer, and you explicitly want to delegate that decision to the VSA issuer.

MarkLodato (commented):
Looks good to me. The examples really help! A few more suggestions.

docs/verification_summary/v1.md
```diff
@@ -184,7 +184,11 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
 >
 > Users MAY add additional, non-SLSA properties to this field provided the
-> values do not conflict with the definition of [SlsaResult].
+> values do not conflict with the definition of [SlsaResult]. Users SHOULD
+> negotiate the meaning of such properties with their intended verifier.
+>
+> This field MAY be absent if the verifier is not attesting to a specific SLSA
+> level.
 >
 > This field MAY be set even if the `policy` field is not set.
```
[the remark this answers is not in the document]
kpk47: Done.

I added this note to address an earlier comment about the relationship between verifiedLevels and policy, but we can drop it now since How to Verify explains how to use optional fields in verification.

docs/verification_summary/v1.md
```diff
@@ -253,22 +257,75 @@ WARNING: This is just for demonstration purposes.
...
+    meets your goal.
+
+Verification mitigates different threats depending on the VSA's contents and the
+verification procudure.
 
 IMPORTANT: A VSA does not protect against compromise of the verifier, such as by
 a malicious insider. Instead, VSA users SHOULD carefully consider which
 verifiers they add to their roots of trust.
 
+### Examples
```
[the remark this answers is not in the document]
kpk47: Done

docs/verification_summary/v1.md
```diff
@@ -253,22 +257,75 @@ WARNING: This is just for demonstration purposes.
...
+
+1.  Suppose party X wants to delegate to party Y the decision for whether to
+   publish artifact A to resource R. Party X verifies that:
```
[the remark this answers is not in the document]
kpk47: Done

docs/verification_summary/v1.md
```diff
@@ -253,22 +257,75 @@ WARNING: This is just for demonstration purposes.
...
+   is at most Z old." Party X verifies that:
+   -The signature on the VSA envelope using Y's public signing key from their
+   preconfigured root of trust.
+   -`subject` is A.
+   -`predicateType` is `https://slsa.dev/verification_summary/v1`.
+   -`verifier.id` is Y.
+   -`resourceUri` is R.
+   -`slsaResult` is `PASSED`.
+   -'verifiedLevels` is `SLSA_BUILD_LEVEL_2` or `SLSA_BUILD_LEVEL_3`.
+   -`timeVerified` is no earlier than the current time - Z.
```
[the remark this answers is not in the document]
kpk47: I'll just drop the reference to timeVerified from the example.

docs/verification_summary/v1.md
TomHennen: > I thought about it a bit more and I think it's do-able to make the level optional in the specs, yet have verifiers make it mandatory in their API / interface. For example, in the slsa-verifier, I'd imagine having a `SLSA_BUILD_LEVEL_ANY` to force users to express their intent about levels and we'd make the level compulsory.

I like this.

> On the generation side, I could imagine having a SLSA_BUILD_LEVEL_UNKNOWN to make it explicit because in general I prefer explicitness, but I don't have a very strong opinion beyond that. It kinda seems awkward to me to have VSA without levels in general, since the SLSA standard so heavily relies on levels in the first place. To some extent, it also encourages provenance verifiers to discard levels.

This could probably be fine.  Maybe `SLSA_BUILD_LEVEL_UNEVALUATED` if we thought people might not want to say it was 'UNKNOWN' (because that seems 'bad'?).

laurentsimon: `SLSA_BUILD_LEVEL_UNEVALUATED` LGTM.

laurentsimon: btw, if adding a unevaluated level feels overkill, feel free to skip. My main worry was around verification.

docs/verification_summary/v1.md
```diff
@@ -132,21 +134,25 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
 
 > Timestamp indicating what time the verification occurred.
 
 <a id="resourceUri"></a>
 `resourceUri` _string ([ResourceURI]), required_
 
 > URI that identifies the resource associated with the artifact being verified.
+>
+> This field is required to prevent confusion attacks. E.g., a VSA indicating
+> that a package can be published as `foo` should not be useable to publish
```
joshuagl, line 147:
```suggestion
> that a package can be published as `foo` should not be usable to publish
```
I think "usable" is preferred in en_US? It's usually me trying to sneak en_GB spellings into the spec :-D

docs/verification_summary/v1.md
```diff
@@ -240,6 +253,90 @@ WARNING: This is just for demonstration purposes.
 
 <div id="slsaresult">
 
+## How to Verify
```
joshuagl, line 256:
```suggestion
## How to verify
```
Nit: we usually only capitalise the initial word in a heading

docs/verification_summary/v1.md
```diff
@@ -240,6 +253,90 @@ WARNING: This is just for demonstration purposes.
...
+verification procudure.
+
+IMPORTANT: A VSA does not protect against compromise of the verifier, such as by
+a malicious insider. Instead, VSA users SHOULD carefully consider which
+verifiers they add to their roots of trust.
+
+### Examples
+
+1.  Suppose party X wants to delegate to party Y the decision for whether to
+    accept  artifact A as resource R. Party X verifies that:
```
joshuagl, line 297:
```suggestion
    accept artifact A as resource R. Party X verifies that:
```

joshuagl (approved):
This is great. The examples in verification are extremely useful, thank you!

TomHennen (approved):

docs/verification_summary/v1.md
kpk47: Okay, I'm glad we were able to sort this out.

I've added `SLSA_BUILD_LEVEL_UNEVALUATED` to `slsaResult` and updated the first example to use it.

docs/verification_summary/v1.md
kpk47: I am switching this field back to required, but I believe is consistent with adding `SLSA_BUILD_LEVEL_UNEVALUATED`. If any producer feels strongly about not setting this field, they can set it to an empty array.

docs/verification_summary/v1.md
```diff
@@ -34,11 +34,13 @@ Understanding of SLSA [Software Attestations](/attestation-model),
 
 A Verification Summary Attestation (VSA) is an attestation that some entity
```
olivekl, line 35: These first two sentences are a little confusing to me. The first one says evaluated against "some policy" -- the "some" implied that it's not a specific policy. Then the second sentence drills down and specifies the policy is "resourceUri."Are VSAs only evaluated against that one policy? If so, are two sentences needed? I went back and forth a few times trying to see if there was another difference between the statements, but I don't think there is.

docs/verification_summary/v1.md
```diff
@@ -34,11 +34,13 @@ Understanding of SLSA [Software Attestations](/attestation-model),
...
-assume that the artifacts met the indicated SLSA level without themselves
-needing to evaluate the artifact or to have access to the attestations the
-`verifier` used to make its determination.
+in-toto attestation [Statement]) by evaluating the artifact and its associated
+attestation(s) against some `policy`.  Specifically, a VSA is a claim by
+`verifier` that `subject` passed the policy for `resourceUri`. Users who trust
+the `verifier` may assume that the artifacts identified by the
+`(subject, resourceUri)` pair met the indicated SLSA level without
+themselves needing to evaluate the artifact or to have access to the
+attestations the `verifier` used to make its determination.
```
olivekl, line 43: There's a lot of terms in this paragraph (and page overall). Is there a link to definitions that I missed? If not, it'd be great to add that at the top of the page.

docs/verification_summary/v1.md
```diff
@@ -34,11 +34,13 @@ Understanding of SLSA [Software Attestations](/attestation-model),
...
 
 The VSA also allows consumers to determine the verified levels of
```
olivekl, line 45: I think this section is saying: if the transitive dependencies have VSAs themselves, then this VSA summarizes them also. And if there's provenance but no VSA for a dependency, this top-level VSA will summarize the provenance, essentially creating transitive VSAs within a top-level VSA. And if there's no provenance for dependencies, the VSA does nothing for them.

Is that correct?

docs/verification_summary/v1.md
```diff
@@ -240,13 +253,103 @@ WARNING: This is just for demonstration purposes.
 
 <div id="slsaresult">
 
+## How to verify
+
+Consumers use VSAs to accomplish goals based on delegated trust. We call the
+process of establishing a VSA's authenticity and determining whether it meets
+the consumer's goals 'verification'. Goals differ, as do levels of confidence
+in VSA producers, so the verification procedure will changes to suit its
```
olivekl, line 261:
```suggestion
in VSA producers, so the verification procedure changes to suit its
```

docs/verification_summary/v1.md
```diff
@@ -240,13 +253,103 @@ WARNING: This is just for demonstration purposes.
...
+    -   `slsaResult` is `PASSED`.
+
+    -   `verifiedLevels` contains `SLSA_BUILD_LEVEL_UNEVALUATED`.
+
+    Note: This example is analogous to traditional code signing. The expected
+    value for `verifiedLevels` is arbitrary but agreed to by the producer and
+    the verifier. The verifier does not need to check additional fields, as X
+    fully delegates the decision to Y.
+
+2.  Suppose party X wants to enforce the rule "Artifact A at resource R must
```
olivekl, line 321: How would you feel about a diagram here, or earlier in the page showing the process generally? Would it be overkill to show a high-level process diagram of who is creating what, who is verifying what, who is consuming that verification, etc?
Maybe not in scope of this PR, though!

AdamZWu (approved):
I like the new "how to verify" section and the examples. Thanks for the revision!

docs/verification_summary/v1.md
```diff
@@ -240,13 +253,103 @@ WARNING: This is just for demonstration purposes.
...
+Consumers use VSAs to accomplish goals based on delegated trust. We call the
+process of establishing a VSA's authenticity and determining whether it meets
+the consumer's goals 'verification'. Goals differ, as do levels of confidence
+in VSA producers, so the verification procedure will changes to suit its
+context. However, there are certain steps that most verification procedures
+have in common.
+
+Verification SHOULD include the following steps:
+
+1.  Verify the signature on the VSA envelope using the preconfigured roots of
```
olivekl, line 267: As I was reading this list I was mentally filling in what each step accomplished or offered in terms of building trust. Would it be overkill for the intended audience if you added a few words to explain each step? I.e.,:

1. Verify the signature on the VSA envelope (establishes that the VSA has not been tampered with)
2. Verify that the statement's subject matches the digest (confirms that the VSA is attached to the correct artifact, of whatever...)
3. etc...

olivekl (commented):
Left a few comments on content; feel free to ping again for a language review if you'd like!

docs/verification_summary/v1.md
AdamZWu: The basic idea is that, the `dependencyLevels` in a VSA summarizes the *policy verification results* for all transitive dependencies.

The policy verification result of a dependency could either come from an "actual" verification, i.e. evaluation the provenance of the dependency against a policy for the dependency Uri; or from a VSA that summarizes a prior policy verification on this dependency.

Effectively, VSA functions as a verification "shortcut" -- if a matching and valid VSA is found, the verifier no longer needs to evaluate the artifact provenance and its entire dependency sub-tree.

---
If a dependency has no VSA nor provenance, what happens really depends on the policy for its Uri.
- If the policy says "evaluation at L0" (effectively bypasses the provenance check), then the recursive verification will pass at L0;
- Otherwise, since a provenance is required but missing, the recursive verification will fail.

The `dependencyLevels` in the dependent's VSA will record the recursive verification results accordingly.

olivekl: Ah, interesting, thanks! @kpk47 would you be open to including some of this language? This explanation makes sense to me, but I'm not sure I'd have gleaned that from the current paragraph.

docs/verification_summary/v1.md
laurentsimon: +1 on having a diagram with names, and use the names consistently. I left a few comments where the use of `users`, `party`, `verifiers` (VSA consumer or VSA producer) is not always obvious to readers.

May even be simpler to call the "verifier" the "VSA producer" :)

iiuc, the examples in the text assume that the consumer is the same as the producer, and that's why the X and Y can negotiate. But that's not always the case: Example Chromium -> Google VSA -> consumers. Can we illustrate this example as well?

docs/verification_summary/v1.md
```diff
@@ -176,6 +182,13 @@ of the other top-level fields, such as `subject`, see [Statement]._
 > level implies all levels below it (e.g. SLSA_BUILD_LEVEL_3 implies
 > SLSA_BUILD_LEVEL_2 and SLSA_BUILD_LEVEL_1), so there is no need to
 > include more than one level per track.
+>
+> Users MAY add additional, non-SLSA properties to this field provided the
+> values do not conflict with the definition of [SlsaResult]. Users SHOULD
```
laurentsimon, line 187: Users -> Producers?

docs/verification_summary/v1.md
```diff
@@ -176,6 +182,13 @@ of the other top-level fields, such as `subject`, see [Statement]._
 > level implies all levels below it (e.g. SLSA_BUILD_LEVEL_3 implies
 > SLSA_BUILD_LEVEL_2 and SLSA_BUILD_LEVEL_1), so there is no need to
 > include more than one level per track.
+>
+> Users MAY add additional, non-SLSA properties to this field provided the
```
laurentsimon, line 186: Users -> Producers and verifiers (VSA producers)?

docs/verification_summary/v1.md
laurentsimon: Other places to update (not currently updated in this PR but you may as well update them):
Users MUST count each dependency only once per SLSA track -> verifiers?
Users MUST NOT include more than one level per SLSA track -> verifiers?

docs/verification_summary/v1.md
```diff
@@ -240,13 +253,103 @@ WARNING: This is just for demonstration purposes.
...
+Verification mitigates different threats depending on the VSA's contents and the
+verification procudure.
+
+IMPORTANT: A VSA does not protect against compromise of the verifier, such as by
+a malicious insider. Instead, VSA users SHOULD carefully consider which
+verifiers they add to their roots of trust.
+
+### Examples
+
+1.  Suppose party X wants to delegate to party Y the decision for whether to
```
laurentsimon, line 298: party X -> producer P (which also happens to be the consumer)?
party Y -> verifier V?

docs/verification_summary/v1.md
```diff
@@ -240,13 +253,103 @@ WARNING: This is just for demonstration purposes.
...
+
+    -   `resourceUri` is R.
+
+    -   `slsaResult` is `PASSED`.
+
+    -   `verifiedLevels` contains `SLSA_BUILD_LEVEL_UNEVALUATED`.
+
+    Note: This example is analogous to traditional code signing. The expected
+    value for `verifiedLevels` is arbitrary but agreed to by the producer and
+    the verifier. The verifier does not need to check additional fields, as X
```
laurentsimon, line 318: confused by this sentence `The verifier does not need to check additional fields` -> the consumer? But in the previous sentence `verifier` is correct.

docs/verification_summary/v1.md
laurentsimon: re: non-SLSA results. Something to keep in mind is what happens when a verifying tool encounters a non-SLSA fields that it does not understand. iiuc, verification will fail open. In x509, they mark extensions as "critical": if a verifier encounters a critical extension that it does not understand, it must fail (fail safe). Is this relevant to VSAs?

It would also be useful to provide some guidance how verifiers should name non-SLSA fields, to avoid collision with (future) SLSA fields.

Base on the two comments above, it could look like:
```json
"claims": [
  {
    "name": "slsa.dev/verifiedLevels"
    "critical": true,
    "value": _list_
  },
  {
    "name": "slsa.dev/resourceUri"
    "critical": false, // Verification is optional
    "value": "some-name"
  },
]
```

laurentsimon: Created https://github.com/slsa-framework/slsa/issues/899 for tracking. Resolving

docs/verification_summary/v1.md
```diff
@@ -240,13 +253,103 @@ WARNING: This is just for demonstration purposes.
...
+## How to verify
+
+Consumers use VSAs to accomplish goals based on delegated trust. We call the
+process of establishing a VSA's authenticity and determining whether it meets
+the consumer's goals 'verification'. Goals differ, as do levels of confidence
+in VSA producers, so the verification procedure will changes to suit its
+context. However, there are certain steps that most verification procedures
+have in common.
+
+Verification SHOULD include the following steps:
```
laurentsimon, line 265: Should some of these steps be a `MUST`? Can we use use `MUST` or `SHOULD` in each verification step?

docs/verification_summary/v1.md
kpk47: I gave it a go, but went slightly higher level and removed the decision points. @olivekl @AdamZWu does my summary make sense to you and match your understanding of the chained verification process?

MarkLodato:
TODO: Are we planning to do a v1.1 release for this, are we going to defer this until v1.1 comes naturally, or are we going to roll this into v1.0?

docs/verification_summary/v1.md
MarkLodato, line 1: Could you update the PR title and description to be more, er, descriptive?

- title: summarize the change rather than just "update for v1.1"
- desc: Could you go into more detail? "Update the introduction" doesn't really explain how the intro was changed or why, and similarly for why the fields are now optional or why the section on verifying VSAs is warranted.

This is also a breaking change (change to required field) so we should use the `BREAKING CHANGE:` footer as per https://www.conventionalcommits.org/en/v1.0.0/.

MarkLodato (changes requested):

docs/verification_summary/v1.md
```diff
@@ -4,22 +4,23 @@ description: SLSA v1.0 specification for a verification summary of artifacts by
...
-the dependencies were verified at.
+Assert that the verifier has verified an artifact or set of artifacts. Optionally
+include details about the verification process, such as the verified SLSA
+level(s) and the verifier's expectations.
 
-This allows software consumers to make a decision about the validity of an
+VSAs allow software consumers to make a decision about the validity of an
 artifact without needing to have access to all of the attestations about the
-artifact or all of its transitive dependencies.  They can use it to delegate
+artifact or all of its transitive dependencies. They can use it to delegate
```
olivekl, line 19:
```suggestion
artifact or all of its transitive dependencies. Consumers can use VSAs to delegate
```

docs/verification_summary/v1.md
```diff
@@ -4,22 +4,23 @@ description: SLSA v1.0 specification for a verification summary of artifacts by
...
 complex policy decisions to some trusted party and then simply trust that
 party's decision regarding the artifact.
 
-It also allows software producers to keep the details of their build pipeline
+They also allow software producers to keep the details of their build pipeline
```
olivekl, line 23:
```suggestion
VSAs also allow software producers to keep the details of their build pipeline
```

docs/verification_summary/v1.md
```diff
@@ -28,29 +29,27 @@ been included).
...
+attestation(s) against the `policy` for `resourceUri`. Consumers who trust
+the `verifier` may assume that the artifacts identified by the
+`(subject, resourceUri)` pair met the indicated SLSA level without
+themselves needing to evaluate the artifact or to have access to the
+attestations the `verifier` used to make its determination.
+
+VSAs can also be chained together to meet higher level goals, such as tracking
+the verified SLSA level(s) for the `subject`'s transitive dependencies. Rather
+than verifying provenance for the artifact and each of its transitive
+dependencies all at once, the verifier can verify each one independently and
```
olivekl, line 49: `dependencies all at once, the verifier can verify each one independently
`
"each one" = provenance? artifact? dependency?

docs/verification_summary/v1.md
```diff
@@ -28,29 +29,27 @@ been included).
...
+produce VSAs. Finally, the verifier combines those VSAs, with the artifact acting
+as the final VSA's `subject` and each transitive dependency being an entry in
+`dependencyLevels`.
```
olivekl, line 52:
```
Finally, the verifier combines those VSAs, with the artifact acting
as the final VSA's `subject` and each transitive dependency being an entry in
`dependencyLevels`.
```
Just for clarity's sake, I'd suggest simplifying the construction after the comma. Something like: "Finally, the verifier combines those VSAs; the artifact is
the final VSA's `subject` and each transitive dependency is an entry in
`dependencyLevels`."

olivekl (approved):
A couple tiny suggestions for clarity, but otherwise LGTM!

docs/verification_summary/v1.md
kpk47: Dependency.

docs/verification_summary/v1.md
kpk47: Done

docs/verification_summary/v1.md
kpk47: Done.

There's a lot of content in this PR. Do you think the summary does it justice?

docs/verification_summary/v1.md
```diff
@@ -240,13 +252,111 @@ WARNING: This is just for demonstration purposes.
...
+
+4.  Verify that the `verifier` matches the public key (or equivalent) used to
+    verify the signature in step 1. This step identifies the VSA producer in
+    cases where their identity is not implicitly revealed in step 1.
+
+5.  Verify that the value for `resourceUri` in the VSA matches the expected
+    value. This step ensures that the consumer is using the VSA for the
+    producer's intended purpose.
+
+6.  Verify that the value for `slsaResult` is `PASSED`. This step ensures the
```
TomHennen, line 285: Instead of `slsaResult` this should be `verificationResult`, right?

docs/verification_summary/v1.md
```diff
@@ -240,13 +252,111 @@ WARNING: This is just for demonstration purposes.
...
+
+    -   `subject` is A.
+
+    -   `predicateType` is `https://slsa.dev/verification_summary/v1`.
+
+    -   `verifier.id` is V.
+
+    -   `resourceUri` is R.
+
+    -   `slsaResult` is `PASSED`.
```
TomHennen, line 319: As above, `verificationResult`

docs/verification_summary/v1.md
```diff
@@ -240,13 +252,111 @@ WARNING: This is just for demonstration purposes.
...
```
TomHennen, line 343: `verificationResult`

docs/verification_summary/v1.md
```diff
@@ -240,13 +252,111 @@ WARNING: This is just for demonstration purposes.
 
 <div id="slsaresult">
```
TomHennen, line 253: Should this div go down near line 352?  Should there be a different div id for 'how to verify'?

TomHennen (changes requested):

docs/verification_summary/v1.md
```diff
@@ -172,10 +177,17 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
-> Users MUST NOT include more than one level per SLSA track. Note that each SLSA
-> level implies all levels below it (e.g. SLSA_BUILD_LEVEL_3 implies
+> VSA producers MUST NOT include more than one level per SLSA track. Note that
+> each SLSA  level implies all levels below it (e.g. SLSA_BUILD_LEVEL_3 implies
 > SLSA_BUILD_LEVEL_2 and SLSA_BUILD_LEVEL_1), so there is no need to
 > include more than one level per track.
+>
+> VSA producers MAY add additional, non-SLSA properties to this field provided
+> the values do not conflict with the definition of [SlsaResult]. VSA Producers
+> SHOULD negotiate the meaning of such properties with their intended verifier.
```
laurentsimon, line 187: verifier -> consumer?

docs/verification_summary/v1.md
MarkLodato: Personally I would go into more detail, particularly to explain the reason for the change. Something like:

- Clarify that VSA can be used to express verification of things other than the SLSA Level. Previously this was technically possible, but it was only mentioned in one sentence (under SlsaResult). Now the purpose, model, and field documentation make this explicit.
- Make fields `policy` and `timeVerified` optional because [...]
- Add new "How to verify" section, including examples. Previously it was unclear to many readers exactly how VSAs were intended to be used, leading to possible mistakes. Now we make this explicit. This also brings the VSA spec up to parity with Provenance, which already has such a section.
- Add a `SLSA_BUILD_LEVEL_UNEVALUATED` entry because [...]
- Explain why `resourceUri` is required.
- Other edits for clarity.

Hmm, this PR is getting a bit big. Might make sense to split...

docs/verification_summary/v1.md
MarkLodato: I actually think SlsaResult should move up here.

docs/verification_summary/v1.md
```diff
@@ -172,10 +177,17 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+>
+> This field MAY be absent if the verifier is not attesting to a specific SLSA
+> level.
```
MarkLodato, line 190: Line 175 says required but here it says it's optional. Which is it?

If required, what does that mean? Is one of the SLSA_ results required? When we have more than one track, is it required to have an entry for each track?

docs/verification_summary/v1.md
```diff
@@ -28,29 +29,27 @@ been included).
...
-all of an artifact’s _transitive_ dependencies.  The verifier does this by
-either a) verifying the provenance of each non-source dependency listed in
-the [resolvedDependencies](/provenance/v1#resolvedDependencies) of the artifact
-being verified (recursively) or b) matching the non-source dependency
-listed in `resolvedDependencies` (`subject.digest` ==
-`resolvedDependencies.digest` and, ideally, `vsa.resourceUri` ==
-`resolvedDependencies.uri`) to a VSA _for that dependency_ and using
-`vsa.verifiedLevels` and `vsa.dependencyLevels`.  Policy verifiers wishing
-to establish minimum requirements on dependencies SLSA levels may use
-`vsa.dependencyLevels` to do so.
```
MarkLodato, line 53: We seem to have lost this level of detail, e.g. checking `vsa.resourceUri` against `provenance.resolvedDependencies.uri`. Should this go somewhere as well? Perhaps a new "How to produce" section?

Alternatively should this detail go in the SLSA Provenance "How to verify" section, since it's more about provenance?

/cc @TomHennen

docs/verification_summary/v1.md
```diff
@@ -4,22 +4,23 @@ description: SLSA v1.0 specification for a verification summary of artifacts by
...
+Verification summary attestations convey high-level information about an
+artifact's verification, allowing  consumers to delegate verification decisions
+to trusted third parties.
 
 ## Purpose
 
-Describe what SLSA level an artifact or set of artifacts was verified at
-and other details about the verification process including what SLSA level
-the dependencies were verified at.
+Assert that the verifier has verified an artifact or set of artifacts. Optionally
```
MarkLodato, line 13: Throughout the doc, "verifier" is ambiguous. Is it the entity producing the VSA (verifying some other artifact) or the entity checking the VSA (verifying the VSA itself). We need more precise terminology. A simple diagram might also help to give a quick visual.

Maybe "VSA Producer" and "VSA Consumer"? (This was in an earlier comment by @laurentsimon. Raising visibility here.)

docs/verification_summary/v1.md
```diff
@@ -132,21 +131,27 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
+> environment.
 
 <a id="policy"></a>
-`policy` _object ([ResourceDescriptor]), required_
+`policy` _object ([ResourceDescriptor]), optional_
 
 > Describes the policy that the `subject` was verified against.
 >
-> The entry MUST contain a `uri`.
+> This field is RECOMMENDED.
```
MarkLodato, line 154: Why is it recommended? (or alternatively just remove the recommendation if we don't have a good reason)

docs/verification_summary/v1.md
MarkLodato, line 1: I'm worried that this new version is in a weird limbo where it's half SLSA-specific and half generic. For example, it's not clear if you need to list a SLSA build level even if you're saying nothing about the build level.

Would it make sense to make it fully generic and perhaps move it to the in-toto repo? Moving it to in-toto would have the side benefit of avoiding a SLSA version number bump.

Note that provenance would still remain within the SLSA repo because it's inextricably tied to the SLSA build model and SLSA build levels. VSA, on the other hand, seems actually quite generic.

docs/verification_summary/v1.md
```diff
@@ -240,13 +252,111 @@ WARNING: This is just for demonstration purposes.
 
 <div id="slsaresult">
 
+## How to verify
```
MarkLodato, line 255: nit: again, it's unclear if this "verify" means produce or consume.

docs/verification_summary/v1.md
```diff
@@ -240,13 +252,111 @@ WARNING: This is just for demonstration purposes.
...
+    `https://slsa.dev/verification_summary/v1`. This step ensures that the
+    in-toto predicate is using this version of the VSA format.
+
+4.  Verify that the `verifier` matches the public key (or equivalent) used to
+    verify the signature in step 1. This step identifies the VSA producer in
+    cases where their identity is not implicitly revealed in step 1.
+
+5.  Verify that the value for `resourceUri` in the VSA matches the expected
+    value. This step ensures that the consumer is using the VSA for the
+    producer's intended purpose.
```
MarkLodato, line 283: This piece probably needs more fleshing out, perhaps in a separate PR. There have been many questions about what the resource URI is, e.g. #891.

docs/verification_summary/v1.md
MarkLodato: I think the confusion results from overlap in `verificationResult` and `verifiedLevels`. The latter also has a `FAILED` entry, and the name implies that it's only for PASSED. Should we deprecate `verificationResult` and just say `verifiedLevels` are only for things that passed?

I'm assuming the original intention was to support saying "I checked at L2 and it failed". If so, is that a foot gun?

docs/verification_summary/v1.md
```diff
@@ -240,13 +252,111 @@ WARNING: This is just for demonstration purposes.
...
+
+    -   `predicateType` is `https://slsa.dev/verification_summary/v1`.
+
+    -   `verifier.id` is V.
+
+    -   `resourceUri` is R.
+
+    -   `slsaResult` is `PASSED`.
+
+    -   `verifiedLevels` contains `SLSA_BUILD_LEVEL_UNEVALUATED`.
```
MarkLodato, line 321: I don't understand this piece. Why does someone check that the build level was unevaluated? If they don't care, shouldn't they just ignore it?

docs/verification_summary/v1.md
```diff
@@ -240,13 +252,111 @@ WARNING: This is just for demonstration purposes.
...
+    threats are addressed by verifying each SLSA level.
+
 ## _SlsaResult (String)_
 
 </div>
 
 The result of evaluating an artifact (or set of artifacts) against SLSA.
 SHOULD be one of these values:
 
+-   SLSA_BUILD_LEVEL_UNEVALUATED
```
MarkLodato, line 359: What is the purpose of this new field? That's not clear to me.

docs/verification_summary/v1.md
```diff
@@ -240,13 +252,111 @@ WARNING: This is just for demonstration purposes.
...
+    -   `slsaResult` is `PASSED`.
+
+    -   `verifiedLevels` is `SLSA_BUILD_LEVEL_2` or `SLSA_BUILD_LEVEL_3`.
+
+    Note: In this example, verifying the VSA mitigates the same threats as
+    verifying the artifact's SLSA provenance. See
+    [Verifying artifacts](/spec/v1.0/verifying-artifacts) for details about which
+    threats are addressed by verifying each SLSA level.
+
 ## _SlsaResult (String)_
```
MarkLodato, line 352: Q: Should this enum name be changed to something not slsa specific?

MarkLodato (changes requested):
Sorry for so many comments.

docs/verification_summary/v1.md
kpk47: I've been wondering the same thing. I don't have an answer, but I've got a couple of questions/thoughts that may help shape the decision.
- Do we expect VSAs that communicate strictly non-SLSA properties (e.g. an attestation for a property explicitly out of SLSA's scope)? If so, we should try to move the VSA spec out of slsa-framework. 
- Does SLSA expect to support any attestations that would be alternatives to VSA? If so, then we should update the spec with a more general description of a summary attestation and its minimum requirements to convey SLSA provenance faithfully. 

If we do keep the VSA spec under slsa-framework, then the community needs to articulate a clearer vision of its role in the SLSA ecosystem. If nothing else, we should update the "How-to SLSA" pages with more VSA examples.

kpk47: +1 to splitting the PR. I'll keep "how to verify" in this one and move the rest to new ones.

docs/verification_summary/v1.md
AdamZWu: Actually I think not mentioning those here is better.

The check of `vsa.resourceUri` and `provenance.resolvedDependencies.uri` really belongs to the topic of policy and verification: i.e. VSA is intended for short-circuiting verification on a certain policy, and the URI matching is *an opinionated way* to check whether the short-circuit is permitted.

In order for that to work, there are some implicit assumptions, namely 1) the policy being evaluated is completely determined by the URIs; 2) there is a 1:1 mapping between the URIs and policies; 3) both `vsa.resourceUri` and `provenance.resolvedDependencies.uri` follow the same specs.

We could either spend a lot of energy fully flesh out all these, and in the end we just described one way it can work ... or, we just leave a high-level description that "the verifier can use VSA to short-circuit policy verification when appropriate" and leave the decision of what is appropriate to the verifiers?

docs/verification_summary/v1.md
TomHennen: So, if `verificationResult != 'PASSED'` then `verifiedLevels` must be empty?  If so, SGTM

docs/verification_summary/v1.md
TomHennen: That's fine too, as long as they go together. :)

MarkLodato:
The v1.1 directory now exists as of #942. Should this change move there? or alternatively split out the editorial from content changes, and apply the former to v1.0 and the latter to v1.1?

kpk47:
I've split this PR into two changes. This one contains the original's editorial changes, moved to the v1.1 directory. I moved the content changes to #964.

docs/spec/v1.1/verification_summary.md
```diff
@@ -21,17 +22,17 @@ attestation] framework:
 
 ## Purpose
 
-Describe what SLSA level an artifact or set of artifacts was verified at
-and other details about the verification process including what SLSA level
-the dependencies were verified at.
+Assert that the VSA producer has verified an artifact or set of artifacts.
+Optionally include details about the verification process, such as the verified
+SLSA level(s) and the verifier's expectations.
```
laurentsimon, line 27: `verifier's expectations` == `policy`? In the rest of the text we say `policy`. Shall we use the same term throughout the doc? It's a bit confusing otherwise, at least for me.

docs/spec/v1.1/verification_summary.md
```diff
@@ -41,23 +42,20 @@ been included).
...
-listed in `resolvedDependencies` (`subject.digest` ==
-`resolvedDependencies.digest` and, ideally, `vsa.resourceUri` ==
-`resolvedDependencies.uri`) to a VSA _for that dependency_ and using
-`vsa.verifiedLevels` and `vsa.dependencyLevels`.  Policy verifiers wishing
-to establish minimum requirements on dependencies SLSA levels may use
-`vsa.dependencyLevels` to do so.
+in-toto attestation [Statement]) by evaluating the artifact and its associated
+attestation(s) against the `policy` for `resourceUri`. Consumers who trust
+the `verifier` may assume that the artifacts identified by the
+`(subject, resourceUri)` pair met the indicated SLSA level without
```
laurentsimon, line 48: to be more general and follow earlier wording, we could say `pair met the details about the verification process`.

docs/spec/v1.1/verification_summary.md
```diff
@@ -41,23 +42,20 @@ been included).
...
+themselves needing to evaluate the artifact or to have access to the
+attestations the `verifier` used to make its determination.
```
laurentsimon, line 50: may add another example that the VSA works in cases where the repository is _not_ public. That's why we can't verify SLSA provenance directly for closed-source software.

docs/spec/v1.1/verification_summary.md
```diff
@@ -41,23 +42,20 @@ been included).
...
+
+VSAs can also be chained together to meet higher level goals, such as tracking
+the verified SLSA level(s) for the `subject`'s transitive dependencies. Rather
```
laurentsimon, line 53: same comment as before: if we're saying that levels and other verified metadata is optional, we could be more generic `tracking details about the verification process for the subject's transitive dependencies`.

docs/spec/v1.1/verification_summary.md
```diff
@@ -41,23 +42,20 @@ been included).
...
+than verifying provenance for the artifact and each of its transitive
+dependencies all at once, the verifier can verify each dependency independently
```
laurentsimon, line 55: If a verifier can verify provenance and generate VSA, why does it need to verify them independently? Is this for security or is it a caching (optimization) mechanism? Is there an assumption in the sentence that maybe there are multiple `verifiers`, one that verified dependencies (say, RedHat) while another consumers it (say, Google) to validate a Google binary? Maybe the sentence can be expanded to motivate both cases...?

docs/spec/v1.1/verification_summary.md
```diff
@@ -41,23 +42,20 @@ been included).
...
+and produce VSAs. Finally, the verifier combines those VSAs; the artifact
+is the final VSA's `subject` and each transitive dependency is an
+entry in `dependencyLevels`.
```
laurentsimon, line 58: same comment: do we want to be generic about "verified metadata" or mention verified levels? If the former, we could have a dedicated section with examples for different types of metadata, starting with a verified level. Just an idea...

docs/spec/v1.1/verification_summary.md
```diff
@@ -41,23 +42,20 @@ been included).
...
-to establish minimum requirements on dependencies SLSA levels may use
-`vsa.dependencyLevels` to do so.
+in-toto attestation [Statement]) by evaluating the artifact and its associated
+attestation(s) against the `policy` for `resourceUri`. Consumers who trust
+the `verifier` may assume that the artifacts identified by the
+`(subject, resourceUri)` pair met the indicated SLSA level without
+themselves needing to evaluate the artifact or to have access to the
+attestations the `verifier` used to make its determination.
+
+VSAs can also be chained together to meet higher level goals, such as tracking
```
laurentsimon, line 52: what does higher-level goal mean?
Not clear what "chained" means. Do VSAs reference each other like in a block chain, ie do they link to a prior VSA in their output format?

docs/spec/v1.1/verification_summary.md
AdamZWu: I think including both may be more useful, i.e. "... met the indicated SLSA level and other criteria in the policy".

We are implementing a producer/consumer of VSA v1.0 and realized that there are two different scenarios:
1. We consume VSA produced by ourselves: in this case we could leverage the extension attributes to fully convey all the information that we are interested in, such as additional verification beyond what is defined in the SLSA BUILD levels;
2. We consume VSA produced by others (and similarly, others consume VSA produced by us): in this case, since the consumer and produced may not have mutual definition on extended checks, we must all resort to the standard, i.e. SLSA BUILD levels.

So, if we are producing a VSA without foreknowledge of who will consume it, we should put in both SLSA levels and the extension attributes in the VSA.

docs/spec/v1.1/verification_summary.md
AdamZWu: I think this part refers to the "divide-and-conquer" strategy the VSA enables.

For a naive example, if a system image has 20 software packages, and each package has 50 dependency libraries. A complete transitive dependency verification would involve verifying 1+20+20*50 = 1021 artifacts, which may consume too much time and violate the verifier's latency SLO.

However, with VSA, what can happen is that, when the 20 software packages get staged in an artifact repo, the repo invokes transitive verification for each, so each verification only checks 1+50 artifacts; and the verification produces VSAs, which propagate to the system image's attestation bundle. Then, when we verify the system image, the verifier will only need to perform at 1+20 checks, because the presence of VSAs of the packages eliminates the need to check the dependency libraries (since VSA is a proof that they are already checked).

Effectively, VSA functions as a "distributed dynamic programming table" and facilitates efficient transitive dependency verification.

docs/spec/v1.1/verification_summary.md
AdamZWu: I think "chained" refers to the following:

- VSA summarizes verification results; and
- VSA can short-circuit verification.

These two properties feed each other, so that a VSA is able to summarize the verification results on all transitive dependencies. A VSA represents cumulative verification along the dependency chains of an artifact.

Maybe it is more accurate to describe VSAs subsumes from one another.

For a native example, a system image with 20 packages, each with 50 dependency libraries.

Without VSA, the attestation bundle of the system image will look like:
```
[system image build provenance]
[package 1 build provenance]
...
[package 20 build provenance]
[package 1 library 1 build provenance]
...
[package 1 library 50 build provenance]
[package 2 library 1 build provenance]
...
...
[package 20 library 50 build provenance]
```

With VSA and staging repo verification, the attestation bundle of the system image will look like:
```
[system image build provenance]
[package 1 VSA]
...
[package 20 VSA]
```

After the system image get verified, the attestation bundle of the system image can be reduced to just:
```
[system image VSA]
```

PR #871, opened 2023-06-01
Adds a Github Actions workflow to automate the deployment of the website

chtiangg:
This pull request is one of the actions item from [Restructure the Versioning System for SLSA](https://github.com/slsa-framework/slsa-proposals/pull/10), specifically the "[Automate Building and Deploying The Website](https://github.com/slsa-framework/slsa-proposals/blob/main/0006/README.md#automate-building-and-deploying-the-website)" section.

I tested this action by running it on my own fork:

- Github Action workflow run: [link](https://github.com/chtiangg/slsa/actions/runs/5146549884)
- Deployed website: https://chtiangg.github.io/slsa (note: there is a "slsa" suffix and it disables the version-selection on the website).

PR #883, opened 2023-06-13
Fix new deploy website action

joshuagl:
In #871 we removed the `environment` specification from the `deploy` job during code review (at [my querying](https://github.com/slsa-framework/slsa/pull/871#discussion_r1214218701)), but it turns out it _is_ required to set the environment (even if the docs for the action don't make that 100% clear).

This PR re-introduces the environment configuration for the deploy job. In addition, it moves the permissions required for the deploy job to that job (PLP).

### DIFF
--- a/.github/workflows/deploy_website.yml
+++ b/.github/workflows/deploy_website.yml
@@ -7,8 +7,6 @@ on:
 
 permissions:
   contents: read
-  pages: write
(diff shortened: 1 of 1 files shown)


joshuagl:
@chtiangg I would appreciate your 👀 here, PTAL.

PR #884, opened 2023-06-13
Update date in 2023-06-13-slsa-github-worfklows-container-based.md

asraa:
Updates date in blog post.

It seems like website deployment is currently broken, so it looks like the version with the old date never published.


joshuagl (approved):
The old site publishing workflow is still active, so the post was published.

I'll merge this change to update the date.

Issue #875, opened 2023-06-06
Resolve ambiguity in ResourceDescriptor `annotations`


MarkLodato:
From https://github.com/slsa-framework/slsa/pull/870:

### Problem

The v1.0 provenance spec is currently in conflict with itself because it defines ResourceDescriptor twice:

- The body of the spec indirectly defines ResourceDescriptor by linking to the in-toto v1 definition, which only allows annotations of JSON type **object**, i.e. `map<string, google.protobuf.Struct>`.
- The summary (cue and proto) define of ResourceDescriptor directly, allowing annotations of **any** JSON type (object, array, string, number, boolean), i.e. `map<string, google.protobuf.Value>`.

This can manifest itself if a producer generates a non-object type but the consumer only accepts non-object type.

If the spec had said that only one was canonical and the other was informative, we could fix this by just updating the informative one. But we have no such language, thus the spec is ambiguous.

Fixing this is backwards incompatible because it is changing a type.

- Making it more restrictive (i.e. only object type) means that existing implementations that generate any type will fail.
- Making it less restrictive (i.e. any JSON type) means that existing implementations that only accept object type will fail.

Thus technically this *should* require a major version bump according to Semantic Versioning. However, a major version bump is fairly disruptive. We believe it will likely cause confusion among consumers, who likely will not notice this. (There are not many implementations of SLSA v1.0 yet.)

An alternative solution is to keep our definition and update in-toto v1 to be more permissive, as per https://github.com/in-toto/attestation/issues/242.

### Decision

In the [2023-06-05 spec meeting](https://docs.google.com/document/d/1kMP62o3KI0IqjPRSNtUqADodBqpEL_wlL1PEOsl6u20/edit#heading=h.4svh1796azw9) meeting, we decided to handle this with a one-off minor version bump because a major version bump is too disruptive. So we won't merge this PR as-is but instead will create a v1.1 branch. As part of that, we will also address future inconsistency by (1) clearly marking one format canonical in the event of a conflict and (2) linking to an immutable version of the in-toto spec to avoid potential changes in in-toto. (We could also avoid conflict by only defining it once, or not inlining the ResourceDescriptor, but that makes the spec harder to read by requiring readers to bounce between two different docs.)

Note: If in-toto decides to change to match SLSA (https://github.com/in-toto/attestation/issues/242) we don't need a version bump, but we will still do (1) and (2) above.

marcelamelara:
[...]

PR #892, opened 2023-06-29
content: schema are informative, spec text is canonical

joshuagl:
As discussed in the [2023-06-05 spec meeting](https://docs.google.com/document/d/1kMP62o3KI0IqjPRSNtUqADodBqpEL_wlL1PEOsl6u20/edit#heading=h.4svh1796azw9) clarify the format of ResourceDescriptor and make clear that the schema are only informative, with the specification text (whether our text or the linked in-toto/attestation spec) being the canonical definition.

- Update protobuf and cue definitions to match the in-toto attestations specification.
- Emphasise that the specification text, not the schema, is the authoritative specification should they be in conflict.
- Link to an immutable version of the in-toto specification to avoid potential changes in in-toto resulting in SemVer breaking changes to SLSA's formats.

Fixes: #875


docs/provenance/schema/v1/provenance.cue
```diff
@@ -40,7 +40,9 @@
     "downloadLocation": string,
     "mediaType": string,
     "content": bytes, // base64-encoded
-    "annotations": object,
+    "annotations": {
+        [string]: value
```
MarkLodato, line 44: In cue, I think `_` represents any type?  So perhaps `[string]: _  # any JSON type`?

docs/provenance/v1.md
```diff
@@ -105,6 +105,12 @@ This predicate follows the in-toto attestation [parsing rules]. Summary:
 
 ## Schema
 
+*NOTE: This schema is informative. The canonical definition of field formats
+and their allowed values is the specification text, including the in-toto
+attestation specification for fields and types defined therein. Where this
+schema disagrees with the text, the text is considered the correct and
+authoritative specification.*
```
MarkLodato, line 112: It's not clear what "this schema" refers to, since the cue, proto, and text are all under the same heading. Perhaps move the cue/proto into a new "Summary" subsection under "Schema" and update it to call it the summary?

I also find the "including the in-toto..." piece confusing. Could we perhaps shorten it a bit? Maybe just "This summary (in cue) is informative. In the event of a disagreement with the text description, the text is authoritative."?

docs/provenance/v1.md
```diff
@@ -119,9 +125,19 @@ features. -->
 
 <details>
 <summary>Protocol buffer schema</summary>
+*NOTE: This schema is informative. The canonical definition of field formats
```
MarkLodato, line 128: nit: need blank line before to make it be parsed as markdown

MarkLodato (commented):
Thanks! I just have a suggestion on clarity

docs/provenance/schema/v1/provenance.cue
joshuagl: Thanks! That seems correct based on [_Cuetorials](https://cuetorials.com/overview/types-and-values/#top-and-bottom):

> “_” is top and matches all value. It is also called the “any” value.

and the [language reference](https://cuelang.org/docs/references/spec/#top).

docs/provenance/v1.md
joshuagl: Fixed, thanks.

docs/provenance/v1.md
joshuagl: Great feedback, thanks. I made the proposed changes.

docs/provenance/v1.md
```diff
@@ -105,8 +105,9 @@ This predicate follows the in-toto attestation [parsing rules]. Summary:
 
 ## Schema
 
-*NOTE: This section describes the fields within `predicate`. For a description
-of the other top-level fields, such as `subject`, see [Statement].*
+### Summary
+
+*NOTE: This summary (in cue) is informative. In the event of a disagreement with the text description, the text is authoritative.*
```
joshuagl, line 110: Nit: wrap
```suggestion
*NOTE: This summary (in cue) is informative. In the event of a
disagreement with the text description, the text is authoritative.*
```

docs/provenance/schema/v1/provenance.cue
```diff
@@ -40,7 +40,9 @@
     "downloadLocation": string,
     "mediaType": string,
     "content": bytes, // base64-encoded
-    "annotations": object,
+    "annotations": {
+        [string]: _  # any JSON type
```
MarkLodato, line 44: Oops, comments start with `//`, not `#`.

```suggestion
        [string]: _  // any JSON type
```

MarkLodato (approved):

MarkLodato:
Is this an `editorial` or a `spec` change? (Affects how # of reviews and waiting period.)

joshuagl:
> Is this an `editorial` or a `spec` change? (Affects how # of reviews and waiting period.)

Great question. My original feeling was editorial, because the edits don't change the intent of the spec assuming you agree with the change to suggesting the text is authoritative. I wavered because I didn't want to feel like I was sneaking this in. But then, we do have the consensus from the meeting. I'll change to editorial.

lehors (approved):
LGTM but so we aren't changing the version number at all. Should there be a new entry added to the change history or a note somewhere? This does seem worth some notification to the reader somehow.

MarkLodato:
> LGTM but so we aren't changing the version number at all. Should there be a new entry added to the change history or a note somewhere? This does seem worth some notification to the reader somehow.

Good point. How about we require a changelog entry for every `content` change?

If so, then might want to consider every `content` change a semver "patch" release, which would only show up in the changelog?

lehors:
Yes, I think having a patch release would make sense.

joshuagl:
I like the idea of rolling a patch release for this (and a few other in-flight changes, see #900).

Are we treating the [what's new?](https://slsa.dev/spec/v1.0/whats-new) page as the changelog, with a new version of the page for each release?

lehors:
I think that would make sense.

MarkLodato:
Friendly ping. Is this going in v1.0 or v1.1? The latter now exists as of #942 FYI.

joshuagl:
Consensus was for this to target v1.1. I've updated the PR to make the changes against the v1.1 folder.

joshuagl:
I've added a very minimal "What's new" page for v1.1 which describes this change: https://deploy-preview-892--slsa.netlify.app/spec/v1.1/whats-new

docs/spec/v1.1/provenance.md
```diff
@@ -122,8 +124,16 @@ features. -->
 <details>
 <summary>Protocol buffer schema</summary>
 
+*NOTE: This summary (in protobuf) is informative. In the event of a
+disagreement with the text description, the text is authoritative.*
+
 Link: [provenance.proto](schema/provenance.proto)
 
+<!-- Note: This protobuf definition prioritises being a human-readable summary
```
MarkLodato, line 132: should this be uncommeted? Might be useful for readers?

MarkLodato (approved):

docs/spec/v1.1/provenance.md
joshuagl: That's a good idea. The current layout has two "NOTE:" sections, one after the other, but I think we have some more NOTE's to merge from recent v1 changes and can handle improving things after that consolidation.

PR #895, opened 2023-06-30
spec-editorial: fix provence->provenance typo in threats

joshuagl:
Each version of the threats page features a typo in an example where provenance is spelled provence, this change fixes each instance of the typo.


MarkLodato (approved):
nit: I think this is `spec-editorial`, not `style`. (`style` is for changes to the look without changes to the wording.)

Issue #266, opened 2022-01-11
Consider using Netlify for hosting


MarkLodato:
To make pull requests easier to review, I think we should consider setting up [Netlify](https://netlify.com).

* At a minimum, we can use a free personal account to enable automatic pull request previews. It's a really nice feature. For example, see https://github.com/MarkLodato/slsa/pull/14 (https://deploy-preview-14--marklodato-slsa.netlify.app/). This requires granting the Netlify app write permission to issues and PRs (but not commit access) so I'd like approval from @slsa-framework/slsa-steering-committee before doing this.

* We can also consider switching our hosting to Netlify instead of GitHub pages. This would have some fairly minor benefits, mostly making [redirects](https://docs.netlify.com/routing/redirects/) easier and allowing other templating engines (or custom Jekyll plugins) without a custom GitHub Actions setup. The cost would be to reconfigure our DNS. Not sure it's worth the effort.

* Finally, we could apply for a free [open source plan](https://www.netlify.com/legal/open-source-policy) if we put a Netlify blub on the homepage or footer.

Personally, I think we should start by enabling the PR previews and see if that's sufficient. I'd be the only admin since we don't have a paid (or open source) plan, but that is probably good enough for preview purposes.

Thoughts?

----

Tasks:

* [x] Configure building of slsa.dev on Netlify (https://slsa.netlify.app)
* [x] Transfer ownership to a Team under the [Netlify Open-Source plan](https://www.netlify.com/legal/open-source-policy)
  * [x] Add “This site is powered by Netlify”, and include a link back to the Netlify home page, or use a [premade badge](https://www.netlify.com/press/#badges).
  * [x] Fill out the [form](https://opensource-form.netlify.com/).
  * [x] Configure the new team (https://app.netlify.com/teams/slsa)
  * [x] Transfer ownership of slsa.netlify.app to the new team
  * [x] Set up access control for project maintainers and OpenSSF admins
* [x] Point slsa.dev domain to Netlify
  * [x] Any prerequisites?
    * Looks like switching is not seamless due to the delay in provisioning Let's Encrypt certificates. Need to figure out how to not drop traffic.
  * [x] Ask David A. Wheeler to ask LF to update the [DNS settings](https://docs.netlify.com/domains-https/custom-domains/configure-external-dns/) once it's ready.
* [ ] Set up warnings and playbooks
  * [x] Write documentation for how to maintain the site
[...]

PR #898, opened 2023-07-05
impl: update README to explain how to use Netlify

MarkLodato:
- Document that we use Netlify instead of GitHub Pages now.
- Update the "developing and testing locally" instructions to use
  `netlify dev` instead of running Jekyll directly so that redirects are
  handled properly.
- Document how deploy previews work.
- Document our production Netlify and DNS configuration.
- Start a short playbook for how to debug. It likely could be expanded,
  but this is a good first start.

Issue: #266


docs/README.md
```diff
@@ -1,33 +1,56 @@
...
-This directory contains sources for https://slsa.dev, served
-via GitHub pages and rendered with Jekyll.
+This directory contains sources for https://slsa.dev, rendered with Jekyll and
+served by [Netlify](https://netlify.com).
 
-## Developing and testing locally
+## Development
 
-1.  Install ruby, bundler, and the dev headers:
+### Developing and testing locally
```
arewm, line 8: Nit: the difference between development and developing isn't clear. It seems redundant here.

arewm (approved):

kpk47 (approved):

docs/README.md
MarkLodato: Good point. Changed to "Building and testing locally". Is that better?

Alternatively, I could move this and "Deploy previews" up one level and drop "Development"?

arewm: I am okay with the current form. It makes sense to have the development vs. production separation as the latter won't be relevant unless you are interested in the actual production configuration.

docs/README.md
```diff
@@ -39,18 +62,74 @@ via GitHub pages and rendered with Jekyll.
...
+slsa.dev      ALIAS  apex-loadbalancer.netlify.com
+www.slsa.dev  CNAME  slsa.netlify.app
+```
+
+## Playbooks
+
+### Something is wrong with the site. How do I debug and/or roll back?
+
+Go to https://app.netlify.com/sites/slsa/deploys?filter=main to see recent
+deployments. You need to be logged into Netlify to see the list.
```
joshuagl, line 123: logged in _and_ a member of the ACL mentioned above?

docs/README.md
```diff
@@ -39,18 +62,74 @@ via GitHub pages and rendered with Jekyll.
...
+
+View previous a version of the site by clicking on a deployment's date:
```
joshuagl, line 125:
```suggestion
View a previous version of the site by clicking on a deployment's date:
```

joshuagl (approved):

docs/README.md
MarkLodato: I can't test it since I am already a member, but IIUC any logged in user can see the list of deployments but only members can perform mutations (e.g. publish). I updated the text to say that. Does that help, or do you suggest alternate wording?

joshuagl: The expanded sentence looks good to me, thank you.

Issue #894, opened 2023-06-29
Discussion: Possibly ambiguous language regarding the use of cache artifacts


adityasaky:
Currently, the [Provenance spec](https://slsa.dev/provenance/v1#model) reads as follows regarding caches:

> During execution, the build process might communicate with the build platform’s control plane and/or build caches. This communication is not captured directly in the provenance, but is instead implied by builder.id and subject to SLSA Requirements. Such communication SHOULD NOT influence the definition of the build; if it does, it SHOULD go in resolvedDependencies instead.

Zeroing in on communicating with the cache specifically, I'm parsing this as "if using a cached artifact changes the build definition, record it in resolvedDependencies." First of all, is this a fair reading of the sentence?

If yes, it raises the question of what changing the build definition means. buildDefinition describes the inputs to the build. The two ways I see that using a cache can impact the inputs to the build is a) as a configuration to use the cache, and b) by treating a cached artifact as an input to the build as it changes the build result in some way.

The first is entirely disconnected from actually using a cached artifact in my mind because a build can be configured to use a cache and end up with zero hits. Further, this configuration wouldn't actually end up in resolvedDependencies, but presumably one of the parameters fields? On the other hand, the second raises other questions about cache behavior.

1. Does a cached artifact qualify as a build input? Possibly yes if the builds are not reproducible, but even then, we have variation between two build executions that each don't use the cache.
2. How do we even know if a cached artifact impacts the build if builds aren't repeated or otherwise tested for reproducibility? (It also raises the possibility of a build being reproducible when using a cached artifact which is not reproducible itself.)

The requirements table calls out the impact of caches for the [isolation requirement](https://slsa.dev/spec/v1.0/requirements#isolation-strength):

> It MUST NOT be possible for one build to inject false entries into a build cache used by another build, also known as “cache poisoning”. In other words, the output of the build MUST be identical whether or not the cache is used.

This seems to indicate that using a cache must not have an impact on builds beyond a configuration to use or not use the cache. Which means cache-specific behavior is never expected to be recorded in Provenance? Regardless of what the intent is, I think some clarifying may be in order! :smile:

arewm:
Interesting, the part that makes the least sense to me is

> This communication is not captured directly in the provenance, but is instead implied by builder.id

I generally read these sections indicating that all cached content use for the build must have an immutable reference (i.e. a hash) of the content to be used for addressing. If the build system leverages a cache in a build, the provenance should include the cached content as it was _originally_ cached.

[...]

PR #901, opened 2023-07-06
editorial: clarify requirements around cache use by the build platform

arewm:
There was an inconsistency between the provenance model and build L3 requirements for how build caches might affect a build:

* The provenance model indicated that communication with a cache MAY influence the definition of a build (and if it does, then the communication SHOULD go in `resolvedDependencies`).
* The build isolation requirements indicated that the build output MUST be identical whether a cache is used.

Clarification includes:

* At build L3, the provenance MUST be identical whether a build platform's cache is used or not.
* This implies that all communication with the build cache is to leverage `byproducts` which should not be present in the provenance.
* Builds are not precluded from using a cache outside the build platform and therefore outside these isolation requirements.

Clarification does NOT include:

* Requirements between cache utilization and reproducible builds as reproducibility is not covered in the build track.

Addresses #894

Signed-off-by: arewm <arewm@users.noreply.github.com>

### DIFF
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -318,7 +318,7 @@ If the build platform leverages a cache for builds, it MUST guarantee the follow
     words, the output of the build MUST be identical whether or not the cache is
     used.
 -   If the build platform is capable of providing the provenance for an external
-    resource without a cache, then the provenance SHOULD remain unchanged if a cache
+    resource without a cache, then the provenance MUST remain unchanged if a cache
     is used. In other words, the output of the provenance MUST be identical whether
     or not the cache is used.
 
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -317,11 +317,12 @@ If the build platform leverages a cache for builds, it MUST guarantee the follow
     cache used by another build, also known as "cache poisoning". In other
     words, the output of the build MUST be identical whether or not the cache is
     used.
--   If the build platform is capable of providing the provenance for an external
-    resource without a cache, then the provenance MUST remain unchanged if a cache
-    is used. In other words, the output of the provenance MUST be identical whether
-    or not the cache is used. Communication with the build cache MUST NOT be present
-    in `resolvedDependencies`.
+-   If the build platform is capable of providing the provenance information for
+    an external resource when a cache is not in use, then the provenance
+    information MUST remain unchanged if a cache is used. In other words, the
+    information in the provenance MUST be identical whether or not the cache is
+    used. Communication with the build cache MUST NOT be represented in
+    `resolvedDependencies`.
 
 There are no sub-requirements on the build itself. Build L3 is limited to
 ensuring that a well-intentioned build runs securely. It does not require that
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -328,8 +328,7 @@ ensuring that a well-intentioned build runs securely. It does not require that
 a build platform prevents a producer from performing a risky or insecure build. In
 particular, the "Isolated" requirement does not prohibit a build from calling
 out to a remote execution service or a "self-hosted runner" that is outside the
-trust boundary of the build platform. Additionally, build L3 does not prohibit
-builds from resolving dependencies outside of the knowledge of the build platform.
+trust boundary of the build platform.
 
 NOTE: This requirement was split into "Isolated" and "Ephemeral Environment"
 in the initial [draft version (v0.1)](../v0.1/requirements.md).
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -317,15 +317,18 @@ If the build platform leverages a cache for builds, it MUST guarantee the follow
     cache used by another build, also known as "cache poisoning". In other
     words, the output of the build MUST be identical whether or not the cache is
     used.
--   The resolved dependencies used to generate the cached artifacts MUST be captured
-    in the provenance.
+-   If the build platform is capable of providing the provenance for an external
+    resource without a cache, then the provenance SHOULD remain unchanged if a cache
+    is used. In other words, the output of the provenance MUST be identical whether
+    or not the cache is used.
 
 There are no sub-requirements on the build itself. Build L3 is limited to
 ensuring that a well-intentioned build runs securely. It does not require that
 a build platform prevents a producer from performing a risky or insecure build. In
 particular, the "Isolated" requirement does not prohibit a build from calling
 out to a remote execution service or a "self-hosted runner" that is outside the
-trust boundary of the build platform.
+trust boundary of the build platform. Additionally, build L3 does not prohibit
+builds from resolving dependencies outside of the knowledge of the build platform.
 
 NOTE: This requirement was split into "Isolated" and "Ephemeral Environment"
 in the initial [draft version (v0.1)](../v0.1/requirements.md).
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -307,13 +307,18 @@ The build platform MUST guarantee the following:
 -   It MUST NOT be possible for one build to persist or influence the build
     environment of a subsequent build. In other words, an ephemeral build
     environment MUST be provisioned for each build.
+-   The build platform MUST NOT open services that allow for remote influence
+    unless all such interactions are captured as `externalParameters` in the
+    provenance.
+
+If the build platform leverages a cache for builds, it MUST guarantee the following:
+
 -   It MUST NOT be possible for one build to inject false entries into a build
     cache used by another build, also known as "cache poisoning". In other
     words, the output of the build MUST be identical whether or not the cache is
     used.
--   The build platform MUST NOT open services that allow for remote influence
-    unless all such interactions are captured as `externalParameters` in the
-    provenance.
+-   The resolved dependencies used to generate the cached artifacts MUST be captured
+    in the provenance.
 
 There are no sub-requirements on the build itself. Build L3 is limited to
 ensuring that a well-intentioned build runs securely. It does not require that
--- a/docs/spec/v1.0/requirements.md
+++ b/docs/spec/v1.0/requirements.md
@@ -320,7 +320,8 @@ If the build platform leverages a cache for builds, it MUST guarantee the follow
 -   If the build platform is capable of providing the provenance for an external
     resource without a cache, then the provenance MUST remain unchanged if a cache
     is used. In other words, the output of the provenance MUST be identical whether
-    or not the cache is used.
+    or not the cache is used. Communication with the build cache MUST NOT be present
+    in `resolvedDependencies`.
 
 There are no sub-requirements on the build itself. Build L3 is limited to
 ensuring that a well-intentioned build runs securely. It does not require that
--- a/docs/spec/v1.0/provenance.md
+++ b/docs/spec/v1.0/provenance.md
@@ -87,8 +87,7 @@ The model is as follows:
     captured directly in the provenance, but is instead implied by `builder.id`
     and subject to [SLSA Requirements](requirements.md). Such
     communication SHOULD NOT influence the definition of the build; if it does,
-    it SHOULD go in `resolvedDependencies` instead. Cache communications MUST NOT
-    be present at SLSA Build L3.
+    it SHOULD go in `resolvedDependencies` instead.
 
 -   Finally, the build process outputs one or more artifacts, identified by
     `subject`.
--- a/docs/spec/v1.0/provenance.md
+++ b/docs/spec/v1.0/provenance.md
@@ -87,7 +87,8 @@ The model is as follows:
     captured directly in the provenance, but is instead implied by `builder.id`
     and subject to [SLSA Requirements](requirements.md). Such
     communication SHOULD NOT influence the definition of the build; if it does,
-    it SHOULD go in `resolvedDependencies` instead.
+    it SHOULD go in `resolvedDependencies` instead. Cache communications MUST NOT
+    be present at SLSA Build L3.
 
 -   Finally, the build process outputs one or more artifacts, identified by
     `subject`.


docs/spec/v1.0/requirements.md
```diff
@@ -301,13 +301,18 @@ The build platform MUST guarantee the following:
...
+
 -   It MUST NOT be possible for one build to inject false entries into a build
     cache used by another build, also known as "cache poisoning". In other
     words, the output of the build MUST be identical whether or not the cache is
     used.
--   The build platform MUST NOT open services that allow for remote influence
-    unless all such interactions are captured as `externalParameters` in the
-    provenance.
+-   The resolved dependencies used to generate the cached artifacts MUST be captured
+    in the provenance.
```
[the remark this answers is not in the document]
arewm: This isn't a statement of the completeness of the resolved dependencies. Instead, it is a statement about the transitiveness of the resolved dependencies used to generate the cached artifacts. 

The effect of this statement is as we indicated in the issue -- the provenance (including the resolved dependencies) should be identical whether a cache is used or not.

I can replace this line with 

```suggestion
-   The provenance generated for the build MUST be identical whether or not the cache is
    used.
```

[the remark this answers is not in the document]
MarkLodato: I still find it confusing. I'm not sure other readers will interpret it properly. Could you be more explicit about the concern that you are attempting to address? Is it that, if a build requests an external resource, and it happens to be served from a cache rather than fetching externally, then the provenance ought to list it as though it were fetched externally?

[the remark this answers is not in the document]
arewm: Yes, I think that is an accurate statement. While this line of illustration quickly gets wordy, I feel like is still important to try to convey in the specification. I was trying to simplify it using a similar model from the previous point

> -   It MUST NOT be possible for one build to inject false entries into a build
    cache used by another build, also known as "cache poisoning". In other
    words, the output of the build MUST be identical whether or not the cache is
    used.

but a construction like this might actually be more parallel -- 

```suggestion
-   If the build platform is capable of providing the provenance for an external resource without a cache,
     then the provenance should remain unchanged if a cache is used. In other words, the output of the provenance
     MUST be identical whether or not the cache is used.
```

Just like the previous point indicated that the build output MUST be the same whether or not the cache is used, this point is indicating that the generated provenance MUST be the same whether the cache is used.

---

Going back to this comment briefly (this is a tangent, but the thought just occurred to me)

> I don't understand this. L3 has no requirement to list resolved dependencies. This would add one only if it the dependency comes from a cache? In other words, if you had not used the cache, it wouldn't be listed, but if you turn it on, now it has to be listed? Am I misunderstanding?

L3 has a requirement that

> Completeness of resolved dependencies is best effort.

As I understand it, this best-effort completeness is because the build platform might not _know_ about all dependencies (i.e. if the build pulls in dependencies without going through the platform itself). The best effort is for the provenance generated for the artifact as a whole, but I think that it would still be reasonable to require the platform to represent a complete set of dependencies that it knows about.

Therefore, if a build uses a build cache, then the build platform should know about all of the resolved dependencies and these dependencies resolved via the cache can therefore be complete in the provenance. 

That being said, we state this as the rest of the specification stands because we don't actually qualify what best effort means. If we do clarify this point then we might be able to be more specific on the cache-based provenance requirements.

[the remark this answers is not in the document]
MarkLodato: Yeah, your updated wording sounds good to me.

> As I understand it, this best-effort completeness is because the build platform might not know about all dependencies

Agreed.

> I think that it would still be reasonable to require the platform to represent a complete set of dependencies that it knows about.

I'm very hesitant to add anything more to L3 because it makes it adds cost to adopt. Perhaps that would make sense for a future L4, as an intermediate level before "fully hermetic"?

> That being said, we state this as the rest of the specification stands because we don't actually qualify what best effort means.

Yeah, since "best effort" is not yet defined, I'm inclined to leave well enough alone?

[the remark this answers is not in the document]
arewm: I will leave well enough alone and just use the latest suggestion from above with an additional comment on the lack of requirements on the build, namely that build L3 does not prohibit builds from resolving dependencies.

I don't currently see enough value in clarifying the completeness of known dependencies in a future L4 as an intermediate before "fully hermetic". I think that the _assumption_ that most at build L3 will be satisfying this is enough and that the next logical step is to a complete set of resolved dependencies.

docs/spec/v1.0/requirements.md
```diff
@@ -301,20 +301,28 @@ The build platform MUST guarantee the following:
...
     environment MUST be provisioned for each build.
+-   The build platform MUST NOT open services that allow for remote influence
+    unless all such interactions are captured as `externalParameters` in the
+    provenance.
+
+If the build platform leverages a cache for builds, it MUST guarantee the following:
+
 -   It MUST NOT be possible for one build to inject false entries into a build
     cache used by another build, also known as "cache poisoning". In other
     words, the output of the build MUST be identical whether or not the cache is
```
adityasaky, line 312: This is not a change in this PR but I think it may need to be revisited from the perspective of L3 vs L4. To ensure the use of a cache results in an identical output, we need the build to be deterministic. https://github.com/slsa-framework/slsa/issues/894#issuecomment-1627467364

docs/spec/v1.0/requirements.md
```diff
@@ -301,20 +301,28 @@ The build platform MUST guarantee the following:
...
     used.
--   The build platform MUST NOT open services that allow for remote influence
-    unless all such interactions are captured as `externalParameters` in the
-    provenance.
+-   If the build platform is capable of providing the provenance for an external
```
adityasaky, line 314: To confirm I'm parsing this right:

> If the build platform is capable of providing the provenance for an external resource without a cache

This is the provenance for the external resource itself, right?

> then the provenance SHOULD remain unchanged if a cache is used

And this is the provenance of the "top level" build that uses the external resource?

>  In other words, the output of the provenance MUST be identical whether or not the cache is used.

If we're building `foo` and the external resource is pre-built `bar` with its own provenance, we would just record `bar` in `resolvedDependencies` regardless of whether we fetched `bar` from some source upstream or a local cache. This allows for optionally verifying recursively the provenance of `bar`. I think this is fair and is separate from cached intermediate artifacts, because regardless of the cache, `bar` doesn't change. What changes is where `bar` comes from, a remote copy or a (relatively) local copy.

On that note, if we're fetching the _source_ of `bar` and building it during the build of `foo`, `bar` then becomes an intermediate artifact. `bar`'s _source_ would be recorded in `resolvedDependencies` but we'd have no provenance for `bar` unless the builder is configured to generate provenance for intermediate builds as well. `bar` can then be cached for future builds of `foo`, but now have the problem of testing if `bar` has an impact on `foo`. And as `bar` is not recorded (it's an intermediate artifact), even if we have its provenance from our builder, we don't make it reachable, though `foo`'s provenance is aware of the existence of the dependency through the source.

I think this digresses from the text I'm referring to but I do wonder if the text must be more specific as to when it applies. The first scenario where `bar` is an external resource with provenance available that may or may not be cached is more straightforward because at least `bar` does not change and `foo`'s provenance indicates a dependency on `bar` itself rather than its source.

docs/spec/v1.0/requirements.md
arewm: I don't think that a deterministic/reproducible build will be within the build track itself. Instead, I think that it will likely be in a new track (some initial discussions: #873). That being said, I think that the wording here is fine. If the build would have been built with a cache or if it would have been built without, the output should be the same. We are talking about a theoretical identicality not a practical one.

docs/spec/v1.0/requirements.md
arewm: According to the terminology,

> | Build caches | An intermediate artifact storage managed by the platform that maps intermediate artifacts to their explicit inputs. A build may share build caches with any subsequent build running on the platform.

This is one of two occurrences of the word `intermediate` in the specification. The other is in the provenance where it says 

> In most cases, [byproducts] SHOULD NOT contain all intermediate files generated during
the build. Instead, this SHOULD only contain files that are likely to be useful
later and that cannot be easily reproduced.

Therefore, your understanding seems to be consistent. Simply pulling external dependencies via some caching mechanism isn't what the specification considers to be a build cache. This might be more like what you were saying -- a mirror or registry of external resources. While that situation isn't the subject of this clarification, I think that its use case is still straightforward as the artifacts' `resolvedDependencies` need to be included in the provenance.

---

> `bar` can then be cached for future builds of `foo`, but now have the problem of testing if `bar` has an impact on `foo`. And as `bar` is not recorded (it's an intermediate artifact), even if we have its provenance from our builder, we don't make it reachable, though `foo`'s provenance is aware of the existence of the dependency through the source.

It feels like you are trying to identify a way for a consumer to verify the claims on the provenance being unchanged when a cache is in play. This is not an action that is required by consumers as they should be _trusting_ the platforms and _verifying_ artifacts [[ref](https://slsa.dev/spec/v1.0/principles#trust-platforms-verify-artifacts)]. The onus is on the build platform (and any certification process they undergo) to ensure that the requirement is met. If desired, a consumer can always request information from the build platform to verify claims, but this is expensive work.

---

Are there any additional changes that can be made in this location or others to better support the clarifications presented in this PR?

docs/spec/v1.0/requirements.md
```diff
@@ -301,20 +301,28 @@ The build platform MUST guarantee the following:
...
+    resource without a cache, then the provenance MUST remain unchanged if a cache
+    is used. In other words, the output of the provenance MUST be identical whether
```
arewm, line 316: @david-a-wheeler, does this address your concern mentioned in https://github.com/slsa-framework/slsa/issues/894#issuecomment-1628937179? @MarkLodato, is this introducing too strong of a statement?

docs/spec/v1.0/requirements.md
adityasaky: > Therefore, your understanding seems to be consistent. Simply pulling external dependencies via some caching mechanism isn't what the specification considers to be a build cache. This might be more like what you were saying -- a mirror or registry of external resources. While that situation isn't the subject of this clarification, I think that its use case is still straightforward as the artifacts' resolvedDependencies need to be included in the provenance.

The text highlighted in the PR here is for external resources and not intermediate artifacts. I think we agree that in that case, the provenance would not change if the external resource is cached (which is more like a local mirror or registry), and not the point of confusion in #894. It is probably good to have it in the text though!

> It feels like you are trying to identify a way for a consumer to verify the claims on the provenance being unchanged when a cache is in play. This is not an action that is required by consumers as they should be trusting the platforms and verifying artifacts [ref]. The onus is on the build platform (and any certification process they undergo) to ensure that the requirement is met.

While I'm also thinking about the verification, I started working on this from the perspective of the build platform, to nail down what must and must not be recorded to meet requirements. If the intent is to make a similar statement for a cache of intermediate artifacts, i.e., "the output of provenance MUST be identical whether of not a cache is used for intermediate artifacts", then the spec is essentially we have bit-for-bit equivalence of the final artifact whether or not the cache is used. Which brings us to "We are talking about a theoretical identicality not a practical one." but I think it's still confusing from the perspective of a developer of a build platform looking to implement the spec as this is not captured in the spec itself. I also suspect we do not want to discuss theoretical identicality / reproducibility etc in the spec yet. This is also why I'm not yet convinced we can move reproducibility out of this track entirely. My suggestion is for the spec to say something to the effect of:

"If the build platform uses a cache for intermediate artifacts generated during a build, unless the platform can guarantee the use of a cached resource has no impact on the final artifact, the provenance MUST record the use of cached resources in `resolvedDependencies`."

This would mean we don't have identical provenance metadata for builds with and without the cache but the platform has no ambiguity in what must be included in the provenance.

MarkLodato:
I'm kind of losing the rationale for this change. Could you update the PR description to explain? I'm no longer clear on the problem in the original text, so it's hard for me to say whether this updated text is an improvement.

arewm:
> I'm kind of losing the rationale for this change. Could you update the PR description to explain? I'm no longer clear on the problem in the original text, so it's hard for me to say whether this updated text is an improvement.

The PR description is updated to restate the problem that needs clarification.

MarkLodato:
Thank you for updating the PR description. That really helps define the problem.

I'm still not sure about the proposed changes. I find the new wording even more confusing than before, without resolving any of the original problems. In particular, I don't understand the bits about external communication and the provenance being identical.

I feel like there are two main problems:

1. The original wording of the Build L3 requirement is unclear when the rule about a "cache" applies. I think we can make an argument that, if there is some sort of automatic caching enabled by default, then it MUST NOT be susceptible to cache poisoning, the build output MUST be the same whether or not the cache is enabled, and (by virtue of the previous statement) the provenance SHOULD be identical whether or not the cache is enabled. When the cache is opt-in or user-provided, such as [actions/cache](https://github.com/actions/cache), then I'm not sure how much we can say at L3.
2. We are trying to provide two different things: _requirements_ at L3 about the minimum necessary level of tamper prevention, and _guidelines_ at all levels about how to construct provenance. That is basically what you're saying in the PR description, if I'm understanding correctly. So maybe we should just be more clear about that?

mlieberman85:
> I feel like the following description to me was most clear in the desired outcome (matching cache vs no cache)
> 
> * If the build platform is capable of providing the provenance for an external resource without a cache,
>   then the provenance should remain unchanged if a cache is used. In other words, the output of the provenance
>   MUST be identical whether or not the cache is used

Agreed, while clarifying that the subject (which exists outside the provenance but within the ite6 statement) doesn't have to be identical unless the builds are bit for bit reproducible.

arewm:
I resolved all open threads in this issue as they affected the readability of this PR. @adityasaky, I think most of the open discussion was between us, so if you want to re-state any of it, then can you open a new thread?

The part referenced above is the core part of this PR. This would involve removing the change to the provenance specification as well as lines 325-326.

docs/spec/v1.0/requirements.md
```diff
@@ -301,20 +301,29 @@ The build platform MUST guarantee the following:
...
+    or not the cache is used. Communication with the build cache MUST NOT be present
+    in `resolvedDependencies`.
```
joshuagl, line 318: WDYT to this set of minor tweaks? 
```suggestion
-   If the build platform is capable of providing the provenance information
    for an external resource when a cache is not in use, then the provenance
    information MUST remain unchanged if a cache is used. In other words, the
    information in the provenance MUST be identical whether or not the cache is
    used. Communication with the build cache MUST NOT be represented in
    `resolvedDependencies`.
```

joshuagl (changes requested):
Thanks for the attention to detail in this thread all and to @arewm for continuing to drive this change forward.

> The part referenced above is the core part of this PR. This would involve removing the change to the provenance specification as well as lines 325-326.

I think removing these hunks helps clarify the change.

arewm:
The PR has been distilled to the core change as previously indicated.

docs/spec/v1.0/requirements.md
```diff
@@ -307,13 +307,22 @@ The build platform MUST guarantee the following:
...
     cache used by another build, also known as "cache poisoning". In other
     words, the output of the build MUST be identical whether or not the cache is
     used.
--   The build platform MUST NOT open services that allow for remote influence
-    unless all such interactions are captured as `externalParameters` in the
-    provenance.
+-   If the build platform is capable of providing the provenance information for
+    an external resource when a cache is not in use, then the provenance
+    information MUST remain unchanged if a cache is used. In other words, the
+    information in the provenance MUST be identical whether or not the cache is
```
adityasaky, line 323: I think this + @mlieberman85's suggestion re the artifact hash addresses the reproducibility issue. There may be a separate question about whether intermediate artifacts that can impact the target artifact should be recorded for completeness but that's out of scope here.

arewm: @mlieberman85 , do you have a suggestion to the content of this PR related to your earlier comment?

mlieberman85: So, this makes sense to me, but I do wonder if it's not clear enough to someone who isn't familiar with what we're implying? Maybe just a clarification that unique identifiers such as checksums are what we're talking about here?

arewm: That seems like it might fit better in the previous bullet. Would a change like adding the last sentence as indicated suffice?

```
-  It MUST NOT be possible for one build to inject false entries into a build
    cache used by another build, also known as "cache poisoning". In other
    words, the output of the build MUST be identical whether or not the cache is
    used. This SHOULD be achieved using unique identifiers in the cache such
    as checksums
```

adityasaky: ```In other words, the output of the build MUST be identical whether or not the cache is used.```

This implicitly requires reproducibility, no? Unless this only applies to the provenance _predicate_, allowing the hash of the produced artifact to still change.

arewm:
I don't think that we definitively agreed on a strict improvement on the wording. I am going to close the PR and discussion can continue in the issue before another attempt of clarification is made.

Issue #902, opened 2023-07-07
PR name lint check fails on types with dashes


arewm:
The GitHub action https://github.com/amannn/action-semantic-pull-request uses `(\w*)` for identifying the type, so dashes are not supported.

This will cause all PRs of the following types to fail the linting:

```
spec-content
spec-editorial
non-spec
```

PR #890, opened 2023-06-27
impl: Update CONTRIBUTING.md.

kpk47:
- Add PR title/description requirements.
- Clarify review requirements and merge etiquette.
- Enable workflow to enforce PR title requirements.

fixes #821

PR #903, opened 2023-07-07
impl: Remove dashes from types in PR name lint

arewm:
resolves: #902

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -92,9 +92,9 @@ Review process:
 
 | Tag | Description | Waiting period | # Approvers |
 |---|---|---|---|
-| `spec-content` | A change to the meaning of the specification | 72h | 3 |
-| `spec-editorial` | A clarification to the specification that does not change its meaning | 24h | 2 |
-| `non-spec` | A change to a non-specification page. | 24h | 2 |
+| `specContent` | A change to the meaning of the specification | 72h | 3 |
+| `specEditorial` | A clarification to the specification that does not change its meaning | 24h | 2 |
+| `nonSpec` | A change to a non-specification page. | 24h | 2 |
 | `style` | A user-visible style or layout change. No context changes. | 0h | 1 |
 | `impl` | A user-invisible change, such as editing a README or the repo configuration. | 0h | 1 |
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,7 +100,7 @@ Review process:
 
 Note: PR authors with write access to the repo count as second or third
 approvers for their own PRs. For example, if the author of a PR with the
-`content` tag has write access to to the repo, then the PR only requires
+`content` tag has write access to the repo, then the PR only requires
 two additional approvers before merging. However, a PR with the `impl` tag
 always requires one reviewer, even if the author has write access.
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
kpk47, line 1: Also change the reference to spec-content on L103.

arewm: ah, thanks!

kpk47 (approved):

CONTRIBUTING.md
```diff
@@ -92,15 +92,15 @@ Review process:
 
 | Tag | Description | Waiting period | # Approvers |
 |---|---|---|---|
-| `spec-content` | A change to the meaning of the specification | 72h | 3 |
-| `spec-editorial` | A clarification to the specification that does not change its meaning | 24h | 2 |
-| `non-spec` | A change to a non-specification page. | 24h | 2 |
+| `specContent` | A change to the meaning of the specification | 72h | 3 |
+| `specEditorial` | A clarification to the specification that does not change its meaning | 24h | 2 |
+| `nonSpec` | A change to a non-specification page. | 24h | 2 |
```
MarkLodato, line 97: I find these hard to read, and I believe most conventional commits tools treat these as case-insensitive. (The actual [specification](https://www.conventionalcommits.org/en/v1.0.0/) is ambiguous about case.)

What about `spec`, `editorial`, and `nonspec`, which are more concise? That the first two were the original names in PR #890, which I think were fine.

kpk47: We didn't use `spec`, `editorial`, and `nonspec` in #890 because there are three categories, but `spec` vs `nonspec` implies only two. How about `content`, `editorial`, and `nonspec`? Or `content`, `editorial`, and `docs`?

joshuagl: I prefer `content`, `editorial`, and `nonspec`.

MarkLodato: `content`, `editorial`, `nonspec` SGTM

lehors: I too am not a fan of the camel case names here. I'd prefer the names @joshuagl suggested.

arewm: sounds good to me, thanks.

joshuagl (approved):
thanks!

MarkLodato (approved):

CONTRIBUTING.md
```diff
@@ -92,15 +92,15 @@ Review process:
...
+| `content` | A change to the meaning of the specification | 72h | 3 |
+| `editorial` | A clarification to the specification that does not change its meaning | 24h | 2 |
+| `nonspec` | A change to a non-specification page. | 24h | 2 |
 | `style` | A user-visible style or layout change. No context changes. | 0h | 1 |
 | `impl` | A user-invisible change, such as editing a README or the repo configuration. | 0h | 1 |
 
 Note: PR authors with write access to the repo count as second or third
 approvers for their own PRs. For example, if the author of a PR with the
-`spec-content` tag has write access to to the repo, then the PR only requires
+`content` tag has write access to to the repo, then the PR only requires
```
lehors, line 103: While you're at it, please, fix the following typo so that I don't have to submit a PR just for that! :-)
```suggestion
`content` tag has write access to the repo, then the PR only requires
```
Thanks.

arewm: done

arewm:
This PR is ready to merge.

Issue #742, opened 2023-03-27
Make adjustments so that changes are made via branches - Meta issue - non-spec related


NicoleSchwartz:
As discussed today it would be great if someone could look at making adjustments so that changes are made via branches. This is a meta issue and not directly related to the specification (independent).

This could be done using GitHub Actions and/or Netlify.

This would be a great issue for those wanting to contribute with CI-CD/DevOps skills but unsure where to start helping!

arewm:
If using GitHub Actions, it seems reasonable that it may be possible to update some portion of a `gh-pages` branch based on the source branch. For example, a `v1` branch might only push the content in a `v1` directory. Changes to a `main` branch would only update the files present outside of a version folder while also putting the versioned content in a `latest` "version."

If using Netlify, it seems like it _should_ be [straight forward](https://docs.netlify.com/site-deploys/overview/#branches-and-deploys) but there might be a breaking change to using subdomains instead of modifying the URL path.

MarkLodato:
Related issue: #266 covers switching to Netlify for hosting. We can still make that change if desired. If we don't, the solution needs to keep working for our Netlify preview links, which we use extensively.

Prior art: https://github.com/iamwillbar/spdx-spec/blob/development/v2.3/.github/workflows/publish_common.yml is how spdx is sets up their GitHub Actions build

MarkLodato:
@chtiangg is working on this now

MarkLodato:
Status update: I'm working on this now while @chtiangg is out of town. However, one small snafu: `git subtree split` eliminates lot of merge commits, so the history becomes basically unreadable. (Currently you can do `git log --first-parent` to make the history at least somewhat readable.) I'll give `git filter-repo` a try...

zachariahcox:
challenges
* netlify preview will be complicated with this flow -- changes targeting old release branches should build the "main" version of the site with the proposed changes replacing a target release. 
* need to update linter scripts (and all other places where we look at markdown)

PR #905, opened 2023-07-07
content: add a verifier version field to VSA

### LINKED ISSUE #809: VSA: can we be more consistent with SLSA fields and VSA fields
#### builder and verifier id.
The builder.id is:
```json
"builder": {
                "id": string,
                "builderDependencies": [ ...#ResourceDescriptor ],
                "version": { ...string },
            },
```

but in VSA, the verifier.id is:
```json
"verifier": {
    "id": "<URI>"
  },
```

Those fields should be aligned in their definition. Right now, there's no way to version the verifier. At least, can we add a version field?

#### resource_uri

The provenance uses ResourceDescriptor, but VSA uses a string `resourceUri` only. Can we use ResourceDescriptor for the resource in the VSA?

kpk47:
Add a verifier version field to the VSA schema.

This change Increases consistency between `provenance.builder` and `vsa.verifier`, and it allows VSA producers/consumers to respond to known flaws in verification tools.

fixes #809


docs/verification_summary/v1.md
```diff
@@ -67,7 +67,11 @@ to establish minimum requirements on dependencies SLSA levels may use
 "predicate": {
   // Required
   "verifier": {
-    "id": "<URI>"
+    "id": "<URI>",
+    "version": {
+      ...
+      "<COMPONENT>": "<VERSION>",
```
MarkLodato, line 73:
```suggestion
      "<COMPONENT>": "<VERSION>",
      ...
```

nit: starting with `...` is confusing

docs/verification_summary/v1.md
```diff
@@ -213,7 +222,12 @@ WARNING: This is just for demonstration purposes.
 "predicateType": "https://slsa.dev/verification_summary/v1",
 "predicate": {
   "verifier": {
-    "id": "https://example.com/publication_verifier"
+    "id": "https://example.com/publication_verifier",
+    "version": {
+      "foo": "foo-1.0",
+      "bar": "bar-3.6",
+      "baz": "baz-alpha00"
```
MarkLodato, line 229: Can we come up with a realistic example of how this might be used? It's not clear to me, so an example would help.

docs/verification_summary/v1.md
```diff
@@ -131,6 +135,11 @@ of the other top-level fields, such as `subject`, see [Statement]._
 
 > URI indicating the verifier’s identity.
 
+<a id="verifier.version"></a>
+`verifier.version` _map (string→string), required_
```
MarkLodato, line 139:
```suggestion
`verifier.version` _map (string→string), optional_
```

MarkLodato (changes requested):
In the PR description, could you explain why this is needed? Issues are often long and meandering, so briefly summarizing the rationale is helpful for reviews and for posterity.

docs/verification_summary/v1.md
kpk47: Done. I switched it to slsa-verifier and the corresponding GHA installer.

lehors:
Given that this is requiring new info shouldn't that trigger some version change of the format?

docs/verification_summary/v1.md
```diff
@@ -213,7 +222,11 @@ WARNING: This is just for demonstration purposes.
 "predicateType": "https://slsa.dev/verification_summary/v1",
 "predicate": {
   "verifier": {
-    "id": "https://example.com/publication_verifier"
+    "id": "https://example.com/publication_verifier",
+    "version": {
```
laurentsimon, line 226: I still feel that using resource descriptor is more flexible in general, especially if we need to update this structure in the future. In this example, `slsa-verifier-linux-amd64` is kinda arbitrary naming. If our goal is to help with vuln management, using more conventional naming, like purl, etc may be beneficial. Resource descriptors offer this flexibility.

MarkLodato (approved):

MarkLodato:
> Given that this is requiring new info shouldn't that trigger some version change of the format?

There are no required fields. (An earlier version said required, but it was a mistake.)

So this doesn't require a major version bump, but based on this week's spec meeting, all `content` changes should have a changelog entry and perhaps a minor version bump.

lehors:
> > Given that this is requiring new info shouldn't that trigger some version change of the format?
> 
> There are no required fields. (An earlier version said required, but it was a mistake.)

Why does line #68 read: "// Required" ?
Shouldn't this comment be deleted then?

MarkLodato:
> Why does line # 68 read: "// Required" ? Shouldn't this comment be deleted then?

The `verifier` (and `verifier.id`) was and still is required, but the new `verifier.version` subfield is not. Thus the following was and still is acceptable:

```json
"verifier": { "id": "..." }
```

But point taken - we should clarify in the docs since the comment can clearly be misread.

docs/verification_summary/v1.md
kpk47: I don't think that `ResourceDescriptor` is the right type for tracking versions since it doesn't appear to have a version field. Also, part of the motivation for this PR is making the `verifier` VSA more consistent with the `builder` provenance field. `builder` uses `map(string->string)`, so I think it's best to keep that type here.

I'm going to keep `map(string->string)` in this PR, and we can address whether or not to change both fields to `ResourceDescriptor` separately.

kpk47:
> > Why does line # 68 read: "// Required" ? Shouldn't this comment be deleted then?
> 
> The `verifier` (and `verifier.id`) was and still is required, but the new `verifier.version` subfield is not. Thus the following was and still is acceptable:
> 
> ```json
> "verifier": { "id": "..." }
> ```
> 
> But point taken - we should clarify in the docs since the comment can clearly be misread.

I don't see a good way to clarify in the text since the fields are already clearly labeled required vs optional. I deleted the comment instead.

kpk47:
> > Given that this is requiring new info shouldn't that trigger some version change of the format?
> 
> There are no required fields. (An earlier version said required, but it was a mistake.)
> 
> So this doesn't require a major version bump, but based on this week's spec meeting, all `content` changes should have a changelog entry and perhaps a minor version bump.

I added a changelog entry with a minor version bump, as required by the versioning rules (https://slsa.dev/spec-stages#versioning). 

Given the version bump, we shouldn't merge this PR until we're ready to have v1.1 on the website.

lehors:
> I don't see a good way to clarify in the text since the fields are already clearly labeled required vs optional. I deleted the comment instead.

That works. Thanks!

lehors:
> I added a changelog entry with a minor version bump, as required by the versioning rules (https://slsa.dev/spec-stages#versioning).
> 
> Given the version bump, we shouldn't merge this PR until we're ready to have v1.1 on the website.

Shouldn't we have this on a branch specific to the next version like we did when we were working on 1.0 before publication?

MarkLodato:
> Shouldn't we have this on a branch specific to the next version like we did when we were working on 1.0 before publication?

Yes, this is blocked by #742.

MarkLodato:
The v1.1 directory now exists as of #942.

kpk47:
I moved the changes into the v1.1 directory, so this PR is ready for re-review.

docs/spec/v1.1/verification_summary.md
```diff
@@ -138,6 +141,11 @@ of the other top-level fields, such as `subject`, see [Statement]._
 
 > URI indicating the verifier’s identity.
 
+<a id="verifier.version"></a>
+`verifier.version` _map (string→string), optional_
```
joshuagl, line 145: nit: I don't think we have any uses of unicode in our Markdown, and instead let the Markdown/browser stack take care of convertion `->` to -> (i.e. in the "Change history" section of provenance.md).

```suggestion
`verifier.version` _map (string->string), optional_
```

joshuagl (commented):
This LGTM, thanks. However, as this is a content change, could you please add a changelog entry (per [contributing](https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md#changelog-entry))?

joshuagl (approved):
LGTM, thanks!

MarkLodato (approved):

PR #907, opened 2023-07-07
editorial: clarify which provenance fields are required

kpk47:
- Add a column to the field table for whether a field is REQUIRED or OPTIONAL

### DIFF
--- a/docs/provenance/v1.md
+++ b/docs/provenance/v1.md
@@ -132,19 +132,17 @@ Link: [provenance.proto](schema/v1/provenance.proto)
 
 [Provenance]: #provenance
 
-REQUIRED for SLSA Build L1: `buildDefinition`, `runDetails`
-
 <table>
-<tr><th>Field<th>Type<th>Description
+<tr><th>Field<th>Type<th>Necessity<th>Description
 
 <tr id="buildDefinition"><td><code>buildDefinition</code>
-<td><a href="#builddefinition">BuildDefinition</a><td>
+<td><a href="#builddefinition">BuildDefinition</a><td>REQUIRED<td>
 
 The input to the build. The accuracy and completeness are implied by
 `runDetails.builder.id`.
 
 <tr id="runDetails"><td><code>runDetails</code>
-<td><a href="#rundetails">RunDetails</a><td>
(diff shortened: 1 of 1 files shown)


docs/provenance/v1.md
```diff
@@ -198,7 +194,7 @@ parameters because the build platform is already trusted, and in many cases it i
 not practical to do so.
 
 <tr id="resolvedDependencies"><td><code>resolvedDependencies</code>
-<td>array (<a href="https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md">ResourceDescriptor</a>)<td>
+<td>array (<a href="https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md">ResourceDescriptor</a>)<td>OPTIONAL<td>
```
joshuagl, line 197: `resolvedDependencies` is not really optional at SLSA Build L2+, however it is "best effort" and may be tenant generated at Build L2 and incomplete at Build L3.

joshuagl (changes requested):
I like the idea. However, in some cases optional vs. required gets fuzzier at higher build levels such as the `resolvedDependencies` case I commented on inline.

Perhaps we could have the table column indicate at what SLSA build level the field is required?

docs/provenance/v1.md
```diff
@@ -132,19 +132,17 @@ Link: [provenance.proto](schema/v1/provenance.proto)
 
 [Provenance]: #provenance
 
-REQUIRED for SLSA Build L1: `buildDefinition`, `runDetails`
-
 <table>
-<tr><th>Field<th>Type<th>Description
+<tr><th>Field<th>Type<th>Necessity<th>Description
```
MarkLodato, line 136: Adding a new column causes the text to squish in to a fairly narrow column, which is even worse on mobile. That's why I hadn't added it originally.

Is there another way we could format this to be easier to read? Perhaps included REQUIRED in the Description or Type column? Not sure if that's better or worse.

#### Before

before

#### After

after

MarkLodato (changes requested):
Could you update the PR description to explain *why* this change is being made? That is not obvious to me, so it's hard to judge what the best solution is.

PR #908, opened 2023-07-08
editorial: better explain the build model

MarkLodato:
Text changes:

-   Use "invokes" instead of "defines" to avoid confusion.
-   Clarify that the invocation can be direct or through a trigger.
-   Clarify that build steps are under the tenant control and that the
    platform isolates between build environments.
-   Use consistent terminology, e.g. adding "build" before "platform".
-   Increase clarity through some minor rewording.

Reformatting:

-   Convert the paragraph to an ordered list and italicize the main use
    of each term, to enhance readability and to make it easier to match
    up with the diagram and the table.
-   Move all text to be below the diagram and above the table, to make
    it easier to follow.

Diagram tweaks:

-   Draw a line from interface to control plane, to match the text.
-   Use "dependencies" instead of "resolved dependencies", for
    consistency with the text.
-   Use the same font size for all inputs and outputs.
-   Use solid lines between control plane, build environment, and
    build caches, to avoid visual clutter. There was no reason for
    these to be drawn differently.
-   Draw Build Caches in black, since there was no particular reason
    to draw it in green.

Issue: #714


docs/spec/v1.0/terminology.md
```diff
@@ -66,18 +66,34 @@ chain.
...
+execution is independent.
+
+1.  A tenant invokes the build by specifying *external parameters* through an
+    interface, either directly or via some trigger. Usually, at least one of
+    these parameters is a reference to a dependency, as noted below.
+2.  The build platform's *control plane* interprets these external parameters,
+    fetches an initial set of *dependencies*, initializes a *build environment*,
+    and then starts the execution within that environment.
+3.  The build then performs arbitrary steps, which might include fetching
+    additional *dependencies*, and then writes one or more *output* artifacts.
```
lehors, line 81:
```suggestion
    additional *dependencies*, and then produces one or more *output* artifacts.
```

docs/spec/v1.0/terminology.md
```diff
@@ -66,18 +66,34 @@ chain.
...
+    The steps within the build environment are under the tenant's control.
+    The build platform isolates between build environments to some
```
lehors, line 83:
```suggestion
    The build platform isolates build environments from one another to some
```

docs/spec/v1.0/terminology.md
```diff
@@ -66,18 +66,34 @@ chain.
...
+    degree (which is measured by the SLSA Build Level).
+4.  Finally, for SLSA Build L2+, the control plane outputs *provenance*
+    describing this whole process.
+
+Notably, there is no formal notion of "source" in the build model, just
+parameters and dependencies. Most build platforms have an explicit "source"
+artifact to be built, which is often a git repository; in the build model, the
```
lehors, line 90: This reads as if it's the source artifact that is to be built.
```suggestion
artifact to build from, which is often a git repository; in the build model, the
```

lehors (commented):
LGTM, but made a few suggested edits.

docs/spec/v1.0/terminology.md
```diff
@@ -66,18 +66,34 @@ chain.
...
+3.  The build then performs arbitrary steps, which might include fetching
+    additional *dependencies*, and then writes one or more *output* artifacts.
+    The steps within the build environment are under the tenant's control.
+    The build platform isolates between build environments to some
+    degree (which is measured by the SLSA Build Level).
+4.  Finally, for SLSA Build L2+, the control plane outputs *provenance*
+    describing this whole process.
+
+Notably, there is no formal notion of "source" in the build model, just
+parameters and dependencies. Most build platforms have an explicit "source"
```
joshuagl, line 89:
```suggestion
external parameters and dependencies. Most build platforms have an explicit "source"
```
should we also be more consistent in referring to parameters as "external parameters"?

docs/spec/v1.0/terminology.md
```diff
@@ -66,18 +66,34 @@ chain.
...
+artifact to be built, which is often a git repository; in the build model, the
+reference to this artifact is a parameter while the artifact itself is a
```
joshuagl, line 91:
```suggestion
reference to this artifact is an external parameter while the artifact itself is a
```
some comment on consistency, and therefore clarity, of use of external parameter over parameter.

docs/spec/v1.0/terminology.md
```diff
@@ -66,18 +66,34 @@ chain.
...
+dependency.
+
+For examples on how this model applies to real-world build platforms, see [index
```
joshuagl, line 94: nit:
```suggestion
For examples of how this model applies to real-world build platforms, see [index
```

joshuagl (changes requested):

lehors (approved):
LGTM

docs/spec/v1.0/terminology.md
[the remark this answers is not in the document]
MarkLodato: Yeah, I wasn't sure how to handle it since all three bullets mentioned "dependency" but none really defined it. (I missed the italics on the first bullet.)

What do you think of the change I just pushed, adding a parenthetical comment to the first bullet explaining that params are literals while dependencies are artifacts, and removing italics from bullets 2 and 3?

If you like that phrasing, I could also add a legend to the diagram explaining that, but that may make it too busy.

joshuagl (approved):

docs/spec/v1.0/terminology.md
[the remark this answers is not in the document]
joshuagl: This phrasing LGTM, thanks! Shall we consider a change to the diagram in a separate PR?

PR #915, opened 2023-07-12
impl: improve readability of CONTRIBUTING

MarkLodato:
Apply some minor editorial changes to improve readability without
changing the meaning:

- Split out "pull request conventions" into a separate section, since it
  was a bit buried in "review and approval." The former is what the user
  does before the PR is sent, while the latter is what to expect after
  it is sent.
- Remind the reader in the "pull request conventions" that every commit
  needs to be signed off but otherwise does not matter due to
  squash-and-merge. This was easy to miss before.
- Explain that the table is in order of precedence and that a maintainer
  will update the tag if needed.
- Apply various wording improvements.

## Note to reviewers

This PR is hard to review as a whole since it moves AND edits the same hunk. I recommend viewing each commit individually. The first commit just moves the text verbatim while the second edits it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,8 +21,11 @@ Alternately you can look for issues with a [`shovel-ready`](https://github.com/s
 
 ## Proposing changes
 
-Unless a change is small enough to be fully discussed in a pull request, we
-recommend the following process to propose and reach agreement on changes:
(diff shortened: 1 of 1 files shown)


joshuagl (approved):
LGTM. Thanks for making it easier to link to the PR conventions!

Issue #727, opened 2023-03-24
"Verifying Artifacts": declare a preference for signing methodologies that rely on transparency logs


di:
>> If we do that, do you think that it would make sense for https://slsa.dev/spec/v1.0/verifying-artifacts to declare a preference for signing methodologies that rely on transparency logs?
>
> Yeah, that sounds like a good idea.

_Originally posted by @MarkLodato in https://github.com/slsa-framework/slsa/pull/673#discussion_r1146812053_

joshuagl:
I'm curious whether folks still think this declaration of preference for TL dependent signing methodologies belongs in the _[verifying artifacts]_ page?

It probably doesn't belong in requirements, because it's not strictly required, but it feels like something that build platforms should be accounting for when designing [provenance generation](https://slsa.dev/spec/v1.0/requirements#provenance-generation). Another document aimed at build platforms is the [distributing provenance](https://slsa.dev/spec/v1.0/distributing-provenance) page, where this preference for TL based signing methodologies was originally included but didn't seem to fit.

My concern is that the [verifying artifacts] page is in the purview of consumers, not build platforms, and thus making recommendations about signature production in the place where we describe verification doesn't feel quite right.

Equally, if consensus is that [verifying artifacts] is the best place to put it, I'll make that change ASAP. We might mention it in the discussion of threat mitigation, specifically when discussing Threat E; or when discussing verification of the envelope's signature.

WDYT?

[verifying artifacts]: https://slsa.dev/spec/v1.0/verifying-artifacts

joshuagl:
Another place we might mention this is the [How to SLSA for implementers](https://slsa.dev/how-to-infra) page under the [Build platform](https://slsa.dev/how-to-infra#build-platform) heading. 

However, given that section provides guided reading for build platform implementers and points to [Verifying build platforms](https://slsa.dev/spec/v1.0/verifying-systems), [Producing artifacts](https://slsa.dev/spec/v1.0/requirements), and [Provenance](https://slsa.dev/provenance/v1) pages of the spec I think we should add this to one of those pages.

I'm leaning toward just stating this preference in the [Authentic](https://slsa.dev/spec/v1.0/requirements#provenance-authentic) requirement as this is probably the main place where signing (as opposed to verifying) is discussed.

arewm:
Why would the preference be for transparency logs instead of just timestamps? https://docs.sigstore.dev/cosign/timestamps/

[...]

PR #916, opened 2023-07-12
editorial: declare a preference for TL-based signing

joshuagl:
We encourage the use of transparency log based signing methodologies because they provide observability which makes it easier to detect key compromise, and constraints (time stamping) which make detected compromises easier to remediate.

We prefer them, rather than require them, because there is no one-size fits all solution given numerous constraints on operational load, confidentiality, etc.

Fixes: #727 

 Signed-off-by: Joshua Lock <joshua.lock@uk.verizon.com>


arewm (approved):
That wording seems reasonable, thanks!

MarkLodato (approved):

PR #925, opened 2023-07-20
impl: use Netlify redirects instead of HTML

MarkLodato:
Switch to Netlify's HTTP-level redirects instead of HTML-level
`<meta refresh>` redirects. (Historical note: we originally used the
latter because GitHub Pages did not support HTTP-level redirects.)

Advantages:

- Preserves URL fragments.
  Example: slsa.dev/levels#build-l3 no longer drops the `#build-l3`.
- Supports directory redirection path components after the redirect.
  Example: slsa.dev/spec/v1/levels now works.
- Supports temporary redirects (302). The problem with `<meta refresh>`
  is that they are always interpreted as permanent, which is a problem
  when we do a version bump (e.g. /levels -> /spec/v1.0/levels).
- Cleans up our directory tree, which now only contains true content.

### DIFF
--- a/docs/_layouts/redirect.html
+++ b/docs/_layouts/redirect.html
@@ -1,7 +1,0 @@
-<!DOCTYPE html>
-<html lang="en-US">
-  <title>Redirecting&hellip;</title>
-  <meta http-equiv="refresh" content="0; url={{ page.redirect_to_url }}">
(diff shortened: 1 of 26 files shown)


joshuagl (approved):
Neat!

PR #936, opened 2023-08-02
impl: move images to versioned spec directory

MarkLodato:
Move (or copy) all of the images referenced by `/spec/v1.0/*` to a new
`/spec/v1.0/images/` directory. This makes it easier to create new
versions of the spec. Also set up redirects so that old URLs continue to
work. (`supply-chain-model.svg` is copied verbatim from /images.)

To keep this PR focused, do not touch older versions (`v0.?` and
`v1.0-rc?`). Those versions won't affect the versioning of the spec
post-v1.0, so they don't matter as much. That said, a future PR could
optionally clean up those directories as well.


joshuagl (approved):
LGTM, thanks

PR #942, opened 2023-08-16
content: create v1.1 directory

MarkLodato:
The spec/v1.1/ directory is a copy of spec/v1.0/, while
_data/nav/v1.1.yml is a copy of v1.0.yml with "1.0" replaced by "1.1".

```bash
$ git ls-tree HEAD:docs/spec | grep 'v1.[01]$'
040000 tree 52008aa2d63ee43de55af463c016ea2b6771433a    v1.0
040000 tree 52008aa2d63ee43de55af463c016ea2b6771433a    v1.1
```


MarkLodato:
I don't think our community guidelines make sense for draft branches. While this is technically "content", it is for a new release that is not yet approved so the full 3 reviewers + 72 hours doesn't make sense.

joshuagl (approved):
LGTM

> I don't think our community guidelines make sense for draft branches. While this is technically "content", it is for a new release that is not yet approved so the full 3 reviewers + 72 hours doesn't make sense.

Yes, agreed. They are only guidelines and have several edge cases. I think it's fine to have only two reviewers/approvers for this change, so long as edits to the spec meet the documented criteria.

PR #944, opened 2023-08-16
editorial: fix typo on threats and mitigations page

joshuagl:
The v1.0 and v1.0-rc2 versions of the page have the same typo: "protects agains"

Signed-off-by: Joshua Lock <joshuagloe@gmail.com>


MarkLodato (approved):
Thanks for the fix! Changed the type to `editorial` since it is a user-visible change.

Issue #943, opened 2023-08-16
Update or remove motivating example


joshuagl:
The [motivating example](https://slsa.dev/example) page isn't included in the site's navigation and has fallen out of date with the v1.0 changes to the specification (switch to tracks and removal of SLSA Build L4).

Should we remove the page, or update it? We have [overview](https://slsa.dev/spec/v1.0/threats-overview) and [detailed](https://slsa.dev/spec/v1.0/threats) documentation of threats and mitigations that form part of the spec, so the example is perhaps not as useful as it was in the past.

Any objection to just removoing it?

MarkLodato:
Removing SGTM. It's still in version history if we need it.

PR #945, opened 2023-08-17
nonspec: remove unused and outdated motivating example

joshuagl:
The motivating example page is not referenced/linked to by any other
content and has fallen out of date with SLSA v1.

Fixes: #943

Signed-off-by: joshua.lock@uk.verizon.com


MarkLodato (commented):
+1 to removing.

Is it worth redirecting to https://github.com/slsa-framework/slsa/blob/a25364a3b312f572e4620c812820249853bb40e5/docs/example.md? In general I prefer keeping URLs working forever, but in this case I don't know if anyone cares.

joshuagl:
I agree with the goal of having URLs continue to work, so I will set up the redirect even though I'm not sure anyone cares.

MarkLodato (approved):

PR #948, opened 2023-08-17
editorial: clarify build platform reqs in levels.md

MarkLodato:
Update levels.md to align with requirements.md and to make it more clear how the requirements on the build platform differ between levels:

Build L1:

- Problem: levels.md incorrectly implied that that the *platform* must generate the provenance.

- Fix: Update levels.md to say that provenance must simply "exist", matching the wording on requirements.md. Also simplify the wording while we're at it.

Build L2:

- Problem: The phrase "ensures the trustworthiness of the provenance" could be misread as being a general requirement for L2 builds, whereas the intention was just as a qualifier on "some equivalent platform." Also, the intent of the build platform generating and signing the provenance was unclear, further exacerbating the problem above.

- Fix: Simplify this bullet to focus on the intent of (a) builds running on dedicated infrastructure and (b) tying the provenance to that infrastructure.

Partially addresses #863.

### DIFF
--- a/docs/spec/v1.0/levels.md
+++ b/docs/spec/v1.0/levels.md
@@ -100,9 +100,8 @@ SLSA---other than tamper protection---without changing their build workflows.
 -   Software producer follows a consistent build process so that others can form
     expectations about what a "correct" build looks like.
 
--   Build platform automatically generates [provenance] describing how the
-    artifact was built, including: what entity built the package, what build
-    process they used, and what the top-level input to the build were.
+-   [Provenance] exists describing how the artifact was built, including the
+    build platform, build process, and top-level inputs.
 
 -   Software producer distributes provenance to consumers, preferably using a
     convention determined by the package ecosystem.
@@ -151,10 +150,9 @@ platform itself required by [Build L3].
 
 All of [Build L1], plus:
 
--   The build runs on a hosted build platform that generates and signs[^sign] the
-    provenance itself. This may be the original build, an after-the-fact
-    reproducible build, or some equivalent platform that ensures the
-    trustworthiness of the provenance.
+-   Build platform runs on dedicated infrastructure, not an individual's
+    workstation, and the provenance is tied to that infrastructure through
+    a digital signature[^sign].
 
 -   Downstream verification of provenance includes validating the authenticity
     of the provenance.
@@ -178,8 +176,9 @@ All of [Build L1], plus:
 </section>
 <section id="build-l3">
 
-[^sign]: Alternate means of verifying the authenticity of the provenance are
-    also acceptable.
+[^sign]: Usually this means that the provenance is signed by a key that is only
+    accessible to the build platform, but alternate means of verifying the
+    authenticity of the provenance are also acceptable.
 
 ### Build L3: Hardened builds
 


MarkLodato:
@behnazh-w does this address your concern from #863?

joshuagl (approved):

sudo-bmitch (approved):

docs/spec/v1.0/levels.md
```diff
@@ -151,10 +150,9 @@ platform itself required by [Build L3].
 
 All of [Build L1], plus:
 
--   The build runs on a hosted build platform that generates and signs[^sign] the
-    provenance itself. This may be the original build, an after-the-fact
-    reproducible build, or some equivalent platform that ensures the
-    trustworthiness of the provenance.
+-   Build platform runs on dedicated infrastructure, not an individual's
+    workstation, and the provenance is tied to that infrastructure through
+    a digital signature[^sign].
```
behnazh-w, line 155: Let's take `npm`'s provenance generation on GitHub Actions as a concrete example to see if it meets L2. The JWT tokens are available to the `step` running the build scripts and publishing the provenance and not a dedicated infrastructure (e.g., a reusable workflow hosted on a trusted repository). As discussed [here](https://github.com/slsa-framework/slsa-github-generator/issues/2344#issuecomment-1660719908), `npm` dependencies can potentially grab the tokens and sign arbitrary packages. So, the digital signature might not be tied to the infrastructure. Is it fair to say that `npm` right now does not meet this L2 requirement?

MarkLodato: No, that is not the correct interpretation.

Consider a GitHub Actions workflow in repository `https://github.com/example/foo` on branch `main` and path `.github/workflow/build.yml`. The requirement is that it MUST NOT be possible to generate (valid) provenance claiming to come from {GitHub Actions, `https://github.com/example/foo`, `main`, `.github/workflow/build.yml`} unless it actually did originate from that workflow. The sigstore-based solution meets this requirement. The workflow can generate provenance with arbitrary `subject` values, but it cannot change the `predicate`.

The attack you're describing is out of scope. If an attacker compromises a workflow to gain code execution, such as by compromising a build-time dependency, then all bets are off. Not only can they can generate provenance for arbitrary binaries, but they can also modify the target artifact itself. The SLSA Build track does not cover this threat. It is up to the individual project owner ("producer" in the SLSA model) to properly secure their build workflow. The SLSA Build Track does make a broad claim that a build is "secure", but a more narrow claim that a subject artifact really did come from a particular workflow.

Does that help clarify?

behnazh-w: I still find potential contradictions especially w.r.t. trust boundaries, and assumptions made on control plane isolation at level 2.

>The attack you're describing is out of scope. If an attacker compromises a workflow to gain code execution, such as by compromising a build-time dependency, then all bets are off. Not only can they can generate provenance for arbitrary binaries, but they can also modify the target artifact itself. 

I'm not sure it is out of scope because it looks like in the `npm` case the control plane is ALWAYS making the secrets available to the untrusted build step, while it shouldn't: if we look at the model described [here](https://slsa.dev/spec/v1.0/provenance#model), the `Control Plane` is isolated from the "untrusted" `Build Process` component and the [Provenance is Authentic requirement](https://slsa.dev/spec/v1.0/requirements) at level 2 states that:

>Accuracy: The provenance MUST be generated by the control plane (i.e. within the trust boundary [identified in the provenance](https://slsa.dev/spec/v1.0/provenance#model)) and not by a tenant of the build platform (i.e. outside the trust boundary).

So, the main issue in the `npm` case seems to be that the `Control Plane` is making secrets available to build step in the GitHub Actions workflow, where the provenance is generated and signed using a sigstore-based solution, and not because dependencies might have malicious behavior. 

The bottom line is that we can't trust the generated provenance claiming that a subject artifact really did come from a particular workflow unless we break the assumption that the `Build Process` can be trusted (while it would have been possible if the build step was running as part of a dedicated reusable workflow, which I would call the `Control Plane`). In other words, if you think this attack I'm describing is out of scope, the [Provenance is Authentic requirement](https://slsa.dev/spec/v1.0/requirements) at level 2 might need to be revisited as well.

MarkLodato: I think the confusion lies in how the Sigstore solution works, paired with slightly overly prescriptive wording in the SLSA spec.

In short:

- GitHub generates a short-lived OIDC token containing the provenance predicate, signed by GitHub.
- Sigstore verifies the OIDC token and converts it to a long-lived x.509 certificate, signed by Sigstore.
- Build Process converts the x.509 certificate into a SLSA Provenance attestation and signs it using the x.509 certificate. The full attestation contains **two copies** of the provenance: one in the x.509 certificate and one in the DSSE payload.
- slsa-verifier checks that the payload agrees with x.509 certificate.

The "secret" that is exposed is the private key of the x.509 certificate. But this cannot be used to sign any provenance unless the contents match the x.509 certificate. The only thing that can be done is to sign arbitrary hashes.

---

If you agree with that, then I think the question is how we update the specification to make it more clear that this solution meets the requirements. I think the problematic statement is this:

> The provenance MUST be generated by the control plane

In the Sigstore case, the "provenance" is really the x.509 certificate and the "control plane" is GitHub + Sigstore. The payload just contains duplicate information, plus the output artifact hash.

Two solutions come to mind:

- Call out this interpretation in the spec to make it more clear.
- Change the wording to be more about the requirement around authenticity, i.e. not that "provenance" is "generated" by the control plane, but rather that it is not possible for anyone but the control plane to alter its contents.

Thoughts?

joshuagl: I like the latter solution, "Change the wording to be more about the requirement around authenticity..."

I also think this kind of applied reasoning about the spec is really informative and I wish we could share it outside of this issue. Could we translate the conversation here into a blog post?

behnazh-w: @MarkLodato
> The "secret" that is exposed is the private key of the x.509 certificate. But this cannot be used to sign any provenance unless the contents match the x.509 certificate. The only thing that can be done is to sign arbitrary hashes.

The secret that I was referring to is the GitHub Actions OIDC token itself because it establishes the identity of the workflow for which the certificate authority generates a certificate. Once `id-token:write` permission is granted to a `job` in a GitHub Actions workflow,  `ACTIONS_ID_TOKEN_REQUEST_URL` and `ACTIONS_ID_TOKEN_REQUEST_TOKEN` environment variables can be used to request an OIDC token. This is also used by `sigstore-js` library that `npm` uses [to request](https://github.com/sigstore/sigstore-js/blob/fca33dc8678ce5502439ac17728c0d64f513014f/packages/sign/src/identity/ci.ts#L50) an OIDC token with the correct scope. Because `npm` requires `id-token:write` in the same job that untrusted code runs, the token can be exfiltrated and a provenance can be signed by a certificate generated for the same identity as the original workflow. See [this example GitHub Actions job](https://docs.npmjs.com/generating-provenance-statements#example-github-actions-workflow) to generate a provenance using the `npm` CLI tool.

>Build Process converts the x.509 certificate into a SLSA Provenance attestation and signs it using the x.509 certificate. 

 I think in this case we don't see a clear separation between `Build Process` containing untrusted code and a `Control Plane`. The clear benefit of this separation is the increased trustworthiness of the produced provenances. But if you think that is only needed in l3, then I agree that this statement needs to change (prefer the latter solution) for l2 (I would definitely keep it for l3):

>The provenance MUST be generated by the control plane

behnazh-w: @joshuagl happy to collaborate on a blog post if it's useful. I find it necessary to do a similar reasoning for GitLab as well.

behnazh-w: >We are not protecting against "bad" build processes (including those who are "bad" because they import a compromised dependency).

I understand this threat model and that a "bad" build process can generate bad artifacts anyway and that is out of scope. But here we talk about generating a provenance, which should be ideally happening outside the build process (using isolated reusable workflows for instance). If that separation happens, the secrets by design will not be available to the untrusted build process.

We can argue that this is too strict for Build L2, and is only required by Build L3. The current requirement of Build L3 agrees with this:

>prevent secret material used to sign the provenance from being accessible to the user-defined build steps.

If we accept this clear distinction, then `npm` would satisfy Build L2 because the build process is generating the provenance, while the generators and builders in the slsa-github-generator would satisfy Build L3 because the build process cannot affect provenance generation by design.

joshuagl: I just reread this discussion thread and think there's actually agreement here. 

I think the apparent disagreement comes from the fact that Mark appears to be discussing [slsa-github-generator](https://github.com/slsa-framework/slsa-github-generator), which uses GitHub reusable workflows + Sigstore to make the generated provenance unforgeable, whereas Behnaz appears to be discussing npm's recently added `--provenance` option to its publish command (`npm publish --provenance`) which is generally called from a producer controlled workflow, as demonstrated in this [example GitHub workflow](https://docs.npmjs.com/generating-provenance-statements#example-github-actions-workflow).

slsa-github-generator's combination of reusable workflows & Sigstore meets SLSA Build L3.
`npm publish --provenance`'s use of Sigstore [meets SLSA Build L2](https://slsa.dev/blog/2023/05/bringing-improved-supply-chain-security-to-the-nodejs-ecosystem).

trishankatdatadog: Exactly, that's where I think the disagreement comes too, thanks for the summary, Joshua

behnazh-w: > Do we think we're good to submit this PR as-is, or are further changes needed? If it's strictly better than the old version but still not perfect, my preference would be to just submit it and then have a separate PR to iterate.

We can merge this PR as is and refine the following definitions and requirements in a separate PR:

- The definition of "dedicated infrastructure" at SLSA Build L2
- Change the `Accuracy` part of [Provenance is Authentic](https://slsa.dev/spec/v1.0/requirements#provenance-authentic) requirement or add a new requirement that does not mandate the following for SLSA Build L2, so that `npm` provenances can satisfy L2.
>The provenance MUST be generated by the control plane (i.e. within the trust boundary [identified in the provenance](https://slsa.dev/spec/v1.0/provenance#model)) and not by a tenant of the build platform (i.e. outside the trust boundary)

MarkLodato: Sounds good. Let's continue discussion on issue #863.

PR #952, opened 2023-08-21
impl: add `fix` PR tag and explain purpose of tags

MarkLodato:
Add a new `fix` PR tag for typos, broken links, and similar. Previously
these were categorized as `editorial`, but that sounded too
significant and required two maintainers to approve plus 24 hours wait.
The new `fix` category is treated the same as `impl`, requiring only one
approval and no wait time.

Also:

- Add a paragraph explaining the purpose of the PR tag.
- Rephrase the preface to the table to provide more clear instructions.
- Create a section heading for the table now that the size of the
  section has grown.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,7 @@ this is not currently enforced).
 PRs are expected to meet the following conventions:
 
 -   PR title is of the form `<tag>: <title>`, where:
-    -   `<tag>` is one of the values in the table below.
+    -   `<tag>` is one of the values in the [table below](#pull-request-tags).
     -   `<title>` concisely explains *what* the PR does.
 -   PR description explains *what* and *why* in a bit more detail, providing
     enough context for a reader to understand the change. See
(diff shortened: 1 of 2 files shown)


joshuagl (approved):

CONTRIBUTING.md
```diff
@@ -81,17 +81,25 @@ PRs are expected to meet the following conventions:
     https://github.com/slsa-framework/slsa/pull/840 (predates our `<tag>`
     convention).
 
-Every PR must use one of the following `<tag>` values. Earlier entries in the
-table have higher precedence. (See [review and approval] for the meaning of
-"waiting period" and "# approvers".) If you are not sure, take a guess and a
-maintainer will update if needed.
+### Pull request tags
```
arewm, line 84: Can we change from the use of "tag" to "type" as this terminology is consistent with the GitHub action that we use.

MarkLodato: actually it's small enough that I'll just do it here

arewm (approved):
Suggest changing from `tag` to `types` in a follow-up PR.

CONTRIBUTING.md
MarkLodato: OK, I'll follow up separately then. (I have it all ready to go.)

MarkLodato: Sent out #954 for this.

PR #954, opened 2023-08-21
impl: use conventional names for parts of a PR

MarkLodato:
In CONTRIBUTING, switch to terminology that is consistent with GitHub
and Conventional Commits / action-semantic-pull-request:

Before:

    <tag>: <title>

    <description>

After:

    <type>: <subject>

    <body>

The whole first line is called the "PR title" (unchanged). Previously
this same term was ambiguously also used for what is now `<subject>`.

Note: Conventional Commits and action-semantic-pull-request disagree on
`<subject>`, with Conventional Commits calling it `<description>`. The
name `<subject>` seems less likely for confusion, so we prefer that one.

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -1,4 +1,4 @@
-name: "Lint PR name"
+name: "PR title follows `<type>: <subject>`"
 
 on:
   pull_request_target:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,37 +63,43 @@ this is not currently enforced).
 
 PRs are expected to meet the following conventions:
 
--   PR title is of the form `<tag>: <title>`, where:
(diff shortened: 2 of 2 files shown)


arewm (changes requested):
Are we unifying on conventional commits or https://github.com/amannn/action-semantic-pull-request? The latter uses `Subject` instead of description. If we are unifying on the GitHub action then we can include a reference to it here as well.

Can you also change the name of the workflow?

MarkLodato:
> Are we unifying on conventional commits or https://github.com/amannn/action-semantic-pull-request? The latter uses `Subject` instead of description. If we are unifying on the GitHub action then we can include a reference to it here as well.

Oh, I didn't realize the discrepancy. Done.

(My default would have been to align with the spec rather than an implementation, but I do think "subject" is better than "description".)

> Can you also change the name of the workflow?

Done. I changed the name to be more descriptive / self-documenting. What do you think?

joshuagl (approved):

PR #955, opened 2023-08-22
impl: relax review requirements during Draft stage

### LINKED ISSUE #953: Update Contributing guidelines
Proposed new line 98

The above are loosely enforced during DRAFT periods and up to maintainer discretion. The above are strictly enforced during CANDIDATE periods.

(i don't have ability to propose changes in this repo)

@MarkLodato

MarkLodato:
As discussed at the 2023-08-21 specification meeting, we do not want to
require a 72h waiting period + 3 reviewers for changes to Draft
specifications, since that will make the iteration speed painfully slow.
Draft specifications are clearly marked as unapproved, and there is
still a Candidate approval process to review the whole spec. Therefore,
change CONTRIBUTING to note that Draft versions are manager discretion
regarding waiting period and number of reviewers.

Fixes #953.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,11 +103,14 @@ maintainer will update if needed. See [review and approval] for the meaning of
 | `style` | A user-visible style or layout change. No content changes. | 0h | 1 |
 | `impl` | A user-invisible change, such as editing a README or the repo configuration. | 0h | 1 |
 
-Note: PR authors with write access to the repo count as second or third
+Note 1: PR authors with write access to the repo count as second or third
 approvers for their own PRs. For example, if the author of a PR with the
 `content` tag has write access to the repo, then the PR only requires
 two additional approvers before merging. However, a PR with the `impl` tag
 always requires one reviewer, even if the author has write access.
+
+Note 2: The "waiting period" and "# reviewers" are relaxed during the Draft
+specification stage, leaving it up to Maintainer discretion.
 
 [squash and merge]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,9 +109,10 @@ approvers for their own PRs. For example, if the author of a PR with the
 two additional approvers before merging. However, a PR with the `impl` tag
 always requires one reviewer, even if the author has write access.
 
-Note 2: The "waiting period" and "# reviewers" are relaxed during the
-[Draft](docs/spec-stages.md) specification stage, leaving it up to Maintainer
-discretion.
+Note 2: If the PR only touches files in the [Draft](docs/spec-stages.md)
+specification stage, then the "waiting period" and "# reviewers" are relaxed and
+up to Maintainer discretion. Files in the Draft stage have a large banner at the
+top of each rendered page, as well as the text "Status: Draft".
 
 [squash and merge]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits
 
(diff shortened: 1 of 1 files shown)


joshuagl (approved):

CONTRIBUTING.md
```diff
@@ -103,12 +103,15 @@ maintainer will update if needed. See [review and approval] for the meaning of
...
 
-Note: PR authors with write access to the repo count as second or third
+Note 1: PR authors with write access to the repo count as second or third
 approvers for their own PRs. For example, if the author of a PR with the
 `content` tag has write access to the repo, then the PR only requires
 two additional approvers before merging. However, a PR with the `impl` tag
 always requires one reviewer, even if the author has write access.
 
+Note 2: The "waiting period" and "# reviewers" are relaxed during the Draft
+specification stage, leaving it up to Maintainer discretion.
```
arewm, line 113: You state that 

> Draft specifications are clearly marked as unapproved

Where is this clearly marked and should we make that clearer here or elsewhere?

joshuagl: we have a banner and a status indicator on every page of the draft spec

image

MarkLodato: Added a link. Do you think that's sufficient, or is more warranted?

MarkLodato: > It would be helpful to also inform potential contributors to see what stage their changes might be in, i.e. via the rendered web page or [versions.yml](https://github.com/slsa-framework/slsa/blob/main/docs/_data/versions.yml#L17).

Sorry, I don't understand. Could you elaborate? The stage is already shown on the website, on the right.

Screenshot 2023-08-22 10 36 25 AM

CONTRIBUTING.md
```diff
@@ -103,12 +103,16 @@ maintainer will update if needed. See [review and approval] for the meaning of
...
-Note: PR authors with write access to the repo count as second or third
+Note 1: PR authors with write access to the repo count as second or third
 approvers for their own PRs. For example, if the author of a PR with the
 `content` tag has write access to the repo, then the PR only requires
 two additional approvers before merging. However, a PR with the `impl` tag
 always requires one reviewer, even if the author has write access.
 
+Note 2: The "waiting period" and "# reviewers" are relaxed during the
+[Draft](docs/spec-stages.md) specification stage, leaving it up to Maintainer
+discretion.
```
arewm, line 114: Ultimately, the need for further clarity hinges on our assumptions about the readers and how much they are able to explore/intuit.

This note has been worded assuming that a contributor has prior knowledge of the repository structure. If that is the only intention (i.e. to be a policy for the individuals that might be merging the changes), then we likely don't need further clarification. 

If we want to inform new contributors about what the waiting period might be for their PRs, then I would view it as helpful to show them how the can determine what the expected wait time might be. I have proposed a change below to help. One critical assumption of this change is that contributors would be able to relate the directory structure to the version and therefore identify its stage since we don't lay that out explicitly in CONTRIBUTING.md.
```suggestion
Note 2: The "waiting period" and "# reviewers" are relaxed during the
[Draft](docs/spec-stages.md) specification stage, leaving it up to Maintainer
discretion. The status of a specification version can be seen in the yaml block
of [versions.yml](main/docs/_data/versions.yml). If a version is in the Draft state
then a header will also be shown on its rendered web pages.
```

While I think that a further change like this would be an improvement, it is ultimately a nit as the stage should be discoverable assuming reasonable deduction.

arewm (approved):

CONTRIBUTING.md
MarkLodato: Just pushed a change to try to address. How about that? I only mentioned the rendered page, since I think that's easier than someone figuring it out from versions.yml.

arewm (approved):

CONTRIBUTING.md
arewm: That works, thanks!

MarkLodato:
I'll submit now, but if you have any suggestions, comment here and I'll fix in a follow-up PR.

PR #964, opened 2023-09-06
content: How to verify VSAs

### LINKED ISSUE #878: Make more VSA fields optional, add verification section

kpk47:
Describe how to verify VSAs.
- Detail verification procedure.
- Add verification examples to clarify verification procedure.
- Change `vsa.timestamp` from `required` to `optional` because it is not part of the verification procedure.

Fixes #878 

BREAKING CHANGE: `timeVerified` switch from `required` to `optional`.


docs/spec/v1.1/verification_summary.md
```diff
@@ -245,6 +245,103 @@ WARNING: This is just for demonstration purposes.
...
+VSA consumers use VSAs to accomplish goals based on delegated trust. We call the
+process of establishing a VSA's authenticity and determining whether it meets
+the consumer's goals 'verification'. Goals differ, as do levels of confidence
+in VSA producers, so the verification procedure changes to suit its context.
+However, there are certain steps that most verification procedures have in
+common.
+
+Verification MUST include the following steps:
+
+1.  Verify the signature on the VSA envelope using the preconfigured roots of
```
joshuagl, line 259: I wonder if we need a write up of roots of trust somewhere that we can link to from here? The concept is covered at a high level in https://slsa.dev/spec/v1.1/verifying-artifacts#step-1-check-slsa-build-level but it would be nice to be able to link the term here to an expansion.

This should/could be a separate PR.

joshuagl (approved):
LGTM. Thank you!

PR #967, opened 2023-09-17
fix: add redirect for removed example.md

jwhb:
fixes link [hypothetical curl example](https://slsa.dev/example.md) to example.md at https://slsa.dev/spec/v1.0/use-cases#motivating-example


joshuagl (approved):
Thanks you.

Issue #803, opened 2023-04-05
VSA: should policy digest be mandatory


laurentsimon:
Only the uri field is mandatory. Why is the digest not? Do we expect the uri to not be hashable in some scenarios?

I also wonder whether we should allow a `version` field. I would expect that policies will evolve over time. Or maybe we expect the policy to be renamed to reflect changes?

laurentsimon:
If resourceDescriptor has a version field, we could use it for the policy too.

MarkLodato:
I don't think we should make it required because it is not necessary for correct use of the VSA. It might not be straightforward to hash it. For example, if the policy is a combination of a per-package config plus some global config, you need to define a canonical serialization of both to represent it.

laurentsimon:
I was suspecting the policy may not be hash-able, thanks for confirming. Fwiw, in your example, the base policy would be part of the final policy, like a base image is part of a final image in a dockerfile. In this case the serialization is kinda free.

MarkLodato:
Here's a concrete example: the "policy" is that the new version must have provenance matching the previous version. There is no recorded policy anywhere; it's just a comparison to the previous version. Technically this creates a policy on the fly based on the previous version's provenance, but it is never serialized anywhere. One _could_ define a canonical serialization and hash it, and perhaps that's a good idea, but it's not necessary for correct operation. Thus making it REQUIRED in the spec seems like an overstep.

I think a good compromise is to change from:

> The entry MUST contain a `uri`.

to:

> The entry MUST contain a `uri` identifying which policy was applied and SHOULD contain a `digest` to indicate the exact version of that policy.

laurentsimon:
SGTM

PR #979, opened 2023-10-10
content: recommend a VSA include the policy's digest

joshuagl:
Recommended that the `digest` field of `ResourceDescriptor` is set in a Verification Summary Attestation's (VSA) `policy` object.

Fixes: #803

### DIFF
--- a/docs/spec/v1.1/verification_summary.md
+++ b/docs/spec/v1.1/verification_summary.md
@@ -83,7 +83,7 @@ to establish minimum requirements on dependencies SLSA levels may use
   "resourceUri": <artifact-URI-in-request>,
   "policy": {
     "uri": "<URI>",
-    "digest": { /* DigestSet */ }
+    "digest": { <digest-of-policy-data> }
   }
   "inputAttestations": [
     {
@@ -162,7 +162,7 @@ of the other top-level fields, such as `subject`, see [Statement]._
 > Describes the policy that the `subject` was verified against.
 >
 > The entry MUST contain a `uri` identifying which policy was applied and
-> SHOULD contain a digest to indicate the exact version of that policy.
+> SHOULD contain a `digest` to indicate the exact version of that policy.
 
 <a id="inputAttestations"></a>
 `inputAttestations` _array ([ResourceDescriptor]), optional_
--- a/docs/spec/v1.1/verification_summary.md
+++ b/docs/spec/v1.1/verification_summary.md
@@ -161,7 +161,8 @@ of the other top-level fields, such as `subject`, see [Statement]._
 
 > Describes the policy that the `subject` was verified against.
 >
-> The entry MUST contain a `uri`.
+> The entry MUST contain a `uri` identifying which policy was applied and
+> SHOULD contain a digest to indicate the exact version of that policy.
 
 <a id="inputAttestations"></a>
 `inputAttestations` _array ([ResourceDescriptor]), optional_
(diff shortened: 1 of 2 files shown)


MarkLodato (approved):
Related: I believe there was a separate issue/PR to make `policy` optional (maybe recommended?) since it's not strictly required. That doesn't impact this PR, but just FYI.

joshuagl:
> Related: I believe there was a separate issue/PR to make `policy` optional (maybe recommended?) since it's not strictly required. That doesn't impact this PR, but just FYI.

Thanks! That matches my mental model (I was surprised when working on this change that the policy was required). I agree, this change still works with that change (tracked as part of https://github.com/slsa-framework/slsa/issues/878 but should probably be a separate issue).

docs/spec/v1.1/verification_summary.md
```diff
@@ -153,7 +153,8 @@ of the other top-level fields, such as `subject`, see [Statement]._
 
 > Describes the policy that the `subject` was verified against.
 >
-> The entry MUST contain a `uri`.
+> The entry MUST contain a `uri` identifying which policy was applied and
+> SHOULD contain a digest to indicate the exact version of that policy.
```
lehors, line 157: I think that's fine anyway because it's a SHOULD but presumably one could have different URIs for different versions, no?

joshuagl: yes, there could absolutely be different URIs for different policy versions

kpk47 (approved):

lehors (approved):

docs/spec/v1.1/verification_summary.md
arewm, line 157: While I feel like `immutable` is a more precise word than `exact`, it might not be as well known. Is there a way that we can strengthen the wording to indicate a strongly persistent exactness?

arewm, line 157: Should it be `digest` to indicate that it is a key value?

arewm, line 157: The `digest` present in the schema above is different from the other digest representations. Is there a reason that we have it represented like this?

```
  "policy": {
    "uri": "<URI>",
    "digest": { /* DigestSet */ }
  }
```

arewm (changes requested):

docs/spec/v1.1/verification_summary.md
joshuagl: We're recommending digest as the mechanism and using exact to describe the purpose of that digest. We could use alternatives, such as: precise or explicit? I think I'd lean to "explicit version of that policy".

joshuagl: Well spotted, I can change that to match.  I don't know of a reason it would be different (I thought perhaps because `policy.digest` optional, but so is an `inputAttestations`'s digest).

I will change `/* DigestSet */` to `<digest-of-policy-data>`.

arewm: Explicit works for me. I am now questioning whether we should use `version` now. When I think of a version, SemVer tags come to mind and that doesn't jive with a digest. 

I cannot offer a _suggestion_ as I originally just commented on the one line, but how about something like
```
The entry MUST contain a `uri` identifying which policy version was applied and
SHOULD contain a digest to indicate the explicit reference of that policy.
```

arewm: Thanks for the change. This can be resolved. :)

arewm: Thanks for the change. This can be resolved. :)

MarkLodato (approved):

docs/spec/v1.1/verification_summary.md
MarkLodato: Personally I find the current wording more readable. I don't think people will be confused by the word "version", and any alternative seems less clear.

PR #980, opened 2023-10-10
editorial: tweak presentation of VSA pages

joshuagl:
- Expand VSA --> Verification Summary in the left-hand nav menu. The VSA acronym only makes sense to readers familiar with the VSA.
- Remove redundant "SLSA" prefix from VSA page titles
- Remove specification version from VSA page descriptions


MarkLodato (approved):
Thanks! Nice!

(Note: I changed tag to `editorial` since it does touch a specification page.)

Issue #975, opened 2023-10-03
Workstream: Hardware Attested Build Environments


chkimes:
This is a tracking issue for incorporating Hardware Attested Platforms, aka Trusted Computing into SLSA. The main idea is to provide greater trust in the build by using trusted computing features like Trusted Execution Environments (TEEs) of modern CPUs to reduce the risk of tampering and to increase transparency.

[Workstream shepherd]: Marcela Melara (@marcelamelara), Pavel Iakovenko (@paveliak)

Working proposal: #1051 
Proposal doc: [here](https://docs.google.com/document/d/1lKmyY2hvgopiGKuIx9Fq2zmu3I6GXXLkc7qMgNYSdgc/edit?pli=1#heading=h.ifksclf4mcni)

Related: We might want to merge with #977 (Build L4, discussing reproducible builds) and/or  #985 (about hardening operations) as discussed in [below](https://github.com/slsa-framework/slsa/issues/975#issuecomment-1757645142).

Sub-issues:

-   [ ] #1115
-   [ ] Add detailed requirements description
-   [ ] Describe implementation options
-   [ ] Describe the verification model
-   [ ] #1110 
-   [ ] #1107
-   [ ] #1122 

[Workstream shepherd]: https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md#workstream-lifecycle

---

In the [2023-09-13 Supply Chain Integrity meeting](https://docs.google.com/document/d/1moVFPn5pLi-uGs840_YBCrwdpHajU0ptFmlL4F9GryQ/edit#heading=h.h2qye2iijr1c), @marcelamelara and I presented on a potential new SLSA track, using cryptographic primitives provided by hardware to validate build environments.

Slides: https://docs.google.com/presentation/d/11cycDxYaoZpuG144pR6atI1_zk2CfZOWlNO_f_HhhyE
Doc: https://docs.google.com/document/d/1l7IKAli-K-uof8VkLuiqV5-hMGS_ecDmBcuc07-ILeQ/edit
Recording: TBD pending upload to YouTube

Some points for discussion, seeding some from the SCI meeting:

1. Is this a new track or an extension of other tracks?
2. I've labeled this as the Build Platform Operations track, however the [Future Directions page](https://slsa.dev/spec/v1.0/future-directions) defines a set of requirements that are likely only verifiable through audit, whereas the attestations defined above are verifiable at runtime. Is this the appropriate track to be defining these in or is there yet another track to distinguish these?

[...]

PR #205, opened 2021-11-03
Document conventions for 0.x

MarkLodato:
This was discussed at today's community meeting. I took a stab at documenting what I think are the recommendations, but please review to see if you agree.

I believe @joshuagl @mlieberman85 were the ones who commented during the meeting, but I may remembering incorrectly.

PR #981, opened 2023-10-10
impl: Create new project lifecycle process

MarkLodato:
As discussed in the 2023-10-09 specification meeting, create a new
"Project lifecycle" for major projects (new version, new track, etc).
The main idea is to assign a "project shepherd" to oversee the project.
This person is responsible for moving the project along. The reason for
creating this is that many projects seem to stagnate without a single
responsible person.

This commit populates the initial list of projects as per our discussion
at the meeting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,13 @@ See https://slsa.dev/community for ways to get involved in SLSA development.
 | Project | [Shepherd] |
 | ------- | ---------- |
 | [Build Level 4] | David A Wheeler (@david-a-wheeler) |
-| [Hardware Attested Builds] | Marcela Melara (@marcelamelara), Chad Kimes (@chkimes) |
+| [Hardware Attested Platforms] | Marcela Melara (@marcelamelara), Chad Kimes (@chkimes) |
 | [Source Track] | Kris K (@kpk47) |
 | [Version 1.1 release] | Joshua Lock (@joshuagl) |
 
 [Shepherd]: CONTRIBUTING.md#project-lifecycle
 [Build Level 4]: https://github.com/slsa-framework/slsa/issues/977
-[Hardware Attested Builds]: https://github.com/slsa-framework/slsa/issues/975
+[Hardware Attested Platforms]: https://github.com/slsa-framework/slsa/issues/975
 [Source Track]: https://github.com/slsa-framework/slsa/issues/956
 [Version 1.1 release]: https://github.com/slsa-framework/slsa/issues/900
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -19,6 +19,25 @@ specification, and overall project management. Other git repositories within the
...
+## How to get involved
+
+See https://slsa.dev/community for ways to get involved in SLSA development.
+
+## Active projects
+
+| Project | [Shepherd] |
+| ------- | ---------- |
+| [Build Level 4] | David A Wheeler (@david-a-wheeler) |
+| [Build Platform Operations Track] | Marcela Melera (@marcelamelera), Chad Kimes (@chkimes) |
```
marcelamelara, line 31: Typo in name/handle
```suggestion
| [Build Platform Operations Track] | Marcela Melara (@marcelamelara), Chad Kimes (@chkimes) |
```

marcelamelara (approved):
LGTM thanks!

joshuagl (approved):

joshuagl (approved):

arewm (approved):

README.md
```diff
@@ -19,6 +19,25 @@ specification, and overall project management. Other git repositories within the
...
+## How to get involved
+
+See https://slsa.dev/community for ways to get involved in SLSA development.
+
+## Active projects
+
+| Project | [Shepherd] |
+| ------- | ---------- |
+| [Build Level 4] | David A Wheeler (@david-a-wheeler) |
+| [Hardware Attested Builds] | Marcela Melara (@marcelamelara), Chad Kimes (@chkimes) |
```
marcelamelara, line 31: I'd recommend renaming this project to something like "Hardware Attested Platforms" or "Verifiable Build Environments" to ensure we're communicating the properties we're trying to achieve more clearly, per my comment in https://github.com/slsa-framework/slsa/issues/975#issuecomment-1758119072

MarkLodato: Done.

lehors:
Hmm, sorry for not responding sooner but I do have to say that the use of the term "project" in this case is unfortunate. "Projects" are a type of "Technical Initiatives" under the [OpenSSF charter](https://openssf.org/about/charter/) for which the TAC has defined a [project lifecycle](https://github.com/ossf/tac/blob/main/process/project-lifecycle.md). It would be best to avoid overload that term here...
Maybe call it "Work item", "Work stream", or simply "Subproject"?

As a side note: I've mentioned before that, technically, SLSA is currently a SIG rather than a Project under the current OpenSSF governance structure. I have however proposed to extend the definition of a Project to encompass specification development (see [TAC PR #205](https://github.com/ossf/tac/pull/205)) and it is broadly supported by TAC members. So, SLSA should indeed officially become a Project soon. :-) But that's one more reason not to create a new type of "project".

MarkLodato:
Ugh. "Subproject" implies that there is a larger notion of "project", and I don't think anyone is going to realize the project is SLSA itself. :-/

Is "project" so bad? Isn't it normal to have a project within a project?

I'm not opposed to "workstream" though.

Any other thoughts?

PR #987, opened 2023-10-13
editorial: update mapping of Nix terminology to the SLSA Package model

marksisson:
Change Nix package ecosystem terminology to match definitions of SLSA Package model terminology. Capitalize Nix to match proper name of package ecosystem. Map SLSA Package registry to Nixpkgs (the largest repository of Nix packages [and also a Nix Flake]) and Flakes (distributed repositories of Nix packages). Map SLSA Package name to attributes of Nixpkgs and Flake outputs. Map SLSA Package artifact to Nix Store Path.

### DIFF
--- a/docs/spec/v1.1/terminology.md
+++ b/docs/spec/v1.1/terminology.md
@@ -253,10 +253,10 @@ additions are welcome!
     <td>Package name
     <td>Package
   <tr>
-    <td><a href="https://nixos.org/guides/how-nix-works.html">nix</a> (e.g. <a href="https://nixos.org/">NixOS</a>)
-    <td><em>?</em>
-    <td><a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-store-object">Store Object</a>?
-    <td>Package or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-derivation">Derivation</a>
+    <td><a href="https://nixos.org/manual/nix">Nix</a> (e.g. <a href="https://nixos.org/">NixOS</a>)
+    <td>Repository (e.g. <a href="https://github.com/NixOS/nixpkgs">Nixpkgs</a>) or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-binary-cache">binary cache</a>
+    <td><a href="https://nixos.org/manual/nix/stable/language/derivations.html">Derivation name</a>
+    <td><a href="https://nixos.org/manual/nix/stable/language/derivations.html">Derivation</a> or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-store-object">store object</a>
   <tr>
     <td colspan=4><em>Storage systems</em>
   <tr>
--- a/docs/spec/v1.0-rc2/terminology.md
+++ b/docs/spec/v1.0-rc2/terminology.md
@@ -219,10 +219,10 @@ additions are welcome!
     <td>Package name
     <td>Package
   <tr>
-    <td><a href="https://nixos.org/guides/how-nix-works.html">nix</a> (e.g. <a href="https://nixos.org/">NixOS</a>)
-    <td><em>?</em>
-    <td><a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-store-object">Store Object</a>?
-    <td>Package or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-derivation">Derivation</a>
+    <td><a href="https://nixos.org/manual/nix">Nix</a> (e.g. <a href="https://nixos.org/">NixOS</a>)
+    <td>Repository (e.g. <a href="https://github.com/NixOS/nixpkgs">Nixpkgs</a>) or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-binary-cache">binary cache</a>
+    <td><a href="https://nixos.org/manual/nix/stable/language/derivations.html">Derivation name</a>
+    <td><a href="https://nixos.org/manual/nix/stable/language/derivations.html">Derivation</a> or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-store-object">store object</a>
   <tr>
     <td colspan=4><em>Storage systems</em>
   <tr>
--- a/docs/spec/v1.0/terminology.md
+++ b/docs/spec/v1.0/terminology.md
@@ -253,10 +253,10 @@ additions are welcome!
     <td>Package name
     <td>Package
   <tr>
-    <td><a href="https://nixos.org/guides/how-nix-works.html">nix</a> (e.g. <a href="https://nixos.org/">NixOS</a>)
-    <td><em>?</em>
-    <td><a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-store-object">Store Object</a>?
-    <td>Package or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-derivation">Derivation</a>
+    <td><a href="https://nixos.org/manual/nix">Nix</a> (e.g. <a href="https://nixos.org/">NixOS</a>)
+    <td>Repository (e.g. <a href="https://github.com/NixOS/nixpkgs">Nixpkgs</a>) or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-binary-cache">binary cache</a>
+    <td><a href="https://nixos.org/manual/nix/stable/language/derivations.html">Derivation name</a>
+    <td><a href="https://nixos.org/manual/nix/stable/language/derivations.html">Derivation</a> or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-store-object">store object</a>
   <tr>
     <td colspan=4><em>Storage systems</em>
   <tr>


docs/spec/v1.1/terminology.md
```diff
@@ -253,10 +253,10 @@ additions are welcome!
     <td>Package name
     <td>Package
   <tr>
-    <td><a href="https://nixos.org/guides/how-nix-works.html">nix</a> (e.g. <a href="https://nixos.org/">NixOS</a>)
-    <td><em>?</em>
-    <td><a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-store-object">Store Object</a>?
-    <td>Package or <a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-derivation">Derivation</a>
+    <td><a href="https://nixos.org/guides/how-nix-works.html">Nix</a> (e.g. <a href="https://nixos.org/">NixOS</a>)
+    <td><a href="https://github.com/NixOS/nixpkgs">Nixpkgs</a> / <a href="https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake">Flakes</a>
```
MarkLodato, line 257: Reading the documentation, it looks like the following might be a closer fit:

- Package ecosystem = [Nix Flakes](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html) (NixOS)
- Package registry = [Registry](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-registry.html) (or maybe "global registry") = the thing that maps identifiers to locations
- Package name = [Flake reference (flakeref)](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#flake-references)
- Package artifact = [Flake](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#description) (but see other comment)

What do you think? The output attribute seems like a sub-artifact, while store path is an identifier not an artifact itself.

docs/spec/v1.1/terminology.md
```diff
@@ -253,10 +253,10 @@ additions are welcome!
...
+    <td><a href="https://github.com/NixOS/nixpkgs">Nixpkgs</a> attribute path / <a href="https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake">Flake</a> output attribute
+    <td><a href="https://nixos.org/manual/nix/stable/glossary.html#gloss-store-path">Store Path</a>
```
MarkLodato, line 259: I'm confused by Nix here. What is the actual artifact (sequence of bytes) distributed as part of Nix? I don't think it's the "Store Path". The old version said "Derivation" which sounded right to me? In other words, when you "fetch" something from Nixpkgs, you actually fetch the derivation?

For Flakes, I think the thing being distributed is called a "flake". Is that right?

MarkLodato (changes requested):
Thank you for the help! I know almost nothing about Nix, and this has taught me a lot. I have a few questions.

/cc @mlieberman85 who added the original definitions

docs/spec/v1.1/terminology.md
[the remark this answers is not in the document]
marksisson: The Nix ecosystem is challenging to describe, please bear with me, I will do my best to describe it.

Nix is the package manager. Nixpkgs is the largest community managed repository of packages. And there are other user managed repositories (e.g. NUR or just individual git repositories). Nixpkgs is a set of freeform Nix expressions (no schema), as is every other Nix (user managed) package repository before Flakes. This made it difficult to consume packages without first looking at the Nix expression to know what attributes to access. Nix Flakes have a schema, standard mechanism for pinning, evaluation caching, and provide for hermetic evaluation. But packages are still defined using Nix expression language (flake or no flake), and evaluated by Nix package manager command line tool.

I am not sure that is all totally clear. But just to say that Nix is one ecosystem, and flakes are just an additional layer.

For example, Nixpkgs is also a Nix Flake. So you can access packages via the Flake schema, or via convention existed before Nix Flakes (i.e. via default.nix file in root of repo, and nested attribute paths).

In looking at how the other ecosystems are documented on the SLSA terminology page, the package ecosystem appears to be the name of the package manager; in this case, Nix is the package manager. Nix is currently supported on any Linux distro and on macOS (so, more than NixOS).

For the Package Registry column, I would say the general term would be "Repository" (in fact Nix Flakes are strongly tied to git repository - i.e. nix will ignore "dirty" files during evaluation; files must be staged or committed; this is part to the pure evaluation, e.g. hermitic build). Since Nixpkgs is the community repository, I guess it would be the one listed here.

For the Package name, this would be an attribute path determined by convention when not using flakes, and would be an attribute under the "packages" attribute of the flake schema when using flakes, i.e. a flake ref (there is no attribute path under "packages", as flakes do not allow nesting attributes below the schema defined attributes).  When you build a package using the traditional Nixpkgs, you specify an attribute from the package set (i.e. the attrset, which is a Nix datatype), which may not necessarily be the same as the package name in the derivation (pname). pname is internal from end user perspective. So, you will either specify attribute path when using traditional nix, or flake ref when using flakes interface.

For Package artifact, when you build a Nix package, there are two underlying steps that occur, instantiation and realization.  The instantiation step takes the Nix expression for a derivation (this is defined using the builtin derivation function), evaluates the expression, and creates a store derivation.  The store derivation is the "build recipe".  These are the .drv files in the nix store (/nix/store).  The realization step takes the store derivation, and builds the package, producing a store path (the output of the build).  This will be in the nix store with a name consisting of the derivation pname and an input-address (calculated from the derivation inputs).  Store paths are usually symlinked to user's directory, or added to user's environment via PATH.


n.b.
Nix package manager can be used on any Linux or macOS. NixOS is a linux distro that is built using the Nix package manager.

docs/spec/v1.1/terminology.md
```diff
@@ -253,10 +253,10 @@ additions are welcome!
     <td>Package name
     <td>Package
   <tr>
```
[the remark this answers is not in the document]
marksisson: Per comments above, Nix should be considered one ecosystem. Maybe I was getting too granular by differentiating traditional nix and nix flakes?

docs/spec/v1.1/terminology.md
marksisson: Flakes can have more than packages. They can output modules, configurations, shells, etc.

A flake can have many packages. Nixpkgs is an example. There is no central authority for finding all flakes (or nix repositories in general).

So, it may still be best to refer to Nixpkgs as the main community repository, when referring to traditional nix or nix flakes (since it is both).

I think Package name should be Flake ref as you have specified when using flakes.

Package artifact would still be a store path.  Nix is still evaluating a nix expression that produces a store derivation, that is then realized to a store path.

docs/spec/v1.1/terminology.md
marksisson: I would say the actual artifact is the store path resulting from the instantiation and realization of the nix derivation.  Derivation is slightly overloaded - there is the derivation expression, which is written using the builtin derivation function, and there is the store derivation, which is the evaluated derivation expression written to the nix store (this is just the "build recipe"). The final output of a package build is not the derivation, it is the artifact(s) of the build.

For Flakes, the flake is just a repository with a flake.nix file.  So, just a well-defined entry point for Nix. In the end, the output attribute set of the flake schema is still a set of name/value pairs with the values being nix expressions that must conform to the types in the schema (e.g. packages must be derivation expressions).

docs/spec/v1.1/terminology.md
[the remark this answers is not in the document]
marksisson: My hope would be to not make Nix package manager look like a splintered ecosystem. It is just a little confusing with two interfaces (the traditional interface and the flake interface). Both interfaces use derivations and produce store paths.

I hope this is helping :exploding_head:

docs/spec/v1.1/terminology.md
marksisson: Also, I should mention Nix is a "from source" package manager. It also support binary substitution via a binary cache. Because store paths are input addressable, as long as the inputs of the derivation have not changed, a binary substitute can be used.  This is based on the assumption that building again would result in the same binary (not exactly the case unless the build is truly reproducible). So when a user "fetches" a "package" using Nix package manager and the Nixpkgs package repository, they are usually getting the binary substitute from the Nix binary cache (this is populated by a continuous integration service called Hydra).

marksisson: Another thing came to mind. Nix derivations can specify multiple outputs. So for example, you can have "binary" and "development" outputs, which would be written to different store paths. Nix installs the first output (or outputs specified in the "meta.outputsToInstall" attribute of the derivation) if no specific output is requested. By convention, the first output are the "binaries" of the package.

This a leading me to think we should use "Nix Package" for the Package artifact column.  From a package consumer perspective, the store path name is probably a low level detail.

marksisson: Maybe this is all we need. It would avoid getting into the interface details of the package manager (traditional vs flake). It is accurate regardless of traditional interface or flake interface.

- Package ecosystem = [Nix](https://nixos.org/manual/nix) (NixOS)
- Package registry = [Nixpkgs](https://github.com/NixOS/nixpkgs)
- Package name = [Derivation name](https://nixos.org/manual/nix/stable/language/derivations.html)
- Package artifact = [Derivation output](https://nixos.org/manual/nix/stable/language/derivations.html)

MarkLodato: Thanks. This is all amazingly helpful!

For this model, we are exclusively interested in the things that **cross the trust boundary** between client and registry. Here's a simple [data flow diagram](https://threat-modeling.com/data-flow-diagrams-in-threat-modeling/) that might help (we should probably add this to the site):

slsa-package-dfd drawio

A package ecosystem is a convention for mapping names to artifacts. A registry is one authority for this mapping. A client queries a registry with a name and gets back an artifact. For example, in NPM, the client requests a name plus some additional qualifiers (e.g. name `react` version `18.2.0`) and they get back a tarball.

I'm still having a hard time wrapping my head around Nix. Let me take a shot at explaining and see if this is anywhere close. When a user runs something like `nix install <name>`, the Nix client does the following:

1. If there is a local pre-built version of `<name>` already on the machine (in the Nix store?), use that.
2. Else if there is a local derivation (?) for `<name>`, build and use that.
3. Else query an external binary cache for `<name>`. If one is found, download and use that.
4. Else query an external registry (repository?) such as Nixpkgs for a derivation of `<name>`. If one is found, download, build, and use that.
5. Else fail.

Does that sound even somewhat correct? If so, it is the things downloaded on steps 3 and 4 that we care about, because those are the ones that cross the trust boundary of local client to remote package registry. SLSA is concerned about tampering in the registry or by a package owner: does the artifact that we got really correspond to what `<name>` ought to be?

Assuming that's all right, what are the things that are actually downloaded in steps 3 and 4? "Binary substitute" and "derivation", respectively?

marksisson: The steps are just a little out of order. You have to start with a nix expression. Here are a revised set of steps:

1. Nix tries to find the name in a set of nix expressions (this is configurable, but we will assume Nixpkgs or the packages output of a flake).  If it cannot match the name to a nix expression that evaluates to a derivation, it fails.
2. If the derivation is already realized in the local nix store, then use that.
3. If the nix configuration allows binary substitution, check binary cache, and if match is found, use that. (you can disable this, so that you always build from source)
4. Else build the derivation (instantiate and realize), and use that. The build occurs in a sandbox, with only inputs/dependencies available.  Downloads of source and transitive closure of dependencies are allowed in sandbox, because they all have content hashes that must match downloaded artifacts, or build fails.
5. Optionally sign and upload derivation outputs to a binary cache.


- download of the nix expression in step 1 (this is the derivation; an expression)
- download of binary substitute in step 3 (outputs of derivation, e.g. executables, libraries, documentation, etc.)
- download of source and transitive closure of dependencies in step 4
- upload in step 5

Package Owner -> Package Registry
- Since the "package" is the derivation expression, and a hosted package repository is most likely a git repo, I guess this boundary is crossed by committing nix derivation expression to attribute set in package repo?

Client -> Package Registry
- client will download set of nix expressions, e.g. the nixpkgs repo or a flake repo

Package Registry -> Client
- I guess this is really binary cache to client

So for Nix, I guess the Package Registry box in the diagram is a combination of the set of nix derivation expressions and the binary cache.

marksisson: In case you are curious about the build sandbox:

- Linux uses cgroups and namespaces
- macOS uses [sandbox profiles](https://www.chromium.org/developers/design-documents/sandbox/osx-sandboxing-design/)

The sandbox is very strict.

marksisson: For comparison to your NPM example above, the client requests a name and possibly specific derivation outputs and they get back populated store path(s) in the nix store.

MarkLodato: Thanks. So it sounds like maybe the answer is:

* Package ecosystem = [Nix](https://nixos.org/manual/nix) (NixOS)
* Package registry = Repository (e.g. [Nixpkgs](https://github.com/NixOS/nixpkgs)) or [Binary cache](https://nixos.org/manual/nix/stable/package-management/basic-package-mgmt)
* Package name = [Derivation name](https://nixos.org/manual/nix/stable/language/derivations.html)
* Package artifact = [Derivation](https://nixos.org/manual/nix/stable/language/derivations.html) or [store object](https://nixos.org/manual/nix/stable/glossary.html#gloss-store-object) 

Does that look right to you?

marksisson: Maybe to help with the derivation concept:

A nix expression is written using the nix expression language.  A nix expression will evaluate to a nix type, such as a set, list, string. A derivation is a special type, that can only be defined with the builtin derivation function.

A day in the life of a derivation would be:
1. evaluation of a nix expression that results in a derivation type
2. instantiate the derivation type to get a store derivation (a .drv file in nix store)
3. realize the store derivation (.drv file) to get a store output path or paths (executable, library, doc, etc)

So, evaluate, instantiate, realize.

marksisson: I think "package" is synonymous with derivation in the Nix ecosystem. I do not think nix community users would think of the binary cache as a package registry?  If we substitute Derivation for Package in the column names, I think it helps.
- Package registry maps to Derviation registry, so e.g. Nixpkgs as you have stated
- Package name maps to Derivation name, which is a direct mapping as you have stated
- Package artifact maps to Derivation artifact, which is the derivation output(s) - so based on that -> Derivation artifact should not equal Derivation.  The derivation outputs are store paths (not store objects - those are in the sqlite database of nix), but store paths might be considered an implementation detail?

n.b.
A store object consists of a file system object, references to other store objects, and other metadata. This is in the database.

marksisson: I guess Package artifact could be "realized Derivation output(s)".  That is really what it is it seems.

MarkLodato: Sorry, I'm getting confused by different levels of abstraction. If we just consider things that are "distributed" and ignore anything that happens locally within a machine, I believe there are two types of distributed artifacts:

1. A Nix expression, materialized as a source file in a git repo called Nixpkgs, that essentially describes how to build a binary. This expression is identified by a "Derivation name".
2. (optional) A pre-compiled binary fetched a binary cache. This binary is identified by a derivation as well?

Is that correct?

marksisson: Yes, I would say that is correct.

I just do not think of the nix derivation expression as a distributed artifact. The nix derivation expression is like a homebrew formula, a pacman PKGBUILD, etc.  E.g. the formula or PKGBUILD are not listed as package artifacts for their respective package ecosystem.

marksisson: Still thinking through this with you...

Package artifact - A file or other immutable object that is intended for distribution.

Using the above definition, this does seem to be exactly a store output path (realized derivation output).  A store output path is an immutable file or immutable directory of files.  The store output paths are the result of building the package outputs (derivation outputs).  These store output paths are what is intended for distribution using a command like nix-copy-closure.

marksisson: I think I am finally coming around to what you are looking for. I am thinking too much in the nix way. And need to be thinking in the SLSA model that you presented earlier. So I think your last proposal for the mapping is correct, but maybe change store object to store path. Does that sound good?

docs/spec/v1.1/terminology.md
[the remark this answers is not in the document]
06kellyjac: With flakes the main difference here is official source inputs and locking + default enforcing purity.

With pre-flakes nix you can still opt in to that purity and get the same effect by using `niv` or manual pinning.

At the end of the day under the hood its still derivations and thats the most relevant part here, no need to go off topic.

For the node ecosystem it'd be like telling a new user "well there's nodejs libraries on npm but you can either use npm, yarn, pnpm, yarnv2 "berry", or even bun" when the main thing to focus on is the package :)

docs/spec/v1.1/terminology.md
marksisson: I just reviewed the Nix definitions of store path and store object, and you are correct that is should be store object. I have updated the commit to reflect your proposed mapping. Thanks a lot for walking through this with me.

MarkLodato (approved):
Thanks for all this! I think the end result looks good.

docs/spec/v1.1/terminology.md
MarkLodato: Yeah, there are effectively two separate "packages" here, the "source" package (Derivation) and the "binary" package (store object). The same thing goes with Homebrew etc (I assume - I'm unfamiliar with them). Either way, end result looks good to me.

MarkLodato:
Friendly ping. Can another maintainer approve? (`editorial` requires 2 approvals)

trishankatdatadog (approved):
Not a NixOS expert but LGTM

MarkLodato:
Let's go with this. If folks have further suggestions, we can always have another PR. Thanks again, @marksisson!

Issue #968, opened 2023-09-20
Values of `verifiedLevels` in VSA v1


AdamZWu:
The current documentation seems to refer to two separate ladders.

- The SLSA build ladder, which currently goes from 0 to 3, and the value is dependent on the builder, not policy verification.
- The "FAILED" value is reserved for denoting policy verification failure, which is separate from the SLSA build ladder, but there are no other values mentioned.

Would like clarification, esp. on the "FAILED" value and its use case.

joshuagl:
The text describing [`SlsaResult`](https://slsa.dev/spec/v1.0/verification_summary#emslsaresult-stringem) states that `SlsaResult` conveys "The result of evaluating an artifact (or set of artifacts) against SLSA."

That is, the VSA describes the SLSA level _verified_  by the VSA's signer, not the level _claimed_ by the artefact's producer.

AdamZWu:
There are two subtle issues with the interpretation @joshuagl:

1. The "FAILED" is over-broad. Current spec only defined a single track (i.e. SLSA BUILD), so it looks like it could work; however, as soon as we add another track, "FAILED" becomes ambiguous -- which track did it fail? or all of them?

Also, "FAILED" doesn't have the "SLSA_" reserved prefix, so it is in a strange position -- it sort of belongs to the SLSA track (or multiple SLSA tracks?) as it is a possible SLSA verification result, OTOH it also sort of belongs to the non-SLSA tracks, if the verifier also evaluates customer ladders.

2. The SLSA BUILD track doesn't seem to need "FAILED" at all. Since the build level 0 is effectively catching everything, technically nothing can fail that level.

Additionally, I think it would be useful to separately convey "what is the build level of the artifact", and "what is the overall policy verification result". So even if an artifact failed the overall policy verification, knowing that it is built in compliance with SLSA BUILD 2 is better than know nothing.

MarkLodato:
Related: #964 by @kpk47 

I agree with @AdamZWu about removing FAILED from `verifiedLevels`.

I also agree it might be a good idea to separate "what did the policy require" from "what did I observe".  The set difference would show you why the thing failed. And while we're at it, it might simplify things to no longer have "SLSA_BUILD_L<n>" imply "SLSA_BUILD_L<n-1>". That way it's purely a set.

[...]

PR #988, opened 2023-10-13
editorial: Clarify DependencyLevels, SlsaResult, and quote level keys.

AdamZWu:
This is related to issues #968, just a cleanup preparing the proposal to remove "FAILED" at v1.1 in a future commit.

### DIFF
--- a/docs/spec/v1.0/verification_summary.md
+++ b/docs/spec/v1.0/verification_summary.md
@@ -180,8 +180,8 @@ of the other top-level fields, such as `subject`, see [Statement]._
 > its dependencies), or "FAILED" if policy verification failed.
 >
 > Users MUST NOT include more than one level per SLSA track. Note that each SLSA
-> level implies all levels below it (e.g. SLSA_BUILD_LEVEL_3 implies
-> SLSA_BUILD_LEVEL_2 and SLSA_BUILD_LEVEL_1), so there is no need to
+> level implies all levels below it (e.g. `SLSA_BUILD_LEVEL_3` implies
+> `SLSA_BUILD_LEVEL_2` and `SLSA_BUILD_LEVEL_1`), so there is no need to
 > include more than one level per track.
 
 <a id="dependencyLevels"></a>
@@ -192,11 +192,14 @@ of the other top-level fields, such as `subject`, see [Statement]._
 > Map from [SlsaResult] to the number of the artifact's _transitive_ dependencies
 > that were verified at the indicated level. Absence of a given level of
 > [SlsaResult] MUST be interpreted as reporting _0_ dependencies at that level.
+> A set but empty `dependencyLevels` object means that the artifact has **no**
+> dependency at all, while an unset or null `dependencyLevels` means that the
+> verifier makes no claims about the artifact's dependencies.
 >
 > Users MUST count each dependency only once per SLSA track, at the highest
-> level verified. For example, if a dependency meets SLSA_BUILD_LEVEL_2,
-> you include it with the count for SLSA_BUILD_LEVEL_2 but not the count for
->SLSA_BUILD_LEVEL_1.
+> level verified. For example, if a dependency meets `SLSA_BUILD_LEVEL_2`,
+> you include it with the count for `SLSA_BUILD_LEVEL_2` but not the count for
+> `SLSA_BUILD_LEVEL_1`.
 
 <a id="slsaVersion"></a>
 `slsaVersion` _string, optional_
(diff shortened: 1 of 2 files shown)


AdamZWu:
Oh, hold on. `SLSA_BUILD_LEVEL_UNEVALUATED` got leaked from v1.1 into v1.0.
Let me fix that.

docs/spec/v1.0/verification_summary.md
```diff
@@ -191,12 +191,15 @@ of the other top-level fields, such as `subject`, see [Statement]._
 >
 > Map from [SlsaResult] to the number of the artifact's _transitive_ dependencies
 > that were verified at the indicated level. Absence of a given level of
-> [SlsaResult] MUST be interpreted as reporting _0_ dependencies at that level.
+> [SlsaResult] MUST be interpreted as reporting _0_ dependencies at that level,
+> e.g. an empty `dependencyLevels` object means that the artifact has **no**
+> dependency at all (vs. a non-present, i.e. "null", `dependencyLevels` means
+> that the verifier does not disclose the dependency stats).
```
MarkLodato, line 197:
```suggestion
> [SlsaResult] MUST be interpreted as reporting _0_ dependencies at that level.
> An set-but-empty `dependencyLevels` object means that the artifact has **no**
> dependency at all, while an unset or null `dependencyLevels` means that the
> verifier makes no claims about the artifact's dependencies.
```

Minor wording suggestion.

docs/spec/v1.0/verification_summary.md
```diff
@@ -191,12 +191,15 @@ of the other top-level fields, such as `subject`, see [Statement]._
...
 >
 > Users MUST count each dependency only once per SLSA track, at the highest
-> level verified. For example, if a dependency meets SLSA_BUILD_LEVEL_2,
-> you include it with the count for SLSA_BUILD_LEVEL_2 but not the count for
->SLSA_BUILD_LEVEL_1.
+> level verified. For example, if a dependency meets `SLSA_BUILD_LEVEL_2`,
+> you include it with the count for `SLSA_BUILD_LEVEL_2` but not the count for
+> the other `SLSA_BUILD` levels.
```
MarkLodato, line 202: I find this change to be less readable. What was the motivation? L1 is the only level below L2, so it seemed clear to me that this was explaining that if it's L2, you don't also report it at L1.

docs/spec/v1.0/verification_summary.md
```diff
@@ -252,17 +255,18 @@ WARNING: This is just for demonstration purposes.
...
--   SLSA_BUILD_LEVEL_0
--   SLSA_BUILD_LEVEL_1
--   SLSA_BUILD_LEVEL_2
--   SLSA_BUILD_LEVEL_3
--   FAILED (Indicates policy evaluation failed)
-
-Note that each SLSA level implies the levels below it. For example,
-SLSA_BUILD_LEVEL_3 means (SLSA_BUILD_LEVEL_1 + SLSA_BUILD_LEVEL_2 +
-SLSA_BUILD_LEVEL_3).
+SHOULD contain one value per SLSA track:
```
MarkLodato, line 258: This looks like a mistaken content change. This PR is just supposed to change formatting and minor wording changes. However, this change affects the schema, but also doesn't work. The field is a scalar String, but the new wording suggest it's a collection of strings.

MarkLodato (changes requested):
Thanks!

docs/spec/v1.0/verification_summary.md
AdamZWu: Oh good point, I think I overlooked that this is already in the context of "for example". :P

docs/spec/v1.0/verification_summary.md
AdamZWu: Oh right I think I got confused. WDYT I revise the description to clarify that the `SlsaResult` is a scalar that comprises values from multiple tracks (currently there are two: the "SLSA BUILD" track and "policy evaluation failure" track) This is to lead to the upcoming removal of the "policy evaluation failure" track values in v1.1.

MarkLodato: I think it's still worse than before. Frankly, this particular field needs to be redesigned in the next version. I don't think we should bother trying to fix the v1.0 wording. Instead, we should put effort towards figuring out what a better API is. So I would revert this part of the change (in v1.0 and v1.1) and leave that to a separate PR.

AdamZWu: Sounds good.

lehors (approved):
LGTM

trishankatdatadog (approved):
LGTM, thx!

PR #990, opened 2023-10-16
impl: rename "project" to "workstream"

MarkLodato:
As pointed out in #981, the term "project" is used for SLSA overall. For
example, see the top of CONTRIBUTING.md. To avoid overloading that term,
rename to "workstream". This seems to imply the same concept.


MarkLodato:
Once this is approved, I'll rename the existing issues and label. Thanks @lehors for pointing out the problem!

joshuagl (approved):

MarkLodato:
I renamed the issues and labels.

PR #993, opened 2023-10-27
content: add principle for defining levels/tracks

MarkLodato:
Document the principles around defining new levels and tracks. As we
expand SLSA, we need to ensure that SLSA remains focused and
understandable. A large part of that is having a consistent and simple
philosophy.

### DIFF
--- a/docs/spec/v1.1/principles.md
+++ b/docs/spec/v1.1/principles.md
@@ -8,8 +8,9 @@ decisions.
 
 ## Simple levels with clear outcomes
 
-Use levels to communicate security state and to encourage a large population to
-improve its security stance over time.
+Use [levels](levels) to communicate security state and to encourage a large
+population to improve its security stance over time. When necessary, split
+levels into separate tracks to recognize progress in unrelated security areas.
 
 **Reasoning:** Levels simplify how to think about security by boiling a complex
 topic into an easy-to-understand number. It is clear that level N is better than
@@ -21,19 +22,20 @@ Guidelines:
 -   **Define levels in terms of concrete security outcomes.** Each level should
     have clear and meaningful security value, such as stopping a particular
     class of threats. Levels should represent security milestones, not just
-    incremental progress.
+    incremental progress. Give each level an easy-to-remember mnemonic, such as
+    "Provenance exists".
 
 -   **Balance level granularity.** Too many levels makes SLSA hard to understand
     and remember, while too few makes each level hard to achieve. Try to bundle
     requirements into levels such that an implementer is likely implement all at
     the same time.
 
--   **Use tracks sparingly.** SLSA has independent sets of levels called
-    *tracks* that measure progress along independent axes. Strive for no more
-    than 3-5 tracks, lest SLSA be too complex to understand. Furthermore, avoid
-    creating overlapping or confusingly similar tracks. When a track is
-    warranted, define a crisp objective of the track, such as "trustworthy
-    provenance" for the [Build track](levels#build-track).
+-   **Use tracks sparingly.** Additional tracks add extra complexity to SLSA, so
+    a new track should be seen as a last resort. Each track should have a clear,
+    distinct purpose with a crisply defined objective, such as trustworthy
+    provenance for the [Build track](levels#build-track). As a rule of thumb, a
+    new track may be warranted the threats it addresses are unrelated to those
+    of other tracks. Try to avoid tracks that sound confusingly similar.
 
 ## Trust platforms, verify artifacts
 
--- a/docs/spec/v1.1/principles.md
+++ b/docs/spec/v1.1/principles.md
@@ -28,7 +28,9 @@ Guidelines:
 -   **Balance level granularity.** Too many levels makes SLSA hard to understand
     and remember; too few makes each level hard to achieve. Collapse levels
     until each step requires a non-trivial but manageable amount of work to
-    implement.
+    implement. Separate levels if they require significant work from multiple
+    distinct parties, such as infrastructure work plus user behavior changes, so
+    long as the intermediate level still has some security value (prior bullet).
 
 -   **Use tracks sparingly.** Additional tracks add extra complexity to SLSA, so
     a new track should be seen as a last resort. Each track should have a clear,
--- a/docs/spec/v1.1/principles.md
+++ b/docs/spec/v1.1/principles.md
@@ -34,9 +34,9 @@ Guidelines:
     a new track should be seen as a last resort. Each track should have a clear,
     distinct purpose with a crisply defined objective, such as trustworthy
     provenance for the [Build track](levels#build-track). As a rule of thumb, a
-    new track may be warranted if the threats it addresses are unrelated to
-    those of other tracks. Try to avoid tracks that sound confusingly similar in
-    either name or objective.
+    new track may be warranted if it addresses threats unrelated to another
+    track. Try to avoid tracks that sound confusingly similar in either name or
+    objective.
 
 ## Trust platforms, verify artifacts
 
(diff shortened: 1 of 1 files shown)


docs/spec/v1.1/principles.md
```diff
@@ -6,6 +6,37 @@ description: An introduction to the guiding principles behind SLSA's design deci
...
+    incremental progress. Give each level an easy-to-remember mnemonic, such as
+    "Provenance exists".
+
+-   **Balance level granularity.** Too many levels makes SLSA hard to understand
+    and remember, while too few makes each level hard to achieve. Try to bundle
+    requirements into levels such that an implementer is likely implement all at
+    the same time.
+
+-   **Use tracks sparingly.** Additional tracks add extra complexity to SLSA, so
+    a new track should be seen as a last resort. Each track should have a clear,
```
arewm, line 34: How should we handle the situation where some changes might need to be added to already-published levels in order to add balanced implementations? This action will require a bumping of the SLSA version, and I assume that we also want to create a balance to prevent that from happening too much.

Would it be preferable to start out with a combined track and then split them with a later major version bump if there was opposition or the reverse -- to start separate and major version bump to combine them later?

docs/spec/v1.1/principles.md
```diff
@@ -6,6 +6,37 @@ description: An introduction to the guiding principles behind SLSA's design deci
...
+Guidelines:
+
+-   **Define levels in terms of concrete security outcomes.** Each level should
+    have clear and meaningful security value, such as stopping a particular
+    class of threats. Levels should represent security milestones, not just
+    incremental progress. Give each level an easy-to-remember mnemonic, such as
+    "Provenance exists".
+
+-   **Balance level granularity.** Too many levels makes SLSA hard to understand
+    and remember, while too few makes each level hard to achieve. Try to bundle
```
arewm, line 29:
```suggestion
    and remember; too few makes each level hard to achieve. Try to bundle
```
The comma confused me when glancing at it as I read the line "and remember, while [...]" so making the separation of opposing concerns clearer with a semicolon.

docs/spec/v1.1/principles.md
```diff
@@ -6,6 +6,37 @@ description: An introduction to the guiding principles behind SLSA's design deci
...
+    requirements into levels such that an implementer is likely implement all at
+    the same time.
```
arewm, line 31: Haven't there been some comments about the difference between Provenance L2/L3? Should we specifically call out that there needs to be _intentional_ implementer actions to clearly differentiate between successive levels?

docs/spec/v1.1/principles.md
```diff
@@ -6,6 +6,37 @@ description: An introduction to the guiding principles behind SLSA's design deci
...
+
+-   **Use tracks sparingly.** Additional tracks add extra complexity to SLSA, so
+    a new track should be seen as a last resort. Each track should have a clear,
+    distinct purpose with a crisply defined objective, such as trustworthy
+    provenance for the [Build track](levels#build-track). As a rule of thumb, a
```
arewm, line 36: How "protected" should this crisp objective be? For example, if a track is created with a specific objective, only to find out later that the objective isn't crisp/precise enough, would a modification of the objective require a version bump?

docs/spec/v1.1/principles.md
```diff
@@ -6,6 +6,37 @@ description: An introduction to the guiding principles behind SLSA's design deci
...
+    new track may be warranted the threats it addresses are unrelated to those
```
arewm, line 37:
```suggestion
    new track may be warranted if the threats it addresses are unrelated to those
```

docs/spec/v1.1/principles.md
```diff
@@ -6,6 +6,37 @@ description: An introduction to the guiding principles behind SLSA's design deci
...
+    of other tracks. Try to avoid tracks that sound confusingly similar.
```
arewm, line 38: confusingly similar in "short name" (do we have a name for that, i.e. build) or objective.

arewm (changes requested):

docs/spec/v1.1/principles.md
MarkLodato: Hmm, I don't have any particular thoughts here. Do you think we need to touch on this, or can we just omit it if we don't have a strong guideline?

docs/spec/v1.1/principles.md
MarkLodato: Hmm, I didn't touch on versioning at all here. Should that instead go in https://slsa.dev/spec-stages#versioning so that all considerations about version bumps are in one place?

docs/spec/v1.1/principles.md
lehors: I don't have a problem with adding requirements to an existing level in a new version of the spec. I think it's easy enough to understand "SLSA 2.0 Build Level 3" vs "SLSA 1.0 Build Level 3" for instance, if we are consistently increasing the level of security provided.
On the other hand removing any requirement from an existing track+level, even through a split to a different track, would translate into something like "SLSA 2.0 Build Level 3" being less secure than "SLSA 1.0 Build Level 3" which would be terribly confusing.

lehors (commented):
Thanks for putting this together.

docs/spec/v1.1/principles.md
```diff
@@ -6,6 +6,38 @@ description: An introduction to the guiding principles behind SLSA's design deci
...
+
+-   **Define levels in terms of concrete security outcomes.** Each level should
+    have clear and meaningful security value, such as stopping a particular
+    class of threats. Levels should represent security milestones, not just
+    incremental progress. Give each level an easy-to-remember mnemonic, such as
+    "Provenance exists".
+
+-   **Balance level granularity.** Too many levels makes SLSA hard to understand
+    and remember; too few makes each level hard to achieve. Try to bundle
+    requirements into levels such that an implementer is likely implement all at
```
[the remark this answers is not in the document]
MarkLodato: fixed

docs/spec/v1.1/principles.md
arewm: I feel like a corollary to reducing the number of tracks would also be reducing the number of versions produced as that adds a different form of complexity to the specification -- more related to keeping up to date with the breaking changes introduced.

docs/spec/v1.1/principles.md
arewm: The current guidance for creating new levels is this: 

> Each level should have clear and meaningful security value, such as stopping a particular class of threats

What we do _not_ want, however, is for levels to be created based on classes of threats without work being done by the necessary party/parties. In order to increase the levels, work needs to be undertaken. 

New levels should be defined based on clearly identified classes of threats that are mitigated _AND_ clearly defined implementations that need to be undertaken. If both are not present, then a new level is likely not needed.

docs/spec/v1.1/principles.md
arewm: It might make sense to go there, yes. It is also relevant for the conventional commits -- for example, would these changes be considered editorial or would they be content changes? The conventional commit considerations would likely be too hard to pre-classify though.

trishankatdatadog (approved):
LGTM, thanks for explicitly clarifying all this!

docs/spec/v1.1/principles.md
MarkLodato: As discussed in today's spec meeting, I'll leave this to a future PR but leave the comment thread open for visibility.

docs/spec/v1.1/principles.md
MarkLodato: (same - will leave to a separate PR)

MarkLodato: OK, I gave an alternate wording a shot. How does this sound?

MarkLodato: FYI I just added a sentence about splitting levels, which I think is in line with @sudo-bmitch's comments at the spec meeting.

arewm (approved):

lehors (approved):
LGTM

PR #1008, opened 2023-12-11
fix: address new markdown-lint warnings triggered on new release

joshuagl:
Markdown lint v0.38.0 has some changes which result in new warnings for MD0030 and MD0045 that weren't triggered prior to the update.

- MD030/list-marker-space: is now being triggered for Markdown inside a comment block
- MD045/no-alt-text: is now being triggered for HTML  tags

Signed-off-by: Joshua Lock <joshua.lock@uk.verizon.com>

### DIFF
--- a/docs/spec/v1.0/terminology.md
+++ b/docs/spec/v1.0/terminology.md
@@ -3,22 +3,23 @@ title: Terminology
 description: Before diving into the SLSA specification levels, we need to establish a core set of terminology and models to describe what we're protecting.
 ---
 <!--- Note on updating docs:
-* Using terms such as "developer," "maintainer," "producer," "author," and
-  "publisher" interchangeably can cause confusion.
-  *  For consistency: Whenever possible, default to "producer," in line with the
-     model of producer--consumer--infrastructure provider. "Maintainer" is reserved
-     for sections specifying the act of continuing to maintain a project after its
-     creation, or when used in a less technical context where it is unlikely to cause
-     confusion. Author is reserved for the act of making source code commits or
-     reviews. Individual is used when the context's focus is specifying a single
-     person (i.e., "an individual's workstation" or "compromised individual").
-* Using terms such as "platform," "system," and "service" interchangeably can cause
-  confusion.
-  * For consistency: Whenever possible, default to "platform." Instead of using "service,"
-    a reference to a "hosted platform" should be used. A reference to some specific
-    software or tools internal to a platform can be made with "platform component" unless
-    there is a more appropriate definition to use directly like "control plane." External
-    self-described services and systems can continue to be called by these terms.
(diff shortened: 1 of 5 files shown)


docs/spec/v1.0/terminology.md
```diff
@@ -3,22 +3,23 @@ title: Terminology
...
+      person (i.e., "an individual's workstation" or "compromised individual").
+*   Using terms such as "platform," "system," and "service" interchangeably can cause
+    confusion.
+  *   For consistency: Whenever possible, default to "platform." Instead of using
+      "service,"a reference to a "hosted platform" should be used. A reference to some
+      specific software or tools internal to a platform can be made with "platform
+      component" unless there is a more appropriate definition to use directly like
+      "control plane." External self-described services and systems can continue to
+      be called by these terms.
 --->
```
MarkLodato, line 23: This is actually a bug introduced in markdownlint v0.32. It shouldn't format HTML comments at all, and it certainly shouldn't only fire one rule but not others. In this case, MD030 and a few others fail to detect HTML comments that end with an extra `-`. The workaround is simple: remove the extra `-`.

I reported the bug here: https://github.com/DavidAnson/markdownlint/issues/1067

So please revert this hunk and replace with:

```suggestion
-->
```

docs/spec/v1.1/terminology.md
```diff
@@ -3,22 +3,23 @@ title: Terminology
...
+      person (i.e., "an individual's workstation" or "compromised individual").
+*   Using terms such as "platform," "system," and "service" interchangeably can cause
+    confusion.
+  *   For consistency: Whenever possible, default to "platform." Instead of using
+      "service,"a reference to a "hosted platform" should be used. A reference to some
+      specific software or tools internal to a platform can be made with "platform
+      component" unless there is a more appropriate definition to use directly like
+      "control plane." External self-described services and systems can continue to
+      be called by these terms.
 --->
```
MarkLodato, line 23: Same: revert this hunk and fix this one ending tag

```suggestion
-->
```

MarkLodato (changes requested):

docs/spec/v1.0/terminology.md
joshuagl: Thanks! I suspected as much and did try to to search for related issues, but didn't find any.

MarkLodato: Technically `--->` does not comply with [CommonMark 0.16-0.30](https://spec.commonmark.org/0.30/#html-comment), though this restriction is slated to be [removed](https://github.com/commonmark/commonmark-spec/commit/d5ddfae696c53f09fd5b6182238de716dddeb40a) in 0.31. 🤷

MarkLodato (approved):
Thanks!

PR #1009, opened 2023-12-12
editorial: Separate producer from platform requirements in Levels

marcelamelara:
This PR adds a separate "Build Platform Requirements" row in each table of levels.md, more clearly defining what SW producers vs. build platform are responsible for at each level. To be clear, the requirements themselves have not changed, but rather the row in which they appear.

This change more closely matches the current separation in requirements.md, and also facilitates adding requirements for each role in higher/newer Build Track levels.

### DIFF
--- a/docs/spec/v1.1/levels.md
+++ b/docs/spec/v1.1/levels.md
@@ -97,6 +97,7 @@ SLSA---other than tamper protection---without changing their build workflows.
 
 <dt>Requirements<dd>
 
+-   Software Producer:
     -   Follow a consistent build process so that others can form
         expectations about what a "correct" build looks like.
     -   Run builds on a build platform that meets Build L1 requirements.
--- a/docs/spec/v1.1/levels.md
+++ b/docs/spec/v1.1/levels.md
@@ -103,7 +103,6 @@ SLSA---other than tamper protection---without changing their build workflows.
     -   Run builds on a build platform that meets Build L1 requirements.
     -   Distribute provenance to consumers, preferably using a convention
         determined by the package ecosystem.
-
 -   Build platform:
     -   Automatically generate [provenance] describing how the artifact was
         built, including: what entity built the package, what build process
@@ -159,7 +158,7 @@ All of [Build L1], plus:
 -   Build platform:
     -   Generate and sign[^sign] the provenance itself. This may be done
         during the original build, an after-the-fact reproducible build, or
-        some equivalent platform that ensures the trustworthiness of the
+        some equivalent system that ensures the trustworthiness of the
         provenance.
 -   Consumer:
     -   Validate the authenticity of the provenance.
--- a/docs/spec/v1.1/levels.md
+++ b/docs/spec/v1.1/levels.md
@@ -95,19 +95,17 @@ but is trivial to bypass or forge.
 Projects and organizations wanting to easily and quickly gain some benefits of
 SLSA---other than tamper protection---without changing their build workflows.
 
-<dt>Software Producer Requirements<dd>
+<dt>Requirements<dd>
 
--   Follow a consistent build process so that others can form expectations
-    about what a "correct" build looks like.
-
--   Distribute provenance to consumers, preferably using a convention
-    determined by the package ecosystem.
-
-<dt>Build Platform Requirements<dd>
-
--   Automatically generate [provenance] describing how the artifact was
-    built, including: what entity built the package, what build process they
-    used, and what the top-level input to the build were.
+-   Software producer:
+    -   Follow a consistent build process so that others can form
+        expectations about what a "correct" build looks like.
+    -   Distribute provenance to consumers, preferably using a convention
+        determined by the package ecosystem.
+-   Build platform:
+    -   Automatically generate [provenance] describing how the artifact was
+        built, including: what entity built the package, what build process
+        they used, and what the top-level input to the build were.
 
 <dt>Benefits<dd>
 
@@ -149,22 +147,20 @@ Projects and organizations wanting to gain moderate security benefits of SLSA by
 switching to a hosted build platform, while waiting for changes to the build
 platform itself required by [Build L3].
 
-<dt>Software Producer Requirements<dd>
+<dt>Requirements<dd>
 
 All of [Build L1], plus:
 
--   Run builds on a hosted build platform that meets Build L2 requirements.
-
--   Enable downstream verifiers of provenance to validate the authenticity
-    of the provenance.
-
-<dt>Build Platform Requirements<dd>
-
-All of [Build L1], plus:
-
--   Generate and sign[^sign] the provenance itself. This may be done during
-    the original build, an after-the-fact reproducible build, or some
-    equivalent platform that ensures the trustworthiness of the provenance.
+-   Software producer:
+    -   Run builds on a hosted build platform that meets Build L2
+        requirements.
+-   Build platform:
+    -   Generate and sign[^sign] the provenance itself. This may be done
+        during the original build, an after-the-fact reproducible build, or
+        some equivalent platform that ensures the trustworthiness of the
+        provenance.
+-   Consumer:
+    -   Validate the authenticity of the provenance.
 
 <dt>Benefits<dd>
 
@@ -204,21 +200,19 @@ strong tamper protection.
 Most software releases. Build L3 usually requires significant changes to
 existing build platforms.
 
-<dt>Software Producer Requirements<dd>
+<dt>Requirements<dd>
 
 All of [Build L2], plus:
 
--   Run builds on a hosted build platform that meets Build L3 requirements.
-
-<dt>Build Platform Requirements<dd>
-
-All of [Build L2], plus:
-
--   Implement strong controls to:
-
-    -   prevent runs from influencing one another, even within the same project.
-    -   prevent secret material used to sign the provenance from being
-        accessible to the user-defined build steps.
+-   Software producer:
+    -   Run builds on a hosted build platform that meets Build L3
+        requirements.
+-   Build platform:
+    -   Implement strong controls to:
+        -   prevent runs from influencing one another, even within the same
+            project.
+        -   prevent secret material used to sign the provenance from being
+            accessible to the user-defined build steps.
 
 <dt>Benefits<dd>
 
--- a/docs/spec/v1.1/levels.md
+++ b/docs/spec/v1.1/levels.md
@@ -95,17 +95,19 @@ but is trivial to bypass or forge.
 Projects and organizations wanting to easily and quickly gain some benefits of
 SLSA---other than tamper protection---without changing their build workflows.
 
-<dt>Requirements<dd>
+<dt>Software Producer Requirements<dd>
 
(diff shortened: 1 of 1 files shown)


marcelamelara:
cc @joshuagl since I added this to the v1.1 directory (please let me know, if this shouldn't be part of those changes)

docs/spec/v1.1/levels.md
```diff
@@ -95,17 +95,19 @@ but is trivial to bypass or forge.
...
-<dt>Requirements<dd>
+<dt>Software Producer Requirements<dd>
+
+-   Follow a consistent build process so that others can form expectations
+    about what a "correct" build looks like.
 
--   Software producer follows a consistent build process so that others can form
-    expectations about what a "correct" build looks like.
+-   Distribute provenance to consumers, preferably using a convention
+    determined by the package ecosystem.
```
jkjell, line 104: Perhaps we could also add something like:

>Additionally, leveraging a [SECURITY-INSIGHTS.yml](https://github.com/ossf/security-insights-spec/blob/main/specification.md) can offer machine-readable information about the expected location to retrieve a provenance document.

jkjell (approved):
I really like the clarity this separation of responsibilities brings!

docs/spec/v1.1/levels.md
kpk47: > Perhaps we could also add something like:
> 
> > Additionally, leveraging a [SECURITY-INSIGHTS.yml](https://github.com/ossf/security-insights-spec/blob/main/specification.md) can offer machine-readable information about the expected location to retrieve a provenance document.

This sounds like an implementation suggestion rather than part of the levels specification, so I don't think it belongs on this page. I'd suggest we move discussion into a new PR since adding it here would change this from an editorial PR to a content one.

kpk47 (approved):

docs/spec/v1.1/levels.md
marcelamelara: Agree with @kpk47 . I like this suggestion, but also think it goes beyond the scope of this PR.

docs/spec/v1.1/levels.md
```diff
@@ -147,18 +149,23 @@ Projects and organizations wanting to gain moderate security benefits of SLSA by
...
 
--   The build runs on a hosted build platform that generates and signs[^sign] the
-    provenance itself. This may be the original build, an after-the-fact
-    reproducible build, or some equivalent platform that ensures the
-    trustworthiness of the provenance.
+-   Run builds on a hosted build platform that meets Build L2 requirements.
 
--   Downstream verification of provenance includes validating the authenticity
+-   Enable downstream verifiers of provenance to validate the authenticity
     of the provenance.
```
MarkLodato, line 159: Hmm, this isn't a producer requirement though. It's more of an "ecosystem" requirement.

That said, the original bullet wasn't clear to begin with. We really need to rework verification for v1.1. The intent of this bullet (which doesn't really come across) is that merely publishing provenance is not sufficient at L2. Something needs to check that provenance, and only then does it "become" L2. But this really needs to be fleshed out...

Not sure how best to align with this PR though.

marcelamelara: I see, that's important nuance. 

As @MarkLodato points out, this isn't within the scope of this PR, but a question that immediately comes to mind with this "ecosystem requirement" is how anyone in the ecosystem would verify that the provenance has been checked? It sounds like a VSA generated by a downstream entity is part of this requirement. 

One way that I can think to resolve this in the context of this PR might be to add a third row starting at L2 that is for "Consumer Requirements", and reword the requirement:

```suggestion
-   Ensure that provenance, including its authenticity, is verified. This may be delegated to a third-party verifier. 
     Note: this requirement will be refined in a future version of the spec.
```

I am a little worried that this change turns this PR into a content PR, rather than merely an editorial PR. I'm also wondering if this PR should be part of v1.0 or v1.1.

MarkLodato: v1.1 sounds seems right to me (let's leave v1.0 stable), as does sticking to only editorial changes in this PR.

Instead of adding separate rows per party, have you considered prefixing each bullet with the party? Would that achieve the same effect without having too many rows? For example:

-   **Software producer:** Run builds on a hosted build platform that meets Build L2 requirements.
-   **Build platform:** Generate and sign the provenance itself. [...]
-   **Consumer:** Validate the authenticity of the provenance.

Here "Consumer" might not be a perfect fit, but perhaps it's good enough for this PR and we can iterate later?

marcelamelara: I'm fine prefixing bullets with the party if we're going beyond producer/platform requirements. That will certainly keep the tables a little more concise.

FWIW, I think that "consumer" does make a lot of sense here because it sets us up for making more explicit verification requirements. For example, IMO, consumers of artifacts should always check the authenticity of the provenance for each consumed artifact (whether that's as a dependency or as an application), or delegate this task to a trusted verifier. That said, there might be other verification scenarios that don't pertain to the consumer at L2, so I agree that this needs to be iterated over some more.

docs/spec/v1.1/levels.md
```diff
@@ -97,15 +97,17 @@ SLSA---other than tamper protection---without changing their build workflows.
...
-    process they used, and what the top-level input to the build were.
+    -   Follow a consistent build process so that others can form
+        expectations about what a "correct" build looks like.
 
--   Software producer distributes provenance to consumers, preferably using a
-    convention determined by the package ecosystem.
+    -   Distribute provenance to consumers, preferably using a convention
+        determined by the package ecosystem.
+
+-   **Build platform:** Automatically generate [provenance] describing how
```
kpk47, line 108: I don't like the inconsistency in this presentation. I think it would be best to standardize on the format
```
**Actor:** Requirement
**Actor2:** Requirement
**Actor2:** Requirement
...
```

marcelamelara: The main goal of this PR is to present a more clearly visible split between the different actors, which IMO a single list of bullets (which is the current format) doesn't do very well. Can we compromise on something like this?

```
- **Actor1**:
   - Requirement1
   - Requirement2
- **Actor2**:
   - Requirement1
- **Actor3**:
   - Requirement1
   - Requirement2
```

docs/spec/v1.1/levels.md
```diff
@@ -97,15 +97,19 @@ SLSA---other than tamper protection---without changing their build workflows.
 
 <dt>Requirements<dd>
 
--   Software producer follows a consistent build process so that others can form
-    expectations about what a "correct" build looks like.
+-   **Software producer:**
```
MarkLodato, line 100: suggestion: Remove the bold on each entry (not longer needed if it's not inline with the text) and remove blank lines between bullets to make it a "compact" list. Here's a screenshot of what it would look like with those changes:

image

MarkLodato (approved):

docs/spec/v1.1/levels.md
marcelamelara: done.

docs/spec/v1.1/levels.md
```diff
@@ -97,15 +97,15 @@ SLSA---other than tamper protection---without changing their build workflows.
...
-    artifact was built, including: what entity built the package, what build
-    process they used, and what the top-level input to the build were.
-
--   Software producer distributes provenance to consumers, preferably using a
-    convention determined by the package ecosystem.
+-   Software producer:
+    -   Follow a consistent build process so that others can form
+        expectations about what a "correct" build looks like.
+    -   Distribute provenance to consumers, preferably using a convention
+        determined by the package ecosystem.
```
arewm, line 104: One key difference between this section and requirements.md is that the specification indicates

> The producer MUST select a build platform that is capable of reaching their desired SLSA Build Level.

This is slightly stronger than following a consistent build process. It is an assumption that is tied in to the fact that producers have to distribute provenance. 

```suggestion
    -   Follow a consistent build process so that others can form
        expectations about what a "correct" build looks like.
    -   Run builds on a hosted build platform that meets Build L1
        requirements.
    -   Distribute provenance to consumers, preferably using a convention
        determined by the package ecosystem.

```

arewm (changes requested):

docs/spec/v1.1/levels.md
marcelamelara: Makes sense to me. I'm ok with adding in the extra bullet point as long as this still counts as a purely editorial change. @MarkLodato ?

MarkLodato: I'm fine with that.

marcelamelara: One fine point to mention though, is that the "hosted" build platform is not an L1 requirement, so that's the only edit I'd make to the extra bullet point.

docs/spec/v1.1/levels.md
```diff
@@ -151,13 +151,16 @@ platform itself required by [Build L3].
...
 
--   The build runs on a hosted build platform that generates and signs[^sign] the
-    provenance itself. This may be the original build, an after-the-fact
-    reproducible build, or some equivalent platform that ensures the
-    trustworthiness of the provenance.
-
--   Downstream verification of provenance includes validating the authenticity
-    of the provenance.
+-   Software producer:
+    -   Run builds on a hosted build platform that meets Build L2
```
marcelamelara, line 155:
```suggestion
    -   Run builds on a **hosted** build platform that meets Build L2
```

docs/spec/v1.1/levels.md
```diff
@@ -151,13 +153,16 @@ platform itself required by [Build L3].
...
-
--   Downstream verification of provenance includes validating the authenticity
-    of the provenance.
+-   Software producer:
+    -   Run builds on a [hosted] build platform that meets Build L2
+        requirements.
+-   Build platform:
+    -   Generate and sign[^sign] the provenance itself. This may be done
+        during the original build, an after-the-fact reproducible build, or
+        some equivalent platform that ensures the trustworthiness of the
```
joshuagl, line 162: Nit: the "platform [...] use platform" jars a little, could we do something like:
```suggestion
        some equivalent system that ensures the trustworthiness of the
```

docs/spec/v1.1/levels.md
```diff
@@ -97,15 +97,17 @@ SLSA---other than tamper protection---without changing their build workflows.
...
-
--   Software producer distributes provenance to consumers, preferably using a
-    convention determined by the package ecosystem.
+-   Software Producer:
+    -   Follow a consistent build process so that others can form
+        expectations about what a "correct" build looks like.
+    -   Run builds on a build platform that meets Build L1 requirements.
+    -   Distribute provenance to consumers, preferably using a convention
+        determined by the package ecosystem.
+
```
joshuagl, line 106:
```suggestion
```
The other entries are a single list with each top-level bullet being an actor and the sub-bullets being requirements on the actor. Removing this line collapses L1 to match the other lists.

joshuagl (approved):
I really like how this has turned out, thanks Marcela and all the reviewers for iterating on this!

marcelamelara:
The last set of recommended changes are in. Thank you all for the feedback!

PR #1025, opened 2024-02-06
impl: use consistent "only-leading" table syntax

### LINKED ISSUE #1023: Markdown style for tables
Our Markdown linter now (as of markdownlint-cli v0.39, #1021) checks Markdown tables for [syntactic consistency](https://github.com/DavidAnson/markdownlint/blob/main/doc/md055.md). The question is whether to use leading and trailing pipes or just leading pipes. Examples:

Leading and trailing pipes:

```markdown
| Key | Value |
| --- | ----- |
| A   | One   |
| B   | Two   |
| C   |       |

| Table with | uneven columns |
| --- | --- |
| A really long line that has a lot of content. | Trailing pipe is hard to see |
| Lines are not the same length. | See? |
| No second column. | |
```

Only leading pipes:

```markdown
| Key | Value
| --- | -----
| A   | One
| B   | Two
| C   |

| Table with | uneven columns
| --- | ---
| A really long line that has a lot of content. | Trailing pipe is hard to see
| Lines are not the same length. | See?
| No second column. |
```

@joshuagl expressed preference for only leading pipes in #1021.

Personally I think smaller tables look slightly better with trailing pipes but larger tables look much worse with them. Given that markdownlint doesn't allow table-by-table decision, I'm also inclined to go with only leading pipes.

I'll send out a PR to fix.

MarkLodato:
Fixes #1023 by using a consistent Markdown syntax for tables with a
leading `|` but no trailing trailing `|`. Change the markdown lint
config to require this in all files. Example:

```markdown
| Key | Value
| --- | -----
| One | A
| Two | B
```


MarkLodato:
I'll leave this open for a day or two in case anyone else wants to offer an opinion.

lehors (approved):
I support this approach.
Thanks.

PR #1032, opened 2024-02-13
impl: fix comparison instructions

MarkLodato:
Commit 74452a8 added instructions to compare builds, but it had a few
bugs:

- Did not work on MacOS due to differences between BSD `sed` (Mac) and
  GNU `sed` (Linux)
  - To edit in-place without a backup file, GNU sed uses `-i` (no
    parameter) and BSD sed uses `-i ''`. There is no cross-platform way
    to do this. So instead, just use a backup file (`-i.bak`) then
    delete it afterward.
  - `\+` is a GNU extension to basic regexp. Switch to extended regexp
    (`-r`) which works on GNU sed and modern BSD sed.
- Incorrectly edited all dates instead of just the first. Only the first
  is the build timestamp (`site.time` in Jekyll). The rest are
  deterministic and should not be scrubbed. Use the `1` sed range to
  fix.
- Did not note that the bash `**/*` syntax requires `shopt -s globstar`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -97,15 +97,18 @@ $ mv _site _site.A
 # Prepare version B
 $ JEKYLL_BUILD_REVISION=COMMIT bundle exec jekyll build
 $ mv _site _site.B
-$ sed -i -e 's@<updated>[^<]\+</updated>@<updated>DATE</updated>@' _site.{A,B}/feed.xml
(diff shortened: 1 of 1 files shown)


lehors (approved):
Yeah, that works for me! :-)

MarkLodato:
Thanks for reporting the issue, @lehors!

Issue #961, opened 2023-08-28
Workstream: Dependency Track


kpk47:
We discussed a dependency track in the community meeting on Aug 28, 2023. It may have significant overlap with the Source Track, so we should discuss them together.

melba-lopez:
Since dependencies does overlap, i just want folks to remember the original thread on source that @marcelamelara and I originally were thinking through  https://github.com/slsa-framework/slsa/issues/463  

Also, dependencies and some of source can also leverage s2c2f controls and maybe SLSA can just require/reference those controls "integrate/handshake" with S2C2F.  @camaleon2016 and i have discussed this in the past and have talked about it from an SCI WG gap assessment.

arewm:
There are many different aspects to dependencies which will be relevant to hardening the supply chain. Instead of creating a track just for dependencies, it might make more sense to figure out how dependencies relate across various other current, proposed, or in development tracks.

As an example, another related property to dependencies is reproducibility (i.e. in terms of ensuring that dependencies are appropriately pins as a basic requirement towards achieving reproducibility). I mentioned this relationship in the [document](https://docs.google.com/document/d/1Jk0yZnkTC3dfp8G5dmO8K9r1Kc7TRX2QVOwcFSKw1OQ/edit) which has been started for the reproducibility (https://github.com/slsa-framework/slsa/issues/873#issuecomment-1754089904).

The initial proposal for reproducibility was to tack it on to the build track, but my proposal was to instead include it as higher ordered levels on top of these basic dependency-based properties.

meder:
Hi everyone,

I'd like to start drawing some boundaries around what the Dependencies Track will cover:

1. Objective here would be to enable fundamental capabilities that lead to and enable reduction of risk from exposure to 3P dependencies.
2. Dependency is defined using [SLSA's terminology](https://slsa.dev/spec/v1.0/terminology#build-model:~:text=initialize%20the%20build.-,Dependencies,-Artifacts%20fetched%20during).
3. 3P qualification means that the dependency is owned by an entity different to the publisher of a software package.
4. Levels should apply to the software packages (SLSA terminology).

Given the breadth of risks, both in nature and impact, the track likely doesn't need to get too specific about things like known vulnerabilities and SLOs, etc and instead focus on improvements and controls that enable effective management of that risk.

Would love to hear others' thoughts and comments and if that aligns with their thinking.

mlieberman85:
[...]

Issue #956, opened 2023-08-22
Workstream: Source Track


kpk47:
This is a tracking issue for creating a Source track. The main idea is to cover properties of how the source code was developed. The exact thrust of this track --- i.e. the threats mitigated by this track --- are still TBD.

[Workstream shepherd]: Kris K (@kpk47)

Sub-issues:

-   [ ] TODO

[Workstream shepherd]: https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md#workstream-lifecycle

---

SLSA v0.1 set requirements on source code management that we removed from v1.0. We should reintroduce those requirements (or something similar) in a Source Track. We will also need to create a format for any source attestations.

A few questions to start discussion:
1. What does the Source level attach to? Is it a project, a repo, a commit, or something else?
2. Which version control systems do we need to consider?
3. What sort of guarantees should the source track make? Traceability (i.e. the source came from this repo), transparency (i.e. this code was written by this organization, person, etc), quality (i.e. the source is trustworthy because x y z), others?
4. Will the source track have the same requirements for open and closed source projects? The same standards of evidence for meeting those requirements?

joshuagl:
cc @TomHennen who has done a lot of thinking about source attestations in the past.

steiza:
This is great! I'm going to take a stab at answering these questions.

> 1. What does the Source level attach to? Is it a project, a repo, a commit, or something else?

I think the top-level object should be a list of changesets. A changeset can be made up of one or more commits. A changeset should have a 1-1 relationship with a review process (e.g. pull request in GitHub terminology or merge request in GitLab terminology), although a changeset might not have a review if it was pushed directly to a remote. A changeset has a 1-1 relationship with a repository, but the changesets described in the source attestations for an artifact could come from more than one repository.


> 2. Which version control systems do we need to consider?

Pragmatically, I would focus on git. At the very least, a graph-based distributed version control system. I worry that we might not be able to come up with a specification that is meaningful **and** covers all the version control systems out there.


> 3. What sort of guarantees should the source track make?

Generally, I'd lean towards objective statements that don't change over time. Instead of "no vulnerabilities found at X time" do things like list components for later determination of vulnerabilities, or a specific command that was run and the output it produced.

We should frame these in terms of what verification policies we'd want to run, and then figure out how to represent them such that the policy can be evaluated.

The shortest list of things I'd like to verify is things like:
- All contributions have one or more strongly authenticated authors
[...]

Issue #287, opened 2022-02-02
Strongly authenticated robots


ribbybibby:
> Strong authentication: Authentication that maps back to a specific person using an authentication mechanism which is resistant to account and credential compromise. For example, 2-factor authentication (2FA) where one factor is a hardware security key (i.e. YubiKey

If you have a robot account that is performing actions in Github, like reviewing or merging PRs, can that robot be considered to have 'strong authentication'? I can configure 2FA for the robot account but the token that the robot ultimately uses could conceivably be compromised and used without any second factor being required.

david-a-wheeler:
> Can a robot count as one or even two persons? (for Two-person reviewed)

We might consider allowing relaxing two-person review to one person if the proposal is from a "trusted robot". Most "trusted robot" changes are (unsurprisingly) very mechanical, like reformatting or renaming.

However, while you should *definitely* use many tools to scan proposals before accepting them, I don't think robots make very good reviewers. Robots typically look at narrow issues, not the issues a human can understand.

The bigger problem is that I think many OSS projects cannot *implement* "two-person reviewed", or if they do they risk long delays of security-relevant functions. That said, obviously it's better to require two-person reviewed if you can do that reliably.

joshuagl:
Relates to #196. The two-party approval requirements will most likely not make 1.0, except the recommendation to use two-party approval for all administrative changes to the build service. Therefore I think we can defer this till post 1.0

zachariahcox:
I believe this topic is covered by https://github.com/slsa-framework/slsa/issues/196

The working theory is that SLSA source track is primarily concerned with the trustworthy generation of provenance attestations.
The implied meaning of the contents of those attestations (and whether or not combinations of rules qualify a revision for various grades) is opinionated and up to the issuer of the vSA.

Marking "closed" for now. Please reopen if you feel that this topic has not been addressed fully! Thanks!

Issue #196, opened 2021-10-27
Code review exemption for changes by trusted robots?


MarkLodato:
For the [two-person reviewed](https://slsa.dev/requirements#two-person-reviewed) requirement at SLSA 4, how should we handle changes authored by robots? It's always an option for two trusted persons to review and approve, but what if we want the changes to be completely automated or to only require a single reviewer? Should there be some exemption in the case that the trusted robot is locked down and honors the two-person principle?

Hixie:
One example of a robot we use in Flutter land is the autoroll bot for rolling one of our repositories into another one of our repositories. We use the same rules for contribution to both repositories and to the bot itself, but it would be impractical to also require those rules for the bot's actions, since the bot is continually rolling (one commit every 3 hours, on average).

justinabrahms:
Another example of one of these robots is my company has several teams which use a robo-committer to bump versions in their pom.xml files on commit.

zachariahcox:
I believe the current thinking is that SLSA is concerned with tamper-resistant provenance attestations. 
EG: It is up to more opinionated "VSA"-type tooling to decide whether or not robots are allowed. 
At the very least, we expect the "evidence" and the claims summarized by that evidence to be issued by different actors.

PR #1037, opened 2024-03-13
content: Add draft of the Source track.

kpk47:
This change adds the working draft of SLSA's Source track. It includes basic terminology, level requirements, and an attestation format.


docs/spec/v1.1/levels.md
```diff
@@ -229,14 +233,118 @@ All of [Build L2], plus:
...
+
+<dt> Benefits <dd>
+Version control is a software development best practice, and discussing its benefits is beyond the scope of this page.
+</dl>
+
+### Source L2: Verified history
+
+<dl class="as-table">
+<dt> Summary <dd>
+The project is stored and managed through a source control platform that ensures the change history's integrity..
```
ramonpetgrave64, line 288:
```suggestion
The project is stored and managed through a source control platform that ensures the change history's integrity.
```

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +233,118 @@ All of [Build L2], plus:
...
+<dt> Intended for <dd>
+Organizations that are unwilling or unable to incorporate code review into their software development practices.
+
+<dt> Requirements <dd>
+
+**[Strong authentication]** User accounts that can modify the source or the project's configuration must use two-factor authentication or its equivalent.
+
+**[Verified timestamps]** Each entry in the change history must contain at least one timestamp that is determined by the source control platform and cannot be modified by clients. It MUST be clear in the change history which timestamps are determined by the source control platform.
+
+**[Retained history]** The change history MUST be preserved as long as the source is hosted on the source control system. The source MAY migrate to another source control system, but the organization MUST retain the change history if possible. It MUST NOT be possible for persons to delete or modify the change history, even with multi-party approval, except by trusted platform admins following an established deletion policy.
```
laurentsimon, line 299: nit: you could provide a use case to motivate it, like if a secret must be deleted from history

docs/spec/v1.1/source-att.md
```diff
@@ -0,0 +1,89 @@
...
+## Schema
+
+| Field | Type | Required at Level | Description
+| --- | --- | --- | ---
+| `projectID` | string | L1+ | Immutable project ID. For example, the SLSA Specification's project ID is `www.github.com/slsa-framework/slsa`.
+| `revisionID` | string | L1+ | Immutable revision ID. In git terminology, this is the commit hash.
+| `digest` | string | L1+ | Digest of the source. In git terminology, this is the repo's tree hash.
+| `timestamp` | string | L2+ | ISO 8601 representation of the time at which the version control platform generated the attestation. This timestamp MUST be determined by the platform.
+| `2FA` | FeatureHistory | L2+ | Whether 2FA was enabled for this project.
+| `historyRetained` | FeatureHistory | L2+ | Whether this project retains its version history.
```
laurentsimon, line 42: So _only_ the platform is able to provide the feature of history retention? What if a company mirrors a project that has no source level, but performs the retention themselves? Should they not be able to create a source attestation themselves?

docs/spec/v1.1/source-att.md
```diff
@@ -0,0 +1,89 @@
...
+A SLSA Source attestation asserts that a project follows certain practices
+pertaining to its source code. Attestations are scoped to a project as a whole
+and generated when creating a complete copy of that project's source. For
+example, a project may produce a source attestation when exporting its source
+for distribution or when a build pipeline fetches the source.
+
+## Provenance generation
+
+The version control platform is completely trusted in SLSA. Attestations for
+Source L2+ MUST be generated by the trusted version control platform.
```
laurentsimon, line 27: this seems a bit constraining. How about relaxing it a bit: the platform MUST provide access to the information (e.g., via API). The platform SHOULD generate the attestation, but other party MAY generate it using platform's API, etc.

docs/spec/v1.1/source-att.md
```diff
@@ -0,0 +1,89 @@
...
+
+| Field | Type | Required at Level | Description
+| --- | --- | --- | ---
+| `projectID` | string | L1+ | Immutable project ID. For example, the SLSA Specification's project ID is `www.github.com/slsa-framework/slsa`.
+| `revisionID` | string | L1+ | Immutable revision ID. In git terminology, this is the commit hash.
+| `digest` | string | L1+ | Digest of the source. In git terminology, this is the repo's tree hash.
+| `timestamp` | string | L2+ | ISO 8601 representation of the time at which the version control platform generated the attestation. This timestamp MUST be determined by the platform.
+| `2FA` | FeatureHistory | L2+ | Whether 2FA was enabled for this project.
+| `historyRetained` | FeatureHistory | L2+ | Whether this project retains its version history.
+| `codeReview` | FeatureHistory | L3+ | Whether this project requires code review.
```
laurentsimon, line 43: These fields are a "summary" of a policy enforced by the platform. However there are many cases where you need to have a different kind of attestation that simply enumerates the list of commits and their reviewers:
1. When the policy (containing _who_ should be allowed to review) is not useable by consumers. That's the case for most small open-source projects. From a consumer's point of view. there is _no_ way to detect whether identities are the same person or not. Yet, we want to enable consumers to know about _who_ reviewed the code, to enable them to make their own judgment.
2. There are a lot of corner cases in practice. Example: someone edits a PR and then LGTM. This is common on dependabot or other automated PRs. user must be able to customize their policy and risk tolerance. 

I think we need to separate _policy_ from the _facts_. Example: in the build track, the builder creates an attestation which is a list of _facts_: the repo, the commit, etc. There is _no_ policy involved, ie no interpretation. That is left to consumers, ie they decide if the right repo was used, etc. The VSA (or other attestation)'s role is for an org (or project) to hide the details.

Similarly, here, there are facts to report: the settings, who reviewed each PR, etc. These fields should be part of the source attestation.

Then come the policy, which is _who_ is able to review. In the current proposal, the attestation is essentially a VSA, not an attestation about facts. The VSA is created by the CVS platform, which is totally fine. But there are cases where this cannot be the case, and we should enable these other use cases.

docs/spec/v1.1/source-reqs.md
```diff
@@ -0,0 +1,104 @@
...
+
+-   **Scope** The Source track is now scoped to Revisions rather than builds.
+Why?: To facilitate verification without anchoring it to a build.
+
+-   **Model** Added a model, definitions, and the concept of verification.
+Why?: SLSA does not yet have a model for version control systems, and we need such a model to be able to discuss them.
+
+## Infrastructure Requirements
+
+The version control platform MUST provide at least:
```
laurentsimon, line 17: The model here seems a bit constraining. Not every CVS is in charge of reviews. Think gerrit on GitHub. can we capture that?

docs/spec/v1.1/source-att.md
ianlewis: +1

I think for code review we should have some other trusted way to determine if code was reviewed and verify that way. `Reviewed-by` trailers are an example, though it may be hard to trust them. Maybe you could also verify via an API who's information can't be trivially modified.

kpk47: Are you proposing attestations for each review/commit/PR? We discussed that option in the Google Doc version but rejected it because it would be cumbersome to manage all those attestations. And since they'd ultimately come from the source platform, we wouldn't gain much since you have to trust the source platform in either case.

laurentsimon: > Are you proposing attestations for each review/commit/PR

No. I'm proposing an attestation that _lists_ PRs. The one you have here is fine as well, but it's an opinionated attestation based on a policy evaluation.

docs/spec/v1.1/source-reqs.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Changes from v0.1
+
+-   **Scope** The Source track is now scoped to Revisions rather than builds.
+Why?: To facilitate verification without anchoring it to a build.
+
+-   **Model** Added a model, definitions, and the concept of verification.
+Why?: SLSA does not yet have a model for version control systems, and we need such a model to be able to discuss them.
+
+## Infrastructure Requirements
```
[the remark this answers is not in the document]
kpk47: I moved these requirements up into Infrastructure Requirements.

docs/spec/v1.1/source-reqs.md
kpk47: @laurentsimon Again, this is about where you draw the box. I'd rather just draw the box around all the trusted source components and call it a day. I would need to understand the purported benefit of further detail to change my mind.

@ianlewis I don't think it's practical to relegate code review checks to end user policy evaluation. See my other comment for more detail. What benefit do you see to commit signing here? Are you proposing that reviewers sign the commits? IIUC such a scheme is impractically difficult to implement because each signature effectively creates a new commit that invalidates the previous one. If we want to go in that direction, we'd probably want [gittuf](https://gittuf.dev/), but I think even that system is too complex for most projects.

docs/spec/v1.1/source-att.md
laurentsimon: All I'm saying is that we're saying _only_ the source platform is allowed to generate these attestations, whereas there are cases where some claims (retentions) may not be possible and projects _may_ be struck at a level below what they could. So I suppose consumers will need to take in these low-level attestations without retention, and store the code themselves, and create a new attestation for it? (something along these lines)

I'm not trying to say this does not work, I'm trying to understand how this fits in the overall picture and how consumers will use it.

docs/spec/v1.1/source-att.md
laurentsimon: You're correct there are timing considerations, but if an consuming entity wants to do that verification, we should not prevent them from doing so. Otherwise they'll just be using scorecard or other mechanisms for all open-source projects. Tracking bugs seems good to me, thanks!

docs/spec/v1.1/source-att.md
kpk47: Ah, I misunderstood your comment. The source track is scoped to how the project maintains itself. If the project is unwilling or unable to retain its source code, then it's L1 until that situation changes. Although it may seem harsh, there is no alternative in the scope of the source track -- we can't tell developers they aren't allowed to delete their code. 

If someone else wants to retain the source code for their own use, that's a dependency concern. If someone wants to republish the source code with their own guarantees layered on top (e.g. retention), then they're effectively creating a new project and should generate its attestations from its source control platform.

laurentsimon: I think that answers my question, thanks

docs/spec/v1.1/source-att.md
```diff
@@ -0,0 +1,91 @@
...
+## Provenance generation
+
+The version control platform is completely trusted in SLSA. Attestations for
+Source L2+ SHOULD be generated by the trusted version control platform.
+Alternatively, other authorized parties MAY generate L2+ attestations using
+APIs provided by the source control platform.
+
+The Source track follows the same provenance generation requirements as the
+Build track. Please see
+[Provenance generation](requirements#provenance-generation) for details.
```
arewm, line 33: This seems odd to put here as the references for what needs to be met are not explicitly defined. The requirements for provenance generation are not tied to the levels of the source track directly and the requirements in the build track are additionally tailored to the generation of the build provenance (i.e. around completeness). 

If you want to cross-reference specific criteria in generating attestations, would it be possible to extract these to a separate and common location and then reference those requirements from both the source and build tracks?

docs/spec/v1.1/source-att.md
arewm: > Otherwise you enforce the policy "were all changes reviewed by someone who can approve changes now" as opposed to "were all changes reviewed by someone who could approve changes at the time."

> You're correct there are timing considerations, but if a consuming entity wants to do that verification, we should not prevent them from doing so.

These timing considerations would be a tool in being able to retract policy-based approvals. These retractions are hard because you don't want to invalidate _all_ prior approvals by an entity. There are many situations where an individual leaves organizations on good terms and all prior validations hold. There are likely fewer situations where an individual leaves and organization and at least _some number_ of their prior validations are retracted.

docs/spec/v1.1/source-att.md
```diff
@@ -0,0 +1,91 @@
...
+| `historyRetained` | FeatureHistory | L2+ | Whether this project retains its version history.
+| `codeReview` | FeatureHistory | L3+ | Whether this project requires code review.
+| `differentPersons` | FeatureHistory | L3+ | Whether this project has a different persons policy.
+
+### FeatureHistory
+
+| Field | Type | Description
+| --- | --- | ---
+| `enabled` | boolean | If true, then the feature was enabled at the time of attestation generation.
+| `since` | string | ISO 8601 representation of the timestamp at which the feature was last enabled.
```
arewm, line 53: How would an entity outside of the version control platform be able to know when the feature was last enabled? They would likely have to query the platform and the current value would easily not have been the value when the commit was made.

docs/spec/v1.1/source-reqs.md
```diff
@@ -0,0 +1,106 @@
...
```
arewm, line 15: Should we define who the infrastructure providing entity is to know how it is applied?

docs/spec/v1.1/source-reqs.md
```diff
@@ -0,0 +1,106 @@
...
+
+-   **[Change history]** There exists a record of the history of changes that went into the revision. Each change MUST contain:
+    -   The immutable reference to the new revision
+    -   The identities of the proposer, reviewers (if any), and merger (if different to the proposer)
+    -   Timestamps of the reviews (if any) and submission
+    -   The change description/justification
+    -   The content of the change
+    -   The parent revisions.
+
+-   **[Immutable reference]** There exists a way to indefinitely reference this particular, immutable revision. This is usually {project identifier + revision ID}. When the revision ID is a digest of the revision, as in git, nothing more is needed. When the revision ID is a number or otherwise not a digest, then the project server MUST guarantee the immutability of the reference.
```
arewm, line 27: Indefinitely feels like the wrong word. Would deterministically be better? This sentence (also above in levels.md) also effectively defining "immutable reference" as "immutable revision" -- is there a different term we can use instead of immutable revision? Or should we drop that?

docs/spec/v1.1/source-reqs.md
```diff
@@ -0,0 +1,106 @@
...
+-   A mechanism for modifying the canonical source through a **revision process**.
+
+The version control platform SHOULD additionally provide:
+
+-   A mechanism for assigning roles and/or permissions to identities.
+-   A mechanism for including code review in the revision process.
+-   Two-factor authentication for the account system (L2+ only).
+-   Audit logs for sensitive actions, such as modifying security controls.
+
+## Levels
```
arewm, line 43: Do we want to mirror the layout for the build levels here? For example, but not limited to the use of tables?

docs/spec/v1.1/source-reqs.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Level 0: No guarantees
+
+Summary: L0 represents the lack of a Source level. Projects are L0 by definition before adopting SLSA.
+
+Intended for: Experimental Projects, or other Projects early in their lifecycle.
+
+Requirements: N/A
+
+Benefits: N/A
```
arewm, line 53: Do we need this explicit mention of level 0? If it is beneficial, should we add it to the build track too?

docs/spec/v1.1/source-reqs.md
arewm, line 43: Are there other entities involved in achieving these levels other than the infrastructure providers? Can we split the levels out?

docs/spec/v1.1/terminology.md
```diff
@@ -64,6 +64,28 @@ chain.
 | Consumer | A party who uses software provided by a producer. The consumer may verify provenance for software they consume or delegate that responsibility to a separate verifier. | A developer who uses open source software distributions. A business that uses a point of sale system.
 | Infrastructure provider | A party who provides software or services to other roles. | A package registry's maintainers. A build platform's maintainers.
 
+### Source model
+
+The Source track is scoped to a single project that is controlled by some organization. That organization determines what Source level should apply to the project and administers technical controls to enforce that level.
+| Term | Description
+| --- | ---
+| Source | An identifiable set of text and binary files and associated metadata usually used as input for the build system (see SLSA Build Track).
```
arewm, line 72: No action. Just an observation... we mention binary files here explicitly. This has the potential to be relevant in any future dependency track that might require the dependencies to meet certain criteria (i.e. provenance).

docs/spec/v1.1/terminology.md
```diff
@@ -64,6 +64,28 @@ chain.
...
+| Revision | The canonical source at a given point in time as identified by the version control system. Git identifies revisions by digest, but other version control systems may use other identifiers.
```
arewm, line 73:
```suggestion
| Revision | The canonical source at a given point in time as identified by the version control system. As an example, git identifies revisions by digest.
```
Do you want to be more specific about what that digest is?

docs/spec/v1.1/terminology.md
```diff
@@ -64,6 +64,28 @@ chain.
...
+| Change | A set of modifications to one or more source files and associated metadata. Change metadata MUST include any information required to situate the change in relation to other changes (e.g. parent Revision).
```
arewm, line 74: Would a word like locate be better than situate? Also `Revision` capitalization is inconsistent with later use.

docs/spec/v1.1/terminology.md
```diff
@@ -64,6 +64,28 @@ chain.
...
+
+| Role | Description
+| --- | ---
+| Administrator | A human who can perform privileged operations on one or more projects. Privileged actions include, but are not limited to, modifying the change history and modifying project- or organization-wide security policies.
+| Merger | The person who applies a change to the source. This person may be the submitter or a different trusted person, depending on the version control platform.
+| Proposer | The human who proposes a particular change to the source.
+| Reviewer | The human who reviews a particular proposed change to the source.
+| Trusted person | A human who is authorized by the organization to propose and approve changes to the source.
+| Trusted robot | A machine that is authorized by the organization to propose and/or approve changes to the source.
+| Untrusted person | A human who has limited access to the project. They MAY be able to read the source. They MAY be able to propose or review changes to the source. They MAY NOT approve changes to the source or perform any privileged actions on the project.
```
arewm, line 87: Can we merge any of these with the earlier `Roles` section or are these explicitly unique to source?

docs/spec/v1.1/terminology.md
```diff
@@ -64,6 +64,28 @@ chain.
 | Consumer | A party who uses software provided by a producer. The consumer may verify provenance for software they consume or delegate that responsibility to a separate verifier. | A developer who uses open source software distributions. A business that uses a point of sale system.
 | Infrastructure provider | A party who provides software or services to other roles. | A package registry's maintainers. A build platform's maintainers.
 
+### Source model
```
arewm, line 67: The build model introduces the space. Do you want to do something similar for the source model?

arewm, line 67: What is the logic behind putting the source model here? Just that the source track would come before the build track?

Should we create a subheading for models and then alphabetize the various models within it?

arewm (changes requested):
Thanks for this extensive draft!

docs/spec/v1.1/source-att.md
```diff
@@ -0,0 +1,91 @@
...
+and generated when creating a complete copy of that project's source. For
+example, a project may produce a source attestation when exporting its source
+for distribution or when a build pipeline fetches the source.
+
+## Provenance generation
+
+The version control platform is completely trusted in SLSA. Attestations for
+Source L2+ SHOULD be generated by the trusted version control platform.
+Alternatively, other authorized parties MAY generate L2+ attestations using
+APIs provided by the source control platform.
```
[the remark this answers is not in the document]
laurentsimon: > Who is an authorized party? Is it acceptable for the build platform to generate the source attestation?

An authorized party is in the eye for the receiver, ie it's whomever the receiver trusts to generate it. For example, an organization like Google could create the attestation using underlying GitHub APIs, and that attestation would be trusted internally at Google. Other orgs might not trust it and might generate their own.

So yes a builder may generate it you shift left.

> Is this only allowable

I think it's allowable. If the trusted platform generates it and you trust it, you need not re-generate it. If the version control does not support the generation but supports API, then another entity would generate it. If you don't trust the platform to generate it, you may also re-generate it (although that only makes sense if we consider "fact" attestations https://github.com/slsa-framework/slsa/pull/1037#discussion_r1526020307, else you have to trust the provider for both identity and enforcement, in which case re-generating it does not seem to provide additional benefits).

docs/spec/v1.1/source-reqs.md
```diff
@@ -0,0 +1,104 @@
...
+
+-   An account system or some other means of identifying persons. This account system MUST support two-factor authentication or an equivalent.
+-   A mechanism for modifying the canonical source through a **revision process**.
+
+The version control platform SHOULD additionally provide:
+
+-   A mechanism for assigning roles and/or permissions to identities.
+-   A mechanism for including code review in the revision process.
+-   Mechanisms for **strongly authenticating** identities, such as 2FA.
+-   Audit logs
```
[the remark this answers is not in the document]
ramonpetgrave64: Yes, maybe `Audit logs against the changes of in the Features`

docs/spec/v1.1/source-reqs.md
laurentsimon: > @laurentsimon Again, this is about where you draw the box. I'd rather just draw the box around all the trusted source components and call it a day. I would need to understand the purported benefit of further detail to change my mind.

My only suggestion would be to clarify the description of `Version Control System` to accommodate 3P review systems. That's what prompted my comment. The current description only mentioned self-contained platforms like Github and Gitlab.

docs/spec/v1.1/source-att.md
```diff
@@ -0,0 +1,91 @@
...
+
+A SLSA Source attestation asserts that a project follows certain practices
+pertaining to its source code. Attestations are scoped to a project as a whole
+and generated when creating a complete copy of that project's source. For
+example, a project may produce a source attestation when exporting its source
+for distribution or when a build pipeline fetches the source.
+
+## Provenance generation
+
+The version control platform is completely trusted in SLSA. Attestations for
```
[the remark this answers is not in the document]
kpk47: I mean that the source platform is central to the source track in the same way the build platform is central to the build track. We trust the platform to record the source, proposed changes, and change history faithfully. 

Does this sentence make more sense to you now that I've renamed the platform to `source control platform`?

docs/spec/v1.1/source-att.md
kpk47: My intention isn't to track per-commit. Rather, we choose a point in the change history and decide that everything after that point should meet the desired SLSA level. In practice, I expect this to mean that projects adopt a SLSA level and then never regress below it, but I don't think that's realistic in practice. The best solution I could come up with is recording how long the different protections have been in effect, which lets consumers write policies that detect level regressions.

docs/spec/v1.1/source-reqs.md
kpk47: Changing "indefinitely" to "deterministic" SGTM, but I disagree on changing "immutable." What we want is to prevent changes to revisions. Git prevents those changes by using the commit digest as the commit ID. The point of this language is to disallow any version control system that has mutable revisions. (I'm not aware of any, but I think writing defensively makes sense here.)

docs/spec/v1.1/source-reqs.md
kpk47: Done.

The change is in terminology.md.

docs/spec/v1.1/source-reqs.md
kpk47: I'm not convinced we need them. The tables in the Build track are for displaying which requirements apply to each level, but we only have one requirement here tied to a level.

kpk47: I think we've got the right split here. The section above is what features the infrastructure provider must support, and this section is which of those features project owners/admins must enable at each level. What other entity should be involved?

docs/spec/v1.1/source-reqs.md
kpk47: Done.

L0 is already in the table in levels.md, so no need to repeat it here.

docs/spec/v1.1/terminology.md
kpk47: Alphabetizing SGTM.

We might want to reorganize this page before finalizing the next release. It's pretty hard to read.

docs/spec/v1.1/terminology.md
kpk47: Done

I clarified that we mean tree hash, not commit hash. I think I noted that fact somewhere else in the PR, but I can't find it at the moment.

docs/spec/v1.1/terminology.md
kpk47: I think situate is the correct word. In my mind, locating doesn't include context but situate does. We want to know (at least) the parent revision, so we want some added context.

docs/spec/v1.1/terminology.md
kpk47: I think this page needs a bit reformatting, or maybe we break it up into several pages. For now, I moved the existing Roles section to a subsection of the build model and kept these roles as a subsection of the source model.

kpk47 (commented):
I still have a few of the meatier comments to address, but I want to check that you think I'm heading in the right direction.

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,128 @@ All of [Build L2], plus:
...
+### Source track
+
+| Source Levels | Requirements | Focus
+| ----------- | ------------ | -----
+| [Source L0] | (none) | (n/a)
+| [Source L1] | Version control | Basic project hygiene
+| [Source L2] | Strong authentication, Retained history, Verified timestamps | Change attribution
+| [Source L3] | Code review, Different persons | Change authorization
+
+### Source L0: No guarantees
```
arewm, line 241: Change Source `LX` and Build `LX` should be an additional subheading lower `####`

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,130 @@ All of [Build L2], plus:
...
+
+<dt> Requirements <dd>
+
+**[Strong authentication]** User accounts that can modify the source or the project's configuration must use two-factor authentication or its equivalent.
+
+**[Verified timestamps]** Each entry in the change history must contain at least one timestamp that is determined by the source control platform and cannot be modified by clients. It MUST be clear in the change history which timestamps are determined by the source control platform.
+
+**[Retained history]** The change history MUST be preserved as long as the source is hosted on the source control system. The source MAY migrate to another source control system, but the organization MUST retain the change history if possible. It MUST NOT be possible for persons to delete or modify the change history, even with multi-party approval, except by trusted platform admins following an established deletion policy. Possible legitimate reasons to modify the change history include deleting information that must not be stored in the change history, such as secret material, personally identifiable information, or illegal materials.
```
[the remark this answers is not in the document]
arewm: That's fair. I don't have any suggestions on how to make this better off hand...

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,130 @@ All of [Build L2], plus:
...
+-   **[Context-specific approvals]** Approvals are for a specific context, such as a repo + target branch + revision in git. Moving fully reviewed content from one context to another still requires review. (Exact definition of "context" depends on the project, and this does not preclude well-understood automatic or reviewless merges, such as cutting a release branch.) Git example: If a fully reviewed commit in one repo is merged into a different repo, or a commit in one branch is merged into a different branch, then the merge still requires review.
+-   **[Atomic change sets]** Changes are recorded in the change history as a single revision that consists of the net delta between the proposed revision and the parent revision. In the case of a nonlinear version control system, where a revision can have more than one parent, the diff must be against the "first common parent" between the parents. In other words, when a feature branch is merged back into the main branch, only the merge itself is in scope.
```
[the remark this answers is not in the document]
arewm: Some communities are very adamant about _never_ squashing commits, instead using commits to describe logical bits of change and then just merging them all in to preserve the blame. While this might introduce some risk (i.e. by creating reachable malicious commits), that risk can be mitigated if all commits are reviewed during the process. 

While squash+merge might be a best practice from a security perspective, I am not sure whether we can be overly prescriptive of it as a requirement for this track/level.

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,130 @@ All of [Build L2], plus:
     source and build process.
 
 </dl>
-</section>
+
+## Source Track
```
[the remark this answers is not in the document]
arewm: New discussion opened

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,130 @@ All of [Build L2], plus:
...
+</dl>
+
+### Source L1: Version controlled
+
+<dl class="as-table">
+<dt> Summary <dd>
+The project source is stored and managed through a modern version control system.
+
+<dt> Intended for <dd>
+Organizations that are unwilling or unable to host their source on a version control platform. If possible, skip to Level 2.
```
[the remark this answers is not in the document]
arewm: Ah, thanks. I should have started the review by looking at the terminology. As long as we are being clear and consistent, it should be fine for the first draft.

docs/spec/v1.1/source-att.md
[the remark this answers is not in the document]
arewm: For the build track, the platform is trusted to the extend that it accurately represents the content represented in the attestation. The platform itself isn't _necessarily_ trusted, but its identity is associated with the attestation and therefore the trust can be assessed by a consumer. 

Maybe this doesn't translate as well for the source track because there isn't a trusted control plane in the same way that there is one for build. Maybe this first sentence just isn't needed as it is assumed that the generator of the attestation will be trusted?

docs/spec/v1.1/source-att.md
```diff
@@ -0,0 +1,91 @@
+---
+title: Source Attestations
+description: Description of the SLSA source attestation specification for verifying how source code is managed.
+layout: standard
+---
+This document describes the project metadata necessary to represent which SLSA
+Source level a project satisfies, if any.
```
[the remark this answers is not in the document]
arewm: It was a challenge to not have an attestation format at all. Instead, just to make the source track about measurable properties. Whomever wants to measure these properties can report to their validity as desired (i.e. a VSA, API, etc.)

docs/spec/v1.1/source-att.md
arewm: Sure. This challenges the ability for an entity that isn't the source control platform to be able to generate the data required for an attestation. Any other entity wouldn't be continuously watching the configurations to be able to know if the features have been switched off. If the systems only check for the presence at the time of a build, for example, then there is no way for that system to know that the setting wasn't disabled temporarily, a commit or two deleted, and then reenabled before the build.

docs/spec/v1.1/source-reqs.md
arewm: I don't have a better word/phrase/definition. I was just always told to not use words in their own definition. I don't have a better phrasing for now, so we can just leave the change at deterministically.

arewm (commented):
This feels like it is going in the right direction. The biggest uncertainty to me is whether we want to introduce an additional attestation for source.

docs/spec/v1.1/levels.md
kpk47: Done

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
kpk47: I wasn't aware that there are communities that feel strongly about merge strategies. I'm not sure what's best here, so I've opened https://github.com/slsa-framework/slsa/issues/1040.

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,130 @@ All of [Build L2], plus:
...
+-   **[Atomic change sets]** Changes are recorded in the change history as a single revision that consists of the net delta between the proposed revision and the parent revision. In the case of a nonlinear version control system, where a revision can have more than one parent, the diff must be against the "first common parent" between the parents. In other words, when a feature branch is merged back into the main branch, only the merge itself is in scope.
+
+Trusted robots MAY be exempted from the code review process. It is RECOMMENDED that trusted robots so exempted be run only software built at Build L3+ from sources that meet Source L3.
+
+**[Different persons]** The organization strives to ensure that no two user accounts correspond to the same person. Should the organization discover that it issued multiple accounts to the same person, it MUST act to rectify the situation. For example, it might revoke project privileges for all but one of the accounts and perform retroactive code reviews on any changes where that person's accounts are the author and/or code reviewer(s).
```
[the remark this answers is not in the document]
kpk47: I've added a definition to terminology.md based on the definition from the Build track (which unfortunately is on requirements.md rather than the terminology page).
```
A package's <dfn>producer</dfn> is the organization that owns and releases the    
software. It might be an open-source project, a company, a team within a          
company, or even an individual. 
````

File: docs/spec/v1.1/source-att.md (hunk 21 above)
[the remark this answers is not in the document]
Comment (kpk47) [reply docs/spec/v1.1/source-att.md]: Good idea. I dropped the first sentence.

File: docs/spec/v1.1/source-attestation.md [hunk 29]
```diff
@@ -0,0 +1,90 @@
...
+
+The Source track follows the same provenance generation requirements as the
+Build track. Please see
+[Provenance generation](requirements#provenance-generation) for details.
+
+## Schema
+
+| Field | Type | Required at Level | Description
+| --- | --- | --- | ---
+| `projectID` | string | L1+ | Immutable project ID. For example, the SLSA Specification's project ID is `www.github.com/slsa-framework/slsa`.
```
adityasaky, line 38: Repositories and usernames / orgs can be renamed, so I don't know if immutable works here.

docs/spec/v1.1/source-attestation.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Source Attestations
```
adityasaky, line 2: I wonder if a generalized VSA (with some tweaks as to what's being "verified") or SCAI (more directly as a series of arbitrary claims) would work here? From my read, this is really just making a series of controls claims about the source control platform. I'd also clarify the name / description to avoid confusion further down the line for attestations that may be more directly source pertinent.

docs/spec/v1.1/source-attestation.md
```diff
@@ -0,0 +1,90 @@
...
+The Source track follows the same provenance generation requirements as the
+Build track. Please see
+[Provenance generation](requirements#provenance-generation) for details.
+
+## Schema
+
+| Field | Type | Required at Level | Description
+| --- | --- | --- | ---
+| `projectID` | string | L1+ | Immutable project ID. For example, the SLSA Specification's project ID is `www.github.com/slsa-framework/slsa`.
+| `revisionID` | string | L1+ | Immutable revision ID. In git terminology, this is the commit hash.
```
adityasaky, line 39: Is this or digest below going to match the subject of the attestation?

adityasaky: Also, could this be a tag object ID or must it always be resolved to a commit?

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,128 @@ All of [Build L2], plus:
...
+<dt> Intended for <dd>
+Organizations that are unwilling or unable to host their source on a source control platform. If possible, skip to Level 2.
+
+<dt> Requirements <dd>
+
+**[Version controlled]** Every change to the source is tracked in a version control system that meets the following requirements
+
+-   **[Change history]** There exists a record of the history of changes that went into the revision. Each change MUST contain:
+    -   The immutable reference to the new revision
+    -   The identities of the submitter, reviewers (if any), and merger (if different to the uploader)
```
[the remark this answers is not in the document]
laurentsimon: I don't think so. In the case of GitHub I would expect it to be a GitHub handle

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,128 @@ All of [Build L2], plus:
...
+-   **[Atomic change sets]** Changes are recorded in the change history as a single revision that consists of the net delta between the proposed revision and the parent revision. In the case of a nonlinear version control system, where a revision can have more than one parent, the diff must be against the "first common parent" between the parents. In other words, when a feature branch is merged back into the main branch, only the merge itself is in scope.
+
+Trusted robots MAY be exempted from the code review process. It is RECOMMENDED that trusted robots so exempted be run only software built at Build L3+ from sources that meet Source L3.
```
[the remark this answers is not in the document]
laurentsimon: I think it means a trusted robot must be running software that is at least SLSA L3 for both build and source track. I suppose we may also need to say that it's necessary but not sufficient?

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
adityasaky: At this level, only a VCS is required without the accompanying host platform, which is why I'm reading this to mean raw commits...I suppose my question is more about email in the commit metadata vs signature.

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,128 @@ All of [Build L2], plus:
...
+Organizations that are unwilling or unable to host their source on a source control platform. If possible, skip to Level 2.
+
+<dt> Requirements <dd>
+
+**[Version controlled]** Every change to the source is tracked in a version control system that meets the following requirements
+
+-   **[Change history]** There exists a record of the history of changes that went into the revision. Each change MUST contain:
+    -   The immutable reference to the new revision
+    -   The identities of the submitter, reviewers (if any), and merger (if different to the uploader)
+    -   Timestamp of the reviews (if any)
```
adityasaky, line 274: With only a VCS and no review / host platform, what does a review look like? should this be taken out at this level?

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
kpk47: I don't think we need to be prescriptive about how to represent identities at L1 -- this level isn't meant to be trustworthy. However, if I were presented with a commit from a L1 project where the email in the commit metadata conflicts with the signature, I'd expect the signature to take priority.

docs/spec/v1.1/levels.md
kpk47: I had considered removing this line until L2, but I couldn't figure out how to do so without creating ambiguity. (E.g. would L2 have a different metadata format? Or maybe it would change a field from optional to required?) It seemed clearer to put the full metadata definition in L1 and note that reviews may not be present.

To your other question: code reviews can happen outside review/host platforms. The review could happen over a mailing list, and the metadata could be a `Reviewed by` line in the commit. Remember, metadata at L1 is not trustworthy.

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,128 @@ All of [Build L2], plus:
...
+    -   Timestamp of applying the change
+    -   The change description/justification
+    -   The content of the change
+    -   The parent revisions.
+
+-   **[Immutable reference]** There exists a way to reference this particular, immutable revision. This is usually {project identifier + revision ID}. When the revision ID is a digest of the revision, as in git, nothing more is needed. When the revision ID is a number or otherwise not a digest, then the project server MUST guarantee the immutability of the reference.
+
+<dt> Notes <dd>
+Most popular version control systems meet these requirement, such as git, Subversion, Mercurial, and Perforce.
```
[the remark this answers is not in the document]
kpk47: Done

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
kpk47: Sorry, the "be" is an extra word. Does it make more sense now?

I also added a note that Build L3+ and Source L3 are not sufficient.

docs/spec/v1.1/source-attestation.md
kpk47: I think one of the great mistakes in SLSA v1.0 was providing an attestation format without specifying what information we're trying to convey. I don't intend to repeat that mistake, so I've intentionally avoided specifying a format here. That specification can come later.

Kindly refrain from calling my work not pertinent, and I'll refrain from calling you impertinent. I've named the attestation to match the track. That's the shorthand we've been using in meetings, and it seems like a logical enough choice to me. Should SLSA need to free up "source" in its namespace in the future, we can deal with it then.

docs/spec/v1.1/source-attestation.md
kpk47: Done

docs/spec/v1.1/source-attestation.md
kpk47: > Is this or digest below going to match the subject of the attestation?

The digest below.

> Also, could this be a tag object ID or must it always be resolved to a commit?

I've been avoiding tying anything to tags because they're mutable. Do we gain something by using tags here?

docs/spec/v1.1/source-attestation.md
adityasaky: Apologies! I did not mean any disrespect or insult with the word "pertinent". I think your work on this has been tremendous and I appreciate it!

To clarify, I was wondering if we wanted to distinguish between attestations for source platform specific claims and those that may apply (I was thinking "pertain" earlier) more specifically to the source itself.

docs/spec/v1.1/source-attestation.md
laurentsimon: note that it's possible to provide an immutable ID: GitHub does support immutable user and repository IDs.

docs/spec/v1.1/source-attestation.md
kpk47: No worries at all, and apologies for my terse response.

> To clarify, I was wondering if we wanted to distinguish between attestations for source platform specific claims and those that may apply (I was thinking "pertain" earlier) more specifically to the source itself. I'm fine with clarifying this later if the need arises.

Thank you for clarifying. I see what you mean now -- this track is more about development practices than it is about the source itself, and "source" is a rather generic name for such a specific track. I'd like the track and attestation names to match because the alternative may be confusing, but I'm not opposed to renaming the track. I opened #1041 to discuss.

adityasaky: > I think one of the great mistakes in SLSA v1.0 was providing an attestation format without specifying what information we're trying to convey. I don't intend to repeat that mistake, so I've intentionally avoided specifying a format here. That specification can come later.

Understood. To me, it does read a lot like an attestation predicate proposal (that you could, for example, create a proto definition for in in-toto/attestation), especially next to the build provenance doc. Should this be tweaked in the meantime until we decide how this information is specifically communicated in an attestation?

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
adityasaky: > Sorry, the "be" is an extra word. Does it make more sense now?

Yep, thanks!

docs/spec/v1.1/source-attestation.md
adityasaky: > The digest below.

got it.

FWIW, I could also see both the revision ID and the tree ID listed in the subject as digests (the algorithm would be `gitCommit` and `gitTree` rather than `sha1` so there wouldn't be a collision). I'm mentioning it now but it can likely wait until the exact format for the predicate is decided, per the other thread.

> I've been avoiding tying anything to tags because they're mutable. Do we gain something by using tags here?

A tag object ID (ie its SHA-1 hash) is immutable while a tag reference is not. I'm primarily asking for cases where a source attestation is created for a release identified by a tag object. If we expect the field to be a commit ID, we should maybe call out that tag objects for a release must be resolved to the commit they point to (also, there's an edge case here because a tag object can point to any git object type, not just commits, so maybe that should be called out too).

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
arewm: +1

There are multiple use cases for identities. 
* You may be interested in associating the identity with an individual that you can reach out to for further discussion, etc. 
* It may not _necessarily_ about the identity itself, but more about adding the entropy to uniquely associate commits based on _some_ consistent identifier so that you can more easily apply policy rules to find other instances of the same committer.

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,128 @@ All of [Build L2], plus:
...
+-   **[Atomic change sets]** Changes are recorded in the change history as a single revision that consists of the net delta between the proposed revision and the parent revision. In the case of a nonlinear version control system, where a revision can have more than one parent, the diff must be against the "first common parent" between the parents. In other words, when a feature branch is merged back into the main branch, only the merge itself is in scope.
+
+Trusted robots MAY be exempted from the code review process. It is RECOMMENDED that trusted robots so exempted run only software built at Build L3+ from sources that meet Source L3. Note: Meeting Build L3+ and Source L3 are not necessarily sufficient evidence that software is safe to run as a trusted robot.
```
arewm, line 334: This first sentence makes sense, but I am not sure where the recommendation is coming from. It seems unrelated to the track and more related to some applications of SLSA.

docs/spec/v1.1/source-attestation.md
arewm: If you are focusing on the specific information that you are trying to convey, do we need to call this an attestation?

This feels like a similar point that I was trying to make in https://github.com/slsa-framework/slsa/pull/1037#discussion_r1532643046 -- Instead of focusing on a format for track attestations, we should instead focus on the ability to track adherence to the levels based on specific measurable properties. 

I was interpreting this section as a proposal for a formal attestation predicate.

docs/spec/v1.1/source-att.md
kpk47: This conversation is now tracked in https://github.com/slsa-framework/slsa/issues/1042, so I'm resolving this thread.

docs/spec/v1.1/source-att.md
kpk47: I'm coming around to the idea that the source track doesn't need its own attestation, so I'm removing this line for now.

kpk47:
@adityasaky @arewm I think we should go back to the drawing board with the source attestation/evidence, so I've removed it from this PR. Let's get the level spec to a place we're happy merging, and then I'll open a new PR where we can continue working on the attestation/evidence requirements.

adityasaky:
Sounds good!

joshuagl (commented):
Apologies for being late to review this. It's fantastic to see the beginnings of a Source track emerge. Thank you for the continued effort on this @kpk47.
This feels equivalent to what we had in v0.1, only compressed into 3 levels instead of 4 and with more affordances for changes from robots. Does that assessment feel fair/accurate to you?

Most of my comments are related to presentation, the requirements feeling so close to the v0.1 version is a great baseline.

The formatting and structure of the Levels changes don’t match the existing Build track levels. 
Build levels are broken down into producer and platform requirements (see change from benefits to focus in https://github.com/slsa-framework/slsa/commit/cc812c33a03ec98f8bdb9b22304a38b130e9f23e) and are a descriptive overview requirements, whereas the newly added source requirements are detailed and specific enough that they run the risk of seeming authoritative and contradictory to the track text.

The formatting and structure of the Source requirements doesn’t match the “Producing artefacts” document, where the requirements are in a table with check marks for which level they apply to (and further, there are tables per requirement category: provenance generation and isolation, I’m not sure this split makes sense for the source track).

Note: we should update future-directions to remove the source track and/or update it to mention what could come next. Threats and mitigations should be updated too.

Final thought, is adding a new level appropriate for a minor release, or should we consider a major release for this?

lehors:
> Final thought, is adding a new level appropriate for a minor release, or should we consider a major release for this?

We've talked about this on Monday's call actually. We agreed to punt on this issue for now and wait until we decide to publish the next version of the spec. We can then discuss what change to the version number is the most appropriate based on what ends up being in the spec. For now we can still work on "1.1" without it being binding.

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,128 @@ All of [Build L2], plus:
...
+
+<dt> Benefits <dd>
+Version control is a software development best practice, and discussing its benefits is beyond the scope of this page.
+</dl>
+
+#### Source L2: Verified history
+
+<dl class="as-table">
+<dt> Summary <dd>
+The project is stored and managed through a source control platform that ensures the change history's integrity.
```
MarkLodato, line 293: I suggest describing things at a higher level of abstraction and explaining the value that this level gives. See Build track as examples. Maybe something like:

```suggestion
The source code and its change history metadata are retained and authenticated to allow trustworthy auditing and analysis of the source code.
```

docs/spec/v1.1/levels.md
MarkLodato: This topic (trusted robots) probably warrants further discussion and more guidance in the spec. Maybe just add a TODO to flesh it out? I doubt we'll get it right on the first try.

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
kpk47: I created an `Outstanding TODO` section at the top and added an item there about identities.

docs/spec/v1.1/levels.md
kpk47: Added a TODO

docs/_data/nav/v1.1.yml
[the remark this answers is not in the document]
kpk47: That makes sense. I've consolidated all the new content into source-requirements.md and reverted all the other files to HEAD.

docs/spec/v1.1/terminology.md
```diff
@@ -295,6 +297,31 @@ Notes:
     *checksum database* guarantees that all clients receive the same artifact
     for a given URL.
 
+### Source model
+
+The Source track is scoped to a single project that is controlled by some organization. That organization determines what Source level should apply to the project and administers technical controls to enforce that level.
+| Term | Description
```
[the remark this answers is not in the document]
kpk47: I reordered the terms conceptually.

docs/spec/v1.1/terminology.md
```diff
@@ -295,6 +297,31 @@ Notes:
...
+| Change History | A record of the history of changes that went into the revision.
+| Organization | A collection of people who collectively create the Source. Examples of organizations include an open-source projects, a company, or a team within a company.
+| Revision | The canonical source at a given point in time as identified by the version control system. As an example, you can identify a git revision by its tree hash.
+| Source | An identifiable set of text and binary files and associated metadata usually used as input for the build system (see SLSA Build Track).
+| Source Control Platform | A service or suite of services for hosting version controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
+| Version Control System | Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
+
+#### Source Roles
+
+| Role | Description
```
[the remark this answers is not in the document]
kpk47: I did my best to reorder the roles conceptually, but it's a little tricky since none of the terms appear in other definitions. I'm open to other orderings.

MarkLodato:
I'm going to take this PR over ([context](https://openssf.slack.com/archives/C03NUSAPKC6/p1714401798123519)). My goal is to get it submitted so that we have something to iterate on. It's not going to be perfect. The intent is that folks can open PRs to address specific points, rather than having one mega-PR for everything.

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+| Role | Description
+| --- | ---
+| Administrator | A human who can perform privileged operations on one or more projects. Privileged actions include, but are not limited to, modifying the change history and modifying project- or organization-wide security policies.
+| Trusted person | A human who is authorized by the organization to propose and approve changes to the source.
+| Trusted robot | Automation with an authentic identity that is authorized by the organization to propose and/or approve changes to the source.
+| Untrusted person | A human who has limited access to the project. They MAY be able to read the source. They MAY be able to propose or review changes to the source. They MAY NOT approve changes to the source or perform any privileged actions on the project.
+| Proposer | The role that proposes a particular change to the source.
+| Reviewer | The role that reviews a particular proposed change to the source.
+| Approver | The role that approves a particular change to the source.
+| Merger | The role that applies a change to the source. This person may be the proposer or a different trusted person, depending on the version control platform.
```
zachariahcox, line 46: There is an actor who "clicked the merge button", but this may or may not be the identity of the actor that authenticated to push the changes. 

For example, I may click the merge button to add a PR to a GitHub Merge Queue, but the Merge Queue's service principal is the identity who eventually pushes the ref update of the main branch.
This is because merge-queue-like systems contain the contents of multiple contributors and change sets.

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Source Platform Requirements
+
+The version control system MUST provide at least:
```
zachariahcox, line 50: This section may be mixing concerns between the VCS and the source control platform (SCP). 

Git commit metadata is forgeable from the perspective of the SCP. 
The SCP typically provides an identity / authz layer which is separate from the git content it manages, and it maintains its own timestamps for when activities happened from its perspective. EG: a commit may claim it was authored last year, but all the SCP knows for sure is that it was uploaded last week. 

Authentic contributions will typically be built in terms of the SCP's identity and timestamp models, not the VCS's.

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+-   Two-factor authentication for the account system (L2+ only).
+-   Audit logs for sensitive actions, such as modifying security controls.
+
+## Levels
+
+### Level 1: Version controlled
+
+Summary: The project source is stored and managed through a modern version control system.
+
+Intended for: Organizations that are unwilling or unable to host their source on a source control platform. If possible, skip to Level 2.
```
zachariahcox, line 82: > host their source on a source control platform

Would this include self-hosted versions of SCPs?

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+Summary: The source code and its change history metadata are retained and authenticated to allow trustworthy auditing and analysis of the source code.
+
+Intended for: Organizations that are unwilling or unable to incorporate code review into their software development practices.
+
+Requirements:
+**[Strong authentication]** User accounts that can modify the source or the project's configuration must use multi-factor authentication or its equivalent.
+
+**[Verified timestamps]** Each entry in the change history must contain at least one timestamp that is determined by the source control platform and cannot be modified by clients. It MUST be clear in the change history which timestamps are determined by the source control platform.
+
+**[Retained history]** The change history MUST be preserved as long as the source is hosted on the source control system. The source MAY migrate to another source control system, but the organization MUST retain the change history if possible. It MUST NOT be possible for persons to delete or modify the change history, even with multi-party approval, except by trusted platform admins following an established deletion policy.
```
zachariahcox, line 101: verified timestamps are provided by the SCP, so when I migrate the source to a new SCP, I think this means that the org must keep the original repo around to keep the "true" timestamps. 
Or else, we may be saying that SCPs must provide a way to export the verified timestamps somehow. We should consider making that clearer here, and I'm not sure which way is better! I'm not sure SCPs would import timestamps like that.

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Level 3: Changes are authorized
+
+Summary: All changes to the source are approved by two trusted persons prior to submission.
+
+Intended for: Enterprise projects and mature open source projects.
+
+Requirements:
+
+**[Code review]** All changes to the source are approved by two trusted persons prior to submission. User accounts that can perform code reviews MUST use two-factor authentication or its equivalent.
```
zachariahcox, line 113: Should a robot approval count or must it always be human?

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+The following combinations of trusted persons are acceptable:
+
+-   Proposer and reviewer are two different trusted persons.
+-   Two different reviewers are trusted persons.
+
+The code review system must meet the following requirements:
+
+-   **[Informed review]** The reviewer is able and encouraged to make an informed decision about what they're approving. The reviewer MUST be presented with a full, meaningful content diff between the proposed revision and the previously reviewed revision. For example, it is not sufficient to just indicate that a file changed without showing the contents.
+-   **[Context-specific approvals]** Approvals are for a specific context, such as a repo + target branch + revision in git. Moving fully reviewed content from one context to another still requires review, except for well-understood automatic processes. For example, you do not need to review each change to cut a release branch, but you do need review when backporting changes from the main branch to an existing release branch.
```
zachariahcox, line 122: > you do not need to review each change to cut a release branch

This seems really tricky. 
If I create a new release ref that points to the tip of main, I must ensure that the tip of main complies with all requirements that govern the release ref. 
It's normal for the requirements to be different between those contexts. 

> Moving fully reviewed content from one context to another still requires review, except for well-understood automatic processes

This makes sense to me. I'm not sure the example expresses this though.

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+-   **[Atomic change sets]** Changes are recorded in the change history as a single revision that consists of the net delta between the proposed revision and the parent revision. In the case of a nonlinear version control system, where a revision can have more than one parent, the diff must be against the "first common parent" between the parents. In other words, when a feature branch is merged back into the main branch, only the merge itself is in scope.
+
+Trusted robots MAY be exempted from the code review process. It is RECOMMENDED that trusted robots so exempted be run only software built at Build L3+ from sources that meet Source L3.
```
zachariahcox, line 125:
```suggestion
Trusted robots MAY be exempted from the code review process. It is RECOMMENDED that an exempted trusted robot artifact be built at Build L3+ from sources that meet Source L3.
```

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+
+The Source track is scoped to a single project that is controlled by some organization. That organization determines what Source level should apply to the project and administers technical controls to enforce that level.
+
+| Term | Description
+| --- | ---
+| Source | An identifiable set of text and binary files and associated metadata usually used as input for the build system (see SLSA Build Track).
+| Organization | A collection of people who collectively create the Source. Examples of organizations include an open-source projects, a company, or a team within a company.
+| Change | A set of modifications to one or more source files and associated metadata. Change metadata MUST include any information required to situate the change in relation to othe| Version Control System | Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
+| Revision | The canonical source at a given point in time as identified by the version control system. As an example, you can identify a git revision by its tree hash.
+r changes (e.g. parent revision).
```
zachariahcox, line 31:
```suggestion
```

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Source model
+
+The Source track is scoped to a single project that is controlled by some organization. That organization determines what Source level should apply to the project and administers technical controls to enforce that level.
+
+| Term | Description
+| --- | ---
+| Source | An identifiable set of text and binary files and associated metadata usually used as input for the build system (see SLSA Build Track).
+| Organization | A collection of people who collectively create the Source. Examples of organizations include an open-source projects, a company, or a team within a company.
+| Change | A set of modifications to one or more source files and associated metadata. Change metadata MUST include any information required to situate the change in relation to othe| Version Control System | Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
```
zachariahcox, line 29:
```suggestion
| Change | A set of modifications to one or more source files and associated metadata. Change metadata MUST include any information required to situate the change in relation to other changes (e.g. parent revision).
| Version Control System | Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
```

zachariahcox (commented):
@chkimes and I collaborated on this review. 
Thanks for putting this together!

docs/spec/v1.1/source-requirements.md
Nikokrock: From my point, it is not always a human (some sources might not be easily read by a human, for example). Or in case of merges, depending on the way you handle your merge processes, a robot might be able to review if the change can be approved or not.

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
...
+| Revision | The canonical source at a given point in time as identified by the version control system. As an example, you can identify a git revision by its tree hash.
```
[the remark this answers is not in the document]
Nikokrock: I agree git does not have a verifiable timestamp but SCP should have one ? (so maybe the example is wrong). Time is an important point in case you detect an attack such as XZ and need to go back to a "safe" point.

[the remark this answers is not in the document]
zachariahcox:
```suggestion
| Revision | A specific state of a repository as identified by the version control system. As an example, you can identify revision of a git repository by its tree id. 
```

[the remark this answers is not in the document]
zachariahcox: @Nikokrock it's true that the SCP should have a timestamp for when a revision was uploaded, but 1) not all SCPs provide that feature, and 2) (to me) that information seems separate from what defines the revision itself. I think I can upload the same revision of a repository to multiple SCPs. They would all have different "receive" timestamps but should be considered "the same code" imo.

docs/spec/v1.1/source-requirements.md
```diff
@@ -0,0 +1,129 @@
+# SLSA Source Track
+
+## Objective
+
+The SLSA Source Track mitigates [Threat A ("Submit unauthorized change")](/spec/v1.0/threats#a-submit-unauthorized-change), scoped to a code repository and the organization that owns that repository. Concretely: an attacker must compromise the accounts of two organization members to publish code in a Source Level 3-conformant repository, and the evidence of those unauthorized changes cannot be destroyed without further attacks.
```
zachariahcox, line 5: (The "Threat A" url references a deleted ref.)

It seems to me that the source track (similar to the build track) should start with generating provenance at the lower levels, then move to implementation of specific policies at higher levels. 
The revision of a consumed repository is a kind of artifact like any other software dependency and as such should come with all its paperwork. The "build process" of a repository revision is the pull request, and the result of a completed pull request should be a new revision accompanied by provenance claims. 

This example appears to be covering the requirements and benefits of a specific policy, similar to the ["Build Level 1" dedicated infrastructure requirement](https://slsa.dev/spec/v1.0/levels#build-l1). 
> Concretely: an attacker must compromise the accounts of two organization members to publish code in a Source Level 3-conformant repository, and the evidence of those unauthorized changes cannot be destroyed without further attacks.


At a minimum, consumers of a repository revision will want to know the following provenance claims: 
1. Who contributed these changes (according to the SCP)? 
2. When did they do so (according to the SCP)? 
3. What policy compliance claims are associated with this revision (according to the SCP)? 


Based on this, it seems to me that a focused adaptation of the excellent ["build track" summary](https://slsa.dev/spec/v1.0/levels#build-track) would work well here. EG:


The SLSA **SOURCE** track describes increasing levels of trustworthiness and completeness in a **repository revision's** provenance. Provenance describes what **Source Control Provider produced the revision**, what process was used, and who the contributors were. The lowest level only requires the provenance to exist, while higher levels provide increasing protection against tampering of the **version control system, the provenance, or the revision**.

The primary purpose of the **source** track is to enable [verification](https://slsa.dev/spec/v1.0/verifying-artifacts) that the **revision followed the expected process**. Consumers have some way of knowing what the expected provenance should look like for a given package and then compare each **source revision's** actual provenance to those expectations. Doing so prevents several classes of [supply chain threats](https://slsa.dev/spec/v1.0/threats).

docs/spec/v1.1/source-requirements.md
joshuagl: We've talked in the past about how we might modify review constraints in the face of robots, I believe it was deemed out of scope for this first PR. Previous discussion includes:

- https://github.com/slsa-framework/slsa/issues/196
- https://github.com/slsa-framework/slsa/issues/956
- https://github.com/slsa-framework/slsa/issues/287

I know some organisations allow robots to automatically merge dependency updates if the change doesn't introduce a breakage. I think that makes sense so long as there are additional controls (i.e., the WIP dependency track in #961) on what changes the robot can introduce.

docs/spec/v1.1/source-requirements.md
TomHennen: Something like that seems reasonable to me.

The one question I have is, you describe the provenance as indicating who the contributors were for a given revision, is that meant to exclude some attestation that describes the state of the repo as a whole?

I think we also need something that describes the state of a the repository as a whole over its lifetime so that consumers don't _need_ to process each specific revision.

Perhaps the intent is to think of 'provenance' broadly here where it might include a couple of different attestation types to achieve those goals?  In which case we'd just need to tweak the language some to make that more clear?

On a related note: do we like the term 'provenance' here?  Will that wind up being confused with https://slsa.dev/provenance

docs/spec/v1.1/source-requirements.md
TomHennen: I think the intent here is whomever pushed the merge button (they eventually cause the Merge Queue to update the main branch).

Do you have any suggestions for alternate language?

docs/spec/v1.1/source-requirements.md
TomHennen: Is the best fix to remove the distinction between the two in this section and state that it's all the responsibility of the SCP who _may_ implement these features via the VCS as appropriate?

docs/spec/v1.1/source-requirements.md
TomHennen: IMO this should only apply to folks who are using 'bare' VCS and that self-hosted SCPs should be eligible for higher levels with the caveat that the trust story will be murkier.  E.g. much easier for a consumer to 'trust' the hosted GitHub/GitLab/etc... solution vs the one I've spun up myself.  But it shouldn't be impossible.

Seem fair?

docs/spec/v1.1/source-requirements.md
TomHennen: Does this come back to the question above of how important the actual time is? (I can see that going either way).

Assuming it _is_ important we would seem to need some ability to track the original times when moving between SCPs?  Should we be evaluating some option for preserving this data that doesn't rely on the SCP itself?  E.g. if we had a per commit attestation a Time Stamp Authority could provide a verified timestamp that could easily migrate to a new SCP.

On a related note, in "MUST retain the change history if possible" I find "MUST" and "if possible" to be contradictory?  Do we resolve this problem by just downgrading "MUST" to "SHOULD"?

TomHennen: So I see your point, it's not strictly contradictory; but I think it is confusing?

In this case it's certainly _possible_ for an SCP to implement some feature that imports the time in a secure fashion and we'll be asking the SCPs to implement other things to implement this spec.  So why bother with the "if possible" qualifier?  Should we be more specific on what cases don't need to retain this information?  Perhaps "importing code from an SCP that does not support or does not have change history".

docs/spec/v1.1/levels.md
```diff
@@ -229,14 +228,128 @@ All of [Build L2], plus:
...
+
+<dt> Requirements <dd>
+N/A
+
+<dt> Benefits <dd>
+N/A
+
+</dl>
+
+#### Source L1: Version controlled
```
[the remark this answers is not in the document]
TomHennen: I think defining L1 as requiring version control would settle the debate about if a source tarball counts as version control (it shouldn't)?

docs/spec/v1.1/source-requirements.md
TomHennen: I'm also realizing that the build track specifies how the information about the build is communicated (provenance) but the source track currently doesn't.  That seems like a gap that should be filled (which would hopefully help address some of these points?)

docs/spec/v1.1/source-requirements.md
lehors: Upon further reflection I think you're right. 
Even if "MUST... if possible" can be interpreted as stronger than SHOULD the reality is that SHOULD already carry the intent we want and there is no real value in getting into the details of the cases where this might not be done.
For the sake of simplicity let's just change it to a SHOULD and drop the "if possible".

docs/spec/v1.1/source-requirements.md
zachariahcox: > is that meant to exclude some attestation that describes the state of the repo as a whole

Yes, there are (typically) many commits in a repo that are invalid, illegal, incomplete or malicious. When you checkout a repo revision for a build, you'll mostly want to know about the version you got. Info about other branches, other revisions, other commits, other points in time is probably not relevant to the choice about whether or not _this_ revision is good to use. 

> something that describes the state of the repository as a whole over its lifetime
Can you talk a little more about how you'd use this kind of data? I guess I'm thinking mostly of ci/cd scenarios: EG, if I fetched a revision of a repo with sha abc123. I'm not sure it would be very relevant to me to know things about the whole repo, like whether the main ref is protected, or when the rules were last changed for the dev branch, etc. Mostly I'll want to know which rules were processed to allow the current revision I have to exist.

TomHennen: > Can you talk a little more about how you'd use this kind of data? I guess I'm thinking mostly of ci/cd scenarios: EG, if I fetched a revision of a repo with sha abc123. I'm not sure it would be very relevant to me to know things about the whole repo, like whether the main ref is protected, or when the rules were last changed for the dev branch, etc. Mostly I'll want to know which rules were processed to allow the current revision I have to exist.

Yes exactly.  And that's what the original (3 years old 😨 ) [Source Control Attestation proposal](https://docs.google.com/document/d/10B-8R3NIGvcPuoIhqb-p18oOe78m6SiiukfE7NyvnYQ/edit?usp=sharing) was trying to get at.

Which isn't to say that it should exclude some ability to reason about any particular commit, they're just separate problems.

As I understand it (and I haven't been paying strict attention to this effort) the latest thinking was that it's useful to have _both_ capabilities.  E.g. 1. some ability for a consumer to determine if the current revision of source is 'good' and 2. some ability for motivated systems or people to determine _why_ 1 is true.

I noticed one of the changes to this PR actually removed the notional attestation format from this PR, so it looks like resolution of that problem is being deferred?

TomHennen: FWIW that doc is shared with https://groups.google.com/g/slsa-discussion.  Everyone should be able to join that group to read it.  My work org doesn't let me share docs with the entire public :(.

docs/spec/v1.1/source-requirements.md
joshuagl: I think self-hosted should be permitted, thinking specifically of projects like freedesktop.org and gnome.org. 

As we're focused on preventing malicious changes through a single account compromise I think an important aspect of the SCP is that the developer account/identity can't make arbitrary/non-observable changes the SCP in order to introduce changes which otherwise wouldn't be allowed. Ideally this would be a achieved through separate ops/admin account, or better yet team, to manage the SCP.

adityasaky (commented):
Just a few comments, I don't expect any of them to block moving forward with this PR with a series of TODOs for us to iterate on.

docs/spec/v1.1/source-requirements.md
TomHennen: Maybe "The _identity of the person_ that applies a change to the source. This person may be the proposer or a different trusted person, depending on the version control platform." ?

TomHennen:
Hey all, in an effort to unblock merging of _this_ PR, so that other issues can be addressed individually (and hopefully more quickly?) I've put together [this document](https://docs.google.com/document/d/13Xt8mA_2b00McGX2vkyhu4GQdFAqtXPu7YXE8ZA6ISE/edit?usp=sharing&resourcekey=0-EqfHF79tUWAKp4PzsE3z1A) (request access or join https://groups.google.com/g/slsa-discussion for access[^1]) to try to tease apart PR blockers from "things that should be addressed before this becomes finalized".

Once we're agreed on TODOs we can turn them into issues and get stuff merging.

Please chime in?

See also [this thread in slsa-specification](https://openssf.slack.com/archives/C03NUSAPKC6/p1718045606647179).

FYI @joshuagl @adityasaky @Nikokrock @zachariahcox

[^1]: Sorry, trying to work with my corp 'firewall' policy.

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
TomHennen: I think the value is also unclear because the spec doesn't yet indicate how the SLSA Source Track metadata is propagated.  Part of the value proposition for Build L1 was that it got people on the path to propagating that metadata?

Maybe there's something similar to do here?

TomHennen:
> I appreciate the effort but I've got to ask: why not opening issues on github instead? Is it that you think commenting on a Google doc is easier/better? (I really don't mean to start a tool war btw)

Exactly.  I really didn't care, I just started there to get things straight in my head.  @joshuagl suggested we can leave it in the doc until we resolve the merge blockers to this PR.

Then I'd better very happy to convert everything to GitHub issues.

docs/spec/v1.1/source-requirements.md
zachariahcox: It seems to me that protected refs should not be writable by actors other than the SCP service principal (side note: this is unfortunately not the case for gh today).
If a protected ref _is_ written by something other than the SCP SP, that's generally a big red flag. 
 
I'd propose "Merger: the human or Trusted Robot that applies a change to the source."

Git digression: 
For most of these security levels, you'd want this to exclusively be the SCP. If a random human was the last pusher to a release ref, you'd probably want to go look through audit logs to see what happened, why did they bypass pr rules, etc.

One downside of this approach is that it makes your push history look very robotic.
We handle mq merges with special ux treatment to help call out that the mq ref was full of mergeable prs.

TomHennen: So I think we might be talking about different things, and maybe we want to capture both?

1. The service principle (am I using this right?) of the entity (user or robot) that actually authorized the changes to the underlying VCS.
2. The service principle of the entity that _triggered_ the change in the first place (which might operate in practice by putting some work on a queue for the trusted robot you mention)

Num 1 is helpful if we care about which accounts have broad scope to tamper with source.

Num 2 is helpful if we want the accounts involved in the review of the change itself.

So in the case you describe (1) would be the SCP SP, (2) would be the account of the person that caused the SCP SP to make that change.

TomHennen:
FYI it was suggested we try to reach agreement on what's needed to submit _this_ PR by the end of the community meeting Monday.  There would still be much more to do to get the final version 'public'.

joshuagl:
I rebased the PR on main and pushed a commit to resolve the linter warnings. Let's figure out remaining minor changes to land this PR and then build on it in future PRs like #1066.

zachariahcox (approved):
This is good progress. 
I vote to merge this one and address the remaining questions with subsequent prs: https://github.com/slsa-framework/slsa/labels/source-track

TomHennen (approved):

TomHennen:
> 

+1

I think we can merge this which will unblock additional changes.

IIUC after Monday's meeting @joshuagl was looking for someone like @mlieberman85 to review, approve, and merge.

WDYT Mike?

joshuagl (approved):
I think merging this PR as-is is a good path forward. I started to cherry-pick @Nikokrock's top commit from #1082, but it failed the linting phase and the edits I had to make touched a lot of lines.

It seems to me the easiest way to progress is to merge this PR and iterate in future smaller PRs, alleviating the need for managing the significant git history in this branch or requiring a maintainer to edit this branch.

@arewm @lehors @mlieberman85 let us know if you disagree with this approach, otherwise I will plan to merge this by EOD (UTC+1) on Friday.

lehors (approved):
LGTM. Thanks.

arewm (approved):
I am fine with this as a first draft since we removed the content on the attestations.

PR #1045, opened 2024-04-10
impl: add a script to easily diff two built sites

MarkLodato:
Whenever we get a pull request to update a dependency, I run a diff to
make sure it doesn't break anything. This script simplifies that process
since it now works on archives download from Netlify, and it no longer
requires copy/pasting commands.

This could easily be turned into a stand-alone script outside SLSA -
seems like a generally useful tool!


joshuagl (approved):
This is very neat!

tools/diff_site
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+USAGE="USAGE: $0 DIR_OR_FILE DIR_OR_FILE"'
+
+Displays the diff of two built site directories, each of which can be a
+local directory (usually `docs/_site`) or a `deploy-*.zip` archive downloaded
+from Netlify. Before diffing, we filter out out innocuous changes such as the
```
lehors, line 6:
```suggestion
from Netlify. Before diffing, we filter out innocuous changes such as the
```

lehors (approved):
I agree this is pretty nifty!

Issue #1039, opened 2024-03-27
Unclear difference between dependency and package threats


NVolcz:
The difference between "(D): Use of compromised dependency" and "(H): Use of compromised package" is unclear.
There is no explanation of what make using a package as a dependency different from using it in any other form or way.

I suggest that we either clarify the difference or merge them into one threat.

https://slsa.dev/spec/v1.0/threats-overview
![image](https://github.com/slsa-framework/slsa/assets/1667690/9524d651-3efa-4dfa-bc18-640dfdc37736)

Below are some parts out of a [conversation on Slack](https://openssf.slack.com/archives/C029E4N3DPF/p1710164125448349):

arewm:
> The presence of signed provenance, trusting the signing entity, and its reference to a specific package (i.e. by immutable reference) can be used. Either during consumption of the generated package or consumption of dependencies for the build

nvolcz:
> Thanks! That makes sense! It sound like it is the same threat but mitigated at different times in the process (edited) @

arewm:
> > but mitigated at different times in the process

> Yes, exactly. The threat model has to have a specific reference perspective. It is presented from the lifecycle of a single package/artifact including its dependencies and its consumers. The same rules can apply to both, but they are different from the perspective of a given artifact. (H) would always exist unless you generate a package/artifact which is never used. (H) for one artifact is (D) for its consumer.

MarkLodato:
Agreed on the confusion. Thanks for pointing it out. We should address this.

(D) is really all of the other threats (A)-(H) recursively, from a different frame of reference. It is saying that a given package could be compromised by doing (A)-(H) from one of its dependencies.

Maybe an example would help. Consider [`event-stream`](https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/), which was compromised through its dependency `flatmap-stream`. The attacker snuck in code into the NPM release of `flatmap-stream` that was not present in the source on GitHub. So that's (F) for `flatmap-stream` and (D) for `event-stream`. Does that help?

MarkLodato:
I also agree that the name of (H) could be improved. We originally said "Trick user into using bad package", which sounded limited to typosquatting, but we really also intend things like modification in transit (e.g. MitM).

laurentsimon:
I also think some of the build threats are not relevant where they should be. For example, imo F and G should be under a "publication threat" rather than build threat.

MarkLodato:
[...]

PR #1046, opened 2024-04-15
content: refine dependency threats

MarkLodato:
Refine the "dependency threats" of the Threats & Mitigations page to
better explain the intent and to differentiate from (H), which sounded
confusingly similar (#1039). More work is needed to refine (H), but that
is left to a separate PR.

-   Explain that "dependency threats" are not distinct threats but
    rather threats to other pieces of software that also affect this
    one. In the diagram, the diagram to color (D) differently to show
    this, mirroring the existing dashed lines, and add "(A-H
    recursively").
-   State that only "build dependencies" are in scope for the threat
    model, matching existing diagram and terminology throughout SLSA.
-   Rename "Use compromised dependency" to "Compromise build
    dependency", both for consistency with other threats, which are from
    the adversary's point of view, and to emphasize that this is
    restricted to build dependencies.
-   Expand the text about dependency threats and give examples.
    Highlight both "include" and "build tool" types of dependencies, and
    also include both an accidental vulnerability and a malicious
    backdoor.


MarkLodato:
I'm interested in feedback from folks to see if this is a step in the right direction.

docs/spec/v1.1/threats.md
```diff
@@ -139,47 +139,81 @@ the source repo does not match the expected value.
...
 
 </details>
 
+### (n/a) Compromise runtime dependency
+
+[runtime dependency]: #na-compromise-runtime-dependency
+
+An adversary compromises a runtime dependency of target artifact, meaning an
+independent software artifact that the target artifact requires for use. A
+runtime dependency cannot, by definition, affect the build process of the target
```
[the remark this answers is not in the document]
MarkLodato: Done. (I think - did I interpret correctly?)

docs/spec/v1.1/threats.md
```diff
@@ -139,47 +139,81 @@ the source repo does not match the expected value.
...
+ability to influence the output is considered a build dependency. This includes:
 
-</details>
+-   Libraries, container base images, bundled files, and so on that are
+    "included" in the target artifact.
+-   Build tools, interpreters, command-line utilities, OS packages, and so on
+    that are "used" during the build process.
+
+**Reminder:** A [runtime dependency] that is loaded a build time is also a build
+dependency!
```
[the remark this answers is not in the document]
MarkLodato: Rephrased and moved to a different section, PTAL

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,105 @@ the source repo does not match the expected value.
...
+
+<details id="runtime-dep"><summary>Use a compromised dependency at runtime <span>(modeled separately)</span></summary>
+
+*Threat:* Load a compromised artifact at runtime, thereby compromising the user
+or environment where the software ran.
 
-*Example:* The artifact dynamically links `libBar` and requires a binary version
-to run. The adversary compromises `libBar`'s build process and inserts malicious
-code. When your artifact runs, it contains the adversary's malicious code.
+*Out of scope:* SLSA's threat model does not explicitly model runtime
```
meder, line 238: I wonder if "deployment time" captures this nuance better.

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,105 @@ the source repo does not match the expected value.
...
+expect that a future version will. In the meantime, you can [apply SLSA
+recursively] to your dependencies in order to reduce the risk of dependency
+threats.
+
+[apply SLSA recursively]: verifying-artifacts.md#step-3-optional-check-dependencies-recursively
+
+### (D) Compromise build dependency
 
-### (D) Use compromised dependency
+[build dependency]: #d-compromise-build-dependency
```
meder, line 163: The diagram highlights that threats A-H apply recursively. Is the goal of this section to list additional threats or the outcomes of those threats(A-H) materializing?

I wonder if we can help readers connect the dots by mapping some of the common OSS dependency threats to SLSA threats A-H (potentially creating new ones), e.g.:

- **Dependency hijacking**: It appears to map to Upload modified package(F) if we expand that a bit in how we define that, but I could also see that there's a missing equivalent to "Submit unauthorized change" (A) for package registries.
- **Dependency confusion/typosquatting**: this appears to map to "Using a compromised dependency" but not very cleanly as dependency is not really compromised, it's just plain malicious and consumers are using it due to a mistake.

docs/spec/v1.1/images/supply-chain-threats.svg
meder, line 1: This will likely be in the followup, but want to get your thoughts on it. Does it make sense to introduce visual 1P vs 3P boundaries in this diagram?

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,105 @@ the source repo does not match the expected value.
...
+ability to influence the output is considered a build dependency. This includes:
 
-*Threat:* The adversary injects malicious code into software required to build
-the artifact.
+-   Libraries, container base images, bundled files, and so on that are
+    "included" in the target artifact.
+-   Build tools, interpreters, command-line utilities, OS packages, and so on
+    that are "used" during the build process.
+-   Runtime dependencies that are loaded during the build process and have the
+    ability to affect the build process.
```
ramonpetgrave64, line 174: These sound the same as Build Dependencies. Could we reframe these as Test Dependencies?

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,105 @@ the source repo does not match the expected value.
...
-<details><summary>Use a compromised build dependency</summary>
+An adversary compromises the target artifact through one of its build
+dependencies. Any artifact that is present in the build environment and has the
+ability to influence the output is considered a build dependency. This includes:
 
-*Threat:* The adversary injects malicious code into software required to build
-the artifact.
+-   Libraries, container base images, bundled files, and so on that are
+    "included" in the target artifact.
+-   Build tools, interpreters, command-line utilities, OS packages, and so on
```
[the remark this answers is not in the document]
MarkLodato: This bulleted list is redundant with the zippies below. I deleted it and added the examples below, specifically calling out compilers. Does that address? If not, do you have another suggestion?

(To me "build toolchain" is less clear, but if you think it's more clear I can add it. I just don't want the list to get too long.)

docs/spec/v1.1/threats.md
MarkLodato: It's not just tests though. I think you also need a copy of a shared library when dynamic linking, so I added that as an example as well.

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,105 @@ the source repo does not match the expected value.
...
+process of MyPackage as part of a test. An adversary compromises Dep such that,
+when run during a build, it injects a backdoor into the output artifact. The
+next time MyPackage is built, it picks up and loads Dep during the build
+process. The malicious code then injects the backdoor into the new MyPackage
 artifact.
 
-*Mitigation:* N/A - This threat is out of scope of SLSA v1.0. However, you can
-mitigate this threat by verifying SLSA provenance for all of your runtime
-dependencies that provide provenance.
+*Mitigation:* In addition to all the mitigations for build tools, you can often
```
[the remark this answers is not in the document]
MarkLodato: Yes, thanks for calling this out. To get this PR landed, I just added a TODO.

docs/spec/v1.1/threats.md
MarkLodato: I think runtime is actually the correct word, though. We need definitions. I interpret "deploy" as "load into an environment to be run". You don't need dependencies to deploy - they'll get loaded just fine. It just won't work when you go to run it. And from a security perspective, you don't get compromised until you actually "run" the thing.

Could you help me understand your concern with "runtime"?

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,105 @@ the source repo does not match the expected value.
...
+
+*Threat:* Load a compromised artifact at runtime, thereby compromising the user
+or environment where the software ran.
 
-*Example:* The artifact dynamically links `libBar` and requires a binary version
-to run. The adversary compromises `libBar`'s build process and inserts malicious
-code. When your artifact runs, it contains the adversary's malicious code.
+*Out of scope:* SLSA's threat model does not explicitly model runtime
+dependencies. Instead, each runtime dependency is considered a distinct piece of
+software with its own threats.
```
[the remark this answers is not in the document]
MarkLodato: Renamed to "artifact" to be more general.

docs/spec/v1.1/threats.md
meder: Ack on expanding the docs to explicitly cover hijacking, confusion and malicious publishers. Issues or TODOs sound good, happy to help with the content.

Hijacking: Agree that it's an outcome of A-H, with F mapping best to some of the high profile attacks we've seen (e.g. ua-parser-js).

Confusion: There are a couple of things to consider here, I think confusion attacks by their nature are not really about compromising anything or performing unauthorized actions but rather creating a malicious package and waiting for the victims to use it. Compromised vs malicious is worth discussing further to see if we need to introduce new threats.

docs/spec/v1.1/threats.md
meder: There are effectively two types of runtime dependencies: build runtime deps and deploy time runtime deps. (D) qualifies them("runtime dependency during the build") but this doesn't.

docs/spec/v1.1/threats.md
meder: My other comment must've gotten lost: but we may need a longer discussion about this on on PR: https://github.com/slsa-framework/slsa/pull/1046#discussion_r1577233486

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,119 @@ the source repo does not match the expected value.
...
+
+<details id="runtime-dep"><summary>Use a compromised dependency at runtime <span>(modeled separately)</span></summary>
+
+*Threat:* Load a compromised artifact at runtime, thereby compromising the user
+or environment where the software ran.
 
-*Example:* The artifact dynamically links `libBar` and requires a binary version
-to run. The adversary compromises `libBar`'s build process and inserts malicious
-code. When your artifact runs, it contains the adversary's malicious code.
+*Out of scope:* SLSA's threat model does not explicitly model runtime
```
jkjell, line 252:
```suggestion
*Mitigation:* N/A - This threat is out of scope of SLSA. SLSA's threat model does not explicitly model runtime
```

Nit: Something similar for consistency with other "out of scope" references on the page.

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,119 @@ the source repo does not match the expected value.
...
+dependencies. Instead, each runtime dependency is considered a distinct artifact
+with its own threats.
 
 </details>
```
jkjell, line 256:
```suggestion
*Example:* **TODO**

</details>
```

jkjell (commented):
Thanks for this work Mark. From my anecdotal view, dependency threats are the most common example of supply chain attacks. Providing this section with more coverage is really important.

meder (approved):

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,119 @@ the source repo does not match the expected value.
...
+
+An adversary compromises the target artifact through one of its build
+dependencies. Any artifact that is present in the build environment and has the
+ability to influence the output is considered a build dependency.
+
+<details id="included-dep"><summary>Include a vulnerable dependency (library, base image, bundled file, etc.)</summary>
+
+*Threat:* Statically link, bundle, or otherwise include an artifact that is
+compromised or has some vulnerability, causing the output artifact to have the
+same vulnerability and the original artifact.
```
joshuagl, line 186: as the original artifact? Or is this saying that the vulnerable artifact is also included? Probably simplest to just cut the sentence 4 words shorter?
```suggestion
same vulnerability.
```

docs/spec/v1.1/threats.md
```diff
@@ -139,44 +139,119 @@ the source repo does not match the expected value.
...
+
+*Example:* The C++ program MyPackage statically links libDep at build time. A
+contributor accidentally introduces a security vulnerability into libDep. The
+next time MyPackage is built, it picks up and includes the vulnerable version of
+libDep, resulting in MyPackage also having the security vulnerability.
+
+*Mitigation:* **TODO**
```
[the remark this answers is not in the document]
joshuagl: Bit of an aside here, but the vulnerability is still present unless there's some code pruning or similar. Importantly, the _threat_ of the vulnerability still exists, because code changes might introduce a call to the vulnerable API or a canny attacker may be able to exploit the vulnerability some other way.

joshuagl (approved):

docs/spec/v1.1/threats.md
MarkLodato: Sorry, I'm getting lost in the conversation. Do you suggest keeping it as-is or changing it, and if the latter, could you recommend a change?

docs/spec/v1.1/threats.md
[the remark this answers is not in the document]
MarkLodato: Hmm, this is a good point. I wonder if this is also "threat" vs "risk" vs "mitigation", which I struggle with. Ideas welcome!

Our taxonomy currently enumerates "threats" that are realized, so to speak. In this case, it describes the case where the output actually is vulnerable. If it's not vulnerable (e.g. unused code path), then it's not a threat. (I think?!)

But there's a larger "risk" (?) that you don't know whether you are vulnerable or not, or you are unable to respond to the threat. For example, you might not be able to find all of the software that you need to patch and re-deploy. And just patching everything that *might* be vulnerable has its own costs (risks?) so that's not great either.

Maybe the answer is that understanding whether or not the output is vulnerable is part of the mitigation to this threat?

joshuagl (approved):

docs/spec/v1.1/threats.md
MarkLodato: Marking as resolved since I think this is addressed.

docs/spec/v1.1/threats.md
MarkLodato: Added a full example.

docs/spec/v1.1/threats.md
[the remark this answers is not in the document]
MarkLodato: Added a TODO at the top of the page.

docs/spec/v1.1/threats.md
```diff
@@ -139,47 +139,81 @@ the source repo does not match the expected value.
...
+
+An adversary compromises a runtime dependency of target artifact, meaning an
+independent software artifact that the target artifact requires for use. A
+runtime dependency cannot, by definition, affect the build process of the target
+artifact. (But remember that any artifact loaded at build time is also
+considered a [build dependency]!) Examples of runtime dependencies include
+dynamically linked libraries or packages, or command-line utilities invoked by
+the target artifact but not bundled with the target artifact.
+
+SLSA's threat model does not explicitly model runtime dependencies. Instead,
```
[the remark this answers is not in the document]
MarkLodato: Added a TODO at the top.

MarkLodato (commented):
Thanks everyone! Since I have a bunch of approvals, I'm going to merge now. Again, this isn't perfect, and there are a ton of TODOs, but hopefully this gives us something that we can further iterate on.

Issue #977, opened 2023-10-09
SLSA Build L4


MarkLodato:
This meta-issue is meant to cover development and release of SLSA Build L4, so that we have a common issue to point to when referring to Build L4.

TODO: link to sub-issues

david-a-wheeler:
Note that I'm currently proposing entries for build levels L4 and L5. See: https://github.com/slsa-framework/slsa/issues/873

arewm:
I mentioned in the comments for the `Reproducible Build requirements in SLSA (an early proposal)` about how I am not convinced that we should add reproducibility onto the SLSA build levels. Where would be the best place to have this conversation? In this issue, in that document, in a community call, elsewhere? I wasn't able to attend a previous call where this discussion happened.

MarkLodato:
What we did in [SLSA Source Track brainstorming](https://docs.google.com/document/d/1nVkvRsxFef2OgVm2CumyLDbt-yYX-A5wtg6dF_sthwo/edit) was for each person to write down some thoughts as separate sections, then people commented on those sections and the original authors refined their ideas. Then once the comments died down, the lead (@kpk47) coalesced the ideas into one proposal. That seems to be working ok?

So if we want that model, then perhaps you could create a section in the doc and write down your thoughts on L4 and/or reproducible builds. That would allow us to critique the argument and you can hone it. It would also leave us something more durable than a docs comments and more readable than a GitHub issue comment. What do you think?

david-a-wheeler:
If you have ways to improve the proposal, currently the Google doc would be the right start.

If you oppose the concept of being able reproduce builds, I guess https://github.com/slsa-framework/slsa/issues/873 would be the place. Are you opposed to being able to reproduce builds, or are you opposed to including them in the "build track", or is it something else? I'm not sure I understand your objection.

david-a-wheeler:
@arewm - I understand! Sorry, I was a little confused about your point. I originally proposed that they be a separate track, but many in the community preferred that they be in the same build track. Please do add commentary.

I think we can separate the issues of (1) what might be usefully added to SLSA and (2) whether or not reproducible builds belongs in a different track. Indeed, as we refine the potential requirements, it may be easier to decide if they belong in the same or different track.

PR #1051, opened 2024-05-03
content: Add attested build environments level requirements

### LINKED ISSUE #975: Workstream: Hardware Attested Build Environments

marcelamelara:
This (draft) PR introduces the following spec changes associated with #975. Per https://github.com/slsa-framework/slsa/issues/975#issuecomment-1773315232 and https://github.com/slsa-framework/slsa/issues/975#issuecomment-1852559183 the spec enhancements are being proposed as a *new Build track level*.  The spec changes introduced in this PR are meant to be complementary to possible requirements being developed in parallel in #977 .

Spec changes:
- [x] Add new level to levels.md
- [x] Add new requirements to requirements.md
- [x] Add related terminology to terminology.md
- [ ] Add example of hardware-based build environment attestation

Part 1 of #975 CC @chkimes


docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+
+<dl class="as-table">
+<dt>Summary<dd>
+
+Compromising a build requires exploiting a vulnerability in the build
+platform's supply chain or physical access to the underlying compute
+platform.
+
+In practice, this means that builds are configured to run on a build platform
+that supports hardware capabilities for attestesting to the state of the
```
deeglaze, line 247:
```suggestion
that supports hardware capabilities for attesting to the state of the
```

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+compute environment executing builds.
+
+<dt>Intended for<dd>
+
+Software releases needing assurances about the integrity of the environment
+used to create the release (e.g., specific compute platform, pre-build
+tamper detection).
+Build L4 usually requires significant changes to existing build platforms.
```
arewm, line 255: There are no prior examples to be able to assess the truthfulness of this statement. Significant changes could be required for any increased level.

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+
+<dt>Requirements<dd>
+
+All of [Build L3], plus:
+
+-   Software producer:
+    -   MUST run builds on a hosted build platform that meets Build L4
+        requirements.
+    -   SHOULD verify the build platform's attestations prior to a build and
+        produce a [verification summary] about the check.
```
arewm, line 265: What would this look like? I would expect that the build system would create a build environment attestation on the artifact. I feel like it is a lot to ask for a producer to verify it and publish the VSA. Or does this not have to be a human/manual process?

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+
+-   Build platform:
+    -   Each build image (i.e., VM or container) made available to software
```
arewm, line 268: The VMs must be built on a SLSA Build L3+ platform as well? What does this mean in practice?

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+        producers MUST be built on a SLSA Build L3+ platform. The generated
+        SLSA Provenance MUST be distributed to allow for independent
+        verification.
```
arewm, line 271: The provenance for _what_ must be distributed the build image? This is already a requirement for build L1-3: https://slsa.dev/spec/v1.0/requirements#distribute-provenance . Is the intention for this provenance to be verified somehow?

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+    -   Distribution of SLSA Provenance for pre-installed software within the
+        build image MAY be best-effort.
```
arewm, line 273: How would this distribution happen? Are we trying to add recursive SLSA into this requirement? Is that too much to do?

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+    -   The boot process of each build environment MUST be measured and
+        attested using a [TCG-compliant measured boot] mechanism. The
+        attestation MUST be authenticated and distributed for independent
+        verification.
```
arewm, line 277: These changes are proposed to be part of the build track but we are now adding additional required attestations into the mix (this attestation + the producer's VSA mentioned earlier). I know that we do not have a 1:1 relationship between attestations and tracks, but this seems to be muddling the simplicity of the build track. Would this proposal be better as its own track?

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+    -   The initial state of the build environment's disk image MUST be
+        integrity measured and attested. The attestation MUST be
+        authenticated and distributed for independent verification.
+    -   Read-write block devices or file system paths MUST be encrypted
+        with a key that is only accessible within the build image.
```
arewm, line 282: Does this need to be included in the attestation? 

Can you clarify the presence of the key? I assume that you are not trying to say that the encryption key should be present in the image used to run the VM/container as then it would be reused for multiple running instances of the build.

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+    -   Before launching a new environment based on a build image (i.e., VM
+        or container instance), its SLSA Provenance MUST be verified.
```
arewm, line 284: What are the criteria for verification here? Is it just the identity or is there more verification needed than that? Are we supposed to verify that it has met some SLSA level?

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+    -   Before making a build environment available for a build request:
+        -   The boot process and state of disk image MUST be verified, and
```
arewm, line 286: This verification feels different than the previously mentioned one. Is it?

docs/spec/v1.1/levels.md
```diff
@@ -26,6 +26,7 @@ tracks without invalidating previous levels.
 | [Build L1]  | Provenance showing how the package was built | Mistakes, documentation
 | [Build L2]  | Signed provenance, generated by a hosted build platform | Tampering after the build
 | [Build L3]  | Hardened build platform | Tampering during the build
+| [Build L4]  | Attested build environment | Tampering before the build
```
arewm, line 29: Can you propose what the addition to requirements.md would look like? I think that is where the _core_ specification is maintained.

arewm (changes requested):
The SLSA build track does not have a requirement for completeness of resolved dependencies as of today (L3). If feels like we should work to achieve that completeness before we further harden the build environment. It may be possible to add complete provenance _alongside_ the other changes, but that feels like too much change from L3 to L4.

docs/spec/v1.1/levels.md
marcelamelara: Thanks for catching this!

docs/spec/v1.1/levels.md
marcelamelara: Fair point. It's true that L3 already has a very similar statement, and we still wanted to be explicit about the fact that this L4 would require additional significant changes on top of L3. We can be more precise in what we mean here: for example, one of the significant requirements is hardware with very particular features (e.g., TPM or TEE support). Would that be more helpful here?

docs/spec/v1.1/levels.md
marcelamelara: The main goal of the proposed requirements is to make tampering _prior_ to a build detectable. That is, at any point from when a build image is itself created/released to the point where a VM is deployed and waiting for a new build job to come in. This means the build platform will actually generate its attestations before and independently of the artifact. We have a figure showing this sequence which I think would be helpful in clarifying things. 

I'll note that the point of using hardware-based integrity measurements and attestation for this is to reduce the amound of manual self-attestation and verification that needs to happen on the part of the build platform and producer. I'll also note that this requirement is a SHOULD, so producer's aren't strictly required to check these attestations.

docs/spec/v1.1/levels.md
marcelamelara: Yes! Edits to requirements.md are part of this PR, and I plan to have a first draft of those ready by the end of this week. I might move some of the content from the current L4 description into that doc.

docs/spec/v1.1/levels.md
marcelamelara: We aren't trying to add recursive SLSA in this requirement. The intent here is to clarify that while SLSA Provenance _is_ required for the build image creation, SLSA Provenance for pre-installed software (e.g., the Linux kernel, packages etc) isn't expected of the build platform.

docs/spec/v1.1/levels.md
marcelamelara: Yes, we're using the creation process of the VM image to bootstrap a first degree of trust in the build environment. Today we're already placing trust in GHA to configure an L3 platform, so say GHA VM images were themselves built on GHA, this requirement would give us assurances that the VM images were built with the same L3 integrity. It's a bit recursive, much like building gcc with gcc is, but the guarantees provided by the rest of the requirements become significantly weakened if we didn't have integrity for the build environment's build process. This is because the Provenance of the VM image gives us the good known value that can be later check against when the VM is then deployed.

marcelamelara: I should also add that this the conditions for meeting requirement are meant to be binary. That is, L4 is achieved iff the VM image is built on a SLSA Build L3+ platform. We believe this is necessary to avoid the issue of resolving what the transitive SLSA level would be.

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+All of [Build L3], plus:
+
+-   Software producer:
+    -   MUST run builds on a hosted build platform that meets Build L4
+        requirements.
+    -   SHOULD verify the build platform's attestations prior to a build and
+        produce a [verification summary] about the check.
+
+-   Build platform:
+    -   Each build image (i.e., VM or container) made available to software
```
MarkLodato, line 268: I recommend removing these requirements. In my mind, the requirement should be relatively simple: there is a hardware-backed attestation (SEV-SNP, TDX, or equivalent) attesting to:
- the entire initial state of the build environment (bootloader, kernel, filesystem, etc)
- all of the inputs to the build

The properties of those inputs, such as the VM, seem outside the scope. There are many inputs to the build, and I would assume that we should verify each of those separately as part of some "transitive SLSA" verification.

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,115 @@ All of [Build L2], plus:
...
+        ID) MUST be generated and cryptographically bound to a valid build
+        environment integrity chain.
+    -   All build platform generated attestations and cryptographic bindings
+        MUST be backed by a hardware root of trust (e.g., [TPM] or [trusted
+        execution environment]). Note: Virtual hardware (e.g., vTPM) MAY be
+        used to meet this requirement.
+    -   Runtime changes to the build environment's disk image SHOULD be
+        observable at runtime by the executing build request.
+
+<dt>Benefits<dd>
```
MarkLodato, line 308: An alternate framing would be that it removes almost all of the build platform from the root of trust. All that's left is the hardware vendor and physical access.

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
marcelamelara: Ah. I can see how this is ambiguous. While the main goal is to provide stronger protections during the build, what we really meant here is tampering before _build execution_, i.e., tampering with the build environment, specifically during the stages when the build environment is created and deployed, but before the tenant's build is actually executed. Perhaps something like "Tampering during build via build environment" would be clearer here?

docs/spec/v1.1/levels.md
marcelamelara: Agreed, I realized looking at this again more recently that this list really enumerates low-level mechanisms for achieving the high-level properties, rather than stating those high-level properties. My thinking is that a lot of the content currently here would ultimately move to the requirements.md description.

> The properties of those inputs, such as the VM, seem outside the scope. There are many inputs to the build, and I would assume that we should verify each of those separately as part of some "transitive SLSA" verification.

I generally agree with the notion that inputs/dependencies to the build should be verified separately. At the same time, the properties of the VM specifically are needed to check the initial state of the build environment. That is, in order for the platform or a strict producer to verify the initial state of the build environment it needs to have good known reference values to check against. The SLSA L3 Provenance for the build image's build provides this good known value, with strong integrity assurances, that can then be bootstrapped to gain trust in the integrity of the initial state of the build environment.

docs/spec/v1.1/levels.md
marcelamelara: At a high level, this is true, and we should probably keep this section more concise than it currently is. We did want to capture some of the nuance of doing this practice, though. For example, you actually need to still place a fair amount of trust in the cloud provider that's hosting the VMs running the builds to not tamper with components like the host OS or the vTPM implementation being used to check the integrity of the build environment.  The other part we wanted to emphasize is the machine-checkable aspect of relying on the attestable hardware, compared to the expectations for verification at L3. Maybe this nuance doesn't need to be covered in such detail here?

docs/spec/v1.1/levels.md
marcelamelara: > there is a hardware-backed attestation (SEV-SNP, TDX, or equivalent) attesting to ... all of the inputs to the build

@MarkLodato Thinking about this some more as I reformulate the requirements. What do you mean by "inputs to the build"? Is this about the completeness of external parameters in the Provenance? If so, this requirement as stated implies that the Provenance itself should be signed by the hardware root of trust.

docs/spec/v1.1/levels.md
marcelamelara: You raise a valid point about preserving the simplicity build track, and the types of attestations being a part of that. The intent here is to actually encapsulate the hardware-based attestations inside an in-toto attestation, to at least ensure we remain within the SLSA attestation model. So at Build L4 there would be two types of in-toto attestations that would be generated: SLSA Provenance and SCAI encapsulating the hardware-based attestations. We think this approach helps us keep the attestation types limited to a select few. I'll add a TODO item to include examples of these attestations to illustrate what we would expect, and that'll hopefully clarify some of these questions.

On the question of converting this workstream into its own track, our original proposal went down that path. But we got some very convincing feedback from some in the community that that would actually introduce more complexity than is necessary in SLSA overall when the Build Track already pertains to properties of the build platform.

docs/spec/v1.1/levels.md
marcelamelara: The main aspect to verify here is the identity/hash of the build image against what's recorded in the Provenance. Checking the SLSA level of the Provenance here gives extra assurances about the trustworthiness of the hash. In practice, this check would follow the standard [verification flow](https://slsa.dev/spec/v1.0/verifying-artifacts#step-1-check-slsa-build-level).

docs/spec/v1.1/requirements.md
```diff
@@ -330,9 +335,77 @@ means that the build ran with no network access. Such a requirement requires
...
+The build platform generated an authenticated attestation to the integrity
+of the entire initial state of the build environment (i.e., VM/container
+image, kernel, and filesystem) was generated at creation time and verified
+at deployment time. The build platform also attested to the build request.
+In other words, tampering with the initial state of the build environment
+MUST be detectable by the platform itself and the build.
+
+The build platform MUST guarantee the following:
+
+-   When creating a new build environment:
```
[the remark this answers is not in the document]
chkimes: I just got to the terminology file, I think just "creating a new build image" would be appropriate given that terminology

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,74 @@ All of [Build L2], plus:
...
+    -   MUST run builds on a hosted build platform that meets Build L4
+        requirements.
+    -   SHOULD verify the [build environment attestations] at the start of a
+        build and produce a [verification summary] about the check.
+
+-   Build platform:
+    -   MUST generate and distribute attestations to good known integrity
+        measurements of the entire initial state of the build environment
+        (i.e., VM/container image, boot process and filesystem). All
+        attestations MUST be authenticated by a hardware root of trust.
```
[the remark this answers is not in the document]
marcelamelara: There's actually [a note](https://github.com/slsa-framework/slsa/pull/1051/files#diff-5f5f50825551c8161c1c73587ac6823a7d81ded3fa19d597147ec06c238bdcf8R393) at the end of the requirements clarifying that virtual hardware also works. Do you think that's enough, or is it worth clarifying up front?

docs/spec/v1.1/requirements.md
```diff
@@ -330,9 +335,77 @@ means that the build ran with no network access. Such a requirement requires
...
 considered in the [future directions](future-directions.md).
 
-<td> <td> <td>✓
+<td> <td> <td>✓<td>✓
+<tr id="hardware-attested">
+<td>Hardware-Attested
+<td>
+
+The build platform generated an authenticated attestation to the integrity
+of the entire initial state of the build environment (i.e., VM/container
```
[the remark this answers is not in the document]
marcelamelara: To clarify, you specifically mean the case where a container is used within a VM, right? My intent here was more to cover the scenario in which container-only build images are used. Whether or not containers are sufficient to achieve L3, I think is a related but separate question.

docs/spec/v1.1/requirements.md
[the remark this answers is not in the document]
marcelamelara: Good catch, this bullet point should say "build image". In terms of unloading the term "image" more, I'm thinking we could use the term "filesystem" whenever the text says "disk image". wdyt?

docs/spec/v1.1/levels.md
deeglaze: This text also presupposes that all builders run in VMs. Are we categorically rejecting remote attestations from physical TPMs on non-virtual machines?

docs/spec/v1.1/requirements.md
```diff
@@ -330,9 +335,77 @@ means that the build ran with no network access. Such a requirement requires
...
+        verified to ensure the guest kernel and filesystem have not been
+        tampered with.
+    -   A unique immutable build environment identifier (e.g.,
+        cryptographic keypair) MUST be generated and cryptographically bound
+        to the build environment via attestation. This *deploy-time
+        attestation* MUST be generated only after build image, boot process
+        and disk image integrity have been verified, and distributed to
+        allow for independent verification.
+-   When accepting a new build request (e.g., GHA build job):
+    -   The build environment's deploy-time attestation and uniqueness of its
```
deeglaze, line 376: What does this mean? I imagine that the SLSA L4 attestation signing key would be deleted from memory after use from a previous build job, so yes you'd need to re-establish the attestation with the key management service, but only AFTER the build request has completed. You don't want the key material in memory while an untrusted build process is taking place, even if it is isolated.

Who does the build environment attestation verification? It seems like the build job is accepting itself from this wording. The build requester would not necessarily be the one with an appropriate attestation verification policy to apply before even initiating a build request, but maybe that's what you meant? Is a build request supposed to happen over an attested channel?

docs/spec/v1.1/requirements.md
```diff
@@ -330,9 +335,77 @@ means that the build ran with no network access. Such a requirement requires
...
+    -   A unique immutable build identifier (e.g., GHA build job
+        ID) MUST be generated and cryptographically bound to the build
+        environment via attestation. This *request-time attestation*
+        MUST be generated only after the deploy-time attestation has been
+        verified, and distributed to allow for independent verification.
+    -   Run-time changes to the build environment's disk image SHOULD be
+        observable at run-time by the executing build. These changes NEED NOT
+        be attested.
+-   Boot, disk, deploy- and request-time attestations MUST be authenticated
+    by a hardware root of trust (e.g., [TPM] or [trusted execution
```
deeglaze, line 388:
```suggestion
    by a DRTM (e.g., [TPM] or [trusted execution
```

docs/spec/v1.1/requirements.md
```diff
@@ -330,9 +335,77 @@ means that the build ran with no network access. Such a requirement requires
...
+        of the build environment has not been tampered with.
+    -   A unique immutable build identifier (e.g., GHA build job
+        ID) MUST be generated and cryptographically bound to the build
+        environment via attestation. This *request-time attestation*
+        MUST be generated only after the deploy-time attestation has been
+        verified, and distributed to allow for independent verification.
+    -   Run-time changes to the build environment's disk image SHOULD be
+        observable at run-time by the executing build. These changes NEED NOT
+        be attested.
+-   Boot, disk, deploy- and request-time attestations MUST be authenticated
```
deeglaze, line 387: I don't think request-time attestation is what's desirable. Although it could be an enhancement for tenant confidentiality, i.e., "I don't want to release my code to a builder I haven't vetted."
For build integrity, we want response-time attestation in order for both input and output resource descriptors to be cryptographically bound to the attestation. This prevents replay attacks given that the attestations are meant to be verified offline (i.e., not during the build process where the environment could respond to a cryptographic challenge for freshness).

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,74 @@ All of [Build L2], plus:
...
+    -   MUST capture all of the inputs to the build in the Provenance.
+    -   MUST verify the build environment integrity against the build
+        environment attestations prior to passing control of the build to
```
deeglaze, line 274: When you say "prior to", do you mean that there is a separator event between boot to the build executor and the executor running any build request? That's my expectation, but this also seems like there could be a temporal prior, where an attestation verification service must do this verification before providing a ticket (for example) to permit the build request to proceed. I don't think the latter is desirable. See my later comment about response-time attestations.

docs/spec/v1.1/terminology.md
```diff
@@ -125,6 +125,39 @@ of build types](/provenance/v1#index-of-build-types).
...
+    *boot process*.
+3.  When a new *build request* is made, the platform assigns the request to
+    a deployed build environment. For SLSA Build L4, the tenant may validate
+    the measurement of the build environment.
+4.  Finally, the *build executior* running within the environment executes
+    the tenant's build steps.
+
+| Primary Term | Description
+| --- | ---
+| Build image | The run-time context within a build environment, such as the VM or container image. Individual components of a build image are provided by both the hosted build platform and tenant, and include the build executor, platform-provided pre-installed guest OS and packages, and the tenant’s build steps and external parameters.
```
deeglaze, line 152: This definition of build image as including parts from a tenant seems contrary to prior uses of the term to mean specifically the image of the build environment.

I would advise checking prior instances of "build image" and disambiguate them with "build environment image" and "tenant build image". A build platform will attest to itself, but it may also mount a container including a tenant's build toolchain for executing a build request, so that's an important distinction to make.

docs/spec/v1.1/terminology.md
```diff
@@ -125,6 +125,39 @@ of build types](/provenance/v1#index-of-build-types).
...
+| Build executor | The platform-provided program dedicated to executing the tenant’s build definition, i.e., running the build, within the build image.
```
deeglaze, line 153: Clarify that the build executor must be a measured component of the build environment image.

docs/spec/v1.1/terminology.md
```diff
@@ -125,6 +125,39 @@ of build types](/provenance/v1#index-of-build-types).
...
+| Build request | The process of assigning a build to a pre-provisioned build environment on a hosted build process.
```
deeglaze, line 154: I always think of requests as messages. Perhaps,

"A user-provided message to the build platform that is used to assign a build to a pre-provisioned build environment on a hosted build process."

But this could be splitting hairs.

deeglaze (changes requested):

marcelamelara:
@mdwood-intel @pdxjohnny PTAL

docs/spec/v1.1/requirements.md
```diff
@@ -32,35 +32,38 @@ build level, implementing any controls as specified by the chosen platform.
   <th>Implementer
   <th>Requirement
   <th>Degree
-  <th>L1<th>L2<th>L3
+  <th>L1<th>L2<th>L3<th>L4
```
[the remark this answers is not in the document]
marcelamelara: This isn't an html style I've seen before either, but this is used through all the docs, apparently, and renders well.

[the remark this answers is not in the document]
lehors: The end tag of many HTML elements is optional, that is the case of the `th`, `td`, and `tr` elements. There is nothing wrong with this. It is very common in HTML written by hand because it saves a lot of typing and possible errors.

https://html.spec.whatwg.org/#optional-tags

docs/spec/v1.1/requirements.md
```diff
@@ -330,9 +335,77 @@ means that the build ran with no network access. Such a requirement requires
 substantial changes to both the build platform and each individual build, and is
 considered in the [future directions](future-directions.md).
 
-<td> <td> <td>✓
+<td> <td> <td>✓<td>✓
+<tr id="hardware-attested">
+<td>Hardware-Attested
+<td>
+
+The build platform generated an authenticated attestation to the integrity
```
joshuagl, line 343: That is, generating an _in-toto_ attestation about the entire initial state of the build environment?

We may run into confusion here between attestation and attestation:

> The process of vouching for the accuracy of information.

from [TCG Glossary](https://trustedcomputinggroup.org/wp-content/uploads/TCG-Glossary-V1.1-Rev-1.0.pdf), and

> An authenticated statement (metadata) about a software artifact or collection of software artifacts.

from [SLSA Terminology](https://slsa.dev/spec/v1.0/terminology)

It might be worth adding a note to clarify to readers who have familiarity with both terms.

docs/spec/v1.1/requirements.md
```diff
@@ -330,9 +335,77 @@ means that the build ran with no network access. Such a requirement requires
...
+of the entire initial state of the build environment (i.e., VM/container
+image, kernel, and filesystem) was generated at creation time and verified
+at deployment time. The build platform also attested to the build request.
+In other words, tampering with the initial state of the build environment
+MUST be detectable by the platform itself and the build.
```
joshuagl, line 348: "an the build."... executor? agent?

docs/spec/v1.1/terminology.md
```diff
@@ -125,6 +125,39 @@ of build types](/provenance/v1#index-of-build-types).
...
+    build process. For SLSA Build L4, the build platform outputs provenance
+    describing this process.
+2.  The hosted build platform deploys a build image on a compute platform to
+    provision a new build environment. For SLSA Build L4, the
+    build platform validates the *measurement* of the environment's
+    *boot process*.
+3.  When a new *build request* is made, the platform assigns the request to
+    a deployed build environment. For SLSA Build L4, the tenant may validate
+    the measurement of the build environment.
+4.  Finally, the *build executior* running within the environment executes
```
joshuagl, line 147:
```suggestion
4.  Finally, the *build executor* running within the environment executes
```

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,74 @@ All of [Build L2], plus:
...
+In practice, this means that builds are configured to run on a build platform
+that supports hardware capabilities for attesting to the state of the
+compute environment executing builds.
+
+<dt>Intended for<dd>
+
+Software releases needing assurances about the integrity of the environment
+used to create the release (e.g., specific compute platform, pre-build
+tamper detection).
+Build L4 usually requires significant changes to existing build platforms.
```
joshuagl, line 255: How about:
```suggestion
Build L4 may require significant changes to existing build platforms.
```
and then list some of the requirements?

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
mkulke: In a TEE featuring a vTPM there should still be a hardware root of trust. The vTPM would be linked to signed HW evidence, for by example by mixing a hash of the vTPMs AKpub into the report-data field of a TEE's attestation report.

So, the original statement sounds fine to me. I assume a HW RoT is key and not something that can be supplanted by sw.

docs/spec/v1.1/requirements.md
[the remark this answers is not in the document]
pdxjohnny: Hibernation and other things probably are affected by bootability of an image. It would be good to ensure container terminology and vm terminology are distinct as it loos like both cases are being captured: https://github.com/slsa-framework/slsa/pull/1051/files#r1612031706

docs/spec/v1.1/levels.md
TomHennen: I'm sympathetic to view that we can just repeat the language from L3 (unless we also want to rewrite the L3 'intended for' section).  Maybe with the caveat "significant changes to existing _L3_ builds platforms"?

I think the requirements below do a fine job of getting into the details.

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
TomHennen: I do wonder if there's a useful middle ground where vTPMs that aren't hardware backed can still provide meaningful value.  If this level requires full hardware backed support would that be too big a step?

docs/spec/v1.1/levels.md
TomHennen: I see the importance of capturing the nuance.  I think what Mark might be getting at is that the benefits as currently enumerated are a bit abstract.

E.g. "Greatly reduces trust in a hosted build platform by increasing observability into the level of integrity of the build environment."

Could this instead be something like "Greatly reduces the TCB of a hosted build platform by preventing tampering with the build execution environment, leaving only the OS and vTPM in scope" or something like that?

TomHennen:
Have we considered that there may be other ways, like reproducible builds, to remove trust in the hosted build platform?  Would that provide equivalent (possibly better?) guarantees? 

Does encoding hardware trust as a discrete build level remove that option from folks?  Perhaps reproducible builds could listed as _a_ way to achieve these goals?  (It would require a fair amount of re-wording...)

docs/spec/v1.1/levels.md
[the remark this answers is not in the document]
lumjjb: I think another potential middle ground is a HSM where authentication requires certain authentication requirements (whether secure/measured boot, validating PCRs, some validated token, etc.)?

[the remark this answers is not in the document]
mkulke: In this case I'd explicitly state here that "software vTPMs" are valid too. sth like: 

> All attestations MUST be authenticated a root of trust. A valid of root of trust might be anchored in a dedicated HW module or might be formed via a chain of cryptographic signatures. 

The latter would be referring to vTPMs that are provisioned by a cloud service provider on tenant VMs and for which you'd get a CSP-provided cert-chain to verify.

docs/spec/v1.1/levels.md
```diff
@@ -231,14 +232,74 @@ All of [Build L2], plus:
...
+
+-   Software producer:
+    -   MUST run builds on a hosted build platform that meets Build L4
+        requirements.
+    -   SHOULD verify the [build environment attestations] at the start of a
+        build and produce a [verification summary] about the check.
+
+-   Build platform:
+    -   MUST generate and distribute attestations to good known integrity
+        measurements of the entire initial state of the build environment
```
EricBurnett, line 269: I'm worried that this definition **precludes any sort of distributed build from ever meeting SLSA L4** - it seems to assume that a build is always hermetic within a single build environment, and so easily described as `host + VM + filesystem + command -> output`. Almost all of the high-trust builds produced by Google, for example, are produced via distributed systems across multiple hosts, where the network, virtual filesystems, etc are inextricably linked to the process. Many could not be produced on any single machine in any reasonable amount of time, if at all. (Needing many TiBs of scratch space in total, many core-days of work, etc.)

I'd have no objection to defining a SLSA framework for hermetic single-host builds, if that is independently useful, but for "L4" where the implicit assumption is that all high-trust builders should aim to achieve it, I'm not yet convinced that this is the correct successor to L3. Though even here, "trust of software" seems still necessary, in that I don't know how to go from a description of a build to a fully realized set of input files without _already_ having to have trusted that some build platform software has interpreted it correctly.

Perhaps the right target is a broadened version of the same: a chain of hardware-rooted build attestations to what software was running on each node leveraged in the build process, with that software in turn attesting what environment it set up locally, etc? This would make it easier to be "tamper evident" for hardware attacks, though still require trusting that the build platform software is sufficiently bug- and backdoor-free.

Failing that, @TomHennen 's comment elsewhere about "independently reproducible" might be a better choice for L4. Though that may also fail on practicality grounds: I'm not sure of many systems that have ever achieved multi-party bitwise reproducibility (I don't think [Debian](https://wiki.debian.org/ReproducibleBuilds) is there yet?), and I don't know what that would look like for closed-source systems that only execute on one build platform in practice.

marcelamelara:
I need to catch up to most comments, but want to respond to a few recent ones first:

> Have we considered that there may be other ways, like reproducible builds, to remove trust in the hosted build platform? Would that provide equivalent (possibly better?) guarantees?

This question keeps coming up, and I've been thinking about this a LOT recently. Our stance so far has been that reproducible builds and (hardware-)attested builds are not mutually exclusive, nor is one better than the other. Both approaches reduce trust in the build platform, but they do so in different ways. My thinking is because they're fundamentally solving different problems. 

The similarities: Reproducible builds gives a producer or consumer assurances that a specific build process is consistent and was _very likely_ not tampered with because you can compare the build results. Relying on trusted hardware can give all parties cryptographically verifiable evidence that the build environment that ran your build wasn't tampered with because you can check the hardware-backed attestations for it. 

The differences: The outcome of using reproducible builds and hardware-attested builds ends up being pretty much the same: "was my build tampered with?" But I see a few fundamental differences in the trust model. In the former, the build platform at an organizational level doesn't need to be fully trusted. The more distinct, independent rebuilders you use, the stronger your assurances are, assuming enough rebuilders aren't colluding.

In the latter, the build platform _system_ doesn't need to be fully trusted. As long as you trust the hardware-based mechanisms, your assurances come from being able to trace a particular build back to a specific compute environment and check the contents of the attestations against the expected good known integrity values for that software.

> Does encoding hardware trust as a discrete build level remove that option from folks? Perhaps reproducible builds could listed as a way to achieve these goals? (It would require a fair amount of re-wording...)

As I describe above, I think the answer here should be no, because they aren't quite solving the same problem. In fact, we're intentionally vague about reproducible builds because we think there are scenarios where one approach may be enough depending on your threat model and/or adoption requirements (as this https://github.com/slsa-framework/slsa/pull/1051#discussion_r1663188034 mentions at the end), and there are scenarios where both may be useful, and we wanted to leave the option open for either scenario.

docs/spec/v1.1/requirements.md
```diff
@@ -330,9 +335,77 @@ means that the build ran with no network access. Such a requirement requires
...
+The build platform MUST guarantee the following:
+
+-   When creating a new build environment:
+    -   The integrity of the build image (i.e., VM or container) MUST be
+        authenticated and verifiable. That is, SLSA Build L3+ Provenance for
+        the build image MUST be generated and distributed to allow for
+        independent verification.
+    -   The boot process of each build environment MUST be measured and
+        attested using a [TCG-compliant measured boot] mechanism. In
+        addition, the initial state of the build environment's disk image
```
deeglaze, line 359: There are reasons for the disk image to not be entirely measured, so this seems like it could be too coarse of a requirement. There are ways security-critical components of user space are measured. They go past the PCRs 0-7 in the TCG-compliant measured boot link.

This whole disk integrity would disallow content-addressed build caches that already have SLSA attestations, such as what Bazel.build uses for performance.

The fact of the matter is that we have bespoke ways of evaluating attestation evidence and specifying policy for their acceptance. There is no standard since OSes consider different things to be security-critical.

Perhaps, "every file read from disk must be integrity-protected by some means traceable to the build environment identity or build request"

TomHennen:
> As I describe above, I think the answer here should be no, because they aren't quite solving the same problem. In fact, we're intentionally vague about reproducible builds because we think there are scenarios where one approach may be enough depending on your threat model and/or adoption requirements (as this [#1051 (comment)](https://github.com/slsa-framework/slsa/pull/1051#discussion_r1663188034) mentions at the end), and there are scenarios where both may be useful, and we wanted to leave the option open for either scenario. How this is reflected in build levels I think is a separate question.

Ah, I guess how to reflect it in the build levels is what I'm most interested in at the moment. :)

The way I'd read defining L4 as requiring hardware attested builds is that it would preclude someone from achieving a similar goal via reproducibility.  While we could then define level 5 as reproducible that would seem to require hardware backed attestations when doing reproducible builds.  Is that desirable?  Is my reading too strict?

The way I think about levels generally is that they're additive and single tracked.  What we seem to have here is a branch of sorts?  I don't know how to handle that within the levels framework.

I could see handling it within a specific level "you have choices on how to can reduce trust in the build platform a) hardware backed attestations, b) reproducible builds, c) both a & b".

To be clear I _do_ see the distinction in how these two approaches reduce the need to trust the build platform; but I wonder how meaningful that distinction would be for downstream users?  Maybe it's a distinction that could be made in whatever attestations are used to communicate level 4-ness?

docs/spec/v1.1/levels.md
bobbiec: >Perhaps the right target is a broadened version of the same: a chain of hardware-rooted build attestations to what software was running on each node leveraged in the build process, with that software in turn attesting what environment it set up locally, etc?

@EricBurnett I agree with this target (and I think it's consistent with the spirit - the build environment is potentially distributed). It would be a small change in the next line to make it more explicit -

```diff
- (i.e., VM/container image, boot process and filesystem)
+ (i.e., VM/container image, boot process, and filesystem, for each host in the build environment)
```

docs/spec/v1.1/terminology.md
marcelamelara: Fair point. This isn't really about the exchanged messages, but about the action of assigning and dispatching a tenant's build process to a pre-deployed build environment. More recently, I've been using the term "build environment dispatch" for this step in the build environment's lifecycle, which is hopefully a little clearer.

docs/spec/v1.1/requirements.md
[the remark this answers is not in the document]
marcelamelara: Let's move this discussion to #1107

docs/spec/v1.1/requirements.md
marcelamelara: I'll be closing this PR soon. Let's move this discussion to #1107 !

marcelamelara:
Per the [7/22 SLSA spec meeting](https://docs.google.com/document/d/1PwhekVB1iDpcgCQRNVN_aesoVdOiTruoebCs896aGxw/edit#heading=h.a7pe3nsqmnuj), these requirements are to be re-worked as a separate track. So, this PR is being superseded by #1115 .

docs/spec/v1.1/levels.md
marcelamelara: @deeglaze Can you please clarify, when you say non-VM, do you mean that a build could be running inside a container (backed by a physical TPM), or a bare metal environment, or either?

deeglaze: I mean a task that is launched by a trusted orchestration daemon that establishes some resource limitations and filesystem access restrictions as Linux allows, but not necessarily following an OCI container description. At Google we have build servers that only run builds but still run on Borg. All the node software is accounted for with Titan TPMs. There’s no strong reason for us to require builds to run in VMs since we have everything measured and can account for the machine access by the identity and access management system and (measured) software-enforced ACLs. The build environment is managed by this production identity system before getting to the build job that then measures the inputs for slsa L3 but without incorporating the measurement into pcrs.

The hardware attestation of prodID and software attestation by borg Id (BCID) protect our build environment integrity in an auditable manner, but without incorporating the entirety of our production ecosystem in the build attestation. The new slsa level should allow for ecosystem measurements to be held back if the operational security and physical security of the servers can ve attested to implicitly with the slsa signing key release mechanism.

Now if you’re saying that the build ecosystem integrity is only a means to an end of requiring the build to be run within confidential computing technologies, I’d say that is not the most important goal to reach. If you want the build ecosystem to satisfy holistic properties like “no humans may access the environment” then you have to talk about more security commitments of the software, not just its measurement.

marcelamelara: Thanks for the detailed explanation!

PR #1055, opened 2024-05-13
impl: refactor diagram in figma (no visual change)

MarkLodato:
Refactor how the diagrams are implemented in Figma so that it's easier
to edit.

There's no visual change. Two SVGs are completely unchanged, and two are
modified but only have lines reordered AFAICT. The two modified SVGs
seem to be a pixel off, for a reason I cannot comprehend.

Explanation: Previously the diagrams were built "up", starting with the
simplest and successively adding more detail. This was a pain to edit,
because you'd first have to find which diagram was the "master" for the
change you wanted to make. Now there's only one "master" with all of the
detail, and the derived diagrams just hide detail that's not wanted.
This is much easier to edit (and understand!)

Also delete some unused diagrams that were in the figma file but never
exported to SVG.


lehors (approved):
LGTM

PR #1057, opened 2024-05-13
content: refactor threat diagram and add overview

MarkLodato:
**Preview: https://deploy-preview-1057--slsa.netlify.app/spec/v1.1/threats**

Refactor the threat diagram to address clarity concerns and to expand it beyond tampering. The intent is that this threat model can be useful for any software supply chain security effort. Many of the threats previously called "out of scope" should be listed here, even if the SLSA ladder does not (yet!) cover them.

NOTE: this PR is a partial solution, but more work is needed. I want to at least merge what we have so far so that others can iterate on it.

This design is the result of much discussion on Slack. Thank you to @adityasaky, @arewm, @david-a-wheeler, @jkjell, @mlieberman85, @marcelamelara, and @trishankatdatadog for your contributions and suggestions.

Summary of major changes:

* Add threat indicators for Producer and Consumer, remove for Dependencies.
* Rename "Package" to "Distribution".
* Rewrite titles to describe the position rather than the tampering threat.

Detailed diagram changes:

* Update the threat markers:
  - Add a threat for Producer to cover malicious intent. For example, if you install malware, it's not tampering---the producer really did intend to write malware, and no amount of code review will "fix" that! (Previously this was called "out of scope".)
  - Add a threat for Consumer to cover...? It makes the diagram nicer and I assume we want something here, but I don't know what that is yet!
  - Remove the threat for Dependency, since it generated a lot of confusion. The intent was that it is recursive, so the hope is that the diagram and text make this clear enough.
  - Re-letter the threat markers accordingly.
  - Update the threat titles to describe the position rather than the tampering threat. The old titles generated a lot of disagreement, and they relied on understanding the non-obvious interpretation of the model. Now, the new titles just describe that model, which is hopefully more clear to everyone.
* Rename "Package" to "Distribution" to better reflect what that box means.
* Move the arrow from Distribution to Dependency to make it clear that (H) is also recursive.
  - Also make the arrow solid instead of dashed, but keep the dashed box around Dependency. The idea is that the use of the dependency is a real input to the build, while Dependency itself is really just another package.
* Move "build params" to "Build threats" instead of "Source threats", and add new "Usage threats" category.
* (minor) Use the same green color throughout, rather than having a slightly different green color for arrows.

Text changes:

* Add an overview section that explains a bit more about the threat model.
* Add Dependency Confusion.
* Update text to match the new diagram (only partially done).


adityasaky:
I prefer the second image, just so it's clear that Z applies to all steps. I second some of the points that came up in today's SLSA spec meeting about Z, FWIW.

Transcribing some comments from Slack:

> > Me: F seems too specific in its naming; you could argue that a compromise of a package upload action to insert another artifact is technically a direct package upload, but those words to me specifically call out something like a developer running twine upload to submit something to PyPI.

> @MarkLodato: Regarding F, yeah, I'd like a better name. Maybe "Misuse of upload credentials" or something like that? My intent was to call out the use upload API to upload arbitrary things, as opposed to tampering with a legitimate build (E) or, if a SLSA policy is in place, tampering with the inputs to build (D).

I think this is better!

>> Me: nit: I also wonder about consistency between F and H with "package" vs "artifact"

> @MarkLodato: Regarding "package" vs "artifact", perhaps it's too subtle. My thinking was that in (H) someone selects a "package identifier" and then receives back an "artifact" (blob of data). But I agree that maybe it's just too confusing

I think this is a good point about the subtlety, but to me that goes back to the confusion we discussed last week about whether the wrong artifact being returned is actually part of G, since this considers the "processing of request for package" operation as well. Maybe I'm overthinking this specific connection though.

MarkLodato:
I updated the [comment](https://github.com/slsa-framework/slsa/pull/1057#issuecomment-2117546134) above with an Option 3 incorporating some feedback from today's meeting.

It's also worth considering Option 0: not adding Z at all. I'd still like to somehow surface the risk of missing visibility, so we'd need an alternate proposal if we don't add a label on the diagram.

MarkLodato:
I think I prefer the new line coming out of the right of Package, since it more clearly indicates that (H) is in scope for dependencies. But now it makes it seem like (I) is not in scope, whereas it really is.

One option for that would be to merge dependency confusion back into (H) like we had before because it makes the diagram nicer, but I don't like that because it's useful to talk about dependency confusion separately. And I don't want to move dependency confusion to J because it happens to consumers too, e.g. `pip install some-typo-package`.

arewm:
Can you merge `Consumer` and `Dependencies` into one if we rename it something like `Consumption`? That makes the explicit connection between the two and we can clarify in text that consumption can be either the final artifact or a dependency input.

MarkLodato:
More diagrams, for the record. (These were all posted in Slack.)

#### Option 4: Add Producer (A) and Consumer (I), remove Dependency (J) and Visibility (Z)

The thinking is that J is confusing since it's really just a manifestation of all the other threats, so we can just visually show with the loop in the diagram. Also Z is confusing since it's not like hte others, so let's just omit it.

supply-chain-threats_no-key_no-JZ

#### Option 5: Same as above, without I

supply-chain-threats_no-key_no-JZ_no-I

#### Option 6: Arrow coming out of consumer

This is from @arewm. The arrow coming out of Consumer is to indicate that the build is just one type of consumer.

supply-chain-threats_arrow-from-consumer

Here's a variant with dashed boxes:

supply-chain-threats_box-dotted-con-to-bld

And now no box on Dependencies:

supply-chain-threats_dotted-con-to-bld

#### Option 7: Same as Option 5 but with describing positions rather than threats

In Slack, Option 5 seemed to be the most popular (or at least I liked it the best!) but there was a lot of disagreement about what the labels should be. Rather than trying to describe threats at each point, we could instead just describe what the point means. This way we can talk about this in the model, and then talk about higher level attacks and point to the locations where they happened. This seemed to get at least some positive feedback.

supply-chain-threats_descriptive

MarkLodato:
OK, I think this is ready for review. As I wrote in the top comment, it's not complete, but I'm hoping that we can at least agree on an intermediate state and merge it with TODOs, so that we can iterate.

mlieberman85 (approved):
LGTM. I agree that generally this is in a reasonable state and we can iterate as needed.

adityasaky (approved):
Not a SLSA maintainer, but LGTM to merge, agree with iterating on TODOs rather than blocking! Thanks for working / coordinating this @MarkLodato!

lehors (approved):
It seems to me that we probably should change Source to Source Control or something like that, but I agree this is an improvement anyway and I'm happy for this to be merged as is.

MarkLodato:
Thanks all. I'll wait for input from a few more folks, since this is a significant change in the main diagram that is used in a lot of places.

Note: I updated the "verification" diagram as well, since H is now out of scope.

docs/spec/v1.1/threats.md
```diff
@@ -24,17 +27,28 @@ that somehow. This isn't specific to dependencies - it applies to the entire
...
-
 <article class="threats">
 
+## Overview
+
 ![Supply Chain Threats](images/supply-chain-threats.svg)
 
-See [Terminology](terminology.md) for an explanation of supply chain model.
+This threat model covers the *software supply chain*, meaning the process by
+which software is produced and consumed. We describe and clusters threats based
```
joshuagl, line 37:
```suggestion
which software is produced and consumed. We describe and cluster threats based
```

docs/spec/v1.1/threats.md
```diff
@@ -24,17 +27,28 @@ that somehow. This isn't specific to dependencies - it applies to the entire
...
+through (I). This is useful because priorities and mitigations mostly cluster
+along those same lines. Keep in mind that dependencies are
+[highly recursive](#dependency-threats), so each dependency has its own threats
+(A) through (I), and the same for *their* dependencies, and so on. For a more
+detailed explanation of the supply chain model, see
+[Terminology](terminology.md).
+
+Importantly, producers and consumers face *aggregate* risk across all of the
+software they produce and consume, respectively. Many organizations produce
+and/or consumer thousands of software packages, both first- and third-party, and
```
joshuagl, line 48:
```suggestion
and/or consume thousands of software packages, both first- and third-party, and
```

docs/spec/v1.1/threats.md
```diff
@@ -48,12 +62,40 @@ SLSA v1.0 does not address source threats, but we anticipate doing so in a
...
+
+<details><summary>Software producer intentionally submits bad code</summary>
+
+*Threat:* Software producer intentionally submits "bad" code, following all
+proper processes.
+
+*Mitigation:* **TODO**
+
+*Example:* A popular extension author sells the rights to a new owner, who then
+modifies the code to secretly mine bitcoin at the users' expense. SLSA does not
```
joshuagl, line 79:
```suggestion
modifies the code to secretly mine Bitcoin at the users' expense. SLSA does not
```
or 'cryptocurrency' ?

docs/spec/v1.1/threats.md
```diff
@@ -785,14 +826,28 @@ lacks a positive attestation showing that some system, such as GitHub, ensured
...
 *Mitigation:* **Outside the scope of SLSA.** That said, some solutions to
 support hermetic and reproducible builds may also reduce the impact of this
 threat.
 
+</details>
+<details><summary>De-list artifact</summary>
+
+*Threat:* The package registry stops serving the artifact.
+
+*Mitigation:* N/A - This threat is out of scope of SLSA v1.0.
```
joshuagl, line 842: Should this be "TODO" like other threats above which are not yet covered by SLSA?

docs/spec/v1.1/threats.md
```diff
@@ -785,14 +826,28 @@ lacks a positive attestation showing that some system, such as GitHub, ensured
...
+
+</details>
+<details><summary>De-list provenance</summary>
+
+*Threat:* The package registry stops serving the provenance.
+
+*Mitigation:* N/A - This threat is out of scope of SLSA v1.0.
```
joshuagl, line 849: Same here, should this be TODO?

joshuagl (approved):
This is excellent, thanks! I really like the major changes and fully agree with them. The minor changes look like an excellent start, too.

I made some minor comments, but this LGTM.

docs/spec/v1.1/threats.md
MarkLodato: went with crypocurrency

docs/spec/v1.1/threats.md
MarkLodato: There are several items listing "out of scope". Instead of addressing them all in this PR, which would be substantial, I instead added a TODO at the top to do that in a future PR.

docs/spec/v1.1/threats.md
MarkLodato: same, see above

joshuagl (approved):

docs/spec/v1.1/threats.md
```diff
@@ -303,10 +337,13 @@ management system to bypass controls.
...
 
-### (C) Build from modified source
+### (D) External build parameters
+
+**TODO:** Move under "Build threats".
 
 An adversary builds from a version of the source code that does not match the
-official source control repository.
+official source control repository, or uses parameters to the build that inject
+behavior that was not intended by the official source.
```
marcelamelara, line 346: Can you give an example of what you mean by "parameters that inject unintended behavior"? For Build L2+ builds, it seems to me that this could happen via (B) if changes to the build definition are introduced. Or do you mean something else here?

docs/spec/v1.1/threats.md
```diff
@@ -662,9 +699,14 @@ controls are in place to prevent abusing admin privileges.
 
 </details>
 
-### (F) Upload modified package
+### (F) Artifact publication
 
-An adversary uploads a package not built from the proper build process.
+An adversary uploads a package artifact that does not reflect the intent of the
+package's official source control repository.
```
marcelamelara, line 705: How are we defining the "official" source control repository? I assume this is the known good source repository that is the "trusted" origin, but I think there's an underlying assumption here that the identity of the "official" repo is somehow apparent or otherwise communicated out of band. I don't think addressing this assumption needs to happen as part of this PR, but this term is used in several other places so I think it would be helpful to make these assumptions more explicit, and describe how they relate to the threat model.

marcelamelara (commented):
Thanks @MarkLodato ! I have a few clarifying comments/questions, but I definitely think this is almost ready.

docs/spec/v1.1/threats.md
```diff
@@ -204,22 +252,8 @@ does not accept this because the version X is not considered reviewed.
 
 </details>
 
-#### (A3) Code review bypasses that are out of scope of SLSA
-
-<details><summary>Software producer intentionally submits bad code</summary>
-
-*Threat:* Software producer intentionally submits "bad" code, following all
-proper processes.
+#### (A3) Render code review ineffective
```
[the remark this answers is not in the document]
MarkLodato: Thanks for catching, fixed.

docs/spec/v1.1/threats.md
```diff
@@ -48,12 +68,40 @@ SLSA v1.0 does not address source threats, but we anticipate doing so in a
...
+*Mitigation:* **TODO**
+
+*Example:* A popular extension author sells the rights to a new owner, who then
+modifies the code to secretly mine cryptocurrency at the users' expense. SLSA
+does not protect against this, though if the extension were open source, regular
+auditing may discourage this from happening.
+
+</details>
+
+**TODO:** More producer threats?
```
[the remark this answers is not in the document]
MarkLodato: Thanks, added as a TODO. (Needs better wording, which is why I left it as a TODO rather than a full item.)

docs/spec/v1.1/threats.md
MarkLodato: See "Build from unofficial parameters" below. Does that help? If so, do you suggest any changes to clarify?

docs/spec/v1.1/threats.md
MarkLodato: Good point, added a TODO.

docs/spec/v1.1/threats.md
```diff
@@ -48,12 +68,40 @@ SLSA v1.0 does not address source threats, but we anticipate doing so in a
...
 
 An adversary introduces a change through the official source control management
 interface without any special administrator privileges.
 
-#### (A1) Submit change without review
+The threats in this category are theoretically mitigated by code review or some
+other quality controls. Contrast this with (A), where such controls are
+ineffective.
+
+**TODO:** Is the split between (A) and (B) clear and valuable?
```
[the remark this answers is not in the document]
MarkLodato: That was not my intent, though I think your description is more in line with what @david-a-wheeler was saying. So we need to align on the model.

I rewrote the paragraph above to hopefully better explain the difference, and add a corresponding paragraph to (A). Does that help?

The reason why I think this is a useful split is because it's grouping by mitigation. In (A) code review won't help, in (B) it will to some degree.

docs/spec/v1.1/threats.md
```diff
@@ -88,16 +91,17 @@ auditing may discourage this from happening.
...
 ### (B) Authoring & reviewing
 
 An adversary introduces a change through the official source control management
 interface without any special administrator privileges.
 
-The threats in this category are theoretically mitigated by code review or some
-other quality controls. Contrast this with (A), where such controls are
-ineffective.
+Threats in this category *can* in theory be mitigated by code review or some
+other quality controls during teh authoring/reviewing process. Contrast this
```
marcelamelara: tiny typo
```suggestion
other quality controls during the authoring/reviewing process. Contrast this
```

docs/spec/v1.1/threats.md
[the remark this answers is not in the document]
marcelamelara: Yes, I think the distinction between (A) and (B) is a little clearer now, thanks! 

The part about the producer practices in (A) is still a bit unclear, though. The [first paragraph](https://github.com/slsa-framework/slsa/pull/1057/files/3a05a18bec31c6c589273f4b42b555945cc46011..c3f63de88420b52ea12ff2fe6bdcb20233845680#diff-8cbad6b869c6a7ad512ed0142e55b3abe92725ebe1507882451f3b83fbbeaf28R74) says "or the producer otherwise uses practices that are not deserving of the consumer's trust", while the [threat description](https://github.com/slsa-framework/slsa/pull/1057/files/3a05a18bec31c6c589273f4b42b555945cc46011..c3f63de88420b52ea12ff2fe6bdcb20233845680#diff-8cbad6b869c6a7ad512ed0142e55b3abe92725ebe1507882451f3b83fbbeaf28R82) below says "Software producer intentionally submits "bad" code, following all proper processes". I think this is why I thought earlier that (A) was about bad first-party code, but it sounds like (A) rather covers producer threats that aren't related to code changes. In that case, I can see compromised producer credentials fitting as a threat category under (A).

That said, I'm fine with merging the current iteration of threats (A) and (B) with possible TODOs for further details and/or examples on the distinction between the two.

marcelamelara (approved):
Thanks for all your work on this @MarkLodato . This lays a good foundation that we can keep refining.

docs/spec/v1.1/threats.md
marcelamelara: Ignore this. I didn't see this typo was already fixed.

docs/spec/v1.1/threats.md
marcelamelara: I see, that does help, thanks. Here's a minor change to the wording that I think clarifies the description.

```suggestion
official source control repository, or changes the build parameters to inject
behavior that was not intended by the official source.
```

docs/spec/v1.1/threats.md
```diff
@@ -662,9 +703,21 @@ controls are in place to prevent abusing admin privileges.
...
+An adversary uploads a package artifact that does not reflect the intent of the
+package's official source control repository.
+
+This is the most direct threat because it is the easiest to pull off. If there
+are no mitigations for this threat, then (D) and (E) are often indistinguishable
+from this threat.
+
+**TODO:** We need to define "official source control repository". Its meaning is
+not obvious. The gist is that each package theoretically has some "official" or
+"canonical" repository from which is "should" be built, and the attack here is
```
marcelamelara, line 717: tiny typo
```suggestion
"canonical" repository from which it "should" be built, and the attack here is
```

docs/spec/v1.1/threats.md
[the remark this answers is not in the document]
MarkLodato: Thanks. I agree it's still a bit fuzzy. I moved the TODO from (B) up to (A) and expanded it a bit to talk about compromised developer credentials, which I agree is unclear where it should go.

I'll submit now and we can hopefully improve over time :)

Thanks for the help!

Issue #1072, opened 2024-06-17
Clarify source-track objective


TomHennen:
(The "Threat A" url references a deleted ref.)

It seems to me that the source track (similar to the build track) should start with generating provenance at the lower levels, then move to implementation of specific policies at higher levels. 
The revision of a consumed repository is a kind of artifact like any other software dependency and as such should come with all its paperwork. The "build process" of a repository revision is the pull request, and the result of a completed pull request should be a new revision accompanied by provenance claims. 

This example appears to be covering the requirements and benefits of a specific policy, similar to the ["Build Level 1" dedicated infrastructure requirement](https://slsa.dev/spec/v1.0/levels#build-l1). 
> Concretely: an attacker must compromise the accounts of two organization members to publish code in a Source Level 3-conformant repository, and the evidence of those unauthorized changes cannot be destroyed without further attacks.

At a minimum, consumers of a repository revision will want to know the following provenance claims: 
1. Who contributed these changes (according to the SCP)? 
2. When did they do so (according to the SCP)? 
3. What policy compliance claims are associated with this revision (according to the SCP)? 

Based on this, it seems to me that a focused adaptation of the excellent ["build track" summary](https://slsa.dev/spec/v1.0/levels#build-track) would work well here. EG:

The SLSA **SOURCE** track describes increasing levels of trustworthiness and completeness in a **repository revision's** provenance. Provenance describes what **Source Control Provider produced the revision**, what process was used, and who the contributors were. The lowest level only requires the provenance to exist, while higher levels provide increasing protection against tampering of the **version control system, the provenance, or the revision**.

The primary purpose of the **source** track is to enable [verification](https://slsa.dev/spec/v1.0/verifying-artifacts) that the **revision followed the expected process**. Consumers have some way of knowing what the expected provenance should look like for a given package and then compare each **source revision's** actual provenance to those expectations. Doing so prevents several classes of [supply chain threats](https://slsa.dev/spec/v1.0/threats).

_Originally posted by @zachariahcox in https://github.com/slsa-framework/slsa/pull/1037#discussion_r1598593653_

TomHennen:
[...]

Issue #758, opened 2023-03-28
Consistency: "attestation" vs "provenance attestation" vs "provenance"


MarkLodato:
We use "provenance" vs "provenance attestation" inconsistently. We also sometimes use a bare "attestation". We should be consistent throughout the spec.

arewm:
An example of the lack of clarity is that the [attestation model](https://slsa.dev/attestation-model) seems to indicate that provenance is part of the full attestation (i.e. its predicate) while the [provenance](https://slsa.dev/provenance/v1) page indicates that Provenance is an attestation.

It may be that provenance is an attestation as in it is a collection of facts, but this terminology might be confusing when compared with the attestation model of software attestations which indicates that it is an authenticated statement. This indicates that provenance may not be an attestation at build L1, but it must be an attestation at build L2 and higher. 

Can we be more precise in the usage of provenance and attestation? Do we need to introduce a new term to help distinguish between the two? Can we just use provenance generally and then clarify it with an attested provenance (i.e. attestation) where that detail is critical? Is attested provenance any better than a provenance attestation?

MarkLodato:
In my mind, provenance is a type of attestation. Usually when we say "provenance" it's implicitly "provenance attestation". The word "attestation" focuses on its form (a signed statement) while "provenance" focuses on its content (metadata about where an artifact came from).

arewm:
I was looking into creating a PR for this and I stopped once I looked at the schema -- https://slsa.dev/provenance/v1#schema

```
    // Standard attestation fields:
    "_type": "https://in-toto.io/Statement/v1",
    "subject": [...],
    // Predicate:
    "predicateType": "https://slsa.dev/provenance/v1?draft",
    "predicate": {
```

The in-toto attestation typed to the SLSA provenance does not have an envelope or bundle, just the statement and predicate. 

I see what you are saying, @MarkLodato . The provenance is an attestation as it conforms to the in-toto attestation format.

Instead of differentiating on provenance vs. attestation, can we differentiate on provenance/provenance attestation vs. authenticated (or maybe signed) attestation/provenance attestation?

While we can try to clarify throughout, I think we should at least address the different terms specifically in Terminology and FAQs.

PR #1083, opened 2024-06-28
content: draft: define source-track objective in terms of revisions and provenance. 

zachariahcox:
fixes https://github.com/slsa-framework/slsa/issues/1072

This PR modifies _draft_ content of the slsa spec. 

## Context
Based on discussion from https://github.com/slsa-framework/slsa/pull/1037

See [discussion here](https://docs.google.com/document/d/13Xt8mA_2b00McGX2vkyhu4GQdFAqtXPu7YXE8ZA6ISE/edit?resourcekey=0-EqfHF79tUWAKp4PzsE3z1A#heading=h.svjr333bawb). 

Copied from [draft proposal here](https://docs.google.com/document/d/13Xt8mA_2b00McGX2vkyhu4GQdFAqtXPu7YXE8ZA6ISE/edit?resourcekey=0-EqfHF79tUWAKp4PzsE3z1A#bookmark=id.4qr65cfy6ufj). 

Google document requires slsa-discussion@googlegroups.com membership.

## Source revision provenance
Repos contain many revisions, most of which are not "official" or otherwise approved for release.  
The goal of the source track is to attest to why a specific revision _was_ approved for release. 

We can think of the SCP / code review tool as “building” the next official revision of a repository using a codified process that involves collecting commits, acquiring reviews, running CI, etc. 
If the change review process is successful, the code review tooling will merge the code changes and attest to the process used to produce the new revision. 

The source provenance attestations associate a specific revision of a repository to security claims and documents (basically build logs) of the process that produced it.

In GitHub terms, a merged pull request and its associated rules evaluation justify why and how a specific git SHA is reachable from a protected branch. 

## Example Scenario
1. A CI system is trying to build some artifact and will download all necessary resources, including repos and packages. 
2. After download, the system will proceed to verify all fetched resources. 
   1. For package artifacts, it takes the hash and looks for build provenance attestations from sigstore or github.
   1. For source artifacts that are not packaged (EG, cloned via git), it takes the revision id and looks for the source provenance from sigstore or github. 
5. Based on the claims in the provenance attestations, the CI system can determine if all resources comply with required policy and choose to proceed.


docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
-
-## Changes from v0.1
-
--   **Scope** The Source track is now scoped to Revisions rather than builds.
-Why?: To facilitate verification without anchoring it to a build.
-
--   **Model** Added a model, definitions, and the concept of verification.
-Why?: SLSA does not yet have a model for version control systems, and we need such a model to be able to discuss them.
-
-## Outstanding TODOs
```
zachariahcox, line 15: happy to add these back, it feels like this content should be (and is?) tracked in issues?

docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
-| Organization | A collection of people who collectively create the Source. Examples of organizations include an open-source projects, a company, or a team within a company.
-| Change | A set of modifications to one or more source files and associated metadata. Change metadata MUST include any information required to situate the change in relation to other changes (e.g. parent revision).
-| Version Control System | Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
-| Revision | The canonical source at a given point in time as identified by the version control system. As an example, you can identify a git revision by its tree hash.
+The SLSA SOURCE track describes increasing levels of trustworthiness and completeness in a repository revision's provenance. 
+Provenance describes which Source Control Provider produced the revision, what process was used, and who were the contributors. 
+The lowest level only requires the provenance to exist, while higher levels provide increasing protection against tampering of the version control system, the provenance, or the revision.
```
zachariahcox, line 7: this PR doesn't actually address the leveling definitions further on in the document.

docs/spec/v1.1/source-requirements.md
TomHennen: I think the idea is that while the source track is still in draft form it's helpful to keep the TODOs here as an indication it's not quite ready yet.

docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
-
-| Term | Description
-| --- | ---
-| Source | An identifiable set of text and binary files and associated metadata usually used as input for the build system (see SLSA Build Track).
-| Organization | A collection of people who collectively create the Source. Examples of organizations include an open-source projects, a company, or a team within a company.
-| Change | A set of modifications to one or more source files and associated metadata. Change metadata MUST include any information required to situate the change in relation to other changes (e.g. parent revision).
-| Version Control System | Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
-| Revision | The canonical source at a given point in time as identified by the version control system. As an example, you can identify a git revision by its tree hash.
+The SLSA SOURCE track describes increasing levels of trustworthiness and completeness in a repository revision's provenance.
```
TomHennen, line 5: open question: should we call this 'provenance' (which is overloaded with the specific format used by SLSA to track _build_ information) or should we call it something else?

If we do continue to use the word 'provenance' here we'll need to update other documentation which only uses 'provenance' when talking about mapping build -> source (and not when talking about source itself).  E.g. https://slsa.dev/spec/v1.0/provenance

We may be able to work around this by eliding the word entirely?

> The SLSA SOURCE track describes increasing levels of trustworthiness and completeness in repository revisions. and their history.
>
> This history describes which Source Control Provider produced the revision, what process was used, and who were the contributors. 
> The lowest level only requires the provenance to exist, while higher levels provide increasing protection against tampering of the version control system, the provenance, or the revision.
> ...

docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
-| Version Control System | Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
-| Revision | The canonical source at a given point in time as identified by the version control system. As an example, you can identify a git revision by its tree hash.
+The SLSA SOURCE track describes increasing levels of trustworthiness and completeness in a repository revision's provenance. 
+Provenance describes which Source Control Provider produced the revision, what process was used, and who were the contributors. 
+The lowest level only requires the provenance to exist, while higher levels provide increasing protection against tampering of the version control system, the provenance, or the revision.
+
+The primary purpose of the source track is to enable verification that a revision followed the expected process. 
+Consumers have some way of knowing what the expected provenance should look like for a given package and then compare each source revision's actual provenance to those expectations.
```
TomHennen, line 10: > The primary purpose of the source track is to enable verification that a revision followed the expected process. 
Consumers have some way of knowing what the expected provenance should look like for a given package and then compare each source revision's actual provenance to those expectations. 

I wonder if we can cut this section?

IIUC it's modified from [the 'Build track' description](https://slsa.dev/spec/v1.0/levels#build-track:~:text=Consumers%20have%20some%20way%20of%20knowing%20what%20the%20expected%20provenance%20should%20look%20like%20for%20a%20given%20package%20and%20then%20compare%20each%20package%20artifact%E2%80%99s%20actual%20provenance%20to%20those%20expectations.%20Doing%20so%20prevents%20several%20classes%20of%20supply%20chain%20threats.), but the interesting thing about the build track is that there's much greater variability in the way a build process is done, so the 'expected' process needs to be defined and communicated somehow and that process may vary from project to project.

However, I think 'source' is different?  The 'expected process' should have much less variability and it's easier to introspect?  Instead of expected process I think we're looking for something like "the source was written or approved of by the project's maintainers" (or something like that)

docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
+Doing so prevents several classes of supply chain threats.
+
+The Source track is scoped to revisions of a single repository that is controlled by an organization. 
+That organization determines the intent of the software in the repository, what Source level should apply to the repository and administers technical controls to enforce that level.
+
+## Definitions
+
+| Term | Description|
+| --- | — |
+| Subject | The subject of the attestation is the immutable identifier of a repository revision. EG: https://github.com/slsa-framework/slsa/commit/306642f21dbdaca2eaafe2df8e98432b4d4f2f02
```
TomHennen, line 20: This seems like an interesting topic of conversation.  When I read 'subject' here I'm thinking of the subject field in [the in-toto statement layer](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md) and I'd imagine that to look a lot like [`resolvedDependencies` does](https://slsa.dev/spec/v1.0/provenance#:~:text=%22resolvedDependencies%22%3A%20%5B%7B%0A%20%20%20%20%22uri%22%3A%20%22git%2Bhttps%3A//github.com/octocat/hello%2Dworld%40refs/heads/main%22%2C%0A%20%20%20%20%22digest%22%3A%20%7B%22gitCommit%22%3A%20%227fd1a60b01f91b314f59955a4e4d4e80d8edf11d%22%7D%0A%7D%5D) so that the build provenance and source attestation can easily be matched.

So I suppose my question is: can we update the example or just remove it until we're more firm on what the attestation format actually is?

docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
+| Source | An identifiable set of text and binary files and associated metadata.
```
TomHennen, line 21: Any particular reason to drop "usually used as input for the build system (see SLSA Build Track)."

docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
+Change metadata MUST include any information required to situate the change in relation to other changes (e.g. parent revision).
+| Version Control System (VCS)| Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
+| Revision | A specific identifier provided by the version control system that identifies a given state of the source. As an example, you can identify a git revision by its tree hash.
 | Change History | A record of the history of changes that went into the revision.
-| Source Control Platform | A service or suite of services for hosting version controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
+| Source Control Platform (SCP) | A service or suite of services for hosting version-controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
+| Source Provenance Attestation | A signed document describing the security claims associated with a revision.
```
TomHennen, line 30: This is probably related to my question about naming above.  We might consider leaving this definition out until we actually have the attestation defined?

I also wouldn't be surprised if we wound up with multiple attestation types.  One for consumers to understand the claim of a given source level in a revision and one that is used as evidence to compute the source level for all revisions.

TomHennen (changes requested):

docs/spec/v1.1/source-requirements.md
zachariahcox: how do we feel about 

```
status: draft
open issues: https://github.com/slsa-framework/slsa/issues?q=is%3Aissue%20state%3Aopen%20label%3Asource-track
```

or something else that keeps it dynamic?

TomHennen: SGTM

docs/spec/v1.1/source-requirements.md
zachariahcox: I was initially thinking that too but have kind of come around. 
While in practice SLSA may make a few specific recommendations (such as around code review process), virtually all consumers will be interested in whether their own policy CI jobs were run properly on the revision (github has _many_ of these). 

To phrase it differently, is the expected verification flow "ensure that all the configured rules passed?" or "ensure that the specific rules I care about passed?" 

From experience, it tends to be a mix: most of the time customers care that "the configured rules were followed" and sometimes they need to know that "definitely this revision passed a specific policy-related-ci."

I feel like this aligns pretty well to an "expectation matching" approach, similar to the build track.

docs/spec/v1.1/source-requirements.md
zachariahcox: Yeah, we can remove the specific example for now. 
However, if we're basing this on revisions, that's just a url to a revision.

docs/spec/v1.1/source-requirements.md
zachariahcox: In the other issue there was a comment thread about there being tons of uses for version control system revisions. 

We could continue to point out that repos could be consumed by other processes.

docs/spec/v1.1/source-requirements.md
zachariahcox: I think this is a specific issue digging into this topic https://github.com/slsa-framework/slsa/issues/758

I would support cutting it for now if it's at all contentious.

docs/spec/v1.1/source-requirements.md
TomHennen: Yeah that SGTM

docs/spec/v1.1/source-requirements.md
TomHennen: Yeah my preference would be to cut and then add back in when we have the actual attestation proposals.

docs/spec/v1.1/source-requirements.md
adityasaky: I suggest we use a different word to avoid confusion.

zachariahcox: @TomHennen @adityasaky I think the objective section of this doc should settle on the term for "the process that created a revision."

I think mechanically this thing would be a "revision creation process attestation" so generally the word provenance seems to work pretty well and carries the right semantics. 
If we choose a new term, it would need to make sense on its own too: You'd fetch the "provenance attestation" for a file artifact and the "<new term> attestation" for the repo artifact, etc. 

Potentially we could just use: "predicateType": "https://slsa.dev/source-provenance/v1" or something else to differentiate the schema. Conceptually, these schemas would hold very similar data, so it might be that most of the existing docs would hold up unless they were discussing specific data elements?

TomHennen: You've convinced me.  How does being deliberate about referring to it as "source provenance" in this track (and then updating the build docs to talk about "build provenance" in another PR) sound?

We can then see how it's received later.

docs/spec/v1.1/source-requirements.md
TomHennen: Oh hmm.  I think we have different ideas in mind here.

I was thinking the source track would identify rules that _must_ be enforced for each level and be silent on any others.  That we'd leave any other rules that might need to be checked to something else.  (But I definitely understand that use case).  The way I'd expect this to work is that downstream consumers of this source probably won't want to care about the specific rules but the repo owner may, and so I'd be inclined to leave this to some sort of policy evaluation that can happen on the repo (the same thing that makes the determination for downstream users).

Note that I _definitely_ see the value in this sort of thing, but given that I think it can be handled separately, and that it would make the source track more complicated (at least I expect it would) how would you feel about deferring some of that pluggability until later (or some orthogonal effort)?

Do you have any particular rules in mind?

docs/spec/v1.1/source-requirements.md
TomHennen: Ok, yeah let's remove for now and then discuss.  My biggest qualm about changing it from what's in resolvedDependencies is that to put it to use we'd have to change the build provenance spec which seems hard.

docs/spec/v1.1/source-requirements.md
adityasaky: is the idea to add more to this PR? I think it might be fine to handle them in separate PRs to keep diffs small and easier to reason about and discuss.

docs/spec/v1.1/source-requirements.md
adityasaky: Given that this is specifically about the _provenance_ about the source changes, I lean towards just capturing what process was followed and leaving it to the verifier to set expectations. So, effectively "ensure that the specific rules I care about passed" in @zachariahcox's comment above.

Source track is indeed a bit different from the build track but IMO there's value in keeping the separation between recording what happened (the source provenance) and what the expectations are when verifying the source provenance. It allows for greater agility wrt which parties we trust for what data. We also don't have to reason about how different expectations measure up against each other. For example, the maintainers may have expectations set A and I may have expectations set B, where B is in fact a subset of A. Instead of reasoning about that for every individual security control making up A and B, we can independently verify A and B are met by inspecting the provenance rather than having this extra bit of indirection.

> Note that I definitely see the value in this sort of thing, but given that I think it can be handled separately, and that it would make the source track more complicated (at least I expect it would) how would you feel about deferring some of that pluggability until later (or some orthogonal effort)?

I know this question is for @zachariahcox but I'm not sure I fully understand this. Could you clarify, perhaps with an example, what you propose we defer? 😄

docs/spec/v1.1/source-requirements.md
adityasaky: I think settling on the term for the process is fine. I also agree that we could define a separate predicate type; I'm not too worried about confusion when this is implemented in attestors / verifiers, but rather in regular conversations where context may be muddier. But let's punt that!

docs/spec/v1.1/source-requirements.md
adityasaky: I think this could be a resource descriptor, like resolvedDependencies, indeed. Note that the attestation framework supports `gitCommit`, `gitTree` etc as types of hashes, so this should be quite straightforward to represent.

docs/spec/v1.1/source-requirements.md
TomHennen: I think the idea is to handle the level definitions in another PR.

docs/spec/v1.1/source-requirements.md
joshuagl: historically we've biased towards leaving todos in the doc so we don't have to context switch away from the editor to work through them, I think it would be useful to leave them/add them back.

docs/spec/v1.1/source-requirements.md
joshuagl: +1 to using "source provenance" and "build provenance" as a starting point.

joshuagl, line 5:
```suggestion
The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's _provenance_. 
```

This matches the style in https://slsa.dev/spec/v1.0/levels

docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
-
-| Term | Description
-| --- | ---
-| Source | An identifiable set of text and binary files and associated metadata usually used as input for the build system (see SLSA Build Track).
-| Organization | A collection of people who collectively create the Source. Examples of organizations include an open-source projects, a company, or a team within a company.
-| Change | A set of modifications to one or more source files and associated metadata. Change metadata MUST include any information required to situate the change in relation to other changes (e.g. parent revision).
-| Version Control System | Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
-| Revision | The canonical source at a given point in time as identified by the version control system. As an example, you can identify a git revision by its tree hash.
+The SLSA SOURCE track describes increasing levels of trustworthiness and completeness in a repository revision's provenance. 
+Provenance describes which Source Control Provider produced the revision, what process was used, and who were the contributors.
```
joshuagl, line 6:
```suggestion
Provenance describes which Source Control Platform (SCP) produced the revision, what process was used, and who were the contributors. 
```

docs/spec/v1.1/source-requirements.md
joshuagl, line 7:
```suggestion
The lowest level only requires the provenance to exist, while higher levels provide increasing protection against tampering with any parts of the Source Control Platform, the provenance, or the revision.
```

docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
+That organization determines the intent of the software in the repository, what Source level should apply to the repository and administers technical controls to enforce that level.
+
+## Definitions
+
+| Term | Description|
+| --- | — |
+| Subject | The subject of the attestation is the immutable identifier of a repository revision. EG: https://github.com/slsa-framework/slsa/commit/306642f21dbdaca2eaafe2df8e98432b4d4f2f02
+| Source | An identifiable set of text and binary files and associated metadata.
+| Organization | A collection of people who collectively create the Source. Examples of organizations include open-source projects, a company, or a team within a company. The organization defines the goals and methods of the repository.
+| Repository | A uniquely identifiable instance of a VCS hosted on an SCP. The repository controls access to the Source in the version control system. The objective of a repository is to reflect the intent of the organization that controls it.
```
joshuagl, line 23: We should move VCS and SCP definitions up so that we have defined terms before we use them.

docs/spec/v1.1/source-requirements.md
```diff
@@ -2,37 +2,34 @@
...
+| Change | A set of modifications to one or more source files and associated metadata. A change is proposed by a “publisher,” and applied to specific revision to create a new revision.
```
joshuagl, line 24: Do we need to define publisher? 🤔

docs/spec/v1.1/source-requirements.md
zachariahcox: Great discussion. I definitely meant @TomHennen 's point 3: how many tools will a customer need to use to verify something? Or baking in an assumption that the attestation format will at least be somewhat extensible.

For teams that need more, I expect the SCP will want to add the additional claims in there somewhere (maybe akin to how the [gh oidc token contains a bunch of useful, gh-specific info](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token)). 

Empirically, to satisfy at least some customers, the revision-creation-process-attestation would need to contain no less than all the info captured by [ruleset insights](https://youtu.be/BYajh38HhCs?t=1025), plus a qualitative assessment as to whether those configured rulesets _also_ combine to fulfill any SLSA levels. 

> how would you feel about deferring some of that pluggability until later (or some orthogonal effort)?

💯 👍 
Does that effectively mean that for the first pass, we're _only_ targeting summaries?
Even with that goal, the wording like "Consumers have some way of knowing what the expected provenance should look like ... and compare it to actual" is pretty accommodating.

docs/spec/v1.1/source-requirements.md
zachariahcox: > to put it to use we'd have to change the build provenance spec which seems hard

I'd guess that you can resolve a dependency that comes with its own provenance attestation though, right? 
Would this case be different?

> I think this could be a resource descriptor, like resolvedDependencies

Nice! Wouldn't this source provenance attestation still have a revision as its subject though?
I'm probably missing some nuance here.

docs/spec/v1.1/source-requirements.md
TomHennen: Oh I see, thanks!  So I think I have a good idea on how to handle nicely.  Perhaps I should put together a proposal for the consumer side attestation that would do what you're looking for?

How do you feel about "Consumers have some way of knowing what security properties the revision meets and how they can trust that determination".

But now that I say that I also see that in the original wording "Consumers have some way of knowing what the expected provenance should look like for a given package and then compare each source revision's actual provenance to those expectations. " I might have gotten confused with the two different mentions of "provenance" :).

I think what that is trying to say is "The primary purpose of the source track is to enable verification that _the process to create a revision meets the expected security properties_.  Consumers have some way of mapping the sources listed in the _build provenance_ to the various _source provenance_ attestations to determine if all sources used during the build meet their requirements."

docs/spec/v1.1/source-requirements.md
TomHennen: Yes I think it would.

It would look something like

```
...
    "subject": [
        {
            "uri": "git+https://github.com/octocat/hello-world@refs/heads/main",
            "digest": {"gitCommit": "7fd1a60b01f91b314f59955a4e4d4e80d8edf11d"}
        }
    ]
...
```

docs/spec/v1.1/source-requirements.md
zachariahcox: Probably we can cut this line altogether, but as written the "lowest level" wouldn't involve SCP yet.

docs/spec/v1.1/source-requirements.md
zachariahcox: this PR will be rebased onto the results of : https://github.com/slsa-framework/slsa/pull/1088/files. 

I'm still not sure I see the utility of keeping these stale topics under VCS, I'll refresh the list when this is put back up for review.

docs/spec/v1.1/source-requirements.md
zachariahcox: >  "uri": "git+https://github.com/octocat/hello-world@refs/heads/main",

this reminds me that we need to define "ref" or "branch" in the spec.

TomHennen: And that reminds me that there are cases where the destination branch / release is often an important part of the context for a 'review'.  So it might not be enough to say commit 'abc123' was reviewed, it might be necessary to know "reviewed in what context".

docs/spec/v1.1/source-requirements.md
zachariahcox: @TomHennen this comment got disconnected post rebase. 👀

docs/spec/v1.1/source-requirements.md
zachariahcox: yeah 💯 , that should be part of the "situation" in the change definition I think.

docs/spec/draft/source-requirements.md
```diff
@@ -18,23 +18,33 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+| Source Control Platform (SCP) | A service or suite of services for hosting version-controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
+| Source Provenance | Verifiable information about a source revision describing where, when and how it was produced.
+| Organization | A collection of people who collectively create the Source. Examples of organizations include open-source projects, a company, or a team within a company. The organization defines the goals and methods of the repository.
+| Repository | A uniquely identifiable instance of a VCS hosted on an SCP. The repository controls access to the Source in the version control system. The objective of a repository is to reflect the intent of the organization that controls it.
+| Change | A set of modifications to the source, or the computed difference between two revisions. A change is proposed by a [publisher](#source-roles), and applied to a specific revision to create a new revision. Change metadata MUST include any information required to situate the change in relation to other changes (e.g. parent revision).
```
zachariahcox, line 43: https://github.com/slsa-framework/slsa/pull/1083#discussion_r1669216998

update to talk about ref context

docs/spec/draft/source-requirements.md
```diff
@@ -18,23 +18,33 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+| Change History | A record of the history of revisions that preceded a specific revision.
```
zachariahcox, line 44: I updated the text here a bit to help it cover git repos too (which don't track change history, technically).

docs/spec/draft/source-requirements.md
```diff
@@ -43,7 +53,7 @@ The Source track is scoped to a single project that is controlled by some organi
 | Trusted robot | Automation with an authentic identity that is authorized by the organization to propose and/or approve changes to the source.
 | Untrusted person | A human who has limited access to the project. They MAY be able to read the source. They MAY be able to propose or review changes to the source. They MAY NOT approve changes to the source or perform any privileged actions on the project.
```
zachariahcox, line 54: we might want to change all instances of "changes to the source" to "new revisions."

docs/spec/draft/source-requirements.md
```diff
@@ -18,23 +18,33 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+| Source Provenance | Verifiable information about a source revision describing where, when and how it was produced.
+| Organization | A collection of people who collectively create the Source. Examples of organizations include open-source projects, a company, or a team within a company. The organization defines the goals and methods of the repository.
+| Repository | A uniquely identifiable instance of a VCS hosted on an SCP. The repository controls access to the Source in the version control system. The objective of a repository is to reflect the intent of the organization that controls it.
+| Change | A set of modifications to the source, or the computed difference between two revisions. A change is proposed by a [publisher](#source-roles), and applied to a specific revision to create a new revision. Change metadata MUST include any information required to situate the change in relation to other changes (e.g. parent revision).
+| Change History | A record of the history of revisions that preceded a specific revision.
+| Branch | A named pointer to a revision. Branches may be modified by authorized actors.
```
zachariahcox, line 45: added a section about branches / refs. needs to move above the "change" definition that will need to use it.

docs/spec/draft/source-requirements.md
```diff
@@ -18,23 +18,33 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance.
+Source provenance describes which Source Control Platform (SCP) produced the revision, when it was generated, what process was used, and who were the contributors.
+Higher levels provide increasing protection against tampering with any parts of the Version Control System, Source Control Platform, or the source provenance.
 
-## Source model
+The primary purpose of the source track is to enable verification that the creation of a revision followed the expected process.
+Consumers have some way of knowing what the expected provenance should look like for a given revision and compare a revision's actual provenance to those expectations.
+Doing so prevents several classes of supply chain threats.
```
TomHennen, line 27:
```suggestion
The primary purpose of the source track is to enable verification that the process to create a revision meets the expected security properties. Consumers have some way of mapping the sources listed in the _build_ provenance to the various _source_ provenance attestations to determine if all sources used during the build meet their requirements.
```

docs/spec/draft/source-requirements.md
```diff
@@ -18,23 +18,33 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance.
+Source provenance describes which Source Control Platform (SCP) produced the revision, when it was generated, what process was used, and who were the contributors.
+Higher levels provide increasing protection against tampering with any parts of the Version Control System, Source Control Platform, or the source provenance.
 
-## Source model
+The primary purpose of the source track is to enable verification that the creation of a revision followed the expected process.
+Consumers have some way of knowing what the expected provenance should look like for a given revision and compare a revision's actual provenance to those expectations.
```
TomHennen, line 26:
```suggestion
The primary purpose of the source track is to enable verification that the process to create a revision meets the expected security properties.
Consumers have some way of mapping the sources listed in the _build_ provenance to the various _source_ provenance attestations to determine if all sources used during the build meet their requirements.
```

docs/spec/v1.1/source-requirements.md
TomHennen: Found it and made a suggestion...

docs/spec/draft/source-requirements.md
```diff
@@ -18,23 +18,33 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
 
 ## Objective
 
-The SLSA Source Track mitigates [Threat A ("Submit unauthorized change")](/spec/v1.0/threats#a-submit-unauthorized-change), scoped to a code repository and the organization that owns that repository. Concretely: an attacker must compromise the accounts of two organization members to publish code in a Source Level 3-conformant repository, and the evidence of those unauthorized changes cannot be destroyed without further attacks.
+The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance.
```
TomHennen, line 21:
```suggestion
The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc...).
```

docs/spec/draft/source-requirements.md
```diff
@@ -18,23 +18,33 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+Source provenance describes which Source Control Platform (SCP) produced the revision, when it was generated, what process was used, and who were the contributors.
```
TomHennen, line 22: This seems out of place here.  When I read it it sounds like it's describing the attestation format but we haven't even finished talking about the objective yet.

I think the reason it's here is to explain 'provenance' for source in the sentence above.  I added a suggestion to just make that a parenthetical to get the point across without sounding like we're describing a specific format yet.

docs/spec/draft/source-requirements.md
```diff
@@ -18,23 +18,33 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+Higher levels provide increasing protection against tampering with any parts of the Version Control System, Source Control Platform, or the source provenance.
```
TomHennen, line 23:
```suggestion
Higher source levels provide increasing protection against tampering with any parts of the Version Control System, Source Control Platform, or the source provenance.
```

TomHennen: Could also sub 'attestations' for 'information'.

docs/spec/draft/source-requirements.md
zachariahcox: the build provenance hasn't been computed yet to be used for this right? When you check source provenance, you're either looking at the build instructions themselves or trying to build some other artifact. 

At least in some cases, I would not expect a build to continue if one of the inputs fails to meet its requirements. This seems true whether the dependency is a package or repository.

docs/spec/draft/source-requirements.md
```diff
@@ -18,23 +18,33 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
-| Source Control Platform | A service or suite of services for hosting version controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
-
-### Source Roles
+| Source | An identifiable set of text and binary files and associated metadata. Source is regularly used as input to a build system (see [SLSA Build Track](requirements.md)).
+| Version Control System (VCS)| Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
+| Revision | A specific state of the source with an identifier provided by the version control system. As an example, you can identify a git revision by its tree hash.
+| Source Control Platform (SCP) | A service or suite of services for hosting version-controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
+| Source Provenance | Verifiable information about a source revision describing where, when and how it was produced.
```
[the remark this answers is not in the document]
TomHennen:
```suggestion
| Source Provenance | Information  about which Source Control Platform (SCP) produced a revision, when it was generated, what process was used, who were the contributors, and what parent revisions it was based on.
```
I'm having second thoughts about cutting this.   In the objective section we seem to be using 'provenance' in the more generic sense (what is this code and what do we know about it).  Would it make sense to describe that concept here separate from whatever attestations carry that information?

It seems helpful to have the generic concept of provenance decoupled from the specific implementation (which would need to be verifiable).

Of course that might be at odds with [the 'provenance' definition here](https://slsa.dev/spec/v1.0/terminology#:~:text=by%20the%20build.-,Provenance,-Attestation%20(metadata)%20describing).

Seems like there should be some way to work it out.

docs/spec/draft/source-requirements.md
```diff
@@ -43,7 +53,7 @@ The Source track is scoped to a single project that is controlled by some organi
 | Trusted robot | Automation with an authentic identity that is authorized by the organization to propose and/or approve changes to the source.
 | Untrusted person | A human who has limited access to the project. They MAY be able to read the source. They MAY be able to propose or review changes to the source. They MAY NOT approve changes to the source or perform any privileged actions on the project.
 | Proposer | The role that proposes a particular change to the source.
-| Reviewer | The role that reviews a particular proposed change to the source.
+| Reviewer | The role that reviews a particular change to the source.
```
TomHennen, line 56:
```suggestion
| Reviewer | A role that reviews a particular change to the source.
```

There can be multiple reviewers so I suggest 'a' instead of 'the'.

docs/spec/draft/source-requirements.md
TomHennen: I see the benefit of changing some of those instances, but we probably want to keep some?

After all, quite a bit of the point is that actors _want_ to make "changes to the source" and one of the reasons we need the source track is to make that safe.  There can perhaps be a section that explains "Changes to the source are realized as new revisions within a repository" or something like that?

docs/spec/draft/source-requirements.md
TomHennen: Oh so I think this gets back to a conversation we were having someplace else.

Sometimes folks will want to examine the source attestations (VSAs?) _after_ the build with the build provenance (that's what we do).  Often the builder doesn't know what the policy for any given package should be so it's helpful to be able to defer until later.

However, I think the important point your making is that _sometimes_ (perhaps often) the builder will be checking these things and it doesn't need to compare it with the build provenance.

In that case we could either drop "Consumers have some way of mapping the sources listed in the _build_ provenance to the various _source_ provenance attestations to determine if all sources used during the build meet their requirements." or we could adjust it to "Consumers can examine the various _source_ provenance attestations to determine if all sources used during the build meet their requirements." (which is generic enough).

We'll probably need to come up with some model and options for "how & when to verify source provenance attestations" but that can happen later.

docs/spec/draft/source-requirements.md
zachariahcox: Let's track it and skip for now.

zachariahcox: https://github.com/slsa-framework/slsa/issues/1095

docs/spec/draft/source-requirements.md
```diff
@@ -18,17 +18,15 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
-Consumers have some way of knowing what the expected provenance should look like for a given revision and compare a revision's actual provenance to those expectations.
-Doing so prevents several classes of supply chain threats.
+The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc...).
 
 The Source track is scoped to revisions of a single repository that is controlled by an organization.
 That organization determines the intent of the software in the repository, what Source level should apply to the repository and administers technical controls to enforce that level.
 
+The primary purpose of the Source track is to enable verification that the creation of a revision followed the expected process.
+Consumers can examine the various source provenance attestations to determine if all sources used during the build meet their requirements
+Doing so prevents several classes of supply chain threats.
```
TomHennen, line 28: We should probably enumerate those at some point.  Is this the time to do so?

docs/spec/draft/source-requirements.md
```diff
@@ -37,12 +35,12 @@ That organization determines the intent of the software in the repository, what
 | Version Control System (VCS)| Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
 | Revision | A specific state of the source with an identifier provided by the version control system. As an example, you can identify a git revision by its tree hash.
 | Source Control Platform (SCP) | A service or suite of services for hosting version-controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
-| Source Provenance | Verifiable information about a source revision describing where, when and how it was produced.
+| Source Provenance | Information  about which Source Control Platform (SCP) produced a revision, when it was generated, what process was used, who were the contributors, and what parent revisions it was based on.
```
TomHennen, line 38:
```suggestion
| Source Provenance | Information  about which Source Control Platform (SCP) produced a revision, when it was generated, what process was used, who the contributors were, and what parent revisions it was based on.
```

TomHennen (changes requested):

docs/spec/v1.1/source-requirements.md
TomHennen: I think this is obsolete with the new wording?

docs/spec/draft/source-requirements.md
zachariahcox: yeah, this is the wording from the levels doc -- I think it's saying "these levels represent bundles of Good Ideas and are widely regarded by experts as effective at preventing classes of supply chain threats. Each increasing level is constructed to prevent specific threat vectors of increasing complexity."

zachariahcox:
```suggestion
```

TomHennen (approved):

docs/spec/draft/source-requirements.md
```diff
@@ -18,34 +18,41 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc...).
 
-## Source model
+The Source track is scoped to revisions of a single repository that is controlled by an organization.
+That organization determines the intent of the software in the repository, what Source level should apply to the repository and administers technical controls to enforce that level.
 
-The Source track is scoped to a single project that is controlled by some organization. That organization determines what Source level should apply to the project and administers technical controls to enforce that level.
+The primary purpose of the Source track is to enable verification that the creation of a revision followed the expected process.
+Consumers can examine the various source provenance attestations to determine if all sources used during the build meet their requirements
```
joshuagl, line 27:
```suggestion
Consumers can examine the various source provenance attestations to determine if all sources used during the build meet their requirements.
```

docs/spec/draft/source-requirements.md
```diff
@@ -18,34 +18,41 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
-| Version Control System | Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
-| Revision | A specific identifier provided by the version control system that identifies a given state of the source. As an example, you can identify a git revision by its tree hash.
-| Change History | A record of the history of changes that went into the revision.
-| Source Control Platform | A service or suite of services for hosting version controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
-
-### Source Roles
+| Source | An identifiable set of text and binary files and associated metadata. Source is regularly used as input to a build system (see [SLSA Build Track](requirements.md)).
+| Version Control System (VCS)| Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
```
joshuagl, line 34: Should git be capitalised here? it's not capitalised in the next entry

docs/spec/draft/source-requirements.md
```diff
@@ -18,34 +18,41 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+| Revision | A specific state of the source with an identifier provided by the version control system. As an example, you can identify a git revision by its tree hash.
+| Source Control Platform (SCP) | A service or suite of services for hosting version-controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
+| Source Provenance | Information  about which Source Control Platform (SCP) produced a revision, when it was generated, what process was used, who the contributors were, and what parent revisions it was based on.
```
joshuagl, line 37:
```suggestion
| Source Provenance | Information about which Source Control Platform (SCP) produced a revision, when it was generated, what process was used, who the contributors were, and what parent revisions it was based on.
```

docs/spec/draft/source-requirements.md
```diff
@@ -18,34 +18,41 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
-| Change History | A record of the history of changes that went into the revision.
-| Source Control Platform | A service or suite of services for hosting version controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
-
-### Source Roles
+| Source | An identifiable set of text and binary files and associated metadata. Source is regularly used as input to a build system (see [SLSA Build Track](requirements.md)).
+| Version Control System (VCS)| Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
+| Revision | A specific state of the source with an identifier provided by the version control system. As an example, you can identify a git revision by its tree hash.
+| Source Control Platform (SCP) | A service or suite of services for hosting version-controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
```
joshuagl, line 36:
```suggestion
| Source Control Platform (SCP) | A service or suite of services (self-hosted or SaaS) for hosting version-controlled software. GitHub and GitLab are examples of source control platforms, as are combinations of tools like Gerrit code reviews with GitHub source control.
```
"Can we self host" is a frequent question, is it worth including in the definition?

docs/spec/draft/source-requirements.md
```diff
@@ -18,34 +18,41 @@ Open issues are tracked with the [source-track](https://github.com/slsa-framewor
...
+| Repository | A uniquely identifiable instance of a VCS hosted on an SCP. The repository controls access to the Source in the version control system. The objective of a repository is to reflect the intent of the organization that controls it.
+| Branch | A named pointer to a revision. Branches may be modified by authorized actors. In git, cloning a repo will download all revisions in the history of the "default" branch to the local machine.
+| Change | A set of modifications to the source in a specific context. As an example, a proposed change to a "releases/1" branch may require higher scrutiny than a change to "users/1".
+| Change History | A record of the history of revisions that preceded a specific revision.
+
+## Source Roles
 
 | Role | Description
 | --- | ---
 | Administrator | A human who can perform privileged operations on one or more projects. Privileged actions include, but are not limited to, modifying the change history and modifying project- or organization-wide security policies.
```
joshuagl, line 48: Nit: projects is not defined.

joshuagl (commented):
A few minor nits, but otherwise LGTM

docs/spec/draft/source-requirements.md
zachariahcox: In this case, I think it's just because it's the beginning of the sentence.

docs/spec/draft/source-requirements.md
zachariahcox: Agree -- this PR only updates the objective (and required definitions). 
I think we're informally using "repository" instead of "project" for this concept, but we should discuss that in the next PR. 

To me, one of the reasons to use "project" is to encompass things like issues, actions, other stuff besides versioning. 
We're not finished, but I suspect this specification will need to focus entirely on source topics, ignoring everything else related to software development.

joshuagl: I agree with your suspicion, I think the other aspects of projects are out of scope for this track. I don't mind handling this in a future PR but we need a mechanism to track that the work still needs doing. I filed https://github.com/slsa-framework/slsa/issues/1099

Issue #241, opened 2021-12-06
Clarify how multiple attestations may be needed to verify "source" requirements


mlieberman85:
Related somewhat to: #129

Even though the provenance spec does allow you to point to source control for "materials," it doesn't allow for the ability to attest to "verified history," "retained indefinitely," or "two-person reviewed." You can implicitly assume if you have a source control uri like git that it's "version controlled."

Another related thing, and I'm unsure if there's an issue for this already but I think the documentation makes it unclear if the expectation is to have multiple attestations associated with an artifact to then allow the client to make a SLSA level judgement or if everything is intended to be included as part of a single attestation. In the case of multiple attestations would a source based one even include a "builder"?

I bring up the above because in secure CI environments you might have separate "builders" performing different tasks for an artifact build. Those builders might have their own identities and the build service might sign provenance based on those identities, e.g. SPIFFE/Spire.

For example, a flow like:

1. builder A downloads source to shared volume
2. builder B downloads dependencies to shared volume
3. builder C with no internet access, etc. (to be hermetic) compiles/packages artifact
4. builder D publishes artifact to artifact repo

TomHennen:
This is what I was hoping to address in https://github.com/in-toto/attestation/issues/47 (before we moved SLSA provenance back to this project).

I had a proposal I was running past @joshuagl but we never actually finished it up.  I've shared with you to get your feedback too.

Should we continue discussion here or in https://github.com/in-toto/attestation/issues/47 ?

mlieberman85:
Right, forgot that was over there. Probably makes sense to continue over there for now but maybe leave this ticket open just so people know to look at it the discussion in-toto/attestation#47?

mlieberman85:
@TomHennen does it make sense to create another github issue for the other thing described above? There might be the need to multiple attestations associated with an artifact based on who is making what claims?

MarkLodato:
IMO it makes sense to have this issue be about the general high level problem ("Another related thing...") while in-toto/attestation#47 is about the specific Source Control predicate.

TomHennen:
On that topic I'd suggest we expect there to be multiple different attestations.  I think that should make it easier to the processes issuing the attestations to have first-hand knowledge about what it is they're signing.

In your example I think it could work like this:

[...]

PR #1094, opened 2024-07-10
content: draft: define how downstream users can verify the SLSA source track level of revisions

### LINKED ISSUE #1042: Source track attestation claims
Tracking issue for https://github.com/slsa-framework/slsa/pull/1037#discussion_r1526020307

Main comments:
- The current claims are a mix of policy (who's allowed to review) and facts (who reviewed).
- The current use case is useful when consumers and producers are the same but does not help for consumption of other (open source) projects

Current suggestion: Separate policy and fact claims, like we do in build track.

### LINKED ISSUE #1071: How to communicate SLSA source track metadata?
Builds have provenance.  Source track doesn’t (it was explicitly removed from this PR).

@Nikokrock says:
> I missed the time when the idea of the provenance for sources was removed. I think that we probably should converge on whether we want them or not in the track. If you had that to L1 it gives L1 more value as the person even if not using a SCP will emit from time to time an attestation that will be public and easily shareable. I still think that the attestation should be link to a revision (i.e based on the hash of the content). This gives the ability to compare two revisions independently from the VCS and SCP used. [tomhennen@google.com](mailto:tomhennen@google.com) I have read the main points of the proposal from 2-3 years ago and I was wondering whether there was a reason it was abandoned ?

In the SLSA specification meeting on June 17th this [was discussed](https://docs.google.com/document/d/1PwhekVB1iDpcgCQRNVN_aesoVdOiTruoebCs896aGxw/edit#bookmark=id.oqoqjt4urxm) (note that this isn't necessarily 'final' we can discuss more here?):
> Split into 1. “how do consumers know the status as communicated by some authority” and 2. “how does that authority know the status is true” (gittuf could likely fill this role)
> 1. We can do ‘now’ with blind trust on the SCP (to unblock immediate progress)
> 2. Would let us remove the need to trust the SCP
>
> We SHOULD NOT have anything in the spec that relies on a specific implementation of a control (e.g. a specific way to validate identity)

TomHennen:
fixes #1071
fixes #1042
refs #241

This PR modifies _draft_ content of the SLSA spec.

## Context

See [discussions here](https://docs.google.com/document/d/13Xt8mA_2b00McGX2vkyhu4GQdFAqtXPu7YXE8ZA6ISE/edit?resourcekey=0-EqfHF79tUWAKp4PzsE3z1A&tab=t.0#heading=h.fhg4lsemfpz2) [and here](https://docs.google.com/document/d/1PwhekVB1iDpcgCQRNVN_aesoVdOiTruoebCs896aGxw/edit#bookmark=id.oqoqjt4urxm).

Google document requires [slsa-discussion@googlegroups.com](mailto:slsa-discussion@googlegroups.com) membership.

## VSA for source

Define how downstream users can verify the SLSA source track level of revisions by using a [VSAs](http://slsa.dev/verification_summary) produced by the Source Control Platform (SCP).

To use these VSAs users do not need to know the specifics of how any given SCP or Version Control System (VCS) meets the SLSA source requirements (which may vary greatly from implementation to implementation).  Instead it is left to the SCP or another trusted 'authority' to make that determination for downstream users.

The question of _how_ the authority ensures those claims to be true is left undefined in this change.

Future updates can include guidance for how to verify source level when combined with [build provenance](https://slsa.dev/provenance).

## Example scenario

1. A user wants to verify https://github.com/slsa-framework/slsa/commit/9a04d1ee393b5be2773b1ce204f61fe0fd02366a is SLSA source level 3.
2. The user 'trusts' GitHub as the authority for source revisions managed by GitHub.
3. The user requests a VSA for https://github.com/slsa-framework/slsa/commit/9a04d1ee393b5be2773b1ce204f61fe0fd02366a from a TBD API
4. The user verifies the VSA following [the standard instructions](https://slsa.dev/spec/draft/verification_summary#how-to-verify) or using [standard tooling](https://github.com/slsa-framework/slsa-verifier?tab=readme-ov-file#verification-summary-attestations-vsa) and looking for `SLSA_SOURCE_LEVEL_2` in the `verifiedLevels` field.


docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,61 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+Source level assertions are issued by the SCP or some other authority that has sufficient evidence to make the determination of a given
+revision's source level.
+
+These assertions are communicated in [Verification Summary Attestations (VSAs)](./verification_summary.md) as follows:
+
+1. `subject.uri` SHOULD be set to a human readable URI of the revision.
+1. `subject.digest` MUST include the revision identifier (e.g. `gitCommit`) and MAY include other digests over the contents of the revision (e.g. `gitTree`, `dirHash`, etc...).
+SCPs that do not use cryptographic digests MUST define a canonical type that is used to identify immutable revisions (e.g. `svn_revision_id`)[^1].
+1. `resourceUri` MUST be set to the URI of the repository, preferably using [SPDX Download Location](https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field) and SHOULD
+provide the branch or refs.  E.g. `git+https://github.com/foo/hello-world@refs/heads/main`.
```
zachariahcox, line 156: what are some of the use cases where you'd need to see historical branch reachability? 

I suspect that most places where refs are needed today they're being used as a proxy measurement for evidence claims. EG: if the revision was reachable from `main` on that date, then it "must have passed" all the rules protecting the main branch.

Problems with that approach include: 
* branches change over time -- a suitable revision may no longer be reachable from this ref due to force push / rebase, etc. 
* rulesets change over time -- conscientious verifiers will still go and check what rules "were in place" when they must have approved this specific revision
* a specific revision may have attestations for multiple "ref contexts", as there are no limits to the number of protected branches nor number of refs that can point to a specific revision.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,61 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+1. `verifiedLevels` MUST include the SLSA source track level the issuer asserts the revision meets. One of `SLSA_SOURCE_LEVEL_0`, `SLSA_SOURCE_LEVEL_1`, `SLSA_SOURCE_LEVEL_2`, `SLSA_SOURCE_LEVEL_3`.
+1. `dependencyLevels` MAY be empty as source revisions are typically terminal nodes in a supply chain.
```
zachariahcox, line 158: I tend to agree with this, but we should call out submodules explicitly here if we go this way.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,61 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+
+Source Level Assertion issuers MAY issue assertions based on their understanding of the underlying system, but SHOULD prefer to issue assertions based on Source Level Evidence appropriate to their SCP.
```
zachariahcox, line 160: I don't quite follow this one. 

I think it's saying: the verification logic should use the evidence provided by SCPs and not its own secret knowledge?

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,61 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+}],
+
+// Predicate
+"predicateType": "https://slsa.dev/verification_summary/v1",
+"predicate": {
+  "verifier": {
+    "id": "https://example.com/source_verifier",
+  },
+  "timeVerified": "1985-04-12T23:20:50.52Z",
+  "resourceUri": "git+https://github.com/foo/hello-world@refs/heads/main",
```
zachariahcox, line 178: It seems to me that this should also be the commit sha url right? 

It would make sense to add "ref: refs/heads/main" in the predicate if that's relevant to consumption later on, but maybe that should be separate? 
For example, if a consumer were to visit that resource uri, it would likely point to a different thing than what was verified here.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,61 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
 Benefits: A compromise of a single human or account does not result in compromise of the project, since all changes require review from two humans.
+
+## Source Attestations
+
+There are two uses for source attestations within the source track:
+
+1. Assertions: Communicate to downstream users what high level security properties a given source revision meets.
+2. Evidence: Provide trustworthy metadata which can be used to determine what high level security properties a given source revision meets.
+
+To provide interoperability and ensure ease of use, it's essential that the 'assertions' are applicabile across all Source Control Platforms.
```
zachariahcox, line 140:
```suggestion
To provide interoperability and ensure ease of use, it's essential that the 'assertions' are applicable across all Source Control Platforms.
```

docs/spec/draft/source-requirements.md
zachariahcox: an alternative resource uri here might be the pull request:

https://api.github.com/repos/sigstore/sigstore-conformance/pulls/95

docs/spec/draft/source-requirements.md
zachariahcox: "assertions" might be useful for handling [merge queue mapping solutions too](https://docs.google.com/document/d/1nVkvRsxFef2OgVm2CumyLDbt-yYX-A5wtg6dF_sthwo/edit?pli=1#heading=h.mis25kwg9h9u).

In many ways, those will be "technically verifiable" but practically you'll have to take the SCP's word for it.

docs/spec/draft/source-requirements.md
TomHennen: Right, I actually realized we need a _list_ of branches (let me make an update).

Rationale:
> Branches often carry semantic meaning that a downstream user may want to verify.  E.g. "It's OK to build from commits that release_* branches pointed to, but not OK to build from personal or 'experimental' branches".  In addition SCP security settings are often per-branch yet builders regularly build from a tag or commit hash.  The result is that it's necessary for the SCP to indicate which branches have ever pointed to a particular revision.  This enables downstream consumers to determine if the revision was from an appropriate context in the repository.

docs/spec/draft/source-requirements.md
TomHennen: Oooh, right.  submodules are interesting.

I believe the advice for SLSA build provenance is that each submodule should be listed separately in the build provenance and so I'd imagine they should each get their own VSA too?

docs/spec/draft/source-requirements.md
TomHennen: That's right.  Though 'secret' is probably going too far.  I could imagine that an SCP might issue a white paper describing how they secure their SCP, going into design detail.  "We enforce these invariants (without a tamper-proof record), so we can confidently assert X, Y, Z."  In that case it's not exactly 'secret'.

Maybe adding 'tamper proof' in there somewhere would make it more clear?

docs/spec/draft/source-requirements.md
TomHennen: No, the idea is that VSAs get matched for that sort of thing in the 'subject' and the `resourceUri` is meant to be a more long-lived thing that users can specify in their policy "this build is only ever supposed to come from this repo".  That repo is unlikely to change over time while the commit hash will change all the time.

I think the [how to verify](https://slsa.dev/spec/draft/verification_summary#how-to-verify) section might help explain?

docs/spec/draft/source-requirements.md
TomHennen: Should we address submodules and other 'indirection' elsewhere in the spec or is it better to keep it aligned with the attestations themselves?

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,60 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
 **[Different persons]** The organization strives to ensure that no two user accounts correspond to the same person. Should the organization discover that it issued multiple accounts to the same person, it MUST act to rectify the situation. For example, it might revoke project privileges for all but one of the accounts and perform retroactive code reviews on any changes where that person's accounts are the author and/or code reviewer(s).
 
 Benefits: A compromise of a single human or account does not result in compromise of the project, since all changes require review from two humans.
+
+## Source Attestations
+
+There are two uses for source attestations within the source track:
+
+1.  Assertions: Communicate to downstream users what high level security properties a given source revision meets.
```
TomHennen, line 137: Should I add these to the terminology doc per our discussion?

> 'assertion' is something that you say with or without evidence.
> evidence is something that can be evaluated by another party
> 'attestation' covers both?

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,60 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+Source Level Assertion issuers MAY issue assertions based on their understanding of the underlying system, but SHOULD prefer to issue assertions based on Source Level Evidence appropriate to their SCP.
+
+#### Example
+
+```json
+"_type": "https://in-toto.io/Statement/v1",
+"subject": [{
+  "uri": "https://github.com/foo/hello-world/commit/9a04d1ee393b5be2773b1ce204f61fe0fd02366a",
+  "digest": {"gitCommit": "9a04d1ee393b5be2773b1ce204f61fe0fd02366a"},
+  "annotations": {"source_branches": ["main", "release_1.0"]}
```
TomHennen, line 170: @zachariahcox should these be `refs/heads/main` and `refs/heads/release_1.0` instead?

docs/spec/draft/source-requirements.md
TomHennen: It's probably worth adding something about this to the spec.  Not sure if that's on your plate.  If not I can give it a go. (I suspect I'll butcher the terminology but hopefully you can correct me)

docs/spec/draft/source-requirements.md
TomHennen: I think there's an argument to be made that the `refs/heads` is a SCP/VCS implementation detail that we don't need to expose to the downstream user and it would be more ergonomic if the issuer elided that bit and just included the branch name as most folks interact with it.

zachariahcox: in my experience yes, we should always fully qualify git refs. We recently chose to unilaterally block refs that look like `refs/heads/refs/heads/main` but it's very easy to trick other tooling into thinking they have `refs/heads/main`.

zachariahcox: maybe we would call the annotation "gitRef" to align with "gitCommit"?

TomHennen: Updated to require the branches to be fully qualified.

I left it as `source_branches` because:

1. We also need to allow other VCS/SCPs to be able to use this (and I don't quite know how I'd want to represent that).
2. I think downstream users probably don't need to distinguish between the different types of branches among different VCS/SCPs.  E.g. I'm hoping they can just do a plain string comparison.  If we had different types per VCS/SCP then downstream users need handle each VCS 'specially'.  At the moment I think standard VSA parsing logic should 'just work' without needing special support for different types.
WDYT?

docs/spec/draft/source-requirements.md
arewm: I feel like tamper proof should be a property/requirement of a VSA. Namely the VSA is an attestation which is tied to and signed by an identity.

TomHennen: Oh yes, the VSA itself is _definitely_ meant to be tamper proof (thanks to the signature).  What I was trying to get at above is that the evidence the SCP relies on when _issuing_ the VSA may not itself be tamper-proof.

E.g. They may just consult the configuration as stored in a database somewhere in consultation with their implicit understanding of the design of the SCP.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,69 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
 **[Different persons]** The organization strives to ensure that no two user accounts correspond to the same person. Should the organization discover that it issued multiple accounts to the same person, it MUST act to rectify the situation. For example, it might revoke project privileges for all but one of the accounts and perform retroactive code reviews on any changes where that person's accounts are the author and/or code reviewer(s).
 
 Benefits: A compromise of a single human or account does not result in compromise of the project, since all changes require review from two humans.
+
+## Source Attestations
+
+There are two uses for source attestations within the source track:
+
+1.  Assertions: Communicate to downstream users what high level security properties a given source revision meets.
+2.  Evidence: Provide trustworthy metadata which can be used to determine what high level security properties a given source revision meets.
```
zachariahcox, line 138: what would be an example of evidence that would appear here?

adityasaky: I think code review attestations could be one example? in-toto also has a test run results predicate type. A signature on that attestation from the CI workflow's identity + the contents of the predicate themselves could be another perhaps? Curious to hear what @TomHennen had in mind. 😄

docs/spec/draft/source-requirements.md
adityasaky: I think this may in fact be a property of how the attestation model is used. You could have one that makes an "assertion" as defined here, like VSA or SCAI, while a lot of other metadata, such as provenance, can be thought of as evidence when accompanying the assertion.

docs/spec/draft/source-requirements.md
TomHennen: Yes right, a code review attestation would be one example.  Or whatever GitHub was thinking for creating when things are merged.

zachariahcox: > a code review attestation would be one example

Ok, great! then I think we can simplify this to just "assertions." 
All the evidence will be stored "somewhere else" and the goal of source attestations is sign a big list of those claims as "assertions."

Potentially git commit metadata might be a kind of evidence that everyone could validate themselves, but we can probably choose to ignore it for now?

zachariahcox: Are we saying that some of the claims should be standardized? eg, something about the number of unique code reviewers. 

```yaml
assertions: 
  # maybe a slsa-specific claim?
  - type: SLSA.source.number-of-reviewers
    context: 
      value: 2
  # or the general case:
  - type: github.source.[name of rule]
    context:
      key: value
```

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,69 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+Source Level Assertion issuers MAY issue assertions based on their understanding of the underlying system, but SHOULD prefer to issue assertions based on Source Level Evidence appropriate to their SCP.
+
+#### Example
+
+```json
+"_type": "https://in-toto.io/Statement/v1",
+"subject": [{
+  "uri": "https://github.com/foo/hello-world/commit/9a04d1ee393b5be2773b1ce204f61fe0fd02366a",
+  "digest": {"gitCommit": "9a04d1ee393b5be2773b1ce204f61fe0fd02366a"},
+  "annotations": {"source_branches": ["refs/heads/main", "refs/heads/release_1.0"]}
```
zachariahcox, line 172: what would this mean? would you need to emit new attestations for this sha if new refs point at it?

Is this effectively attesting that a PR was completed successfully for this commit to be added to each of these refs?

docs/spec/draft/source-requirements.md
adityasaky: From the PoV of attestation-verifier (https://github.com/in-toto/attestation-verifier), this would match on the predicate type and then look for specific properties. I imagine the SOURCE properties in the VSA have SLSA_SOURCE or so prepended as you show it here.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,69 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+
+Source level assertions are issued by the SCP or some other authority that has sufficient evidence to make the determination of a given
+revision's source level.
+
+These assertions are communicated in [Verification Summary Attestations (VSAs)](./verification_summary.md) as follows:
+
+1.  `subject.uri` SHOULD be set to a human readable URI of the revision.
+2.  `subject.digest` MUST include the revision identifier (e.g. `gitCommit`) and MAY include other digests over the contents of the revision (e.g. `gitTree`, `dirHash`, etc...).
+SCPs that do not use cryptographic digests MUST define a canonical type that is used to identify immutable revisions (e.g. `svn_revision_id`)[^1].
+3.  `subject.annotations.source_branches` SHOULD be set to a list of branches that pointed to this revision at any point in their history.
```
zachariahcox, line 155: > at any point in their history

I think this is supposed to help consumers who only want `releases/*` refs? 

They would be able to see if this revision was reachable from any release ref when this attestation was minted. 

For a single revision / subject, a normal git late branching flow would keep reminting these things when important branches point to it.
I'd be nice not to have to do that.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,69 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+2.  Evidence: Provide trustworthy metadata which can be used to determine what high level security properties a given source revision meets.
+
+To provide interoperability and ensure ease of use, it's essential that the 'assertions' are applicable across all Source Control Platforms.
+Due to the significant differences in how SCPs operate and how they may chose to meet the Source Track requirements it is preferable to
+allow for flexibility with 'evidence' attestations.  To that end SLSA leaves 'evidence' attestations undefined and up to the SCPs to determine
+what works best in their environment.
+
+### Source Level Assertions
+
+Source level assertions are issued by the SCP or some other authority that has sufficient evidence to make the determination of a given
```
zachariahcox, line 147: > Source level assertions are issued by the SCP

At least for off-the-shelf SCPs, they would almost certainly never claim a user's process complied with any specific standard 😅 
The platform can only attest what it sees and leave the conclusions to lawyers.

```suggestion
Source level assertions are issued by some authority that has sufficient evidence to make the determination of a given
```

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,69 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+  },
+  "verificationResult": "PASSED",
+  "verifiedLevels": ["SLSA_SOURCE_LEVEL_3"],
+}
+```
+
+#### How to verify
+
+-   VSAs for source revisions MUST follow [the standard method of VSA verification](./verification_summary.md#how-to-verify).
+-   Users SHOULD check that an allowed branch is listed in `subject.annotations.source_branches` to ensure the revision is from an appropriate context within the repository.
```
zachariahcox, line 193: branches are mutable -- this revision will persist for as long as it's hosted. 

I could see potentially that a consumer might want to know _if_ a revision was reachable from specific refs at mint-time, but this information is extremely short-lived. It's just as likely that by consume time, none of the mentioned refs can reach this revision. 

I would push back super hard on teams that require this data too heavily -- do you have any specific examples of this scenario?

docs/spec/draft/source-requirements.md
TomHennen: > Are we saying that some of the claims should be standardized? eg, something about the number of unique code reviewers.

I think it would be nice to provide some standard options for folks, but I don't intend to do so in this PR (and maybe we shouldn't even try for the source track v1?)

docs/spec/draft/source-requirements.md
TomHennen: > I think this is supposed to help consumers who only want releases/* refs?

Something like that yes.

> For a single revision / subject, a normal git late branching flow would keep reminting these things when important branches point to it.
I'd be nice not to have to do that.

Could they be minted on-demand?

If not, do you have any other thoughts about how to capture this information (or perhaps we should see if we can make it inconsequential?)

docs/spec/draft/source-requirements.md
TomHennen: > what would this mean? would you need to emit new attestations for this sha if new refs point at it?

Yeah it would. :/

> Is this attesting that a PR was completed successfully for this commit to be added to each of these refs?

Yes?  It would mean the PR was merged to that destination.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,69 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+"_type": "https://in-toto.io/Statement/v1",
+"subject": [{
+  "uri": "https://github.com/foo/hello-world/commit/9a04d1ee393b5be2773b1ce204f61fe0fd02366a",
+  "digest": {"gitCommit": "9a04d1ee393b5be2773b1ce204f61fe0fd02366a"},
+  "annotations": {"source_branches": ["refs/heads/main", "refs/heads/release_1.0"]}
+}],
+
+"predicateType": "https://slsa.dev/verification_summary/v1",
+"predicate": {
+  "verifier": {
```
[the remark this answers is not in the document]
TomHennen: I'm not sure I follow what you're suggesting here (maybe I'm just getting confused by the yaml which is a foreign language to me).

Is the question: "instead of JSON, can you format this example as YAML?"

Or is it "can we change the existing VSA predicate type to something else?"

docs/spec/draft/source-requirements.md
TomHennen: So, I don't intend for this to mean "branch a _currently_ points to commit x" but rather "branch a pointed to x at some point in its history".  E.g. "commit x was the tip of branch a at some point"

I'd like to be able to write policy that can ensure that something came from "main" or "v1.x" and exclude things that came from "experimental", "testing", etc...

docs/spec/draft/source-requirements.md
adityasaky: sorry, I didn't finish that thought. I suppose it was a question of whether we generalize this for all of SLSA or just the source track for now. To unblock this, maybe we could address this here and table this for a discussion at either the SLSA level or perhaps even the in-toto layer for predicate types.

I also have a small nit because if an assertion could also be without evidence, then all evidence which is in the form of an attestation is also an assertion. So I wonder if an assertion is effectively the same as an attestation? Do we need a different word?

arewm: To the original question
> Should I add these to the terminology doc per our discussion?

While these are proposed differences in the source track, I think that they exist as a broader concept than just the source track. To me, if feels like these changes should be targeted elsewhere. I do also worry about the added complexity that this differentiation might add.

TomHennen: That's fair.  Let me see if I can cover it without defining any new terms.  I think it should help that #1097 is getting firmed up too.

docs/spec/draft/source-requirements.md
zachariahcox: > Could they be minted on-demand?

that sounds tricky! like, do you include forks? user branches? etc. Probably we need to avoid needing it. 

I think the intent is to say: I can deploy this revision because X, where X is the set of rules required to land in the /refs/heads/release/ refspec on this date. 

Ideally, you'd be able to reverify the qualifications for X and not need to use the refname as a place holder. 
If you do need the refname, I think we'd pretty much always need to explain _why_ so we can know if any mapping is required when the ref rules change over time.

docs/spec/draft/source-requirements.md
TomHennen: How's this?  Now it just talks about "Summary attestations" (e.g. VSAs and how we use them) and "detailed attestations" (everything else).

CC @zachariahcox

docs/spec/draft/source-requirements.md
TomHennen: Oh so we typically use the branch name to convey semantic between the folks managing the code and the folks writing the individual policies.

The automated rules that get applied to a 'experimental' and 'release' branch might very well be the same, but the code you put in them would be different.  Being able to convey downstream if something was good enough for the 'release' branch is very helpful!  These names will likely differ from team to team as they set up their branches and development flows differently.

So I suppose I view the refname(?) as orthogonal to the actual rules the SCP is enforcing at any point in time.

Does that explanation help explain why we'd want such a thing?  (Even if we do decide it's too hard to actually implement)

docs/spec/draft/source-requirements.md
TomHennen: Do you think that would be hard to do?  It seems like it would be straightforward but perhaps I'm missing something interesting about the internals of the system?

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,86 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+
+1.  `subject.uri` SHOULD be set to a human readable URI of the revision.
+2.  `subject.digest` MUST include the revision identifier (e.g. `gitCommit`) and MAY include other digests over the contents of the revision (e.g. `gitTree`, `dirHash`, etc...).
+SCPs that do not use cryptographic digests MUST define a canonical type that is used to identify immutable revisions (e.g. `svn_revision_id`)[^1].
+3.  `subject.annotations.source_branches` SHOULD be set to a list of branches that pointed to this revision at any point in their history.
+    -   git branches MUST be fully qualified (e.g. `refs/head/main`) to reduce the likelyhood of confusing downstream tooling.
+4.  `resourceUri` MUST be set to the URI of the repository, preferably using [SPDX Download Location](https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field).
+E.g. `git+https://github.com/foo/hello-world`.
+5.  `verifiedLevels` MUST include the SLSA source track level the issuer asserts the revision meets. One of `SLSA_SOURCE_LEVEL_0`, `SLSA_SOURCE_LEVEL_1`, `SLSA_SOURCE_LEVEL_2`, `SLSA_SOURCE_LEVEL_3`.
```
marcelamelara, line 159: This is another place where I think the definitions and intent are a little inconsistent. A nitpicky comment as well: we don't use the term "issuer" anywhere else AFAIK, so we might want to refer to this as the "verifier" or "attester" instead.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,86 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+```
+
+#### How to verify
+
+-   VSAs for source revisions MUST follow [the standard method of VSA verification](./verification_summary.md#how-to-verify).
+-   Users SHOULD check that an allowed branch is listed in `subject.annotations.source_branches` to ensure the revision is from an appropriate context within the repository.
+-   Users SHOULD check that the expected `SLSA_SOURCE_LEVEL_` is listed within `verifiedLevels`.
+-   Users MUST ignore any unrecognized values in `verifiedLevels`.
+
+### Detailed attestations
```
marcelamelara, line 198: See my comment about re: "detailed" vs. "full" attestations.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,86 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
 **[Different persons]** The organization strives to ensure that no two user accounts correspond to the same person. Should the organization discover that it issued multiple accounts to the same person, it MUST act to rectify the situation. For example, it might revoke project privileges for all but one of the accounts and perform retroactive code reviews on any changes where that person's accounts are the author and/or code reviewer(s).
 
 Benefits: A compromise of a single human or account does not result in compromise of the project, since all changes require review from two humans.
+
+## Source Attestations
+
+There are two broad categories of source attestations within the source track:
```
marcelamelara, line 135: Are attestations only ever at the granularity of revisions? That seems to be the assumption throughout this section, but we might want to make that a bit more explicit upfront.

docs/spec/draft/source-requirements.md
marcelamelara: This is a case where having additional context/attribute information in the predicate to describe the action that was taken or the meaning of the attestation would help.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,86 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+#### How to verify
+
+-   VSAs for source revisions MUST follow [the standard method of VSA verification](./verification_summary.md#how-to-verify).
+-   Users SHOULD check that an allowed branch is listed in `subject.annotations.source_branches` to ensure the revision is from an appropriate context within the repository.
+-   Users SHOULD check that the expected `SLSA_SOURCE_LEVEL_` is listed within `verifiedLevels`.
+-   Users MUST ignore any unrecognized values in `verifiedLevels`.
+
+### Detailed attestations
+
+Source level evidence is the tamper-proof (ideally signed in-toto attestations) that can be used to determine what SLSA Source Level a given revision meets.
```
marcelamelara, line 200: Nit: "Source level evidence" can be misread to mean evidence at the source-level, which isn't what I think we mean here. I suggest rewording this to something like "Evidence about the Source track level".

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,86 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+This evidence can be used by an authority as the basis for issuing a [Summary Attestation](#summary-attestation).
+
+SCPs and VCSes may have different methods of operating that necessetate different forms of evidence.
```
marcelamelara, line 203: Typo
```suggestion
SCPs and VCSes may have different methods of operating that necessitate different forms of evidence.
```

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,86 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+E.g. GitHub based workflows may need different evidence than Gerrit based workflows, which would both likely be different from workflows that
+operate over Subversion repositories.
+
+Examples of evidence:
+
+-   A "code review" attestation which describes the basics of any code review that took place.
+-   An "authentication" attestation which describes how the actors involved in any revision were authenticated.
+
+TODO: Can we define or recommend any canonical formats?
```
marcelamelara, line 212: Might be worth reviving https://github.com/in-toto/attestation/pull/151 and looking at whether we can use/revise other vetted in-toto predicates for these examples.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,86 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+This evidence can be used by an authority as the basis for issuing a [Summary Attestation](#summary-attestation).
+
+SCPs and VCSes may have different methods of operating that necessetate different forms of evidence.
+E.g. GitHub based workflows may need different evidence than Gerrit based workflows, which would both likely be different from workflows that
+operate over Subversion repositories.
+
+Examples of evidence:
+
+-   A "code review" attestation which describes the basics of any code review that took place.
+-   An "authentication" attestation which describes how the actors involved in any revision were authenticated.
```
marcelamelara, line 210: A source SBOM may be another form of evidence to add here. There are also folks looking at capturing SBOM diffs, which might be useful in this context.

marcelamelara (commented):
Thanks @TomHennen ! Most of my comments are around making the definitions a little crisper, but otherwise this is coming together really nicely.

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,86 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
 **[Different persons]** The organization strives to ensure that no two user accounts correspond to the same person. Should the organization discover that it issued multiple accounts to the same person, it MUST act to rectify the situation. For example, it might revoke project privileges for all but one of the accounts and perform retroactive code reviews on any changes where that person's accounts are the author and/or code reviewer(s).
 
 Benefits: A compromise of a single human or account does not result in compromise of the project, since all changes require review from two humans.
+
+## Source Attestations
+
+There are two broad categories of source attestations within the source track:
+
+1.  Summary attestations: Used to communicate to downstream users what high level security properties a given source revision meets.
+2.  Detailed attestations: Provide trustworthy, tamper-proof, metadata which can be used to determine what high level security properties a given source revision meets.
```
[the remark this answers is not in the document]
TomHennen: Went with "Full attestation" thanks for the suggestion.

docs/spec/draft/source-requirements.md
TomHennen: Now "full"

docs/spec/draft/source-requirements.md
TomHennen: Reworded.  Does this help?

docs/spec/draft/source-requirements.md
TomHennen: Fixed

docs/spec/draft/source-requirements.md
TomHennen: Sure.  Added that and a number of other examples.

docs/spec/draft/source-requirements.md
TomHennen: Yes I definitely think getting a code review predicate would be helpful (cc @adityasaky).  Until that lands I've added links to other canonical things that might be interesting for revisions.

docs/spec/draft/source-requirements.md
TomHennen: So in the VSA, what's described here, I don't think we really want to distinguish between the actions.  The goal is that a single VSA can convey all the high level properties that a given revision meets and who verified them.

In this case the meaning of the attestation comes from the predicateType and verifier.id (this identity verified this revision).

Would you agree with that?

Maybe the tricky thing here is how to capture the semantic relationship between branches and revisions?  I suppose another way to do that would be to define some way to ask an API "tell me all the branches that pointed to this revision" and then each time a  branch is pointed at something a new attestation that _just_ says "branch `b` pointed at revision `r`" is created and then the user finds the VSA for that revision (which doesn't contain the branch information at all) elsewhere?

zachariahcox: > Do you think that would be hard to do? It seems like it would be straightforward but perhaps I'm missing something interesting about the internals of the system?

it is "hard" to move a protected ref to point to a sha, but "easy" (no restrictions) to point new / unprotected refs at a sha. 
Mechanically, we're probably only going to be able to mint attestations for merged PRs. By default, with no fancy stuff, if there are future merged PRs that point to the same sha, we'd get two attestations. We wouldn't re-issue the first attestation. 
 
SCPs don't keep track of all the SHAs that have ever been pointed-to by a ref and would represent a huge data set to do so. 
@TomHennen I think assuming semantics around ref names will be tricky -- we should consider writing a custom claim for systems that need that. "merged_to_refs_heads_main" or equivalent.

docs/spec/draft/source-requirements.md
marcelamelara: Thanks for the updates! I can't seem to be able to resolve these comments...

docs/spec/draft/source-requirements.md
```diff
@@ -163,24 +163,108 @@ This is a typical setup for teams who leverage code review tools.
...
+Organizations that want strong guarantees and auditability of their change management processes.
+
+Benefits:
+Provides reliable information to policy enforcement tools.
+
+Requirements:
+
+#### Source attestations
+
+A source attestation contains information about how a specific revision was created and how it came to exist in its present context.
```
marcelamelara, line 199: What does "exist in the present context" mean?

docs/spec/draft/source-requirements.md
```diff
@@ -163,24 +163,108 @@ This is a typical setup for teams who leverage code review tools.
...
+They are associated with the revision identifier delivered to consumers and are a statement of fact from the perspective of the SCP.
```
marcelamelara, line 200: A nitpicky comment: Can we be a bit more explicit or detailed about why the SCP is the "right" perspective for these attestations?

docs/spec/draft/source-requirements.md
```diff
@@ -163,24 +163,108 @@ This is a typical setup for teams who leverage code review tools.
...
+If a consumer is authorized to access source on a particular branch, they MUST be able to fetch the source attestation documents for revisions in the history of that branch.
+
+It is possible that an SCP can make no claims about a particular revision.
+For example, this would happen if the revision was created on another SCP, or if the revision was not the result of an accepted change management process.
+
+#### Change management process
+
+The repo must define how the content of a [branch](#definitions) is allowed to change.
+This is typically done via the configuration of branch protection rules.
+It MUST NOT be possible to modify the content of a branch without following its documented process.
```
marcelamelara, line 211: What does "documented process" mean?

docs/spec/draft/source-requirements.md
```diff
@@ -163,24 +163,108 @@ This is a typical setup for teams who leverage code review tools.
...
+
+SLSA Source Level 2 ensures that all changes are recorded and attributable to an actor.
+SLSA Source Level 3 ensures that the documented process was followed.
+
+The change management tool MUST be able to authoritatively state that each new revision reachable from the protected branch represents only the changes reviewed via the process.
```
marcelamelara, line 216: This requirement isn't very clear. Is this alluding to code review?

docs/spec/draft/source-requirements.md
```diff
@@ -163,24 +163,108 @@ This is a typical setup for teams who leverage code review tools.
...
+
+The change management tool MUST provide at a minimum:
```
marcelamelara, line 218: Seems like something is missing after this.

docs/spec/draft/source-requirements.md
marcelamelara: Yes, it's better, thanks!

docs/spec/draft/source-requirements.md
```diff
@@ -230,19 +317,27 @@ but at SLSA Source Level 3 MUST used tamper-proof [detailed attestations](#detai
 -   Users SHOULD check that the expected `SLSA_SOURCE_LEVEL_` is listed within `verifiedLevels`.
 -   Users MUST ignore any unrecognized values in `verifiedLevels`.
 
-### Detailed attestations
+### Full attestations
 
-Source level evidence is the tamper-proof (ideally signed in-toto attestations) that can be used to determine what SLSA Source Level a given revision meets.
+Full attestations provide tamper-proof evidence (ideally signed in-toto attestations) that can be used to determine
```
marcelamelara, line 322: Nit: Add a link to the attestation model?

marcelamelara (commented):
Thanks for the updates! My main high-level suggestion is that the source track ought to make the roles and requirements of the VCS vs. SCP vs. producer clearer. In the Build track, the distinction between what the hosted build platform vs the producer is responsible for is called out. In the current source track spec, I feel like there are a lot of assumptions/expectations about the SCP, the producer and the VCS that we aren't including right now. So it might be helpful to draw a clearer separation between who is responsible for achieving which requirements.

docs/spec/draft/source-requirements.md
```diff
@@ -244,3 +244,101 @@ Branches may have differing security postures, and a change can be approved for
...
+
+## Communicating source levels
+
+SLSA source level details are communicated using attestations.
+These attestations either refer to a source revision itself or provide context needed to evaluate an attestation that _does_ refer to a revision.
+
+There are two broad categories of source attestations within the source track:
+
+1.  Summary attestations: Used to communicate to downstream users what high level security properties a given source revision meets.
+2.  Full attestations: Provide trustworthy, tamper-proof, metadata with the necessary information to determine what high level security properties a given source revision has.
```
zachariahcox, line 256: number 1 makes sense to me.
Maybe "full" attestations should be "evidence" or "claims" attestations? 
The issuer records what it _saw_ but cannot say what it means due to conflicts of interest. 

IE, Github might be a claims issuer, but something else would be responsible for deciding if those claims "add up" to slsa level 3.

docs/spec/draft/source-requirements.md
```diff
@@ -244,3 +244,101 @@ Branches may have differing security postures, and a change can be approved for
...
+
+To provide interoperability and ensure ease of use, it's essential that the summary attestations are applicable across all Source Control Platforms.
+Due to the significant differences in how SCPs operate and how they may chose to meet the Source Track requirements it is preferable to
+allow for flexibility with the full attestations.  To that end SLSA leaves full attestations undefined and up to the SCPs to determine
+what works best in their environment.
```
zachariahcox, line 261: there might be a good middle ground here. we could recommend a claims writing format, but put few restrictions on what you put in the document. 
This would have nice corollaries with parts of the build attestation format where there are required, optional, and issuer-specific fields.

docs/spec/draft/source-requirements.md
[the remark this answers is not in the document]
zachariahcox: > Is the question: "instead of JSON, can you format this example as YAML?"

haha, no, this was just a poorly written question from me. 

> "can we change the existing VSA predicate type to something else?"

yes I didn't quite grok that there is already a predicate defined for these things somewhere else. I was wondering if we could unify the evidence and "summary" attestation format. 
The evidence attestations will already require some user-provided strings (like check names). 
I was thinking it might make sense to make the verification bits just another kind of evidence assertion.

docs/spec/draft/source-requirements.md
zachariahcox: I think the scenario is: 

1. a cd job is queued to deploy from main
2. on the server, main is evaluated and resolved to sha abc123
1. 5 seconds later on the deployment runner, main no longer resolves to sha abc123 because main has since moved.
3. the deployment runner must determine somehow whether is actually ok to deploy abc123 

options include: 

1. we could say that abc123 is "currently reachable from main" or even first-parent-reachable.
2. we could say that abc123 passed all the required policies to be deployed regardless of what currently points to it. 
4. we could say that abc123 is deployable because it was at one point pointed-to directly by main. 

Each of these approaches has some pros and cons.

docs/spec/draft/source-requirements.md
zachariahcox: what does "full" imply in this context?

docs/spec/draft/source-requirements.md
TomHennen: I'd be happy with that.  I think we'd had something like 'evidence' before but it was somewhat confusing.  Perhaps that's been sorted now?

@marcelamelara suggested 'full'.

I think @arewm had some concerns about the original wording, so maybe he has thoughts here too?

docs/spec/draft/source-requirements.md
TomHennen: Something like 1) "full attestations SHOULD use in-toto attestations to record their claims"

But I think you're thinking more along the lines of 2) "we should define a predicate type for these attestations and leave specific fields 'loose'".

I'd be happy with 2, but perhaps we can leave it to another PR?  I think that's going to be a _whole_ discussion.

docs/spec/draft/source-requirements.md
TomHennen: >  By default, with no fancy stuff, if there are future merged PRs that point to the same sha, we'd get two attestations. We wouldn't re-issue the first attestation.

I'd be fine with this (especially if we leave it open so that folks _can_ attest to multiple refs in the same attestation if they want).

> Maybe you'd only need ref names for if -certain- refs pointed to the sha?

Yes I think recording the protected refs would make sense?  Is that sufficient?

> @TomHennen we could consider writing a custom claim for systems that need those semantics. "merged_to_refs_heads_main" or equivalent. What do you think?

Making it optional would be fine with me, I think that specific name is too GitHub specific though. :)

docs/spec/draft/source-requirements.md
TomHennen: My preference is for 3 since that seems to match the intent of the cd job and would be least surprising for users.

Some helpful context: we have some cd systems that are triggered just by a commit sha, and _not_ by a branch.  So the CD system doesn't necessarily know what branch it's building.

Imagine if we did 1:

There are always going to be race conditions.  People likely change what main points to for all sorts of reasons.  It would probably be bad if they had to coordinate those changes with whatever the cd job was doing.

Imagine if we did 2:

Two branches with entirely different semantics but the same security policies could be substituted for one another.  A 'code review' policy is especially interesting because it is all about context and human judgement.  The machine doesn't evaluate if the code is good or not, it evaluates if some persons said it was good.  Those persons make that decision based on the context they have, and that context includes the branch something is intended to go to.  Imagine if a team decided they wanted to see what would happen if they replaced compiler enabled memory protections with some other new technique.  They keep the security policies the same, but the folks reviewing those changes see them in the context of merging to the long lasting `mem_performance_testing` branch.  With option 2, someone could trigger a build from `mem_performance_testing` branch and get it deployed to production just by referencing a sha from `mem_performance_testing`.  That would result in something potentially insecure being deployed.

docs/spec/draft/source-requirements.md
marcelamelara: I suggested "full" as a slightly clearer contrast to "summary"; what we had before was "detailed", which wasn't particularly clear. I think using the term "evidence attestation" would probably be ok, though I personally find "claim attestation" to be a little tautological since every attestation _is_ a claim.

docs/spec/draft/source-requirements.md
TomHennen: I thought I'd responded to this but it seems to have been lost...

Added a link to https://github.com/in-toto/attestation/blob/main/spec/v1/README.md

Let's address the name question in the other thread...

On the topic of who is issuing the attestations, I added

> These differences also mean that depending on the SCP and the repo's configuration the attestor may
vary from implementation to implementation, often because entities with the knowledge to issue them
may vary.  The authority that issues [summary-attestations](#summary-attestation) MUST understand
which entity should issue each full attestation type and ensure the full attestations come from the
appropriate attestor.

How's that?

TomHennen:
> one thing that occurred to me today while reading this is I've not yet seen any proposals for writing an attestation when a ref is moved but no revision is created. I'm not sure what the content would be like, -- we might not know much about the provenance of the revision at the ref update point.

My thought it: it copies the contents of the prior 'summary attestation' for that ref.

TomHennen:
Some notes from a chat with @zachariahcox (thanks Zac for taking these)

* potentially a scs [("source control system")] is the suite of services [e.g. we don't currently have a way to talk about the entire system as a whole]
* where do attestations come from? pr tool, source host tool, merge queue tool
* ideally all attestations for a thing would be in the same place, co-located with the artifact they describe.
* could we do that with github? we have a download-as-zip system
* we have ways to store files in git, like a separate repo or branch
* google tracks a per-package ruleset, describing what was supposed to have happened.
* at the end of a package build pipeline, the package-policy service issues a vsa for the package. in this mode, the vsa is generated a separately deployed service before the package is published.
* vsa service will be opinionated about which builders are allowed to achieve slsa level x, but then allow project maintainers to configure further settings.
## source branches
* some builders fetch branches
* do we save a new attestation whenever a ref is updated? Or whenever a revision is created? both of these may be supported by a pull request.
* on the experiment track, you may disable memory protection, but not on main.
* at google, global policies are insufficient, branch names are important context





[10:23](https://openssf.slack.com/archives/D078AG6PLJC/p1724336611292529)
more thoughts on the ref naming idea:
it's a pretty common approach to build a "build.yaml" workflow to validate changes.
The contents of that workflow are sometimes conditional based on the name of the current ref.
:point_up: in that scenario, the claim at the end would be "successful build.yaml workflow runid", but the meaning of that claim has become branch specific due to conditionals or other build inputs.
[10:24](https://openssf.slack.com/archives/D078AG6PLJC/p1724336676546369)
to your point, philosophical stuff aside, I do think that we should keep our attestations ref-name / context aware. I suspect our real question is where and how to incorporate those names and how we recommend using them :+1:
10:25
the "reviewers don't take their role seriously unless the target branch is main" problem is also very real!

docs/spec/draft/source-requirements.md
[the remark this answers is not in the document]
TomHennen: I'll assume we can close, but please reopen if you think otherwise. :)

docs/spec/draft/source-requirements.md
TomHennen: In our last discussion I think we agreed that having some way to reference the branch names is useful, so I think this can be resolved?

docs/spec/draft/source-requirements.md
TomHennen: I think we decided that ref names are useful to have, so I'll close this thread (since it's basically about 'why do we care').

docs/spec/draft/source-requirements.md
```diff
@@ -129,3 +129,86 @@ Trusted robots MAY be exempted from the code review process. It is RECOMMENDED t
...
+Due to the significant differences in how SCPs operate and how they may chose to meet the Source Track requirements it is preferable to
+allow for flexibility with the detailed attestations.  To that end SLSA leaves detailed attestations undefined and up to the SCPs to determine
+what works best in their environment.
+
+### Summary attestation
+
+Summary attestations are issued by some authority that has sufficient evidence to make the determination of a given
+revision's source level.
+
+The source track issues summary attestations using [Verification Summary Attestations (VSAs)](./verification_summary.md) as follows:
```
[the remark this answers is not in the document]
TomHennen: Somehow I missed this comment earlier.

I definitely don't intend to exclude SCAI as a 'full attestation' type here.  The verifier making the summary attestation should be able to use SCAI as evidence when it wants, and delegate trust to the various actors as it deems necessary.

So!  To try and clarify that I added a SCAI example to the "Full Attestation" section.

But I'm not sure if that's exactly what you were getting at.  Can you provide more feedback on this?

docs/spec/draft/source-requirements.md
zachariahcox: My personal opinion is that most of these "full" attestations are source provenance, and we should do the work to make sure we can use that term. It's a clear term to encapsulate "what the issuer saw related to the creation of this revision". 

"provenance" as a term doesn't apply to ref-update-only attestations, mentioned elsewhere. 
I understand that those kinds of attestations may have conventional meaning at some sites, critically ref-reachability when there are no new revisions being created is a very subtle boundary to define. 
When we need it, it's probably better to tie it to the code change review context rather than the "ref." 

I recommend "summary" and "provenance" attestations if we can swing it. 
If we cannot deal with this source-track use of "provenance", I recommend "summary" and "claims" (or "metadata?") attestations. @marcelamelara is right that "claims attestations" feels tautological. I'm not sure how to work around this -- we want to attest _evidence_ without attempting to say what it _means_.

docs/spec/draft/source-requirements.md
zachariahcox: we can definitely reference the branch names in the revision-creation claims -- it's not likely to be feasible to mint attestations on demand or on every ref update, but potentially on every closed pull request? 
@TomHennen maybe we can focus the discussion just on that part going forward?

trishankatdatadog (approved):
Personally, I like the idea of a VSA for the Source track. Approving as such, but will leave to community to decide.

docs/spec/draft/source-requirements.md
mlieberman85: I like "summary"

TomHennen: Discussed in the spec meeting.  Calling them 'provenance' attestations seems fine, but we'll need to make it clear that unlike the build track, there can be many different types of provenance attestations which will require some additional terminology work.

I'll send a PR and we can tweak the language elsewhere.

docs/spec/draft/source-requirements.md
TomHennen: Sure, let me make a proposal on how this happens.  Might be worth merging with other language that @zachariahcox added (?) about which branches on 'consumable'.  I'll have to go look for the specifics. (suggestion per @adityasaky ).

Another suggestion: _don't_ define how it's done here but instead leave it up to the implementing systems to define when they set these things.  However I think it's probably helpful to have _some_ minimum path that a system like GitHub could use (as a sort of 'existence proof').

WDYT @zachariahcox ?

TomHennen: Ok, I've made a concrete proposal. PTAL?

docs/spec/draft/source-requirements.md
TomHennen: Done.  I'll resolve this thread now.

docs/spec/draft/source-requirements.md
TomHennen: This might have been resolved by the recent change clarifying when source_branch is specified.  Thoughts?

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,122 @@ Branches may have differing security postures, and a change can be approved for
...
+
+-   VSAs for source revisions MUST follow [the standard method of VSA verification](./verification_summary.md#how-to-verify).
+-   Users SHOULD check that an allowed branch is listed in `subject.annotations.source_branches` to ensure the revision is from an appropriate context within the repository.
+-   Users SHOULD check that the expected `SLSA_SOURCE_LEVEL_` is listed within `verifiedLevels`.
+-   Users MUST ignore any unrecognized values in `verifiedLevels`.
+
+### Provenance attestations
+
+Source provenance attestations provide tamper-proof evidence (ideally signed [in-toto attestations](https://github.com/in-toto/attestation/blob/main/README.md))
+that can be used to determine what SLSA Source Level or other high level properties a given revision meets
```
zachariahcox, line 328:
```suggestion
that can be used to determine what SLSA Source Level or other high level properties a given revision meets.
```

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,122 @@ Branches may have differing security postures, and a change can be approved for
...
+This evidence can be used by an authority as the basis for issuing a [Summary Attestation](#summary-attestation).
+
+SCPs and VCSes may have different methods of operating that necessitate different forms of evidence.
+E.g. GitHub based workflows may need different evidence than Gerrit based workflows, which would both likely be different from workflows that
```
zachariahcox, line 332:
```suggestion
E.g. GitHub-based workflows may need different evidence than Gerrit-based workflows, which would both likely be different from workflows that
```

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,122 @@ Branches may have differing security postures, and a change can be approved for
...
+operate over Subversion repositories.
+
+These differences also mean that depending on the SCP and the repo's configuration the attestor may
```
zachariahcox, line 335: do we use the word "attestor" or stick with "issuer"? is there any connotation difference between the two?

docs/spec/draft/source-requirements.md
zachariahcox: for git repos, a revision is just a tree id in a big database. 
Preventing _trees_ from existing can only happen at push receive time and causes a ton of developer friction. 

Especially when teams rely on pull requests, the "reachable from this ref" security boundary is the most practical. It models whether a commit / tree id is included in a specific subset. That subset has certain properties. I think that's what we're saying by context here, but it might need some additional definition here!

docs/spec/draft/source-requirements.md
zachariahcox: maybe it should say "_this_ documented process?" It's supposed to reference: 

>The repo must define how the content of a [branch](#definitions) is allowed to change.

docs/spec/draft/source-requirements.md
```diff
@@ -163,24 +163,108 @@ This is a typical setup for teams who leverage code review tools.
...
+On VCS like git, the organization MUST enable branch protections that prohibit updating the branch to point to revisions that are not direct descendants of the current revision.
+At a minimum, this typically means preventing "force pushes" and "branch deletion."
+
+It MUST NOT be possible to delete the entire repository (including all branches) and replace it with different source.
+Exceptions are allowed via the [safe expunging process](#safe-expunging-process).
+
+#### Identity Management
+
+There exists an identity management system or some other means of identifying actors.
+This system may be a federated authentication system (AAD, Google, Okta, GitHub, etc) or custom implementation (gittuf, gpg-signatures on commits, etc).
```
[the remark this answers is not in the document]
zachariahcox: I hope we don't need to define things down to that level! 
It would be nice if we can just leave it at 

* Something or somebody stores and deals with source revisions: let's call that thing the "scp"
* The SCP needs to explain how it identifies the actors who do things and record what it saw them do.
* SCPs issue attestations of the above. VSAs can use them or not.

docs/spec/draft/source-requirements.md
zachariahcox: Are you thinking something like: 
"The SCP is responsible for storing and recording changes to the source."

docs/spec/draft/source-requirements.md
zachariahcox: yeah, we're trying not to say pull request. 

In practice, I tend to doubt there will be any practical attestation flow at scale that doesn't rely heavily on code review (making it an excellent context in which to structure work), but code review is not technically a requirement for the technology to be useful.

docs/spec/draft/source-requirements.md
zachariahcox: ### strong authentication: is one of the minimums. maybe it doesn't render very well?

docs/spec/draft/source-requirements.md
```diff
@@ -163,24 +163,108 @@ This is a typical setup for teams who leverage code review tools.
 
 ##### Continuity
 
-Benefits: A compromise of a single human or account does not result in compromise of the project, since all changes require review from two humans.
+For all branches intended for consumption, whenever a branch is updated to point to a new revision, that revision MUST document how it related to the previous revision.
+Exceptions are allowed via the [safe expunging process](#safe-expunging-process).
```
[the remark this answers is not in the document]
zachariahcox: I have the same concerns. At GitHub, we have regular requests to help customers remove sensitive data. Doing so in a git repo will break most systems that key off the commit id. 
It's not a great outcome, but it seems reasonable to support teams with legal obligations to do so.

docs/spec/draft/source-requirements.md
TomHennen: Oh good point.  Updated the language to fix that.  I also found an errant use of 'full' attestation that should have been 'provenance'.

docs/spec/draft/source-requirements.md
TomHennen: Ok, I think we're good here.  The `source_branches` field is a 'SHOULD' which I think does a good enough job of marking it optional.

We can address later if not.

docs/spec/draft/source-requirements.md
TomHennen: Tracking in https://github.com/slsa-framework/slsa/issues/1134

TomHennen:
> Thanks for the updates! My main high-level suggestion is that the source track ought to make the roles and requirements of the VCS vs. SCP vs. producer clearer. In the Build track, the distinction between what the hosted build platform vs the producer is responsible for is called out. In the current source track spec, I feel like there are a lot of assumptions/expectations about the SCP, the producer and the VCS that we aren't including right now. So it might be helpful to draw a clearer separation between who is responsible for achieving which requirements.

Good feedback, added to https://github.com/slsa-framework/slsa/issues/1128#issuecomment-2360198417 to be sure we address it before we release.

docs/spec/draft/source-requirements.md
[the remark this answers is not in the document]
TomHennen: Tracking in https://github.com/slsa-framework/slsa/issues/1135

docs/spec/draft/source-requirements.md
```diff
@@ -163,24 +163,108 @@ This is a typical setup for teams who leverage code review tools.
...
 
-## Source Attestations
+On VCS like git, the organization MUST enable branch protections that prohibit updating the branch to point to revisions that are not direct descendants of the current revision.
+At a minimum, this typically means preventing "force pushes" and "branch deletion."
```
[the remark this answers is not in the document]
TomHennen: Tracking in https://github.com/slsa-framework/slsa/issues/1136

docs/spec/draft/source-requirements.md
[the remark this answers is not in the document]
TomHennen: Tracking as a part of https://github.com/slsa-framework/slsa/issues/1133

docs/spec/draft/source-requirements.md
```diff
@@ -163,24 +163,108 @@ This is a typical setup for teams who leverage code review tools.
...
+
+Summary:
+A consumer can ask the SCP for everything it knows about a specific revision of a repository.
+The information is provided in a documented and tamper-resistant format.
+
+Intended for:
+Organizations that want strong guarantees and auditability of their change management processes.
+
+Benefits:
+Provides reliable information to policy enforcement tools.
```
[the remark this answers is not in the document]
TomHennen: Tracking in https://github.com/slsa-framework/slsa/issues/1137

docs/spec/draft/source-requirements.md
TomHennen: Tracking in https://github.com/slsa-framework/slsa/issues/1138

docs/spec/draft/source-requirements.md
TomHennen: Tracking in https://github.com/slsa-framework/slsa/issues/1139

docs/spec/draft/source-requirements.md
TomHennen: Tracking in https://github.com/slsa-framework/slsa/issues/1140

docs/spec/draft/source-requirements.md
TomHennen: Tracking in https://github.com/slsa-framework/slsa/issues/1141

TomHennen:
Ok, I think all outstanding comments have been addressed or turned into tracking issues.

Can we get one more approval and merge this draft?

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,122 @@ Branches may have differing security postures, and a change can be approved for
...
+These attestations either refer to a source revision itself or provide context needed to evaluate an attestation that _does_ refer to a revision.
+
+There are two broad categories of source attestations within the source track:
+
+1.  Summary attestations: Used to communicate to downstream users what high level security properties a given source revision meets.
+2.  Provenance attestations: Provide trustworthy, tamper-proof, metadata with the necessary information to determine what high level security properties a given source revision has.
+
+To provide interoperability and ensure ease of use, it's essential that the summary attestations are applicable across all Source Control Platforms.
+Due to the significant differences in how SCPs operate and how they may chose to meet the Source Track requirements it is preferable to
+allow for flexibility with the full attestations.  To that end SLSA leaves full attestations undefined and up to the SCPs to determine
```
adityasaky, line 255:
```suggestion
allow for flexibility with the full attestations.  To that end SLSA leaves provenance attestations undefined and up to the SCPs to determine
```

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,122 @@ Branches may have differing security postures, and a change can be approved for
...
+    -   git branches MUST be fully qualified (e.g. `refs/head/main`) to reduce the likelyhood of confusing downstream tooling.
+4.  `resourceUri` MUST be set to the URI of the repository, preferably using [SPDX Download Location](https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field).
+E.g. `git+https://github.com/foo/hello-world`.
+5.  `verifiedLevels` MUST include the SLSA source track level the verifier asserts the revision meets. One of `SLSA_SOURCE_LEVEL_0`, `SLSA_SOURCE_LEVEL_1`, `SLSA_SOURCE_LEVEL_2`, `SLSA_SOURCE_LEVEL_3`.
+MAY include additional properties as asserted by the verifier.  The verifier MUST include _only_ the highest SLSA source level met by the revision.
+6.  `dependencyLevels` MAY be empty as source revisions are typically terminal nodes in a supply chain.
+
+Verifiers MAY issue these attestations based on their understanding of the underlying system (e.g. based on design docs, security reviews, etc...),
+but at SLSA Source Level 3 MUST used tamper-proof [full attestations](#full-attestations) appropriate to their SCP when making the assessment.
```
adityasaky, line 278:
```suggestion
but at SLSA Source Level 3 MUST used tamper-proof [provenance attestations](#provenance-attestations) appropriate to their SCP when making the assessment.
```

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,122 @@ Branches may have differing security postures, and a change can be approved for
...
+-   A [Vuln Scan attestation](https://github.com/in-toto/attestation/blob/main/spec/predicates/vuln.md)
+    which describes the results of a vulnerability scan over the contents of the revision.
+-   A [Test Results attestation](https://github.com/in-toto/attestation/blob/main/spec/predicates/test-result.md)
+ which describes the results of any tests run on the revision.
+-   An [SPDX attestation](https://github.com/in-toto/attestation/blob/main/spec/predicates/spdx.md)
+ which provides a software bill of materials for the revision.
+-   A [SCAI attestation](https://github.com/in-toto/attestation/blob/main/spec/predicates/scai.md) used to
+ describe which source quality tools were run on the revision.
+
+TODO: Can we define or recommend any canonical formats?
```
adityasaky, line 358: I think this can be removed now!

adityasaky (approved):
Some nits, but I think we can merge this! Thanks @TomHennen!

docs/spec/draft/source-requirements.md

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,120 @@ Branches may have differing security postures, and a change can be approved for
...
+    -   git branches MUST be fully qualified (e.g. `refs/head/main`) to reduce the likelyhood of confusing downstream tooling.
+4.  `resourceUri` MUST be set to the URI of the repository, preferably using [SPDX Download Location](https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field).
+E.g. `git+https://github.com/foo/hello-world`.
+5.  `verifiedLevels` MUST include the SLSA source track level the verifier asserts the revision meets. One of `SLSA_SOURCE_LEVEL_0`, `SLSA_SOURCE_LEVEL_1`, `SLSA_SOURCE_LEVEL_2`, `SLSA_SOURCE_LEVEL_3`.
+MAY include additional properties as asserted by the verifier.  The verifier MUST include _only_ the highest SLSA source level met by the revision.
+6.  `dependencyLevels` MAY be empty as source revisions are typically terminal nodes in a supply chain.
+
+Verifiers MAY issue these attestations based on their understanding of the underlying system (e.g. based on design docs, security reviews, etc...),
+but at SLSA Source Level 3 MUST used tamper-proof [provenance attestations](#provenance-attestations) appropriate to their SCP when making the assessment.
```
marcelamelara, line 278: Tiny typo
```suggestion
but at SLSA Source Level 3 MUST use tamper-proof [provenance attestations](#provenance-attestations) appropriate to their SCP when making the assessment.
```

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,120 @@ Branches may have differing security postures, and a change can be approved for
...
+allow for flexibility with the full attestations.  To that end SLSA leaves provenance attestations undefined and up to the SCPs to determine
+what works best in their environment.
+
+### Summary attestation
+
+Summary attestations are issued by some authority that has sufficient evidence to make the determination of a given
+revision's source level.  Summary attestations convey properties about the revision as a whole and summarize properties computed over all
+the changes that contributed to that revision over its history.
+
+The source track issues summary attestations using [Verification Summary Attestations (VSAs)](./verification_summary.md) as follows:
```
marcelamelara, line 264: Eventually, I think this list should be moved closer to the VSA spec itself so that users can reference the schema and these specific requirements together.

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,120 @@ Branches may have differing security postures, and a change can be approved for
...
+
+Verifiers MAY issue these attestations based on their understanding of the underlying system (e.g. based on design docs, security reviews, etc...),
+but at SLSA Source Level 3 MUST used tamper-proof [provenance attestations](#provenance-attestations) appropriate to their SCP when making the assessment.
+
+The SLSA source track MAY create additional tags to include in `verifiedLevels` which attest
+to other properties of a revision (e.g. if it was code reviewed).  All SLSA source tags will start with `SLSA_SOURCE_`.
+
+#### Populating source_branches
+
+The summary attestation issuer may chose to populate `source_branches` in any way they wish.
```
marcelamelara, line 285: Tiny typo
```suggestion
The summary attestation issuer may choose to populate `source_branches` in any way they wish.
```

docs/spec/draft/source-requirements.md
```diff
@@ -237,3 +239,120 @@ Branches may have differing security postures, and a change can be approved for
...
+that can be used to determine what SLSA Source Level or other high level properties a given revision meets.
+This evidence can be used by an authority as the basis for issuing a [Summary Attestation](#summary-attestation).
+
+SCPs and VCSes may have different methods of operating that necessitate different forms of evidence.
+E.g. GitHub-based workflows may need different evidence than Gerrit-based workflows, which would both likely be different from workflows that
+operate over Subversion repositories.
+
+These differences also mean that depending on the SCP and the repo's configuration the issuers of
+provenance attestations may vary from implementation to implementation, often because entities with
+the knowledge to issue them may vary.  The authority that issues
```
marcelamelara, line 337: I'm wondering if we can be a tad more concrete than "Entities with the knowledge to issue them". Is this about the SCP's components, or are these entities more at an organizational level? Or something else I might be missing?

marcelamelara (approved):
Thanks @TomHennen ! I only have minor comments at this point; they can either be resolved quickly or in a future PR.

docs/spec/draft/source-requirements.md
TomHennen: Yeah I think that could make sense.  Filed https://github.com/slsa-framework/slsa/issues/1148

docs/spec/draft/source-requirements.md
TomHennen: Yeah that's a good point.  Perhaps we can address this when we define 'Source Control System' (see #1128)  where I have a proposal that also covers 'Source Attestation Issuer'.

Issue #1107, opened 2024-08-06
What is the best term to describe a build runtime environment's persistent storage


marcelamelara:
One of the requirements of #975 is to integrity measure the build environment's disk/filesystem at deployment time.

There has been [some](https://github.com/slsa-framework/slsa/pull/1051#discussion_r1663296966) [discussion](https://github.com/slsa-framework/slsa/pull/1051#discussion_r1611940068) on scope, nuance and different environment types. This issue moves the discussion to a dedicated place.

To begin resolving this, there seems to be an overall desire to 1) to determine the specific storage resources of the build environment that need to be integrity measured, and 2) to support VM and non-VM build environments. 

I'd prefer to identify the minimum set of resources to measure, and to find language that will allow us to achieve both without having to dive into implementations or special case each type of build environment (VM vs container vs bare metal vs etc).

What we want is to make sure that the initial state of the build runtime contains **at a minimum** the expected OS, pre-installed applications and build executor. One suggestion is to use the term "storage volume", since it covers both bootable and non-bootable persistent storage. This term also doesn't immediately exclude build caches, if the build platform wishes to additionally integrity-check cached content.

Issue #1010, opened 2023-12-12
Clarifications for using tracks sparingly


arewm:
In https://slsa.dev/spec/v1.1/principles, for the guidelines to use tracks sparingly, we state
> **Use tracks sparingly**. Additional tracks add extra complexity to SLSA, so a new track should be seen as a last resort. Each track should have a clear, distinct purpose with a crisply defined objective, such as trustworthy provenance for the [Build track](https://slsa.dev/spec/v1.1/levels#build-track). As a rule of thumb, a new track may be warranted if it addresses threats unrelated to another track. Try to avoid tracks that sound confusingly similar in either name or objective.

I have two lines of questioning

## Tracks and attestations

We state
> Each track should have a clear, distinct purpose with a crisply defined objective, such as trustworthy provenance for the Build track

This example with provenance and a build track creates a mapping in my mind that a track should have an associated attestation. This has been discussed generically in the specification calls in the past, but we have never come out and stated this as a pattern/anti-pattern. Should each track be associated with some specific type of attestation? If so, would the suitability of a specific attestation for informing the facts of some supply chain guarantees be a guideline as to whether requirements should be combined into an existing track instead of creating an additional track?

## Threat mitigation redundancy

I read 
> a new track may be warranted if it addresses threats unrelated to another track

to imply that if a proposed new track addresses threats unrelated to another track that it is likely not warranted. As supply chain security requires a defense-in-depth approach, I feel like this would be an anti-pattern. If we can identify multiple ways to protect against the same threat (and each of these ways can be implemented/tracked in their own rite), then we would have a stronger footing to stand on.

MarkLodato:
(This is a really good question. It took me a long time to think through the answer, and I'm still not 100% confident.)

> Should each track be associated with some specific type of attestation?

I think the answer is "not necessarily," but it probably depends on how you interpret "be associated with" and "some specific type of attestation". 😁 

As a reminder, the higher level principle is that each level/track focuses on a concrete security outcome.

The Build track is a bit unique because having trustworthy build metadata *is* the desired outcome. Here an "attestation" is needed because (1) this build metadata is not usually persisted in any other way and (2) we want to pass the attestation around so that downstream systems can make fast, reliable, and secure decisions based on that metadata.

I don't think an attestation is a defining characteristic of a track, but it might be the common way it's implemented in practice.

Let's consider the future Source track as an example. I can see three ways of implementing that have varying degrees of "attestations":

[...]

PR #1115, opened 2024-08-15
content: draft: Add attested build environment track

marcelamelara:
This PR introduces the following spec changes associated with https://github.com/slsa-framework/slsa/issues/975. The spec enhancements are being proposed as the new "build Environment track".

Spec changes: Adds new high-level build environment terminology and levels.

Part 1 of https://github.com/slsa-framework/slsa/issues/975 CC @paveliak

### DIFF
--- a/docs/spec/draft/attested-build-env-levels.md
+++ b/docs/spec/draft/attested-build-env-levels.md
@@ -168,7 +168,7 @@ integrity for build environments at the time of build image distrbution.
 <dt>Requirements<dd>
 
 -   Build Image Producer:
-    -   MUST automatially generate SLSA [Build L2] or higher
+    -   MUST automatically generate SLSA [Build L2] or higher
     Provenance for created build images (i.e., VM or container images).
     -   MUST allow independent automatic verification of a build image's SLSA
     Provenance. If the build image artifact cannot be published, for example
--- a/docs/spec/draft/attested-build-env-levels.md
+++ b/docs/spec/draft/attested-build-env-levels.md
@@ -29,18 +29,21 @@ execution context. In this track, provenance describes how a [build image]
 was created, how the [hosted] build platform deployed a build image in its
 environment, and the compute platform they used.
 
-| Track/Level   | Requirements | Focus
-| ------------- | ------------ | -----
-| [BuildEnv L0] | (none)       | (n/a)
-| [BuildEnv L1] | Signed build image provenance exists | Tampering during build image distribution
-| [BuildEnv L2] | Attested build environment instantiation | Tampering via the build platform's control plane
-| [BuildEnv L3] | Hardware-authenticated build environment | Tampering via the compute platform's host interface
-| [BuildEnv L4] | Runtime monitored build environment | Tampering by the build platform or compute platform during the build
+| Track/Level   | Requirements | Focus | Trust Root
+| ------------- | ------------ | ----- | ----------
+| [BuildEnv L0] | (none)       | (n/a) | (n/a)
+| [BuildEnv L1] | Signed build image provenance exists | Tampering during build image distribution | Signed build image provenance
+| [BuildEnv L2] | Attested build environment instantiation | Tampering via the build platform's control plane | The compute platform's host interface
+| [BuildEnv L3] | Hardware-attested build environment | Tampering via the compute platform's host interface | The compute platform's hardware
 
-> [!IMPORTANT]
-> The Environment track currently requires a [hosted] build platform.
-> A future version of this track may generalize requirements to cover local
-> build environments (e.g., developer laptop).
+> :warning:
+> The Build Environment track currently requires a [hosted] build platform.
+> A future version of this track may generalize requirements to cover bare-metal
+> build environments.
+
+> :grey_exclamation:
+> We may consider the addition of an L4 to the Build Environment track, which
+> covers hardware-attested runtime integrity checking during a build.
 
 ### Build environment model
 
@@ -55,16 +58,15 @@ A typical build environment will go through the following lifecycle:
 1.  *Build image creation*: A build image producer creates different build
     images through dedicated build process. For the SLSA Environment track,
     the build image producer outputs provenance describing this process.
-    2.  *Build environment instantiation*: The hosted build platform calls
-    into the *host interface* to create a new build environment from a given
-    build image. The *build agent* begins to wait for an incoming build
-    dispatch.
-    **[TODO: revise]** For the SLSA Environment track, the hosted build
-    platform attests to the *measurement* of the environment's *initial
+2.  *Build environment instantiation*: The hosted build platform calls
+    into the *host interface* to create a new instance of a build environment
+    from a given build image. The *build agent* begins to wait for an incoming
+    build dispatch. For the SLSA Environment track, the host interface in the
+    compute platform attests to the integrity of the environment's *initial
     state* during its boot process.
 3.  *Build dispatch*: When the tenant dispatches a new build, the hosted
     build platform assigns the build to a created build environment.
-    **[TODO: revise]** For the SLSA Environment track, the build platform
+    For the SLSA Environment track, the build platform
     attests to the binding between a build environment and *build ID*.
 4.  *Build execution*: Finally, the *build executor* running within the
     environment executes the tenant's build definition.
@@ -84,9 +86,9 @@ and roles:
 | Build dispatch | The process of assigning a tenant's build to a pre-deployed build environment on a hosted build platform.
 | Compute platform | The compute system and infrastructure underlying a build platform, i.e., the host system (hypervisor and/or OS) and hardware. In practice, the compute platform and the build platform may be managed by the same or distinct organizations.
 | Host interface | The component in the compute platform that the hosted build platform uses to request resources for deploying new build environments, i.e., the VMM/hypervisor or container orchestrator.
-| Boot process | In the context of builds, the process of loading and executing the layers of firmware and/or software needed to start up a build environment on the build platform.
+| Boot process | In the context of builds, the process of loading and executing the layers of firmware and/or software needed to start up a build environment on the host compute platform.
 | Measurement | The cryptographic hash of some component or system state in the build environment, including software binaries, configuration, or initialized run-time data.
-| Quote | Hardware-signed data that contains one or more hardware-generated measurements. Quotes may additionally include nonces for replay protection, firmware information, or other platform metadata.
+| Quote | (Virtual) hardware-signed data that contains one or more (virtual) hardware-generated measurements. Quotes may additionally include nonces for replay protection, firmware information, or other platform metadata.
 | Reference value | A specific measurement used as the good known value for a given build environment component or state.
 
 TODO: Disambiguate similar terms (e.g., image, build job, build runner)
@@ -178,8 +180,10 @@ integrity for build environments at the time of build image distrbution.
 
 -   Build Platform:
     -   MUST meet SLSA [Build L2] requirements.
-    -   Prior to deployment of a new build environment, the SLSA Provenance
-    for the selected build image SHOULD be automatically verified.
+    -   Prior to the instantiation of a new build environment, the SLSA Provenance
+    for the selected build image MUST be automatically verified.
+    A signed attestation to the result of the SLSA Provenance verification MUST
+    be generated and distributed (e.g., via a [VSA]).
 
 <dt>Benefits<dd>
 
@@ -196,9 +200,9 @@ source and build process.
 <dl class="as-table">
 <dt>Summary<dd>
 
-The build environment is measured and authenticated prior to dispatching
-any builds, attesting to the integrity of initial state of the environment
-when it's deployed by the build platform.
+The build environment is measured and authenticated by the compute platform
+upon instantiation, attesting to the integrity of the initial state
+of the environment prior to executing a build.
 
 <dt>Intended for<dd>
 
@@ -212,32 +216,46 @@ All of [BuildEnv L1], plus:
 -   Build Image Producer:
     -   Build images MUST be created via a SLSA [Build L3] or higher build
     process.
-    -   **[TODO: revise]** MUST add support in the build image to:
-        -   Automatically check build image components against their
-        reference values during build environment startup.
-        In VM-based images, this can be achieved by enabling a [trusted boot]
-        process. In container-based build images, a dedicated program MUST be
-        used to perform these checks.[^1]
-        -   Automatically generate and distribute a signed attestation that
-        the initial state of a given build environment instance has been
-        verified against the corresponding build image, incl. its SLSA
-        Provenance (e.g., using [SCAI] or a [VSA]).
     -   MUST automatically generate and distribute signed reference values
     for the following build image components: bootloader or equivalent,
-    guest kernel, build agent, build executor, and root filesystem.
+    guest kernel, build agent, build executor, and root filesystem (e.g.,
+    via the image's SLSA Provenance, or [SCAI]).
     Additional build image components whose initial state is to be checked
     MAY be also measured.
+    -   The running build environment MUST be capable of:
+        -   Upon completion of the boot process: Automatically interfacing
+        with the host interface to obtain a signed quote for the
+        environment's initial state.
+        -   Upon build dispatch: Automatically generating and distributing
+        a signed attestation that binds its boot process quote to the
+        assigned build ID (e.g., using [SCAI]).
 
--   **[TODO: revise]** Build Platform Requirements:
+-   Build Platform Requirements:
     -   MUST meet SLSA [Build L3] requirements.
-    -   Prior to deployment of a new build environment, the SLSA Provenance
-    for the selected build image MUST be automatically verified. A signed
-    attestation to the verification of the build image's SLSA Provenance
-    MUST be automatically generated and distributed (e.g., via a [VSA]).
-    -   Prior to dispatching a tenant's build to a deployed environment, its
-    initial state attestation MUST be automatically verified. A signed
-    attestation binding the tenant's build ID to the verified initial state
-    of the selected build environment MUST be generated and distributed.
+    -   Prior to dispatching a tenant's build to an instantiated environment,
+    its boot process quote MUST be automatically verified. A signed
+    attestation to the result of the verification MUST be generated and
+    distributed (e.g., via a [VSA]).
+
+-   Compute Platform Requirements:
+    -   The host interface MUST be capable of generating signed quotes for
+    the build environment's system state.
+    In a VM-based environment, this MUST be achieved by enabling a feature
+    like [vTPM], or equivalent, in the hypervisor.
+    For container-based environments, the container orchestrator MAY need
+    modifications to produce these attestations.
+    -   The host interface MUST validate the measurements of the build image
+    components against their references values during the build environment's
+    boot process.
+    In a VM-based environment, this MUST be achieved by enabling a process
+    like [Secure Boot], or equivalent, in the hypervisor.
+    For container-based environments, the container orchestrator MAY need
+    modifications to perform these checks.[^1]
+    -   Prior to instantiating a new build environment, the host interface
+    MUST automatically verify the SLSA Provenance for the selected build
+    image. A signed attestation to the verification of the build image's
+    SLSA Provenance MUST be automatically generated and distributed (e.g.,
+    via a [VSA]).
 
 <dt>Benefits<dd>
 
@@ -251,58 +269,61 @@ adversary.
 </section>
 <section id="buildenv-l3">
 
-### BuildEnv L3: Hardware-authenticated build environment
+### BuildEnv L3: Hardware-attested build environment
 
 <dl class="as-table">
 <dt>Summary<dd>
 
-The initial state of the build environment is measured and autenticated by
-trusted hardware, and the build ID is verifiably bound to the
-deployed build environment, attesting to the integrity of the environment
-when a tenant's build is first dispatched.
+The initial state of the build's host environment is measured
+and autenticated by trusted hardware, attesting to the integrity
+of the build environment's underlying compute stack prior to executing
+a build.
 
 <dt>Intended for<dd>
 
-Organizations wanting strong assurances that their build was dispatched in
-a known good environment.
+Organizations wanting strong assurances that their build ran on a good
+known host environment.
 
 <dt>Requirements<dd>
 
 All of [BuildEnv L2], plus:
 
-**[TODO: These requirements need to be re-formulated.]**
-
 -   Build Image Producer:
-    -   MUST add support in the build image to:
-        -   Use trusted hardware to check build image component reference
-        values and integrity of the build environment startup.
-        -   Automatically generate and distribute a signed attestation
-        that a given build ID was dispatched in the corresponding build
-        environment instance, incl. its initial state attestation (e.g.,
-        using [SCAI] or a [VSA]).
-        -   Use trusted hardware to sign all build image-generated
-        attestations.
+    -   Upon completion of the boot process: The running build environment
+    MUST be capable of automatically interfacing with the *trusted hardware*
+    component to obtain a signed quote for the host interface's boot process
+    quote and the environment's initial state.
+    -   Upon build dispatch: The generated dispatch attestation MUST include
+    the host interface's boot process quote.
 
--   Build Platform Requirements: TODO
+-   Build Platform Requirements:
+    -   Prior to dispatching a tenant's build to an instantiated environment,
+    the *host interface's* boot process quote MUST be automatically verified.
+    A signed attestation to the result of the verification MUST be generated
+    and distributed (e.g., via a [VSA]).
 
--   Compute Platform Requirements: TODO
+-   Compute Platform Requirements:
+    -   MUST have trusted hardware capabilities, i.e., built-in physical
+    hardware features for generating measurements and quotes for its system
+    state. This SHOULD be achieved using a feature like [TPM],
+    [confidential computing], or equivalent.
+    -   MUST enable validation of the host interface's boot process against
+    its reference values. This MUST be achieved by enabling a process
+    like [Secure Boot], or equivalent, in the host platform.
 
 <dt>Benefits<dd>
 
-Provides hardware-authenticated evidence that a build running on a hosted
-build platform was dispatched in the expected build environment, even in the
-face of a compromised host interface (hypervisor/container orchestrator).
+Provides hardware-authenticated evidence that a build ran in the expected host
+environment, even in the face of a compromised host interface
+(hypervisor/container orchestrator).
 
 </dl>
 
 </section>
-<section id="buildenv-l4">
 
-### BuildEnv L4: Runtime monitored build environment
+## Considerations for Distributed Builds
 
 TODO
-
-</section>
 
 <!-- Footnotes -->
 
@@ -317,11 +338,11 @@ TODO
 [BuildEnv L1]: #buildenv-l1
 [BuildEnv L2]: #buildenv-l2
 [BuildEnv L3]: #buildenv-l3
-[BuildEnv L4]: #buildenv-l4
 [SCAI]: https://github.com/in-toto/attestation/blob/main/spec/predicates/scai.md
+[Secure Boot]: https://wiki.debian.org/SecureBoot#What_is_UEFI_Secure_Boot.3F
 [VSA]: verification_summary.md
 [build image]: #definitions
 [build model]: terminology.md#build-model
 [hosted]: requirements.md#isolation-strength
 [several classes]: #build-environment-threats
-[trusted boot]: https://csrc.nist.gov/glossary/term/trusted_boot
+[vTPM]: https://trustedcomputinggroup.org/about/what-is-a-virtual-trusted-platform-module-vtpm/
--- a/docs/spec/draft/attested-build-env-levels.md
+++ b/docs/spec/draft/attested-build-env-levels.md
@@ -56,7 +56,7 @@ environment and build platform.
 A typical build environment will go through the following lifecycle:
 
 1.  *Build image creation*: A build image producer creates different build
-    images through dedicated build process. For the SLSA Environment track,
+    images through a dedicated build process. For the SLSA Environment track,
     the build image producer outputs provenance describing this process.
 2.  *Build environment instantiation*: The hosted build platform calls
     into the *host interface* to create a new instance of a build environment
@@ -80,7 +80,7 @@ and roles:
 | --- | ---
 | Build ID | An immutable identifier assigned uniquely to a specific execution of a tenant's build. In practice, the build ID may be a cryptographic key or other unique and immutable identfier (e.g., a UUID) associated with the build execution.
 | Build image | The template for a build environment, such as a VM or container image. Individual components of a build image include the bootable storage volume containing the build executor, a dedicated build agent, and pre-installed guest OS and packages.
-| Build image producer | The party that creates and distributes build images. In practice, the build image producer may be the hosted build platform or a third party in a BYO build image setting.
+| Build image producer | The party that creates and distributes build images. In practice, the build image producer may be the hosted build platform or a third party in a bring-your-own (BYO) build image setting.
 | Build executor | A platform-provided program dedicated to executing the tenant’s build definition, i.e., running the build, within the build environment. The build executor must be included in the build image's measurement.
 | Build agent | A program that interacts with the hosted build platform's control plane from within a running build environment. The build agent must be included in the build image's measurement.
 | Build dispatch | The process of assigning a tenant's build to a pre-deployed build environment on a hosted build platform.
--- a/docs/spec/draft/attested-build-env-levels.md
+++ b/docs/spec/draft/attested-build-env-levels.md
@@ -33,9 +33,9 @@ environment, and the compute platform they used.
 | ------------- | ------------ | -----
 | [BuildEnv L0] | (none)       | (n/a)
 | [BuildEnv L1] | Signed build image provenance exists | Tampering during build image distribution
-| [BuildEnv L2] | Attested build environment deployment | Tampering via the build platform's control plane
+| [BuildEnv L2] | Attested build environment instantiation | Tampering via the build platform's control plane
 | [BuildEnv L3] | Hardware-authenticated build environment | Tampering via the compute platform's host interface
-| [BuildEnv L4] | Encrypted build environment | Tampering and data leaks by the build platform or compute platform during the build
+| [BuildEnv L4] | Runtime monitored build environment | Tampering by the build platform or compute platform during the build
 
 > [!IMPORTANT]
 > The Environment track currently requires a [hosted] build platform.
@@ -55,16 +55,17 @@ A typical build environment will go through the following lifecycle:
 1.  *Build image creation*: A build image producer creates different build
     images through dedicated build process. For the SLSA Environment track,
     the build image producer outputs provenance describing this process.
-2.  *Build environment deployment*: The hosted build platform calls into the
-    *host interface* to deploy a new build environment from a given build
-    image on the underlying compute platform.
-    For the SLSA Environment track, the hosted build platform attests to the
-    *measurement* of the environment's *initial state* during its boot
-    process.
+    2.  *Build environment instantiation*: The hosted build platform calls
+    into the *host interface* to create a new build environment from a given
+    build image. The *build agent* begins to wait for an incoming build
+    dispatch.
+    **[TODO: revise]** For the SLSA Environment track, the hosted build
+    platform attests to the *measurement* of the environment's *initial
+    state* during its boot process.
 3.  *Build dispatch*: When the tenant dispatches a new build, the hosted
-    build platform assigns the build to a deployed build environment. For
-    the SLSA Environment track, the build platform attests to the binding
-    between a build environment and *build ID*.
+    build platform assigns the build to a created build environment.
+    **[TODO: revise]** For the SLSA Environment track, the build platform
+    attests to the binding between a build environment and *build ID*.
 4.  *Build execution*: Finally, the *build executor* running within the
     environment executes the tenant's build definition.
 
@@ -76,10 +77,10 @@ and roles:
 | Primary Term | Description
 | --- | ---
 | Build ID | An immutable identifier assigned uniquely to a specific execution of a tenant's build. In practice, the build ID may be a cryptographic key or other unique and immutable identfier (e.g., a UUID) associated with the build execution.
-| Build image | The template for a build environment, such as a VM or container image. Individual components of a build image include the bootable storage volume containing the build executor, a dedicated build platform client, and pre-installed guest OS and packages.
+| Build image | The template for a build environment, such as a VM or container image. Individual components of a build image include the bootable storage volume containing the build executor, a dedicated build agent, and pre-installed guest OS and packages.
 | Build image producer | The party that creates and distributes build images. In practice, the build image producer may be the hosted build platform or a third party in a BYO build image setting.
 | Build executor | A platform-provided program dedicated to executing the tenant’s build definition, i.e., running the build, within the build environment. The build executor must be included in the build image's measurement.
-| Build platform client | A platform-provided program that interfaces with the hosted build platform's control plane from within a running build environment. The build platform client must be included in the build image's measurement.
+| Build agent | A program that interacts with the hosted build platform's control plane from within a running build environment. The build agent must be included in the build image's measurement.
 | Build dispatch | The process of assigning a tenant's build to a pre-deployed build environment on a hosted build platform.
 | Compute platform | The compute system and infrastructure underlying a build platform, i.e., the host system (hypervisor and/or OS) and hardware. In practice, the compute platform and the build platform may be managed by the same or distinct organizations.
 | Host interface | The component in the compute platform that the hosted build platform uses to request resources for deploying new build environments, i.e., the VMM/hypervisor or container orchestrator.
@@ -211,7 +212,7 @@ All of [BuildEnv L1], plus:
 -   Build Image Producer:
     -   Build images MUST be created via a SLSA [Build L3] or higher build
     process.
-    -   MUST add support in the build image to:
+    -   **[TODO: revise]** MUST add support in the build image to:
         -   Automatically check build image components against their
         reference values during build environment startup.
         In VM-based images, this can be achieved by enabling a [trusted boot]
@@ -223,11 +224,11 @@ All of [BuildEnv L1], plus:
         Provenance (e.g., using [SCAI] or a [VSA]).
     -   MUST automatically generate and distribute signed reference values
     for the following build image components: bootloader or equivalent,
-    guest kernel, build platform client, build executor, and root filesystem.
+    guest kernel, build agent, build executor, and root filesystem.
     Additional build image components whose initial state is to be checked
     MAY be also measured.
 
--   Build Platform Requirements:
+-   **[TODO: revise]** Build Platform Requirements:
     -   MUST meet SLSA [Build L3] requirements.
     -   Prior to deployment of a new build environment, the SLSA Provenance
     for the selected build image MUST be automatically verified. A signed
@@ -269,7 +270,7 @@ a known good environment.
 
 All of [BuildEnv L2], plus:
 
-**TODO:** These requirements need to be re-formulated.
+**[TODO: These requirements need to be re-formulated.]**
 
 -   Build Image Producer:
     -   MUST add support in the build image to:
@@ -297,7 +298,7 @@ face of a compromised host interface (hypervisor/container orchestrator).
 </section>
 <section id="buildenv-l4">
 
-### BuildEnv L4: Encrypted build environment
+### BuildEnv L4: Runtime monitored build environment
 
 TODO
 
--- a/docs/spec/draft/attested-build-env-levels.md
+++ b/docs/spec/draft/attested-build-env-levels.md
@@ -1,0 +1,235 @@
+---
+title: Build Environment track
+description: This page gives an overview of the SLSA Build Environment track and its levels, describing their security objectives and general requirements.
+---
+
+
+## Track overview
+
+The SLSA Build Environment track describes increasing levels of integrity and
+trustworthiness of the <dfn>provenance</dfn> of a build's execution context.
+In this track, provenance describes how the [hosted] build platform built the
+base [build image], what build environment they deployed, and the hardware
+platform they used.
+
+| Track/Level   | Requirements | Focus |
+| ------------- | ------------ | ----- |
+| [Environment L0]  | (none)       | (n/a)    |
+| [Environment L1]  | Signed build image provenance exists | Tampering during build image distribution |
+| [Environment L2]  | Attested build environment deployment | Tampering via the build platform's control plane |
+| [Environment L3]  | Hardware-authenticated build environment | Tampering via the compute platform's host interface |
+| [Environment L4]  | Encrypted build environment | Tampering and data leaks by the build platform or compute platform during the build |
+
+> [!IMPORTANT]
+> The Environment track currently requires a [hosted] build platform.
+> A future version of this track may generalize requirements to cover local
+> build environments (e.g., developer laptop).
+
+### Build environment model
+
+<p align="center"><img src="images/build-env-model.svg" alt="Model Build Environment"></p>
+
+The Build Environment track expands upon the [build model] by explicitily
+separating the *build image* and *compute platform* from the abstract build environment
+and build platform.
+
+A typical build environment will go through the following lifecycle:
+
+1.  *Build image creation*: A hosted build platform creates different build
+    images through a separate build process. For the SLSA HABE Track, the
+    hosted build platform outputs SLSA provenance describing this process.
+2.  *Build environment deployment*: The hosted build platform calls into the
+    *host interface* to deploy a new build environment from a given build image
+    on the underlying compute platform.
+    For the SLSA HABE Track, the build platform attests to the *measurement*
+    of the environment's *boot process*.
+3.  *Build dispatch*: When the tenant dispatches a new build, the hosted
+    build platform assigns the build to a deployed build environment. For
+    the SLSA HABE Track, the build platform attests to the binding between
+    a build environment and *build ID*.
+4.  *Build execution*: Finally, the *build executor* running within the
+    environment executes the tenant's build definition.
+
+| Primary Term | Description
+| --- | ---
+| Build ID | An immutable identifier assigned uniquely to a specific execution of a tenant's build. In practice, the build ID may be a cryptographic key or other unique and immutable identfier (e.g., a UUID) associated with the build execution.
+| Build image | The template for a build runtime environment, such as a VM or container image. Individual components of a build image are provided by the hosted build platform, and include the bootable storage volume containing the build executor, a dedicated build platform client, and platform-provided pre-installed guest OS and packages.
+| Build executor | The platform-provided program dedicated to executing the tenant’s build definition, i.e., running the build, within the build image. The build executor must be included in the build image's measurement.
+| Build dispatch | The process of assigning a tenant's build to a pre-deployed build environment on a hosted build platform.
+| Compute platform | The compute system and infrastructure, i.e., the host system (hypervisor and/or OS) and hardware, underlying a build platform. In practice, the compute platform and the build platform may be managed by the same or distinct organizations.
+| Boot process | In the context of builds, the process of loading and executing the layers of firmware and software needed to start up a build environmenton the build platform.
+| Measurement | The cryptographic hash of some system state in the build environment, including software binaries, configuration, or initialized run-time data. Software layers that are commonly measured include the bootloader, kernel, and kernel cmdline.
+
+TODO: Disambiguate similar terms (e.g., image, build job, build runner)
+
+## Environment levels
+
+The lowest level only requires SLSA Build L2 (or higher) Provenance to exist
+for the build image, while higher levels provide increasing auditability of
+the build environment's properties and integrity of the generated provenance
+attestations. The highest levels introduce further requirements for hardware-
+assisted hardening aimed at reducing the trusted computing base of a build.
+
+The primary purpose of the Environment track is to enable [auditing] that a
+build was run in the expected compute environment. The build platform and
+producers can check attestations generated by the build platform against the
+expected properties for a given build environment. This enables any party to
+detect several classes of supply chain threats originating in the build
+environment [TODO: Link here].
+
+As in the Build track, the exact implementation of this track is determined
+by the build platform provider, whether they are a commercial CI/CD service,
+or enterprise organization. While this track describes
+general [TODO: minimum requirements], this track does not dictate the
+following implementation-specific details: the type of build environment
+environment, accepted attestation formats, the type of technologies used to
+meet L3/4 requirements, what SLSA [Build] level the *built artifacts* meet,
+how attestations are distributed, how build environments are identified, and
+what happens on failure.
+
+<section id="environment-l0">
+
+### Environment L0: No guarantees
+
+<dl class="as-table">
+<dt>Summary<dd>
+
+No requirements---L0 represents the lack of any sort of build environment provenance.
+
+<dt>Intended for<dd>
+
+Development or test builds of software that are built and run on the same
+machine, such as unit tests.
+
+<dt>Requirements<dd>
+
+n/a
+
+<dt>Benefits<dd>
+
+n/a
+
+</dl>
+</section>
+
+<section id="environment-l1">
+
+### Environment L1: Signed build image provenance exists
+
+<dl class="as-table">
+<dt>Summary<dd>
+
+The build image (i.e., VM or container image) used to instantiate the build
+environment has SLSA provenance showing how the image was built.
+
+<dt>Intended for<dd>
+
+Build platforms and organizations wanting to ensure a baseline level of
+integrity for build environments at the time of build image generation.
+
+<dt>Software Producer Requirements<dd>
+
+-   MUST run builds using a build image that was built by a hosted build
+    platform that meets Environment L1 requirements.
+
+-   SHOULD verify the build image's SLSA Build provenance for the selected
+    build image, and distribute evidence of the verification to consumers
+    (e.g., using a [VSA])
+
+<dt>Build Platform Requirements<dd>
+
+-   MUST automatially generate and distribute SLSA [Build L1] or higher
+    Provenance for its supplied build images (i.e., VM or container images).
+
+-   The initial state of the build environment's root storage volume
+    MUST be integrity measured and signed.
+
+-   MUST execute builds on infrastructure that meets SLSA [Build L2].
+
+<dt>Benefits<dd>
+
+-   Provides evidence that a build image provided by a hosted build platform
+    was built from the advertised source and build process.
+
+</dl>
+
+</section>
+<section id="environment-l2">
+
+### Environment L2: Attested build environment instantiation
+
+<dl class="as-table">
+<dt>Summary<dd>
+
+The deployed build environment is integrity measured and authenticated
+prior to giving access to the tenant, attesting to the initial state of the
+environment.
+
+<dt>Intended for<dd>
+
+Organizations wanting to ensure that a specific build is running
+in the expected build environment.
+
+All of [Environment L1], plus:
+
+<dt>Software Producer Requirements<dd>
+
+-   MUST run builds in a build environment on a hosted build platform that
+    meets Environment L2 requirements.
+
+-   SHOULD verify the build image's Build SLSA provenance for the selected
+    build image, and distribute evidence of the verification to consumers
+    (e.g., using a [VSA])
+
+<dt>Build Platform Requirements<dd>
+
+-   MUST automatially generate and distribute SLSA [Build L2] or higher
+    Provenance for its supplied build images (i.e., VM or container images).
+
+-   The boot process of each build environment MUST be cryptographically measured
+    and signed.
+
+-   MUST distribute the boot and root storage attestations to allow for
+    independent verification.
+
+-   MUST uniquely bind a build image to a build environment instance.
+
+-   MUST uniquely and immutably bind an instance of a build to its build
+    environment.
+
+-   MUST execute builds in an environment that meets SLSA [Build L3].
+
+<dt>Benefits<dd>
+
+-   Provides evidence that a build environment deployed by a hosted build
+    platform was initialized from the expected build image.
+
+</dl>
+
+</section>
+<section id="environment-l3">
+
+### Environment L3: Hardware-authenticated build environment
+TODO
+
+</section>
+<section id="environment-l4">
+
+### Environment L4: Encrypted build environment
+
+TODO
+
+</section>
+
+<!-- Link definitions -->
+
+[Build L1]: levels.md#build-l1
+[Build L2]: levels.md#build-l2
+[Build L3]: levels.md#build-l3
+[Environment L0]: #environment-l0
+[Environment L1]: #environment-l1
+[Environment L2]: #environment-l2
+[Environment L3]: #environment-l3
+[Environment L4]: #environment-l4
+[future versions]: future-directions.md
+[hosted]: requirements.md#isolation-strength
--- a/docs/spec/draft/attested-build-env-levels.md
+++ b/docs/spec/draft/attested-build-env-levels.md
@@ -100,7 +100,7 @@ TODO
 ## BuildEnv levels
 
 The primary purpose of the Build Environment (BuildEnv) track is to enable
-[auditing] that a build was run in the expected execution context.
+auditing that a build was run in the expected execution context.
 
 The lowest level only requires SLSA [Build L2] Provenance to
 exist for the build image, while higher levels provide increasing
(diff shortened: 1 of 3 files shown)


TomHennen:
FYI if you want to make the linter happy easily [this might help](https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md#markdown-style).  The linter caused me a ton of trouble until I figured this out.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+| Build dispatch | The process of assigning a tenant's build to a pre-deployed build environment on a hosted build platform.
+| Compute platform | The compute system and infrastructure, i.e., the host system (hypervisor and/or OS) and hardware, underlying a build platform. In practice, the compute platform and the build platform may be managed by the same or distinct organizations.
+| Boot process | In the context of builds, the process of loading and executing the layers of firmware and software needed to start up a build environmenton the build platform.
+| Measurement | The cryptographic hash of some system state in the build environment, including software binaries, configuration, or initialized run-time data. Software layers that are commonly measured include the bootloader, kernel, and kernel cmdline.
+
+TODO: Disambiguate similar terms (e.g., image, build job, build runner)
+
+## Environment levels
+
+The lowest level only requires SLSA Build L2 (or higher) Provenance to exist
```
TomHennen, line 67: nit: environment Level 1 seems to only require SLSA Build L1.

I can see the advantage of making L1 require SLSA Build L2 (but that might also be confusing. :)).

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+-   MUST run builds using a build image that was built by a hosted build
+    platform that meets Environment L1 requirements.
+
+-   SHOULD verify the build image's SLSA Build provenance for the selected
+    build image, and distribute evidence of the verification to consumers
+    (e.g., using a [VSA])
+
+<dt>Build Platform Requirements<dd>
+
+-   MUST automatially generate and distribute SLSA [Build L1] or higher
```
TomHennen, line 141: Can you clarify what you mean by 'distribute'?  Provenance has a lot of metadata in it and some business might be nervous about publishing the actual build provenance.  However they'd likely be totally fine with publishing VSAs for those images that show they were verified.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+    Provenance for its supplied build images (i.e., VM or container images).
+
+-   The initial state of the build environment's root storage volume
+    MUST be integrity measured and signed.
+
+-   MUST execute builds on infrastructure that meets SLSA [Build L2].
```
TomHennen, line 147: What is the difference between the "supplied build images" and the infrastructure itself being Build L2?  I suppose it might be that "all the other software in the service is Build L2"?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+The deployed build environment is integrity measured and authenticated
+prior to giving access to the tenant, attesting to the initial state of the
+environment.
+
+<dt>Intended for<dd>
+
+Organizations wanting to ensure that a specific build is running
+in the expected build environment.
+
+All of [Environment L1], plus:
```
TomHennen, line 173: nit: I think some of these requirements are duplicated from L1.  I can also see why we wouldn't focus on removing the duplication _now_ and would wait until it's closer to done.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+
+[Build L1]: levels.md#build-l1
+[Build L2]: levels.md#build-l2
+[Build L3]: levels.md#build-l3
+[Environment L0]: #environment-l0
+[Environment L1]: #environment-l1
+[Environment L2]: #environment-l2
+[Environment L3]: #environment-l3
+[Environment L4]: #environment-l4
+[future versions]: future-directions.md
```
TomHennen, line 234:
```suggestion
[VSA]: verification_summary.md
[future versions]: future-directions.md
```

TomHennen (commented):
I like where this is heading :)

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Yeah, this one has been tricky for us to articulate. Here's what we're trying to tease out. The build platform has some separate (internal) build process for the build images they supply to their users (e.g., GHA runner images), so the platform is essentially acting as a software producer. So for the VM images, we're saying that their SLSA Build Provenance needs to be at least Build L1. At the same time, since tenants will be running builds using these VM images on the hosted build platform, this implies that the build should produce at least Build L2 Provenance for any artifacts built on it.

So, put differently, we're trying to explicitly codify both the "hosted build platform" requirement as well as the need to collect SLSA Provenance for the VM/container images that that platform puts out there to run builds. 

A couple options I'm considering to address this confusion could be 1) SLSA Provenance for build images should also start at Build L2, so the VM images and artifacts built through the hosted platform have the same provenance level, or 2) we focus only on the "platform as a producer" aspect in this track, and say upfront that this track applies to Build L2 or L3 build platforms. I'm leaning towards option 2 because our intent is mainly to bolster the integrity guarantees for artifacts built in attested compute environments. WDYT?

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Fair point. Something that we haven't clearly expressed yet is that for higher levels, we have an expectation that the build platform will verify the Provenance and other attestations for the build images before taking certain actions. I'll amend L1 to support VSAs or full SLSA Provenance, as long as there's evidence about the build image provenance.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Yeah, we're mostly being extra explicit here. FWIW, I pretty much copied the build track levels file, which also has a lot of this duplication :)

docs/spec/draft/attested-build-env-levels.md
TomHennen: Do you think it might help if we considered the producer of the build images as potentially distinct from the build platform itself?  Would that help make the language more clear?

Then a) "the components of build platform must be SLSA Build Level 2+" and b) "images used to run user provided build processes must be Build Level X+, regardless of who produced them".  'b' seems important because surely users will want to bring their own image sometimes?

I could also imagine removing 'b' entirely from the track, and instead handling that as some sort of extension of the build track itself or however [the transitivity problem](https://slsa.dev/spec/v1.0/faq#q-why-is-slsa-not-transitive) is solved.

docs/spec/draft/attested-build-env-levels.md
TomHennen: Ha yeah.  Let me resolve this and we can leave it be for now.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: That's an interesting suggestion, thanks! I think that's already sort of where I was heading with my option 2 above, but I think it makes sense to more cleanly separate the two roles. The only part that's still trickier in my mind is disentangling the build image from the broader build environment once it's running on the platform, e.g. waiting in a runner pool to be dispatched, but I'll give this framing a try.

TomHennen: Oh something else to consider is that "components of the build platform" might wind up being _very large_ in a SaaS environment.  So it might help to be quite explicit about what is and is not in scope.

marcelamelara: I made some changes, PTAL!

docs/spec/draft/attested-build-env-levels.md
marcelamelara: With the split of build image producer and build platform, this does actually become L1 :) Hopefully that's slightly less confusing.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,320 @@
...
+        Provenance (e.g., using [SCAI] or a [VSA]).
+    -   MUST automatically generate and distribute signed reference values
+    for the following build image components: bootloader or equivalent,
+    guest kernel, build platform client, build executor, and root filesystem.
+    Additional build image components whose initial state is to be checked
+    MAY be also measured.
+
+-   Build Platform Requirements:
+    -   MUST meet SLSA [Build L3] requirements.
+    -   Prior to deployment of a new build environment, a signed attestation
```
paveliak, line 227: How this is different from L1 which states that `Prior to deployment of a new build environment, the SLSA Provenance for the selected build image MUST be automatically verified.` Does it add a "distribution" requirement or there is more to it?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,320 @@
...
+-   Build Image Producer:
+    -   MUST automatially generate SLSA [Build L1] or higher
+    Provenance for created build images (i.e., VM or container images).
+    -   MUST allow independent automatic verification of a build image's SLSA
+    Provenance. If the full Provenance document cannot be distributed, for
+    example due to intellectual property concerns, a [VSA] asserting the
+    build image's SLSA Provenance MUST be distributed instead.
+
+-   Build Platform:
+    -   MUST meet SLSA [Build L2] requirements.
```
haydentherapper, line 174: Why did you choose to have build image producer and build platform requirements differ with respect to the SLSA Build level? If they differ, then the overall security of the system ultimately degrades to the lowest SLSA level in my opinion. Given the diff between Build L1 and Build L2 is primarily signed provenance, could we bump up the Build Image Producer requirement to Build L2? (And similarly, for BuildEnv L2 as well, could both be Build L3?)

marcelamelara: This is something we've gone back and forth on, and we originally had the requirement to start at Build L3, but got some pushback about how that could set the bar too high for BuildEnv L1. We've also heard some confusion/concerns around the dependency on the build track in the buildEnv track. So the current choice to lower the requirement down to Build L1 for BuildEnv L1 was mostly to make the barrier to entry for the build image producer lower. Ultimately, I agree with you that starting at Build L2 provenance for BuildEnv L1 images makes more sense from an integrity standpoint, but I also think you raise a broader question about the transitivity of SLSA levels.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: At L2, the intent is to add the requirement to generate and distribute an attestation for the verification, while in L1, it's sufficient to have the verification, but it's not required to attest to this step. I agree the difference isn't clear enough right now. Maybe it makes more sense to remove the verification requirement from L1 since there's no evidence that the build platform perform it anyway?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,320 @@
...
+
+</section>
+<section id="buildenv-l2">
+
+### BuildEnv L2: Attested build environment instantiation
+
+<dl class="as-table">
+<dt>Summary<dd>
+
+The build environment is measured and authenticated prior to dispatching
```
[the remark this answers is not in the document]
marcelamelara: Thanks for catching this.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,320 @@
...
+    image on the underlying compute platform.
+    For the SLSA Environment track, the hosted build platform attests to the
+    *measurement* of the environment's *initial state* during its boot
+    process.
+3.  *Build dispatch*: When the tenant dispatches a new build, the hosted
+    build platform assigns the build to a deployed build environment. For
+    the SLSA Environment track, the build platform attests to the binding
+    between a build environment and *build ID*.
+4.  *Build execution*: Finally, the *build executor* running within the
+environment executes the tenant's build definition.
```
kommendorkapten, line 69: Nit but the indentation is missing on this line 😄

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,320 @@
...
+
+Build platforms and organizations wanting to ensure a baseline level of
+integrity for build environments at the time of build image distrbution.
+
+<dt>Requirements<dd>
+
+-   Build Image Producer:
+    -   MUST automatially generate SLSA [Build L1] or higher
+    Provenance for created build images (i.e., VM or container images).
+    -   MUST allow independent automatic verification of a build image's SLSA
```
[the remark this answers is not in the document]
marcelamelara: Interesting, this is definitely a different scenario. I think the way to handle this is for the build image producer to separately put out an explicit attestation that hash X corresponds to a certain build image. We can then articulate this in the spec as an additional requirement like "The build image producer MUST distribute sufficient information about the build image artifact to allow for verification of its provenance, either in the form of the artifact itself or an immutable identifier (e.g., an attestation to its hash)." WDYT?

docs/spec/draft/attested-build-env-levels.md
marcelamelara: I reconsidered this comment, and I actually think that the argument that the security (and therefore SLSA level) of the system degrades to the lowest level isn't really accurate in this case. This track intends to build upon the integrity of the Build track because all of the requirements stated here are needed in addition to the basic Build L2 requirements in the platform. The said, there's still a reason to bump the Provenance level at L1, but it's because the focus of L2 needs to be about attesting to the "deployment chain".

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: I'd made some changes to explicitly incorporate this scenario. PTAL!

TomHennen (commented):
Thanks for these changes!

This generally LGTM, I would also like to see what @deeglaze thinks about it too.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,348 @@
...
+        -   Upon completion of the boot process: Automatically interfacing
+        with the host interface to obtain a signed quote for the
+        environment's initial state.
+        -   Upon build dispatch: Automatically generating and distributing
+        a signed attestation that binds its boot process quote to the
+        assigned build ID (e.g., using [SCAI]).
+
+-   Build Platform Requirements:
+    -   MUST meet SLSA [Build L3] requirements.
+    -   Prior to dispatching a tenant's build to an instantiated environment,
```
deeglaze, line 235: You can verify a quote's signature without caring about the contents it measured. I think we need to say that the measurements in the quote are verified against the signed reference values provided by the build producer for the instantiated build image.

I'm interested in how you plan on identifying what a booted environment is as a whole to compare against its components. I don't think there is a commonly accepted mechanism for this.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,348 @@
...
+    guest kernel, build agent, build executor, and root filesystem (e.g.,
+    via the image's SLSA Provenance, or [SCAI]).
+    Additional build image components whose initial state is to be checked
+    MAY be also measured.
+    -   The running build environment MUST be capable of:
+        -   Upon completion of the boot process: Automatically interfacing
+        with the host interface to obtain a signed quote for the
+        environment's initial state.
+        -   Upon build dispatch: Automatically generating and distributing
+        a signed attestation that binds its boot process quote to the
```
deeglaze, line 230: The binding should be to the boot, build job description and build job outputs, no?

If the build environment is meant to be from a clean slate every time, then the job description and outputs can be measured to integrity registers. I'm unclear about build ID. Is this not an in-toto signing key? If it is, I'd expect to see the key only released after the build is complete, following a quote binding to a challenge to get the key released to the build environment to sign the in-toto attestation.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,348 @@
...
+
+### Definitions
+
+The Build Environment track specifies the following supply chain components
+and roles:
+
+| Primary Term | Description
+| --- | ---
+| Build ID | An immutable identifier assigned uniquely to a specific execution of a tenant's build. In practice, the build ID may be a cryptographic key or other unique and immutable identfier (e.g., a UUID) associated with the build execution.
+| Build image | The template for a build environment, such as a VM or container image. Individual components of a build image include the bootable storage volume containing the build executor, a dedicated build agent, and pre-installed guest OS and packages.
```
thirumalareddym, line 82: When a container image is used as build environment, in practice those container images are executed on a VM which gets created using a VM image. In that scenario, does the requirements documented here for build image would apply for both the host VM image and container image?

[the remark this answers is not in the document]
marcelamelara: Yes, we're not currently covering builds that are distributed in any way, we have a big TODO for this. I opened #1150 to have the discussion specifically about multiple container images.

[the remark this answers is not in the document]
marcelamelara: Good question. I've actually seen cases in which the containers may _not_ be run inside of VMs, hence the current wording. I agree the current spec is ambiguous about the case you describe. I'll open an issue to address the question of container-in-VMs in the spec.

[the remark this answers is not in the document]
marcelamelara: See #1151

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+
+</section>
+<section id="environment-l2">
+
+### Environment L2: Attested build environment instantiation
+
+<dl class="as-table">
+<dt>Summary<dd>
+
+The deployed build environment is integrity measured and authenticated
```
deeglaze:
```suggestion
The deployed build environment is integrity-measured and authenticated
```

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+prior to giving access to the tenant, attesting to the initial state of the
+environment.
+
+<dt>Intended for<dd>
+
+Organizations wanting to ensure that a specific build is running
```
deeglaze:
```suggestion
Organizations wanting to ensure that an attestation policy–approved build is running
```

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+<dt>Build Platform Requirements<dd>
+
+-   MUST automatially generate and distribute SLSA [Build L2] or higher
+    Provenance for its supplied build images (i.e., VM or container images).
+
+-   The boot process of each build environment MUST be cryptographically measured
+    and signed.
+
+-   MUST distribute the boot and root storage attestations to allow for
+    independent verification.
```
deeglaze: I don't think this part ought to be required. The boot is measured and authenticated before giving access to the tenant, so perhaps we need only for the build agent to use credentials that root back to the attestation server, and that the certificate binds the agent to an immutable label for the policy the environment was evaluated under. If the build image has a unique name (such as a CoBOM tag UUID), then that could be included as well.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+
+-   MUST uniquely bind a build image to a build environment instance.
```
deeglaze: As in a build environment can only be identified as a uniquely identified single build image? What if there are two copies of the image with the same measurements but different endorsers? I'm not sure what "uniquely bind" means.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,235 @@
...
+build was run in the expected compute environment. The build platform and
+producers can check attestations generated by the build platform against the
+expected properties for a given build environment. This enables any party to
+detect several classes of supply chain threats originating in the build
+environment [TODO: Link here].
+
+As in the Build track, the exact implementation of this track is determined
+by the build platform provider, whether they are a commercial CI/CD service,
+or enterprise organization. While this track describes
+general [TODO: minimum requirements], this track does not dictate the
```
deeglaze: Whereas the track does not dictate the implementation-specific details, is there going to be a co-developed reference implementation of such formats/technologies to help folks build the analogy?

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: We've been going back and forth on what the best term is for the "storage volume" that covers both the container and VM use cases (there's an [open issue](#1107) for this discussion). For now, I _think_ we've landed on "root filesystem", as you noted we're using below, and it seems I missed to update this section. I'm going to make a note of your other comments in that issue.

You raise a very valid point re: the volume that contains the build executor and build agents. For now, at L2 we're requiring that those components be measured (and their measurements checked against their reference values ofc), but we aren't assuming that they be part of the bootable volume or even root fs.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,348 @@
...
+<dt>Requirements<dd>
+
+All of [BuildEnv L1], plus:
+
+-   Build Image Producer:
+    -   Build images MUST be created via a SLSA [Build L3] or higher build
+    process.
+    -   MUST automatically generate and distribute signed reference values
+    for the following build image components: bootloader or equivalent,
+    guest kernel, build agent, build executor, and root filesystem (e.g.,
```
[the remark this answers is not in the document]
marcelamelara: The intent for "root filesystem" is to be the `/` mount point. I suppose for windows environments, that would typically be `C:` ?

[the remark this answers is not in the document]
marcelamelara: Resolving the second part of this comment via #1164.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: The build hasn't executed at this point, so there are no build outputs to bind yet. The build ID is typically a unique opaque identifier generated by the build platform's control plane (e.g., GitHub's job ID), and is assigned to a build environment at dispatch time. We anticipate the build environment and build outputs will be indirectly bound via this build ID, which is included both in this "dispatch" quote and also in the resulting SLSA Build Provenance when the build completes.

What do you mean by build job outputs here? The return status, or the actual built artifacts. The former could certainly be easily measured into the integrity registers, the latter is more challenging in the case of very large built artifacts.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,348 @@
...
+    due to intellectual property concerns, an attestation asserting the
+    expected hash value of the build image MUST be generated and distributed
+    instead (e.g., using [SCAI] or a [Release Attestation]). If the full
+    Provenance document cannot be disclosed, a [VSA] asserting the build
+    image's SLSA Provenance MUST be distributed instead.
+
+-   Build Platform:
+    -   MUST meet SLSA [Build L2] requirements.
+    -   Prior to the instantiation of a new build environment, the SLSA Provenance
+    for the selected build image MUST be automatically verified.
```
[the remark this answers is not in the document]
marcelamelara: For now, we're implicitly defining the following simple policy: "Does the build image (via its hash) match my expectations (via its Provenance)"? We have an outstanding TODO to describe the verification process for all of the attestation types generated through this track, that will be in a separate doc. 

Beyond that, we're being very careful not to be overly prescriptive in the verification policies that SLSA users will ultimately define. We're not actually precluding an implementer of this track from bundling the build image and artifact Provenance attestations together (this is something in-toto readily supports), but because the build image isn't fully representative of the _instantiated_ build environment (e.g., because more components may be loaded at startup time, and each environment is uniquely identified by the build platform, even if it comes from the same base image), we can't simply treat the build environment as an input into the build.

docs/spec/draft/attested-build-env-levels.md
deeglaze: Build job outputs I'm just talking about the in-toto ResourceDescriptions of the built artifacts. The build integrity gets bound to the build environment integrity before permitting the signing of the build attestation as meeting a certain level. 

I'm misunderstanding the build ID and treating it as synonymous with the credentials to sign the build. Is there no SLSA-defined role for the build attester? I'm guessing the build agent will take on that role, but I haven't seen specific wording about signing. The RATS architecture would assign the role of Attester, I believe.

Even if the key is not released to the build environment itself and the in-toto attestation is meant to be signed remotely, either the key release request or the signing request that uses the L2-rated key ought to have measurements of both the build job description (measured before starting the build) and the build results description (the output resource descriptors) to permit use of the key. This can be accomplished with an appropriate attestation appraisal policy.

Both inputs and outputs get bound to the request to remove the need for a challenge, since there's nothing left to do if you replay in an attack environment. I don't think it's reasonable to require that the whole of the attestation request to the verification server gets stored. You just need an immutable label for the state of the appraisal policy (including the appraisal implementation provenance itself) and knowledge base of approved reference values.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Absolutely. We've been prototyping this here: https://github.com/chkimes/image-attestation

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,348 @@
...
+    -   Upon completion of the boot process: The running build environment
+    MUST be capable of automatically interfacing with the *trusted hardware*
+    component to obtain a signed quote for the host interface's boot process
+    quote and the environment's initial state.
+    -   Upon build dispatch: The generated dispatch attestation MUST include
+    the host interface's boot process quote.
+
+-   Build Platform Requirements:
+    -   Prior to dispatching a tenant's build to an instantiated environment,
+    the *host interface's* boot process quote MUST be automatically verified.
```
zachariahcox, line 301: for GHA, i guess this would be the runner verifying the host's boot attestation before downloading any customer-generated payload?

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
paveliak: Great question. It depends on how we draw the boundaries of the build environment. Speaking about CloudBuild do we stop at the build executor and treat all individual steps (including the docker images) as a tenant context? Or we measure them as the part of Build environment?

Ultimately CloudBuild runs in an environment that uses different kind of builders for each step. We can say that this _is_ the build environment and individual steps are the payload that may use builders that run things inside a Docker container or in a separate virtual machine or elsewhere. And those builders may in turn spin up other builders. Where do we stop?

I think the scope of this track is to focus on the top level environment, i.e. effectively measure CloudBuild, because as it was said in the beginning of this document we measure the Build Platform. Technically individual builders can also be treated as build environment providers and hence this track could be applied to them as well.

Perhaps it could be summarize with this analogy of nested environments. Practically this track could be applied at any nested level, at least for L1. L2 and L3 get more complicated because  the more we go deeper into the nesting levels the larger "build platform" surface becomes.
```
[ top level environment [docker run [docker in docker run [...]]]]

```

docs/spec/draft/attested-build-env-levels.md
marcelamelara: The short answer is no, it's not the runner performing the verification, it's the control plane of the CI platform.

Here's the general flow we're envisioning: 
1) The hypervisor (i.e., the host interface) boots up a new VM with some measured boot mechanism (commonly something like Secure Boot). This will generate a vTPM-signed attestation for the VM's boot process (plus, some extensions we introduce in this track), which the control plane retains for later verification.
2) When a new build request comes in, the control plane performs an authentication handshake with the build agent (not runner) in the VM. As part of this handshake, the agent will quote the vTPM again, and send the new attestation to the control plane.
3) The control plane will only dispatch a tenant's build into the VM if the content of the new quote matches the expected boot-time quote. 

I will caveat this by saying the intended audience for this doc is software developers/CISOs who need the broad strokes of the requirements. A separate "requirements" doc will specify the properties and protocols in greater detail for actual build or CI platforms looking to implement the track.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: This is an old requirement we took out in a more recent commit. The way GHA operates today is that each VM is assigned a authentication token/ID. So, the idea here was to bind that ID to the build image via an integrity measurement when the VM first starts up, so that you could validate later that a particular build job ran inside of the intended VM.

Again, we took out this specific piece, because it's strictly implementation-specific, and to simplify the requirements of L2. Now, we only require that the build _job_ ID be bound to the build image (as we were discussing above).

docs/spec/draft/attested-build-env-levels.md
marcelamelara: I think you were looking at an older commit, we've revised this requirement so that it's no longer required.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Got it. The build ID is not the Provenance signing key. The build ID is a random identifier that the control plane assigns to the build process. On GHA, for example, this is the 11-digit job number, whose primary purpose is to denote different runs of the same build process. The attestation signing credential is also issued by the control plane, but is typically mapped to the individual _user_ that triggered a build process, at least in the typical OIDC token-based PKI used by Sigstore and GHA.

To your point, SLSA doesn't currently follow the RATS architecture, so there isn't an explicit build attester role. Part of the reason for that is that the Build track doesn't even have a notion of an attested build at every level. 

> Both inputs and outputs get bound to the request to remove the need for a challenge, since there's nothing left to do if you replay in an attack environment. I don't think it's reasonable to require that the whole of the attestation request to the verification server gets stored. You just need an immutable label for the state of the appraisal policy (including the appraisal implementation provenance itself) and knowledge base of approved reference values.

Two thoughts here. First, I think it's definitely possible to make build environments less susceptible to replay attacks. It's actually something I've started thinking about for a while, and I think replay prevention needs to eventually become part of this BuildEnv track. 

Second, I agree that the ideal realization of attested builds would cover provenance for the appraisal  implementation, track the state of the appraisal policy, etc. as you suggest. I'll also say that handling policy is outside of the current scope of SLSA overall. If it's something that interests you, we do have an ongoing policy WG through the in-toto project where we're starting to look at exactly these kinds of problems. Please let me know if you'd like to get connect with that group.

marcelamelara: Tracking the various points raised in this thread in separate issues #1167 #1168 #1169

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: Tracking this in #1170.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: Tracking in #1169

docs/spec/draft/attested-build-env-levels.md
marcelamelara: This was been revised in a later commit.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: This was revised in a later commit.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Commit 209f139 should begin to address this. Tracking the greater details in #1169.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,352 @@
...
+TODO: Disambiguate similar terms (e.g., image, build job, build runner)
+
+### Build environment threats
+
+TODO
+
+## BuildEnv levels
+
+The primary purpose of the Build Environment (BuildEnv) track is to enable
+[auditing] that a build was run in the expected execution context.
```
TomHennen, line 103: this link doesn't seem to work

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,352 @@
...
+<dt>Requirements<dd>
+
+-   Build Image Producer:
+    -   MUST automatically generate SLSA [Build L2] or higher
+    Provenance for created build images (i.e., VM or container images).
+    -   MUST allow independent automatic verification of a build image's SLSA
+    Provenance. If the build image artifact cannot be published, for example
+    due to intellectual property concerns, an attestation asserting the
+    expected hash value of the build image MUST be generated and distributed
+    instead (e.g., using [SCAI] or a [Release Attestation]). If the full
```
TomHennen, line 177: [Release attestation] link doesn't seem to work.

TomHennen (changes requested):
This doesn't show up in the side bar on draft mode and I can't seem direct link to it either.  TBH I'm not entirely sure how all that works.

I'd expect to see it listed on the left here
https://deploy-preview-1115--slsa.netlify.app/spec/draft

And maybe to link to it here?

Other than these few things I think I'm ready to approve and iterate.

marcelamelara:
> This doesn't show up in the side bar on draft mode and I can't seem direct link to it either. TBH I'm not entirely sure how all that works.
> 
> I'd expect to see it listed on the left here
https://deploy-preview-1115--slsa.netlify.app/spec/draft

Thanks @TomHennen ! 

It's in a separate file at https://deploy-preview-1115--slsa.netlify.app/spec/draft/attested-build-env-levels. I'll add a link to it from the Build Track levels.md file, and fix the other links, too!

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Thanks for catching this. I'll remove it since the page this is pointing to doesn't exist yet.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Good catch, thanks!

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+
+TODO
+
+## BuildEnv levels
+
+The primary purpose of the Build Environment (BuildEnv) track is to enable
+auditing that a build was run in the expected execution context.
+
+The lowest level only requires SLSA [Build L2] Provenance to
+exist for the build image, while higher levels provide increasing
```
arewm, line 106: I read this initially as build L2 for the artifacts being generated, not for the build image. I think that this is because I was looking for a connection/commentary on how the BuildEnv and Build tracks support each other. Should we explicitly state this somewhere?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+computing base of a build.
+
+Software producers and third-party auditors can check attestations generated
+by the build image producer and build platform against the expected
+properties for a given build environment. This enables any party to detect
+[several classes] of supply chain threats originating in the build
+environment.
+
+As in the Build track, the exact implementation of this track is determined
+by the build platform provider, whether they are a commercial CI/CD service,
```
arewm, line 119:
```suggestion
by the build platform provider, whether they are a commercial CI/CD service
```

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+or enterprise organization. While this track describes general minimum
+requirements, this track does not dictate the following
+implementation-specific details: the type of build environment, accepted
+attestation formats, the type of technologies used to meet L3/4 requirements,
```
arewm, line 123: This is BuildEnv L3/L4? Should we include L4 here since there is no L4 yet?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+-   Build Image Producer:
+    -   MUST automatically generate SLSA [Build L2] or higher
+    Provenance for created build images (i.e., VM or container images).
+    -   MUST allow independent automatic verification of a build image's SLSA
+    Provenance. If the build image artifact cannot be published, for example
+    due to intellectual property concerns, an attestation asserting the
+    expected hash value of the build image MUST be generated and distributed
+    instead (e.g., using [SCAI] or a [Release Attestation]). If the full
+    Provenance document cannot be disclosed, a [VSA] asserting the build
+    image's SLSA Provenance MUST be distributed instead.
```
arewm, line 179: This makes a reference to a broader conversation that has been going on (i.e. claiming/proving/verifying SLSA for closed build environments). I feel like we should expose it in another location and just reference that here.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+
+-   Build Platform:
+    -   MUST meet SLSA [Build L2] requirements.
+    -   Prior to the instantiation of a new build environment, the SLSA
+    Provenance for the selected build image MUST be automatically verified.
+    A signed attestation to the result of the SLSA Provenance verification
+    MUST be generated and distributed (e.g., via a [VSA]).
```
arewm, line 186: What is the subject for this attestation? The artifact produced? 

Does this mean that the build platform will need to:
1. Verify the provenance of the build image
2. Detect that an artifact has been produced
3. Produce an attestation on that artifact with the VSA (i.e. whether L2 is passed/failed)

Some questions that come up:
* Should this identity be different from that of the build platform or can it be the same?
* Why does the verification need to happen before instantiation? Can it instead be verified after the fact (i.e. using the image recorded in a Tekton Chains-generated provenance if an immutable reference is used)?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+
+Provides evidence that a build image was built from the advertised
+source and build process.
+
+</dl>
+
+</section>
+<section id="buildenv-l2">
+
+### BuildEnv L2: Attested build environment instantiation
```
arewm, line 198: Starting with this level, it seems like the criteria are very different for VMs and containers. They were grouped previously for simplicity, but would it make sense to separate them here as the amount of mutually relevant information is smaller?

Containers do still have some infrastructure behind them (i.e. VMs/nodes running the containers/pods). How do those requirements apply to that infrastructure?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+        a signed attestation that binds its boot process quote to the
+        assigned build ID (e.g., using [SCAI]).
+
+-   Build Platform Requirements:
+    -   MUST meet SLSA [Build L3] requirements.
+    -   Prior to dispatching a tenant's build to an instantiated environment,
+    a signed quote MUST be automatically requested from the build agent,
+    and the contained measurements verified against their boot process
+    reference values. A signed attestation to the result of the verification
+    MUST be generated and distributed (e.g., via a [VSA]).
```
arewm, line 239: Are you saying that there must be an agent? This might not always be the case. 

If a workload is pushed to a VM via SSH (i.e. a build script), would the agent be the ssh daemon or some process invoked in the build script?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+</section>
+<section id="buildenv-l3">
+
+### BuildEnv L3: Hardware-attested build environment
+
+<dl class="as-table">
+<dt>Summary<dd>
+
+The initial state of the build's host environment is measured
+and autenticated by trusted hardware, attesting to the integrity
```
arewm, line 279:
```suggestion
and authenticated by trusted hardware, attesting to the integrity
```

arewm (changes requested):
Thanks for this PR! You have an immense amount of detail present in the change which can definitely be helpful in nuanced situations like these.

I feel like the page needs to be split up so that we can move information to locations that share the level of abstraction/detail, context, and overview. The build track, for example, has information split across:
* [Terminology](https://slsa.dev/spec/v1.0/terminology) documenting the models and definitions
* [Security levels](https://slsa.dev/spec/v1.0/levels) giving a high level overview of the levels, tracks, requirements, benefits, etc.
* [Producing artifacts](https://slsa.dev/spec/v1.0/requirements) giving a more detailed breakdown of the requirements with a discussion of the many nuances

Would it make sense to create a similar split of the information for the build environment track?

marcelamelara:
> I feel like the page needs to be split up so that we can move information to locations that share the level of abstraction/detail, context, and overview. The build track, for example, has information split across: Terminology, security levels, producing artifacts _[truncated for brevity]_


>
> Would it make sense to create a similar split of the information for the build environment track?

This is somewhat what we're already doing (and we have an even more detailed requirements and verification docs in the works). But I agree that we could split things more than we currently are. That being said, we're following a similar approach to the source track right now, which also coalesces all of these pieces (terminology, levels, detailed requirements) into a single page while in draft. 

More generally, from a website navigation standpoint, I think we may need to rethink how we want to organize pages, because I do think that having terminology, levels and requirements for each track on separate pages will become quite difficult/unwieldy to navigate for consumers of SLSA. The split we have right now works well because we have a single track and all of these pages are placed under the "Core specification" section of the website (see [left navigation pane](https://slsa.dev/)). With multiple tracks, we'll probably want to reorganize what falls under core spec versus track-specific sections.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+description: This page gives an overview of the SLSA Build Environment track and its levels, describing their security objectives and general requirements.
+---
+
+## Rationale
+
+Today's hosted build platforms play a central role in an artifact's supply
+chain. Whether it's a cloud-hosted service like GitHub Actions or an internal
+enterprise CI/CD system, the build platform has a privileged level of access
+to artifacts and sensitive operations during a build (e.g., access to
+producer secrets, provenance signing).
```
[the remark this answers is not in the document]
marcelamelara: Good point! We started using the term "hosted" mostly to stay consistent with the current spec, but are certainly not beholden to it if we eventually decide to use a different term. My current take (given our work on this track) is that "hosted" probably _is_ the right term, but that the current definition needs more context, so I'm definitely fine with using this track to help resolve #947. Happy to chime in on that issue to move things along.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+
+| Track/Level   | Requirements | Focus | Trust Root
+| ------------- | ------------ | ----- | ----------
+| [BuildEnv L0] | (none)       | (n/a) | (n/a)
+| [BuildEnv L1] | Signed build image provenance exists | Tampering during build image distribution | Signed build image provenance
+| [BuildEnv L2] | Attested build environment instantiation | Tampering via the build platform's control plane | The compute platform's host interface
+| [BuildEnv L3] | Hardware-attested build environment | Tampering via the compute platform's host interface | The compute platform's hardware
+
+> :warning:
+> The Build Environment track currently requires a [hosted] build platform.
```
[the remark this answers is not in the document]
marcelamelara: Should we just get rid of L0? It's a no-op level anyway.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+> A future version of this track may generalize requirements to cover bare-metal
+> build environments.
+
+> :grey_exclamation:
+> We may consider the addition of an L4 to the Build Environment track, which
+> covers hardware-attested runtime integrity checking during a build.
+
+### Build environment model
+
+<p align="center"><img src="images/build-env-model.svg" alt="Model Build Environment"></p>
```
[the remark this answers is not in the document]
marcelamelara: Yes, we're tracking figure creation in #1165.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+> :warning:
+> The Build Environment track currently requires a [hosted] build platform.
+> A future version of this track may generalize requirements to cover bare-metal
+> build environments.
+
+> :grey_exclamation:
+> We may consider the addition of an L4 to the Build Environment track, which
+> covers hardware-attested runtime integrity checking during a build.
+
+### Build environment model
```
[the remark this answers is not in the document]
marcelamelara: I'd like to, I mostly was trying to leave all BuidEnv track related changes in one place until, but if we are relatively comfortable with these new definitions (though we may continue iterating over them), I'm happy to move them over in a dedicated PR.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+possible to validate the integrity and trace the provenance of core build
+platform components.
+
+## Track overview
+
+The SLSA Build Environment (BuildEnv) track describes increasing levels of
+integrity and trustworthiness of the <dfn>provenance</dfn> of a build's
+execution context. In this track, provenance describes how a [build image]
+was created, how the [hosted] build platform deployed a build image in its
+environment, and the compute platform they used.
```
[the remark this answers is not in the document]
marcelamelara: We are (as is the source track), and the reason is that SLSA Build Provenance is really well suited for describing a build process (inputs, outputs, build ID), but not other aspects about the state of a running environment. That is, we're really expanding what constitutes the full provenance of an artifact: SLSA Build Provenance + build environment provenance 
 (+ source provenance) etc. There already exist other metadata formats that _can_ describe some of those build environment attributes we want to validate (e.g., SCAI), so we're proposing to use those rather than change the Build Provenance format. 

Perhaps what this track needs is an [added definition](https://github.com/slsa-framework/slsa/blob/main/docs/spec/draft/source-requirements.md#definitions) and a [more detailed description](https://github.com/slsa-framework/slsa/blob/main/docs/spec/draft/source-requirements.md#communicating-source-levels) of what build environment provenance is, like the source track currently does?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+
+The Build Environment track expands upon the [build model] by explicitily
+separating the *build image* and *compute platform* from the abstract build
+environment and build platform.
+
+A typical build environment will go through the following lifecycle:
+
+1.  *Build image creation*: A build image producer creates different build
+    images through a dedicated build process. For the SLSA Environment track,
+    the build image producer outputs provenance describing this process.
```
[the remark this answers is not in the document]
marcelamelara: Yes, this is just stating that the build image should itself have SLSA Build Provenance. The additional provenance formats come in at higher levels of the BuildEnv track.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+
+### Definitions
+
+The Build Environment track specifies the following supply chain components
+and roles:
+
+| Primary Term | Description
+| --- | ---
+| Build ID | An immutable identifier assigned uniquely to a specific execution of a tenant's build. In practice, the build ID may be a cryptographic key or other unique and immutable identfier (e.g., a UUID) associated with the build execution.
+| Build image | The template for a build environment, such as a VM or container image. Individual components of a build image include the root filesystem, pre-installed guest OS and packages, the build executor, and the build agent.
```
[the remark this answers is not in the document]
marcelamelara: The build image is the VM/disk image used to actually spin up a VM.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+| Build image producer | The party that creates and distributes build images. In practice, the build image producer may be the hosted build platform or a third party in a bring-your-own (BYO) build image setting.
+| Build executor | A platform-provided program dedicated to executing the tenant’s build definition, i.e., running the build, within the build environment. The build executor must be included in the build image's measurement.
+| Build agent | A program that interacts with the hosted build platform's control plane from within a running build environment. The build agent must be included in the build image's measurement.
```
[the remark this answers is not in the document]
marcelamelara: I think because we're assuming for the entire track that the build environment is running on a hosted build platform (as opposed to a bare metal/local builds), we do expect all environments to have a build agent in them. I'm happy to open a discussion issue for this expectation though, because it's very possible we're missing some important common use case here.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Oh, this is a typo. We had an L4, but took it out because its focus wasn't clear to us. I'll fix this, thanks for catching this.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+| Build dispatch | The process of assigning a tenant's build to a pre-deployed build environment on a hosted build platform.
+| Compute platform | The compute system and infrastructure underlying a build platform, i.e., the host system (hypervisor and/or OS) and hardware. In practice, the compute platform and the build platform may be managed by the same or distinct organizations.
+| Host interface | The component in the compute platform that the hosted build platform uses to request resources for deploying new build environments, i.e., the VMM/hypervisor or container orchestrator.
+| Boot process | In the context of builds, the process of loading and executing the layers of firmware and/or software needed to start up a build environment on the host compute platform.
+| Measurement | The cryptographic hash of some component or system state in the build environment, including software binaries, configuration, or initialized run-time data.
+| Quote | (Virtual) hardware-signed data that contains one or more (virtual) hardware-generated measurements. Quotes may additionally include nonces for replay protection, firmware information, or other platform metadata.
```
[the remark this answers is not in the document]
marcelamelara: It is. The most "official" definition for "quote" I could think of is in section 9.5.3.1 in [this doc](https://trustedcomputinggroup.org/wp-content/uploads/TPM-2.0-1.83-Part-1-Architecture.pdf), which is TPM-specific, but the term is used by many other hardware-based approaches now as well.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+    compute platform attests to the integrity of the environment's *initial
+    state* during its boot process.
+3.  *Build dispatch*: When the tenant dispatches a new build, the hosted
+    build platform assigns the build to a created build environment.
+    For the SLSA Environment track, the build platform
+    attests to the binding between a build environment and *build ID*.
+4.  *Build execution*: Finally, the *build executor* running within the
+    environment executes the tenant's build definition.
+
+### Definitions
```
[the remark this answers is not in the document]
marcelamelara: Agreed, it'd be a little more reader-friendly if you could at least link to the individual definitions. I do think that adding the figure at the top will help a ton because then the reader would at least be able to visually see how the components fit (as [in the current build model[(https://slsa.dev/spec/v1.0/terminology#build-model)). But there are different levels of to these definitions too: some are components, and some are security mechanisms. Maybe we could split those up?

And in general, the terminology page may benefit from some re-organization between what are standard components and what are security/integrity mechanisms.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+separating the *build image* and *compute platform* from the abstract build
+environment and build platform.
+
+A typical build environment will go through the following lifecycle:
+
+1.  *Build image creation*: A build image producer creates different build
+    images through a dedicated build process. For the SLSA Environment track,
+    the build image producer outputs provenance describing this process.
+2.  *Build environment instantiation*: The hosted build platform calls
+    into the *host interface* to create a new instance of a build environment
```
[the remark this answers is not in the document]
marcelamelara: big +1. I think I'll handle this in a separate PR just to keep this one about the content.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: Actually, I changed my mind. I'll move the terms to the terminology page in this PR.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: I'll try to address this as part of #1177 .

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: Tracking this in #1177 .

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+    attests to the binding between a build environment and *build ID*.
+4.  *Build execution*: Finally, the *build executor* running within the
+    environment executes the tenant's build definition.
+
+### Definitions
+
+The Build Environment track specifies the following supply chain components
+and roles:
+
+| Primary Term | Description
```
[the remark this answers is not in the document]
marcelamelara: "Build environment" is already defined in the current build model, but this comment just enforces the need to move all of the terminology definitions in one place.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: CC @paveliak : Thoughts?

docs/spec/draft/attested-build-env-levels.md
marcelamelara: I'm not sure I'm aware of this discussion, is there an issue tracking this?

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
arewm: In https://slsa.dev/spec/v1.0/levels, we have L0. In https://slsa.dev/spec/v1.0/requirements we do not. This table more closely resembles the former so it makes sense to keep it for consistency.

docs/spec/draft/attested-build-env-levels.md
arewm: #974 is at least one issue that I am aware of.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
arewm: I duplicated this comment later: https://github.com/slsa-framework/slsa/pull/1115/files/4610a2c6f55458a35479c81c9e37bc1a2421d3c9#r1790448929

One use case that I am aware of is to provision a VM, copy scripts over, and then run them remotely (i.e. via ssh).

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
arewm: > That is, we're really expanding what constitutes the full provenance of an artifact: SLSA Build Provenance + build environment provenance (+ source provenance) etc

I like something along these lines. While this isn't the first time, it is the first time that the repetition occurred to me. I'll follow up with a comment in #1010 as I think that issue is related as well.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
arewm: I think that we are in agreement then?
> the build image producer outputs provenance describing this process.

This describes the fact that the build image should have a SLSA build provenance attestation generated.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: We're using the definition of build platform according to our current [build model](https://slsa.dev/spec/v1.0/terminology#build-model), which includes the infrastructure, to my understanding. What is the distinction between infrastructure and build platform you're drawing? Do you mean the compute platform, per the definition in this doc?

But perhaps we're interpreting the L3 requirements differently? My understanding based on the [Unforgeable](https://slsa.dev/spec/v1.0/requirements#provenance-unforgeable) requirement at L3 is that the build platform is trusted to handle these sensitive operations, but it must not be the case that the _tenant_/build itself can access the secrets and provenance signing.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: Tracking this in #1185 .

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Tracking this in #1185

docs/spec/draft/attested-build-env-levels.md
marcelamelara: I think the source track has started a pretty good discussion on this, we might want to eventually convert that section into its one page of the spec.

TomHennen: Added a note to https://github.com/slsa-framework/slsa/issues/1148#issuecomment-2411165979 to try and capture this.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,304 @@
...
+    due to intellectual property concerns, an attestation asserting the
+    expected hash value of the build image MUST be generated and distributed
+    instead (e.g., using [SCAI] or a [Release Attestation]). If the full
+    Provenance document cannot be disclosed, a [VSA] asserting the build
+    image's SLSA Provenance MUST be distributed instead.
+
+-   Build Platform:
+    -   MUST meet SLSA [Build L2] requirements.
+    -   Prior to the instantiation of a new build environment, the SLSA
+    Provenance for the selected build image MUST be automatically verified.
```
zachariahcox, line 136: > MUST be automatically verified.
If it must be, do we have to say what about the image is verified? 
Just the VSA's "build level 2" claim about the build image?

docs/spec/draft/attested-build-env-levels.md
zachariahcox: > SLSA Provenance verification MUST be generated and distributed

This is a VSA'd VSA! 
Neat. 

In practice, I wonder if it would typically be a mega VSA that checks provenance on the entire SBOM and the build image.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,304 @@
...
+    A signed attestation to the result of the SLSA Provenance verification
+    MUST be generated and distributed (e.g., via a [VSA]).
+
+<dt>Benefits<dd>
+
+Provides evidence that a build image was built from the advertised
```
zachariahcox, line 142: is it also the case that we're including this build image as one of the "components" or dependencies of any artifacts built using the image? 
It seems to me that the answer should be "yes", but just checking!

docs/spec/draft/attested-build-env-levels.md
marcelamelara: It could be a VSA'd VSA :D Or a regular SLSA -> VSA step, depending on what's available from the build image producer. 

> What is the subject for this attestation? The artifact produced?

I'm expecting the subject for this attestation will be the same as the subject for the _verified_ SLSA Provenance, i.e., the build image that us being instantiated.

> Why does the verification need to happen before instantiation? 

We want to be able to ensure that a build _won't_ run in a problematic/unexpected/compromised environment before the build runs, rather than detect issues with the build image used to spin up the VM/container after the fact.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: You raise a really valid point. I think we need to think more carefully about the different requirements for containers vs VMs. I'll open up an issue to track this.

marcelamelara: Tracking this in #1192

docs/spec/draft/attested-build-env-levels.md
marcelamelara: I tagged #974 issue to associate it with this track, so we're going to use that to track any related to discussion to this.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: I think we can use #974 to track this. @arewm please let me know if you disagree.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: Just so I know whether this is resolved, would you like to see anything additional here? Should I add a link to the specification doc?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -0,0 +1,353 @@
...
+### Definitions
+
+The Build Environment track specifies the following supply chain components
+and roles:
+
+| Primary Term | Description
+| --- | ---
+| Build ID | An immutable identifier assigned uniquely to a specific execution of a tenant's build. In practice, the build ID may be a cryptographic key or other unique and immutable identfier (e.g., a UUID) associated with the build execution.
+| Build image | The template for a build environment, such as a VM or container image. Individual components of a build image include the root filesystem, pre-installed guest OS and packages, the build executor, and the build agent.
+| Build image producer | The party that creates and distributes build images. In practice, the build image producer may be the hosted build platform or a third party in a bring-your-own (BYO) build image setting.
```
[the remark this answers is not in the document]
marcelamelara: Which "party" are you referring to here? The new role of build image producer or the third party in the BYO case?

docs/spec/draft/attested-build-env-levels.md
marcelamelara: We're trying pretty hard not to introduce a circular dependency between the two tracks right now. So in the per-level requirements, we do state which Build track level the artifacts needs to be (via the build platform requirements), but the BuildEnv track doesn't currently support the Build track.

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Agreed, the answer should be yes. I'm tracking this in #1195

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Currently we say that it's the SLSA Provenance what needs to be verified, i.e. it's signature + contents. You're right though, that it could be the VSA if the build image Provenance isn't published. I'm tracking this change in #1196

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: Tracking the windows environment question in #1198

TomHennen (approved):
This is an amazing start.

I think we can continue to iterate on this draft in other issues & PRs.

Thanks Marcela!

TomHennen:
FYI updated title to indicate this is draft!

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
arewm: Sounds good.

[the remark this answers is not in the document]
arewm: Ah, right. https://slsa.dev/spec/draft/terminology 

The build platform includes the control plane which is where this information resides.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
arewm: You can add a link. I went ahead and resolved the thread.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
arewm: The party that creates and distributes the build images. When related to the build track, I would assume that this is the producer, but the working seems like it is a specific build platform.

docs/spec/draft/attested-build-env-levels.md
arewm: I don't think that we need to add a circular dependency. This track can depend on the build track, but the build track doesn't need to depend on the build environment track.

That doesn't mean that the build environment track cannot support increasing confidence in the isolation that the build platform claims.

docs/spec/draft/attested-build-env-levels.md
arewm: I don't see any need to have a VSA'd VSA. One could simply produce multiple VSAs as well.

> We want to be able to ensure that a build won't run in a problematic/unexpected/compromised environment before the build runs, rather than detect issues with the build image used to spin up the VM/container after the fact.

Is this not just an implementation detail? We want to know whether an artifact was built in an environment provisioned by a SLSA conformant build image. It doesn't matter whether this verification happens before instantiation or after as long as we can trust the post-verification to be able to determine the suitability of the build environment. 

This would, of course, likely be harder to do for verifying the environment of the trusted control plane. Are there any common tools/processes that can perform this verification AND generate the VSA before instantiation?

docs/spec/draft/attested-build-env-levels.md
marcelamelara: Tracking this in #1210

docs/spec/draft/attested-build-env-levels.md
marcelamelara: I'm tracking this as part of #1169

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
marcelamelara: Tracking this in #1211

arewm (approved):
Thanks for this draft content. Many of the topics that I have raised have moved to issues. Partially for my reference (because there were so many threads), I think the conversation has moved here:
* Potential clarification for hosted terminology: #947 
* Which parts of the build system should this apply to: #1211
* Clarifications of where/how to use provenance vs VSA: #974; #1148 
* Missing figure: #1165 
* Consistency of terminology/definitions: #1177 (additionally, build image producer should be explicit about the party/parties involved https://github.com/slsa-framework/slsa/pull/1115#discussion_r1790394534)
* Agent-less build environments: #1185 
* Connection between build and build environment track: #1210 
* Attestation flow including how/when to verify: #1169
* VM and container differentiation: #1192

PR #1123, opened 2024-08-27
editorial: Address some of the TODOs in threats and mitigation section

lehors:
This patch focuses on merely editorial changes such as moving sections around and changing 'out of scope' to 'not currently addressed'.

There are a lot more TODOs to go through but that's a first step...

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -15,9 +15,6 @@ The examples on this page are meant to:
 -   Help implementers better understand what they are protecting against so that
     they can better design and implement controls.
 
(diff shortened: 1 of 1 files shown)


lehors:
> I sure wish GitHub PRs did a better job of showing moved content.

Well, to be fair, our setup would be quite challenging for any source control system. I should have indicated that my plan is to make all the changes to the current draft first, because that's what I expect to carry all the history. I think of the version specific folders are essentially just snapshots so I plan to just update the 1.1 folder with 1 PR when we are ready to get 1.1-CR2 out.

marcelamelara (approved):
LGTM thanks!

PR #1124, opened 2024-08-27
impl: Fix Jekyll front matter

lehors:
I noticed that the jekyll build was producing an error because the `description` variable in the front matter of the `verifying-source.md` file contained multiple lines. This fixes the format so that there is no longer any error and as a consequence the page title is now displayed correctly.

However, I must admit that although every page contains a description I cannot find whether it is actually used anywhere...


arewm (approved):
The descriptions are used in page previews, i.e. https://github.com/slsa-framework/slsa/pull/695.

PR #1131, opened 2024-09-09
fix: Fix linter errors from blog post.

TomHennen:
1. Only check for duplicate headings in siblings. It's useful to have allow duplicated headings that aren't in siblings to let docs have parallel construction across the entire doc.
2. Remove absolute references to slsa.dev


joshuagl (approved):
I'm on the fence about this one. I probably would have opted to suppress the rule only for this blog post with code like:
```

PR #1143, opened 2024-09-20
content: draft: reliable -> authenticatable and auditable

### LINKED ISSUE #1137: Revisit 'reliable' language for L3
> "Reliable" may be too strong a qualifier here. It's still up to the consumer to decide if they deem the information as such, so we might want to instead use a term like "auditable" or "verifiable" here.

_Originally posted by @marcelamelara in https://github.com/slsa-framework/slsa/pull/1094#discussion_r1722231751_

TomHennen:
Clarify what type of data source level 3 provides to policy enforcement tools.

'authenticatable and auditable' is more easily understood by the community than 'reliable'.  It's also less prone to misinterpretation than 'verifiable' which might sound like a much more thorough process has been done.

fixes #1137

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -110,7 +110,7 @@ Intended for:
 Organizations that want strong guarantees and auditability of their change management processes.
 
 Benefits:
-Provides reliable information to policy enforcement tools.
+Provides authenticatable and auditable information to policy enforcement tools.
 
 ## System Requirements
 


adityasaky (approved):

marcelamelara (approved):

TomHennen:
Need one more maintainer to approve.  Maybe @arewm can help.

docs/spec/draft/source-requirements.md
```diff
@@ -183,7 +183,7 @@ Intended for:
 Organizations that want strong guarantees and auditability of their change management processes.
 
 Benefits:
-Provides reliable information to policy enforcement tools.
+Provides verifiable information to policy enforcement tools.
```
zachariahcox, line 186: I think we chose the word "reliable" specifically to avoid "verifiable."
I suspect these will never be fully _verifiable_ -- the data required to verify contributor info for instance is enormous and probably PII. 

I think the best we can do is make the claims "authentic," IE: we know who made the claims.

marcelamelara: My main concern with using "reliable" is that it's a subjective term, much how "trustworthy" is in the eye of the auditor/verifier. I do think "authentic" could work, as might "auditable".  @zachariahcox is right that the info may not ever be fully verifiable, but my view is that we're aiming to enable auditing (if not verification) of the process.

TomHennen: I like 'auditable'

TomHennen: Went with "authenticatable and auditable"

marcelamelara (approved):
LGTM thanks for the revisions @TomHennen !

trishankatdatadog (approved):

mlieberman85 (approved):
LGTM

PR #1152, opened 2024-09-24
Comment out non critical TODOs

lehors:
This is the next step into disposing of the TODOs in the threats and mitigation section.
It only leaves in a smaller set of TODOs that really need to be addressed for publication of SLSA 1.1.

The non TODOs are merely commented out so that they can still be considered in the future.

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -15,6 +15,7 @@ The examples on this page are meant to:
 -   Help implementers better understand what they are protecting against so that
     they can better design and implement controls.
 
+<!--
 **TODO:** Expand this threat model to also cover "unknowns". Not sure if that is
 a "threat" or a "risk". Example: If libFoo is compromised, how do you know if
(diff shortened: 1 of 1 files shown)


marcelamelara (approved):
LGTM, thanks @lehors ! How many critical TODOs are left at this point? They aren't displayed in the diff.

TomHennen (approved):
Thanks!

lehors:
> LGTM, thanks @lehors ! How many critical TODOs are left at this point? They aren't displayed in the diff.

Thanks @marcelamelara. Unfortunately it's not so easy to spot the TODOs that are left. You can see them on the threats and mitigations page of the deploy preview, after selecting the Draft version, if you open all the twisties and search for "TODO". I see 7 of them.
Most of them are about describing how a particular threat is being mitigated. This will take someone with more expertise than I have I'm afraid.

I'm happy to open an issue for each TODO left so that it makes it easier to track.

arewm (approved):

PR #1157, opened 2024-09-27
content: Update statement about compliance with SSDF

### LINKED ISSUE #1155: SSDF is no longer "Forthcoming"
just caught this in the about page:
>  A way to measure your efforts toward compliance with forthcoming Executive Order standards in the [Secure Software Development Framework (SSDF)](https://csrc.nist.gov/Projects/ssdf)

That looks like its no longer "Forthcoming" (since ~2022) -- but I haven't personally dived too deep into SSDF specifically

This is currently the case both in the 1.1 RC and the working draft version

lehors:
Close #1155.

### DIFF
--- a/docs/spec/draft/about.md
+++ b/docs/spec/draft/about.md
@@ -19,8 +19,7 @@ SLSA offers:
 -   A common vocabulary to talk about software supply chain security
 -   A way to secure your incoming supply chain by evaluating the trustworthiness of the artifacts you consume
 -   An actionable checklist to improve your own software's security
--   A way to measure your efforts toward compliance with forthcoming
-    Executive Order standards in the [Secure Software Development Framework (SSDF)](https://csrc.nist.gov/Projects/ssdf)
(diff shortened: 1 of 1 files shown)


arewm (approved):

arewm:
nit: Is this more of a `nonspec` change? Otherwise it would require an addition to the CHANGELOG. https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md#pull-request-types

lehors:
> nit: Is this more of a `nonspec` change? Otherwise it would require an addition to the CHANGELOG. https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md#pull-request-types

Indeed, I hesitated about the label to use for that matter. It's technically part of the spec but it's really in a non normative part. I don't think our policy is fine-grained enough to differentiate this from other more substantial changes so we're in unchartered territory. I decided to label it as content but I don't think it's worth an entry in the CHANGELOG.

TomHennen (approved):

PR #1166, opened 2024-09-30
content: source-track: rename SCP to SCS, replace open issues section with links to project and label queries

zachariahcox:
This change is in response to the 9.30 slsa specification meeting on this topic. 

A SCS is the full suite of services and ideas relied upon by the organization to create source revisions.

VCS stuff should mostly fall out of the discussion

Repositories can be used as the concept used when we need to talk about authN and authZ w.r.t. authentic  contributions.


docs/spec/draft/source-requirements.md
```diff
@@ -121,18 +110,18 @@ The source MUST have a location where the "official" revisions are stored and ma
 #### Revisions are immutable and uniquely identifiable
 
 This requirement ensures that a consumer can determine that the source revision they have is the same as a canonical revision.
-The combination of SCP and VCS MUST provide a deterministic way to identify a particular revision.
```
zachariahcox, line 124: @marcelamelara I definitely like having a single term for "the combination of". I'm not sure it really helps bring clarity to what we _mean_ by it, but "the system must provide X" is at least less clunky to read.

docs/spec/draft/verifying-source.md
```diff
@@ -168,7 +168,7 @@ It is not sufficient to indicate that a file changed without showing the content
 
 Require a squash merge strategy for the protected branch.
 
-To guarantee that only commits representing reviewed diffs are cloned, the SCP MUST rebase (or "squash") the reviewed diff into a single new commit (the "squashed" commit) that has only a single parent (the revision previously pointed-to by the protected branch).
+To guarantee that only commits representing reviewed diffs are cloned, the SCS MUST rebase (or "squash") the reviewed diff into a single new commit (the "squashed" commit) that has only a single parent (the revision previously pointed-to by the protected branch).
```
zachariahcox, line 171: edited this use case for completeness, but this one is "the tool that generates commits on behalf of the code review tool". 
For github (and probably other all-in-one systems, these are the same tool.

Issue #1142, opened 2024-09-20
Add 'profiles' for the Source Track


TomHennen:
@adityasaky suggested that it might make the source track more clear if we

1. Define 'profiles' for source control systems as a combination of VCS, SCP, attestation issuers, etc...
2. Each profile should explain how it would meet the various source level requirements
3. Remove many of the examples that are currently inlined with the requirements themselves

@adityasaky mentions this would be made easier once we define a source control system #1128 .

TomHennen:
This is probably a good way to address @marcelamelara's comment here https://github.com/slsa-framework/slsa/issues/1128#issuecomment-2360198417

TomHennen:
I'm wondering if we think this is a blocker for releasing the source track.  It seems like a good amount of work that we haven't made much progress on.  I also think that perhaps it's not required right now?  (Or we can address it in another point release)

adityasaky:
I think this really applies for presenting different ways of adopting the source track. I guess the question is if we expect to cut a release that includes the source track even if there aren't 1+ ways people can adopt it? (i.e., integrations into github, changes needed in gittuf for attestation pieces, etc.)

TomHennen:
I know some folks (myself included) were hoping to get source track out the door soon.  Since we haven't had any work done on profiles, I don't see how we can get both things (e.g. release soon with profiles).

Right now I don't think there's anything that limits source track adoption to just one way, it's more that we don't have any specific documented ways _to_ adopt it.

Perhaps we can:

1. Release without profiles.
2. Start implementing some methods of achieving it.
3. Document those methods as 'profiles'?
4. Cut a new release?

adityasaky:
I think not blocking on this is fine. I also wonder if this is really a cross-slsa issue rather than limited to the source track. We could use this idea to communicate "known ways" to implement any SLSA track. Eg: using github-attest-provenance in gha is an example for the build track.

marcelamelara:
How much more work is needed to define "profiles"? My interpretation of most of the early feedback on this topic was more related to clearly defining the roles and SCS model, which the Source track has already largely done.

adityasaky:
> My interpretation of most of the early feedback on this topic was more related to clearly defining the roles and SCS model, which the Source track has already largely done.

[...]

Issue #1136, opened 2024-09-19
Is 'branch protection' only applicable for cloud-hosted SCPs?


TomHennen:
To my knowledge, branch protection is a feature of specific cloud-hosted SCPs, not git/VCS, so I suggest making this requirement for Continuity more generally about the intent/objectives. Then we can maybe just include an example like "This can be achieved using the branch protection features in services like GitHub, for example."

_Originally posted by @marcelamelara in https://github.com/slsa-framework/slsa/pull/1094#discussion_r1722209349_

TomHennen:
FWIW I think that GitTuf might make this possible outside of a hosted SCP.  @adityasaky would know for sure.

zachariahcox:
good question! 
It's not branded in exactly the same way (because it's way more powerful), but git supports this kind of feature via [pre-receive githooks](https://git-scm.com/docs/githooks/2.27.0#pre-receive).

adityasaky:
I think this may in part be addressed with #1128 and #1142. I think #1142 in particular would allow us to set requirements that may be achieved wildly differently depending on what constitutes the source control system as a whole.

As for the rest of it, I think the requirement could use further clarification. The full set of branch protections (whether via an SCP, pre-receive hooks, etc.) could go quite a bit further than disallowing force pushes and deletions (which are stated as the baseline in the spec atm), so I think clarification is in order as to what the requirement exactly is. Maybe we could reuse this issue for that? 😄

adityasaky:
@zachariahcox I opened https://github.com/slsa-framework/slsa/pull/1175, let me know what you think!

bureado:
The wording in e1ac8548564bc671b66871fe61e21771ff8211f8 LGTM.

But honest Q: if the subject of the VSA already binds the (immutable) revision to the (mutable) human-friendly references at the time of VSA creation as part of the VSA's subject, what's the benefit of having the predicate trying to say the equivalent of "remote reported they implement a receive hook which enforces rejection of operations modifying linear history of a subset of branches deemed for consumption"? (e.g., I saw a `ref_protected` value in a JWS claim, or `GET /branches/<>/protection` at a given timestamp X, etc.)

[...]

PR #1175, opened 2024-10-03
content: draft: Clarify branch protection in source track

adityasaky:
Re: https://github.com/slsa-framework/slsa/issues/1136

I'm trying to separate "branch protection" which is typically associated with SCP features from the property we want for the branches intended for consumption. I've also applied the SCS terminology we adopted in another PR.

The framing of how this can be accomplished is still a bit TBD, if we decide to adopt something like #1142 suggests, we could move the Git + GitHub/GitLab piece out? That's also where we could discuss server side pre-receive hooks etc.?


TomHennen (approved):

marcelamelara (approved):
LGTM, this is a clearer description of the objective, thanks!

TomHennen:
If there are no objections let's plan to merge this today.

Issue #1187, opened 2024-10-14
Update threats.md to discuss SLSA Source Track


TomHennen:
There's an entire section in threats.md about the SLSA Source Track.  Before we finish the track we should update threats.md to talk about the source track specifically.

TomHennen:
I'm going to close this in favor of #1354

PR #1188, opened 2024-10-14
content: Add mitigation for malicious source platform admin.

### LINKED ISSUE #1179: TODO: Need mitigation description for "Platform admin abuses privileges" threat

TomHennen:
Add mitigation for malicious source platform admin.

We didn't have any guidance for this threat.  There are a number of ways we may be able to address this in the future via the SLSA Source Track and/or tools like gittuf.  However, SLSA doesn't currently address them.  This entire section is already labeled as not being handled by SLSA but does still include other mitigations.

I'm using the same language we have for "Compromise build platform admin", which seems like the same sort of threat, and should work 'fine' until we have something better.

Filed #1187 to track

Fixes #1179.

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -323,7 +323,9 @@ has no solution on GitHub.
 *Threat:* Platform administrator abuses their privileges to bypass controls or
 to push a malicious version of the software.
 
-*Mitigation:* TODO
+*Mitigation:* The source platform must have controls in place to prevent and
+detect abusive behavior from administrators (e.g. two-person approvals, audit
+logging).
 
 *Example 1:* GitHostingService employee uses an internal tool to push changes to
 the MyPackage source repo.


arewm (approved):
We have a build platform operations track in the future directions: https://slsa.dev/spec/v1.0/future-directions

Should we make a reference to that?

docs/spec/draft/threats.md
```diff
@@ -323,7 +323,9 @@ has no solution on GitHub.
 *Threat:* Platform administrator abuses their privileges to bypass controls or
 to push a malicious version of the software.
 
-*Mitigation:* TODO
+*Mitigation:* The source platform must have controls in place to prevent and
+detect abusive behavior from administrators (e.g. two-person approvals, audit
```
adityasaky, line 327: I would lean on audit logging more here than two-person approvals. I know platforms have ways to not allow admins to bypass protections but there remains a gap (to my knowledge) in that administrators can still unilaterally change the policy itself, say to remove two-person approval requirements or other protections. Audit logging would a way to detect / mitigate this, until platforms support multiple-approvals for policy changes.

adityasaky: Also: logging would apply to the infrastructure supporting the platform software as well, while two-person approval would not?

TomHennen: The two person approval here isn't related to changes to the source code, but rather requiring two persons to make changes to the underlying infrastructure.

Should I just make that more clear?

adityasaky: Yes, that'd help! My mind went to the "default" two party approval in the source context.

TomHennen: Done.  Also added example solutions.

docs/spec/draft/threats.md
```diff
@@ -323,20 +323,30 @@ has no solution on GitHub.
...
-*Mitigation:* TODO
+*Mitigation:* The source platform must have controls in place to prevent and
+detect abusive behavior from administrators (e.g. two-person approvals for
+changes to the infrastructure, audit logging). A future [Platform
+Operations Track](/future-directions#build-platform-operations-track) may
+provide more specific guidance on how to secure the underlying platform.
 
 *Example 1:* GitHostingService employee uses an internal tool to push changes to
-the MyPackage source repo.
+the MyPackage source repo. Solution 1: employee cannot push changes without
```
adityasaky, line 333:
```suggestion
the MyPackage source repo. Solution 1: employee cannot make changes to the MyPackage repository's data without
```
"push" again reminded me of the more typical two-party review, while a change to the data layer of the service to modify a repository is an infra change.

docs/spec/draft/threats.md
```diff
@@ -323,20 +323,30 @@ has no solution on GitHub.
...
+getting a co-worker to approve. Solution 2: changes are audited and the employee
+would be detected by internal security teams.
 
 *Example 2:* GitHostingService employee uses an internal tool to push a
 malicious version of the server to serve malicious versions of MyPackage sources
 to a specific CI/CD client but the regular version to everyone else, in order to
-hide tracks.
+hide tracks. Solution 1: employee cannot push malicious version without
```
adityasaky, line 340: I think example 2 and 3 need more detailed treatments. While the text proposed as solutions are not meant to be complete (they're specified as examples above) and the text above indicates more guidance for platforms may be published separately, I wonder if a reader may still think the solutions specified here suffice. For the case detailed here, I imagine we might want something closer to recursive SLSA where an employee cannot deploy arbitrary versions of the server software but instead it must be built and deployed using SLSA-compliant processes.

Maybe there shouldn't be inline solutions for each example?

docs/spec/draft/threats.md
adityasaky: This can be ignored if we drop inline examples as I propose in my other comment.

docs/spec/draft/threats.md
TomHennen: Yeah, I'd be happy to drop the inline solutions.  SLSA doesn't really cover this right now anyways, so we really just need to gesture to the right thing to do.

docs/spec/draft/threats.md
TomHennen: Removed!

docs/spec/draft/threats.md
TomHennen: Removed!

adityasaky (approved):
LGTM!

zachariahcox (approved):
great start. 

any package registry admin prior art would be pretty applicable here. 
We could also talk about use of externally verifiable data for certain use cases (like signed refs) or signed source provenance docs. 

like two-set-of-eyes rules, it makes attacks more expensive and difficult to accomplish.

TomHennen:
> great start.
> 
> any package registry admin prior art would be pretty applicable here. We could also talk about use of externally verifiable data for certain use cases (like signed refs) or signed source provenance docs.
> 
> like two-set-of-eyes rules, it makes attacks more expensive and difficult to accomplish.

FWIW the goal wasn't really to address this threat with SLSA, but just mention it as a thing folks should consider.  That doesn't mean we _can't_ address it in the future, just get rid of the TODO.

So, how complete did we want this to be for 1.1?

@lehors any thoughts

arewm (approved):

trishankatdatadog (approved):

lehors:
> > great start.
> > any package registry admin prior art would be pretty applicable here. We could also talk about use of externally verifiable data for certain use cases (like signed refs) or signed source provenance docs.
> > like two-set-of-eyes rules, it makes attacks more expensive and difficult to accomplish.
> 
> FWIW the goal wasn't really to address this threat with SLSA, but just mention it as a thing folks should consider. That doesn't mean we _can't_ address it in the future, just get rid of the TODO.
> 
> So, how complete did we want this to be for 1.1?
> 
> @lehors any thoughts

As far as I'm concerned, I think what you're proposing is enough for 1.1. This can always be improved as we move forward.

lehors (approved):

Issue #1180, opened 2024-10-09
TODO: Need to fill out threat and mitigation of "(G) Distribution channel"


lehors:

michaelwinser:
One mitigation to distribution channel risk is to build directly from source. If you can't be sure of the trustworthiness of the packages, package repositories and indicies, then building direclty from source is the only way to be sure.

PR #1190, opened 2024-10-14
content: draft: Flesh out 'Distribution Channel' threats.

TomHennen:
These threats generally match the threats for 'Artifact Publication' with
the twist that the consumer must do the verification instead.

Explicitly calling out that modification in transit to the consumer is in scope for this threat.

Consumer verification may be simplified if a VSA was issued at publication time.

fixes https://github.com/slsa-framework/slsa/issues/1180

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -664,7 +664,82 @@ cryptographic signature is no longer valid.
 An adversary modifies the package on the package registry using an
 administrative interface or through a compromise of the infrastructure.
 
-**TODO:**
+The distribution channel threats and mitigations look very similar to the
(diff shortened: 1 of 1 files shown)


TomHennen:
> FWIW I'm considering moving these things to usage and just pointing most of (G) there. Any thoughts welcome.

I take it back.  The 'Usage' section says "A usage threat is a potential for an adversary to exploit behavior of the consumer."  Verification would seem to be out of scope.

MarkLodato (approved):

lehors (approved):
Nicely done! Thanks @TomHennen

TomHennen:
Taking into account some review that @zachariahcox did on #1191 I think we're good to go here.

Issue #1182, opened 2024-10-09
TODO: Need to fill out description of "(I) Usage" in threat and mitigation section


lehors:

TomHennen:
My intuition is that the threats and mitigations from 'G' (#1190) will address the usage threats too.  I could duplicate those or just mention it.

Something else that we might want to include is "Improper/insecure use of legitimate software"

PR #1191, opened 2024-10-14
content: draft: Flesh out "Usage" threat

TomHennen:
There are two ways to look at the usage threat:

1. Can the attacker modify the software being delivered to a consumer.
2. Can the consumer use the software insecurly allowing an attacker
   to take advantage of that insecurity to exploit them.

IMO 1 has the same solutions as 'G' (PR https://github.com/slsa-framework/slsa/pull/1190).  I could repeat them
here under usage, but instead I've updated 'G' to include modification
in transit, and I've had 'Usage' address 2 above (albeit by just
deferring to CISA's work in this area).

fixes https://github.com/slsa-framework/slsa/issues/1182

NOTE: this PR is based on top of #1190 since the solution presented in 1190 obviates the need for addressing that here.

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -664,7 +664,82 @@ cryptographic signature is no longer valid.
 An adversary modifies the package on the package registry using an
 administrative interface or through a compromise of the infrastructure.
 
-**TODO:**
+The distribution channel threats and mitigations look very similar to the
+Artifact Publication (F) threats and mitigations with the main difference
+being that these threats are mitigated by having the _consumer_ perform
+verification.
+
+The consumer's actions may be simplified if (F) produces a [VSA][vsa].
+In this case the consumer may replace any provenance verification with
+[VSA verification][vsa_verification].
+
+<details><summary>Build with untrusted CI/CD <span>(expectations)</span></summary>
+
+*Threat:* Replace the package with one built using an unofficial CI/CD pipeline
+that does not build in the correct way.
+
+*Mitigation:* Verifier requires provenance showing that the builder matched an
+expected value or a VSA for corresponding `resourceUri`.
+
+*Example:* MyPackage is expected to be built on Google Cloud Build, which is
+trusted up to Build L3. Adversary builds on SomeOtherBuildPlatform, which is only
+trusted up to Build L2, and then exploits SomeOtherBuildPlatform to inject
+malicious behavior. Adversary then replaces the original package within the
+repository with the malicious package. Solution: Verifier rejects because
+builder is not as expected.
+
+</details>
+<details><summary>Upload package without provenance or VSA <span>(Build L1)</span></summary>
+
+*Threat:* Replace the original package with a malicious one without provenance.
+
+*Mitigation:* Verifier requires provenance or a VSA before accepting the package.
+
+*Example:* Adversary replaces MyPackage with a malicious version of MyPackage
+on the package repository and deletes existing provenance. Solution: Verifier
+rejects because provenance is missing.
+
+</details>
+<details><summary>Tamper with artifact after upload <span>(Build L1)</span></summary>
+
+*Threat:* Take a benign version of the package, modify it in some way, then
+replace it while retaining the original provenance or VSA.
+
+*Mitigation:* Verifier checks that the provenance or VSA's `subject` matches
+the hash of the package.
+
+*Example:* Adversary performs a proper build, modifies the artifact, then
+replaces the modified version of the package in the repository and retains the
+original provenance. Solution: Verifier rejects because the hash of the
+artifact does not match the `subject` found within the provenance.
+
+</details>
+<details><summary>Tamper with provenance or VSA <span>(Build L2)</span></summary>
+
+*Threat:* Perform a build that would not meet expectations, then modify the
+provenance or VSA to make the expectations checks pass.
+
+*Mitigation:* Verifier only accepts provenance with a valid [cryptographic
+signature][authentic] or equivalent proving that the provenance came from an
+acceptable builder.
+
+*Example 1:* MyPackage is expected to be built by GitHub Actions from the
+`good/my-package` repo. Adversary builds with GitHub Actions from the
+`evil/my-package` repo and then modifies the provenance so that the source looks
+like it came from `good/my-package`. Solution: Verifier rejects because the
+cryptographic signature is no longer valid.
+
+*Example 2:* Verifier expects VSAs to be issued by TheRepository. Adversary
+builds a malicious package and then modifies the original VSA's `subject`
+field to match the digest of the malicious package. Solution: Verifier rejects
+because the cryptographic signature is no longer valid.
+
+*Example 3:* Verifier expects VSAs to be issued by TheRepository. Adversary
+builds a malicious package and then issues a VSA of their own from
+EvilRepository for the malicious package. Solution: Verifier rejects because
+they only accept VSAs from TheRepository which the EvilRepository cannot
+issue since it does not have the corresponding signing key.
+</details>
 
 ## Usage threats
 
@@ -938,3 +1013,5 @@ collision resistance.
 [isolated]: requirements.md#isolated
 [unforgeable]: requirements.md#provenance-unforgeable
 [supply chain threats]: threats-overview
+[vsa]: verification_summary
+[vsa_verification]: verification_summary#how-to-verify
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -662,7 +662,8 @@ cryptographic signature is no longer valid.
 ### (G) Distribution channel
 
 An adversary modifies the package on the package registry using an
-administrative interface or through a compromise of the infrastructure.
+administrative interface or through a compromise of the infrastructure
+including modification of the package in transit to the consumer.
 
 The distribution channel threats and mitigations look very similar to the
 Artifact Publication (F) threats and mitigations with the main difference
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -788,7 +788,16 @@ solutions.
 
 ### (I) Usage
 
-**TODO:** What should we put here?
+The consumer uses a package in an unsafe manner.
+
+<details><summary>Improper usage</summary>
+
+*Threat:* The software can be used in an insecure manner, allowing an
+adversary to compromise the consumer.
+
+*Mitigation:* This threat is not addressed by SLSA, but may be addressed by
+efforts like [Secure by Design][secure-by-design].
+</details>
 
 ## Dependency threats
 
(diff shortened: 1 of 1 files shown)


TomHennen:
Hey @MarkLodato, if you have time can you take a  quick look at 'I' in this PR and double-check my logic?  I'd like to make sure I'm not missing anything (or if there's a good reason to include much of 'G' here too).

Thanks!

docs/spec/draft/threats.md
```diff
@@ -662,7 +662,8 @@ cryptographic signature is no longer valid.
 ### (G) Distribution channel
 
 An adversary modifies the package on the package registry using an
-administrative interface or through a compromise of the infrastructure.
+administrative interface or through a compromise of the infrastructure
+including modification of the package in transit to the consumer.
```
MarkLodato, line 666: nit: this should probably go in #1190 instead

docs/spec/draft/threats.md
```diff
@@ -787,7 +788,16 @@ solutions.
...
+The consumer uses a package in an unsafe manner.
+
+<details><summary>Improper usage</summary>
+
+*Threat:* The software can be used in an insecure manner, allowing an
+adversary to compromise the consumer.
+
+*Mitigation:* This threat is not addressed by SLSA, but may be addressed by
+efforts like [Secure by Design][secure-by-design].
+</details>
```
MarkLodato, line 800: need blank line before this one

MarkLodato (approved):
Yeah, I think this is good enough. I didn't have a strong notion of what (I) would be; we added it mostly for symmetry.

MarkLodato:
Oh, note that this is based on #1190 so that one needs to be merged first.

docs/spec/draft/threats.md
TomHennen: Done

lehors (approved):

mlieberman85 (approved):
LGTM

docs/spec/draft/threats.md
```diff
@@ -662,9 +662,99 @@ cryptographic signature is no longer valid.
...
+The consumer's actions may be simplified if (F) produces a [VSA][vsa].
+In this case the consumer may replace provenance verification with
+[VSA verification][vsa_verification].
+
+<details><summary>Build with untrusted CI/CD <span>(expectations)</span></summary>
+
+*Threat:* Replace the package with one built using an unofficial CI/CD pipeline
+that does not build in the correct way.
+
+*Mitigation:* Verifier requires provenance showing that the builder matched an
```
zachariahcox, line 682: This is a great place to expand. Thanks for starting this.

I think the mitigation for this is essentially: 
* Have a way to know what "correct" looks like for this resourceUri. 
   * If you have access to such a thing, you can verify the artifact yourself by inspecting the data in the provenance attestation. 
   * If you do not, you must rely on a VSA with access to definition of "correct"
   * If you do not have either, you _cannot know_ whether the artifact is safe to use, even if it was SLSA Build L3

I think in the example below "build L3" is conflated with "built in the correct way" which is not necessarily the case.

docs/spec/draft/threats.md
```diff
@@ -662,9 +662,99 @@ cryptographic signature is no longer valid.
...
+builder is not as expected.
+
+</details>
+<details><summary>Issue VSA from untrusted intermediary <span>(expectations)</span></summary>
+
+*Threat:* Have an unofficial intermediary issue a VSA for a malicious package.
+
+*Mitigation*: Verifier requires VSAs to be issued by a trusted intermediary.
+
+*Example:* Verifier expects VSAs to be issued by TheRepository. Adversary
```
zachariahcox, line 699: implicitly here I think, the consumer has _already decided ahead of time_ which issuers it trusts for all possible packages it plans to consume.

docs/spec/draft/threats.md
```diff
@@ -662,9 +662,99 @@ cryptographic signature is no longer valid.
...
+*Threat:* Take a benign version of the package, modify it in some way, then
+replace it while retaining the original provenance or VSA.
+
+*Mitigation:* Verifier checks that the provenance or VSA's `subject` matches
+the hash of the package.
+
+*Example:* Adversary performs a proper build, modifies the artifact, then
+replaces the modified version of the package in the repository and retains the
+original provenance. Solution: Verifier rejects because the hash of the
+artifact does not match the `subject` found within the provenance.
```
zachariahcox, line 728: These are really clearly written. 

I wonder if we could simply this into a "basic consumer verification workflow psuedo code"

1. download all the deps and their attestations somehow
2. make sure they all the docs match their resources.
3. make sure the docs "say enough of the right things to pass my policies", either directly or via a VSA issuer I trust
4. continue build

zachariahcox: because that ☝️ algorithm is pretty specific, we should also call out that none of this works unless "the verifier tool MUST be resistant to tampering too." 

Zip-slip attacks are particularly dangerous when they attack the verifier itself. 

We could add another example: "Adversary modifies payload to replace verification tooling during unzip"

docs/spec/draft/threats.md
TomHennen: Filed #1212 to suggest we provide more guidance on how that resourceUri is set in the first place (which makes knowing the expected value easier).

I'm not sure if you're suggesting any other changes here?

docs/spec/draft/threats.md
TomHennen: Right. Looking at the existing threats I think the process taken to determine trust is components or people is out of scope for this page (none of the talk of 'trusted X' describes when or how that trust is established).  Something related is discussed a little bit [here](https://slsa.dev/spec/v1.0/terminology#verification-model).

Are you hoping for something to be called out specifically here?

docs/spec/draft/threats.md
TomHennen: I think we have verification pseudo code [here for provenance](https://slsa.dev/spec/v1.0/verifying-artifacts#step-2-check-expectations:~:text=Given%20an%20artifact%20and%20its%20provenance%3A) and [here for VSAs](https://slsa.dev/spec/draft/verification_summary#how-to-verify) so I don't know that it needs to be replicated here?  Instead, as I understand it, the goal is just to define how these various threats are mitigated by SLSA, the details of which can be detailed elsewhere.

WDYT?

docs/spec/draft/threats.md
TomHennen: Clarify that Build L3 is harder to compromise in the way described here.

TomHennen: Also filed #1215 where we can make this more clear.

docs/spec/draft/threats.md
TomHennen: Chatting with Zac I don't think he needs anything to change here.

docs/spec/draft/threats.md
TomHennen: Chatting with Zac I don't think we need to do anything here at the moment.

TomHennen:
I think we're good to here.  Issues have been filed where we might want to make some other things more clear.

docs/spec/draft/threats.md
zachariahcox: We could link to those, but they both kind of gloss over policy and correctness. 

[this doc](https://slsa.dev/spec/v1.0/verifying-artifacts#step-2-check-expectations:~:text=Given%20an%20artifact%20and%20its%20provenance%3A) is a little vague still I think. 

[This step ](https://slsa.dev/spec/v1.0/verifying-artifacts#step-2-check-expectations:~:text=You%20SHOULD%20compare%20the%20provenance%20against%20expected%20values%20for%20at%20least%20the%20following%20fields%3A)is ignoring the hard part of distributing the policy that should also be checked when verifying artifacts.

docs/spec/draft/threats.md
zachariahcox:
```suggestion
*Example:* Verifier decided ahead of time that it will only trust `TheRepository` to issue attestations about package `Foo`. Adversary builds a malicious version of `Foo` ...
```

docs/spec/draft/threats.md
```diff
@@ -662,9 +662,99 @@ cryptographic signature is no longer valid.
...
+
+</details>
+<details><summary>Tamper with provenance or VSA <span>(Build L2)</span></summary>
+
+*Threat:* Perform a build that would not meet expectations, then modify the
```
zachariahcox, line 733:
```suggestion
*Threat:* Issue an attestation that purposefully misrepresents the subject.
```

docs/spec/draft/threats.md
```diff
@@ -662,9 +662,99 @@ cryptographic signature is no longer valid.
...
+`evil/my-package` repo and then modifies the provenance so that the source looks
+like it came from `good/my-package`. Solution: Verifier rejects because the
+cryptographic signature is no longer valid.
+
+*Example 2:* Verifier expects VSAs to be issued by TheRepository. Adversary
+builds a malicious package and then modifies the original VSA's `subject`
+field to match the digest of the malicious package. Solution: Verifier rejects
+because the cryptographic signature is no longer valid.
+
+*Example 3:* Adversary uploads a malicious package to `repo/evil-package`,
```
zachariahcox, line 751: I think slsa doesn't help with this. 

This is basically saying "slsa worked perfectly, but the artifact was not suitable for use because of external policy."
There wasn't really any tampering involved, right?

TomHennen: SLSA should help here because the downstream consumer has an expectation for what the resourceUri is and they [validate that expectation](https://slsa.dev/spec/draft/verification_summary#how-to-verify:~:text=Verify%20that%20the%20value%20for%20resourceUri%20in%20the%20VSA%20matches%20the%20expected%20value.%20This%20step%20ensures%20that%20the%20consumer%20is%20using%20the%20VSA%20for%20the%20producer%E2%80%99s%20intended%20purpose.) as expected by other threats [example](https://slsa.dev/spec/draft/threats#f-artifact-publication:~:text=Mitigation%3A%20Verifier%20requires%20provenance%20showing%20that%20the%20builder%20matched%20an%20expected%20value.).

_How_ the consumer establishes that expectation for VSAs is being clarified in https://github.com/slsa-framework/slsa/pull/1220, but when checking provenance it's [a bit more wishy-washy](https://slsa.dev/spec/draft/verifying-artifacts#step-2-check-expectations).

docs/spec/draft/threats.md
TomHennen: I don't think this is quite right.  In example 1 and 2 the threat described is that an existing attestation is tampered with, the mitigation described detects these problems because the attacker cannot modify the valid attestations without invalidating the expected signatures.

However, I think 'example 3' should probably be captured in a threat by itself as that deals with _expectations_ mismatching which is usually captured elsewhere.

TomHennen: Filed #1223 to track.

Issue #1069, opened 2024-06-17
Structure & formatting don't match the build track


TomHennen:
Apologies for being late to review this. It's fantastic to see the beginnings of a Source track emerge. Thank you for the continued effort on this @kpk47.

This feels equivalent to what we had in v0.1, only compressed into 3 levels instead of 4 and with more affordances for changes from robots. Does that assessment feel fair/accurate to you?

Most of my comments are related to presentation, the requirements feeling so close to the v0.1 version is a great baseline.

The formatting and structure of the Levels changes don’t match the existing Build track levels. 
Build levels are broken down into producer and platform requirements (see change from benefits to focus in https://github.com/slsa-framework/slsa/commit/cc812c33a03ec98f8bdb9b22304a38b130e9f23e) and are a descriptive overview requirements, whereas the newly added source requirements are detailed and specific enough that they run the risk of seeming authoritative and contradictory to the track text.

The formatting and structure of the Source requirements doesn’t match the “Producing artefacts” document, where the requirements are in a table with check marks for which level they apply to (and further, there are tables per requirement category: provenance generation and isolation, I’m not sure this split makes sense for the source track).

Note: we should update future-directions to remove the source track and/or update it to mention what could come next. Threats and mitigations should be updated too.

Final thought, is adding a new level appropriate for a minor release, or should we consider a major release for this?

_Originally posted by @joshuagl in https://github.com/slsa-framework/slsa/pull/1037#pullrequestreview-1993634758_

TomHennen:
Also see [the discussion here](https://docs.google.com/document/d/13Xt8mA_2b00McGX2vkyhu4GQdFAqtXPu7YXE8ZA6ISE/edit?resourcekey=0-EqfHF79tUWAKp4PzsE3z1A&tab=t.0#heading=h.mva6v1s2xfam).

I believe @Nikokrock is working on fixing this.

zachariahcox:
I don't have permission to edit this title @TomHennen but we should probably rename this: "source track formatting should follow the slsa project standards"

PR #1171, opened 2024-10-01
content: draft: source-track: rename SCP to SCS, replace issue links with project board links. 

zachariahcox:
fixes: https://github.com/slsa-framework/slsa/issues/1128

(cleaned up version of https://github.com/slsa-framework/slsa/pull/1166)

This change is in response to the 9.30 slsa specification meeting on this topic.

A SCS is the full suite of services and ideas relied upon by the organization to create source revisions.

VCS stuff should mostly fall out of the discussion

Repositories can be used as the concept used when we need to talk about authN and authZ w.r.t. authentic contributions.

PR #1194, opened 2024-10-15
style: Put detailed requirements into table form

TomHennen:
This follows the pattern from requirements.md where each requirement is listed along with what level it applies at.

I separate 'platform requirements' from 'change management tool' requirements to mirror the 'Provenance generation' / "Isolation strength" from requirements.md.

Future PRs will add a summary like the 'Build levels' table from requirements.md.

Refs #1069

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -112,7 +112,7 @@ Organizations that want strong guarantees and auditability of their change manag
 Benefits:
 Provides reliable information to policy enforcement tools.
 
-## Platform Requirements
+## System Requirements
 
 Many examples in this document use the [git version control system](https://git-scm.com/), but use of git is not a requirement to meet any level on the SLSA source track.
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -79,8 +79,6 @@ When a repo is transferred to a new organization ("donated"), or if a repo must 
 
 ## Levels
 
-Many examples in this document use the [git version control system](https://git-scm.com/), but use of git is not a requirement to meet any level on the SLSA source track.
-
 ### Level 1: Version controlled
 
 Summary:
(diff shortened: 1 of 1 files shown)


docs/spec/draft/source-requirements.md
```diff
@@ -91,9 +89,37 @@ Intended for: Organizations currently storing source in non-standard ways who wa
...
+A consumer can ask the SCS for everything it knows about a specific revision of a repository.
+The information is provided in a documented and tamper-resistant format.
+
+Intended for:
+Organizations that want strong guarantees and auditability of their change management processes.
+
+Benefits:
+Provides reliable information to policy enforcement tools.
+
+## Platform Requirements
```
zachariahcox, line 115: I think we would call this "system" requirements after our change to SCS, right?

zachariahcox: generally, I don't mind separating these out in this way, but there would need to be some overlap I guess. 

EG: Probably the platform also needs strong auth (which is only mentioned below in the "change tool" parts). 
Maybe not if all the source revisions are stored in artifactory and not git?

docs/spec/draft/source-requirements.md
```diff
@@ -196,21 +203,26 @@ If a consumer is authorized to access source on a particular branch, they MUST b
...
+## Change management tool requirements
 
 The change management tool MUST be able to authoritatively state that each new revision reachable from the protected branch represents only the changes managed via the process.
 
-The change management tool MUST provide at a minimum:
+<table>
+<tr><th>Requirement<th>Description<th>L1<th>L2<th>L3
 
-##### Strong Authentication
+<tr id="strong-authentication"><td>Strong Authentication<td>
```
zachariahcox, line 225: I think the platform would also need strong auth too at l3, right?

zachariahcox (approved):
I think this is looking great. I even like splitting the two concerns apart, but it does seem like both the platform and the change tool need the authn/z bits.

lehors (approved):
Very nice improvement!

docs/spec/draft/source-requirements.md
TomHennen: Agreed, filed https://github.com/slsa-framework/slsa/issues/1199 to fix since that's a content change and I don't want to mix it up with a style fix. :)

docs/spec/draft/source-requirements.md
lehors: I apologize if I'm out of touch on that one but for what it's worth, for the Build track we started with "system" and then switched to "platform" somewhere down the line, which is why the "Verifying build platform" section is in the `verifying-systems.md` file. :-)

TomHennen: Right, this change was the result of a lot of discussion around how 'platform' implies the entire thing may need to be centralized/hosted.  Whereas 'system' is more broad and makes space for decentralized approaches like gittuf. 

The initial change was in #1171.

PR #1200, opened 2024-10-16
content: draft: move strong auth to the system section

### LINKED ISSUE #1199: Source 'systems' need strong auth too, not just change management tools
In [this PR comment](https://github.com/slsa-framework/slsa/pull/1194/files#r1802054524) @zachariahcox says

> I think the platform [system] would also need strong auth too at l3, right?

And I think that's right.  Filing this issue to fix.

TomHennen:
We don't want to restrict strong authentication to the change management tool, but instead want it used throughout the system.  Moving the strong auth requirement there to reflect that desire.

fixes #1199.


mlieberman85 (approved):
LGTM

PR #1201, opened 2024-10-16
nonspec: Make requirements for draft reviews more explicit

TomHennen:
Currently CONTRIBUTING.md says that PRs that affect draft files have 'relaxed' approval requirements but does not specify what those relaxed requirements are.

This PR makes the default requirements for draft PRs explicit (only 1 reviewer, no waiting period) in an attempt to speed progress on draft changes. Approving reviewers will have the opportunity to indicate if they want additional reviewers prior to submitting these draft PRs.

We will still have strict requirements (3 approvers, 72h waiting period) when moving content from draft to rc or final status. This should allow thorough review of changes in their finished state without hampering progress on new tracks.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,12 +138,11 @@ always requires one reviewer, even if the author has write access.
 
 Note 2: If the PR only touches files in the [Draft](docs/spec-stages.md)
 specification stage, then the "waiting period" and "# reviewers" are relaxed and
-up to Maintainer discretion. By default PRs touching files in draft need only
-be approved by one maintainer and have no waiting period. If the first approving
-maintainer would like additional maintainers to approve then will indicate
-this desire in a comment and assign additional maintainers as approvers of the PR.
-Files in the Draft stage have a large banner at the top of each rendered page, as
-well as the text "Status: Draft".
+up to Maintainer discretion (including the PR author if they're a maintainer). Note
+that a relaxed number of reviewers and waiting period may result in more back
+and forth with the expanded set of reviewers as drafts are finalized.
+Files in the Draft stage have a large banner at the top of each rendered page,
+as well as the text "Status: Draft".
 
 [squash and merge]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,8 +138,12 @@ always requires one reviewer, even if the author has write access.
 
 Note 2: If the PR only touches files in the [Draft](docs/spec-stages.md)
 specification stage, then the "waiting period" and "# reviewers" are relaxed and
-up to Maintainer discretion. Files in the Draft stage have a large banner at the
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
No strong opinions. I can see where this PR is coming from, but maybe two instead of three? This is so that the big reviews later going from draft to stable is not too surprising, but two would still be slower than one.

TomHennen:
> No strong opinions. I can see where this PR is coming from, but maybe two instead of three? This is so that the big reviews later going from draft to stable is not too surprising, but two would still be slower than one.

Yeah that's definitely the tradeoff.  I'd hope that the first approver could recognize situations like that that may require additional guidance and indicate it?  It think that's generally the guidance at the moment.  "up to Maintainer discretion" isn't quite specific enough about who that maintainer is.  Should the PR author be allowed to make this distinction?  That might also help with the goal without being as prescriptive.

E.g. We could say "up to Maintainer's discretion (including the PR author if they're a maintainer)"

Then the author could decide if they want to take the 'risk' of merging something that might generate a lot more feedback later.

mlieberman85 (approved):
I think this is a good start. We can iterate if we need to change some of the details

TomHennen:
> > E.g. We could say "up to Maintainer's discretion (including the PR author if they're a maintainer)"
> 
> Yes, pls, thx!

Done!

marcelamelara (approved):
Thanks @TomHennen ! I'm sure we'll run into some special cases, but this seems like it covers the most general ones.

lehors:
For what it's worth, I read the original text as meaning that for Drafts there are no requirements per se and "up to Maintainer discretion", in which case there is no need to define the requirements any further. But I'm happy with the suggested change.

lehors (approved):

MarkLodato (approved):
Is it also a requirement to have the "draft:" in the PR description, like you've been doing lately? Is that checked automatically? (That tag is helpful.)

TomHennen:
> Is it also a requirement to have the "draft:" in the PR description, like you've been doing lately? Is that checked automatically? (That tag is helpful.)

It isn't checked automatically.  I could add that 'draft' tag as a requirement.  Not sure how to check it automatically though.    Would we want just 'draft:' or like I've been doing "[whatever]: draft: [description]"

zachariahcox (approved):

arewm (approved):
Are there any requirements that would be imposed if we added `draft` as an option here? https://github.com/slsa-framework/slsa/blob/main/.github/workflows/conventional-commits.yml#L22-L29

Do we have any process documented for moving content out of draft to ensure that appropriate reviews are given?

TomHennen:
> Are there any requirements that would be imposed if we added `draft` as an option here? https://github.com/slsa-framework/slsa/blob/main/.github/workflows/conventional-commits.yml#L22-L29

I think that just checks that the right prefix was added but doesn't do anything else AFAIK.  Maybe someone else does?

> Do we have any process documented for moving content out of draft to ensure that appropriate reviews are given?

We do!  It's [discussed here](https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md#slsa-versions-management).

TomHennen:
Ok, I think we have enough approvals and waited long enough, so I'm going to merge.  Thanks all!

PR #1203, opened 2024-10-16
content: draft: Harden 'safe-expunging-process'

### LINKED ISSUE #1135: Harden 'safe-expunging-process'
In https://github.com/slsa-framework/slsa/pull/1094#discussion_r1722170324 @marcelamelara said 

> I understand the practical reasons for needing to make exceptions in specific edge cases, but I also worry that the safe expunging process may still be worded too broadly. As in, the level of trustworthiness in a source repo at L2 still isn't super high, so what's to stop a rogue/malicious repo admin from abusing the safe expunging exception, especially since there's no documentation requirement? I'm wondering if it might make sense to raise the level at which such exceptions are permitted to make sure certain controls are in place and/or narrow the scope of the safe expunging process.

Let's make sure we're happy with this process before release.

TomHennen:
fixes #1135

Hardens the 'safe-expunging-process' by:

1. Suggesting that SCSs should document and log changes when possible.
2. SCSs should use multi-party approval when possible

Also clarifies that some of these changes may need to be kept private to comply with local laws.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -61,7 +61,7 @@ If a branch has been identified as consumable branch, force pushes to that branc
 
 Different organizations and tech stacks may have different approaches to the problem.
 
-Safe Expunging Scenarios:
+Safe Expunging Scenarios (non-exhaustive):
 
 ### Legal Takedowns
 
@@ -78,13 +78,13 @@ A team may decide that all reachable commits in the history of a revision need t
 In git VCS, compliance with this new policy will require history to be rewritten (commit metadata is included in the computation of the revision id).
 Policies in this category include things like commit signatures, author / committer formatting restrictions, closed-issue-linkage, etc.
 
-SCSs SHOULD make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
+Organizations MUST make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
 
 ### Repository renames
 
 When a repo is transferred to a new organization ("donated"), or if a repo must be renamed or otherwise have its url changed within the same org, attestations for previous revisions of this repo will no longer be matched because the combination of the repository id and the revision id will have changed.
 
-SCSs SHOULD make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
+Organizations MUST make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
 
 ## Levels
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -46,7 +46,7 @@ Consumers can examine the various source provenance attestations to determine if
 
 ## Safe Expunging Process
 
-SCSs MAY allow _administrators_ to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
+SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
 This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements as well as administrative
 changes within a repository (see below for more information on the various scenarios).
 When used as an attack, this is called “repo hijacking” (or “repo-jacking”) and is one of the primary threats source provenance attestations protect against.
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -49,8 +49,8 @@ Consumers can examine the various source provenance attestations to determine if
 SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
 This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements.
 
-Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact. 
-In version control systems like git, removal of a revision changes the object id of all subsequent revisions that were built on top of it. 
+Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact.
+In version control systems like git, removal of a revision changes the object id of all subsequent revisions that were built on top of it.
 Although there is no "safe" way to do it, it sometimes necessary and there are steps you can take to mitigate the damage.
 
 When used as an attack, this is called “repo hijacking” (or “repo-jacking”) and is one of the primary threats source provenance attestations protect against.
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -50,7 +50,7 @@ SCSs MAY allow the organization to expunge (remove) content from a repository an
 This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements.
 
 Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact.
-In version control systems like git, removal of a revision changes the object id of all subsequent revisions that were built on top of it.
+In version control systems like Git, removal of a revision changes the object IDs of all subsequent revisions that were built on top of it.
 Although there is no "safe" way to do it, it sometimes necessary and there are steps you can take to mitigate the damage.
 
 When used as an attack, this is called “repo hijacking” (or “repo-jacking”) and is one of the primary threats source provenance attestations protect against.
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -46,26 +46,31 @@ Consumers can examine the various source provenance attestations to determine if
 
 ## Safe Expunging Process
 
-Administrators have the ability to expunge (remove) content from a repository and its change history without leaving a record of the removed content.
-This includes changing files, history, or changing references in git and is used to accommodate legal or privacy compliance requirements.
+SCSs MAY allow _administrators_ to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
+This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements as well as administrative
+changes within a repository (see below for more information on the various scenarios).
 When used as an attack, this is called “repo hijacking” (or “repo-jacking”) and is one of the primary threats source provenance attestations protect against.
+As such great care must be taken by SCSs when implementing these changes for legitimate purposes.
+
+SCSs SHOULD require multi-party approval for any changes made under this process.
+
+The Safe Expunging Process MUST be documented and describe how requests and actions are tracked and SHOULD log the fact that content was removed.
 
 On the git VCS, force pushes allow you to remove data from a branch.
 If a branch has been identified as consumable branch, force pushes to that branch must follow the safe expunging process.
 
-TODO: Determine how organizations can provide transparency around this process.
-At a minimum the organization would need to declare why data was removed from the branch.
-
-The goal of this sections is to document that this process is allowed.
 Different organizations and tech stacks may have different approaches to the problem.
 
-Scenarios that need to be addressed:
+Safe Expunging Scenarios:
 
 ### Legal Takedowns
 
-A DMCA takedown request will be addressed by following an agreed-upon process.
+A legal takedown request will be addressed by following an agreed-upon process.
 That process should be documented itself and followed.
 It may be the case that the specific set of commits targeted by the takedown can be expunged in ways that do not impact revisions.
+
+This process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
+and regulations which may require the change to be kept private to the extent possible.
 
 ### Commit metadata rewrites
 
@@ -73,9 +78,13 @@ A team may decide that all reachable commits in the history of a revision need t
 In git VCS, compliance with this new policy will require history to be rewritten (commit metadata is included in the computation of the revision id).
 Policies in this category include things like commit signatures, author / committer formatting restrictions, closed-issue-linkage, etc.
 
+SCSs SHOULD make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
+
 ### Repository renames
 
 When a repo is transferred to a new organization ("donated"), or if a repo must be renamed or otherwise have its url changed within the same org, attestations for previous revisions of this repo will no longer be matched because the combination of the repository id and the revision id will have changed.
+
+SCSs SHOULD make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
 
 ## Levels
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -75,7 +75,7 @@ The takedown request MUST be addressed by following an agreed-upon process.
 
 It may be the case that the specific set of commits targeted by the takedown can be expunged in ways that do not impact revisions.
 
-This process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
+The application of the safe expunging process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
 and regulations which may require the change to be kept private to the extent possible.
 
 ## Levels
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -46,37 +46,31 @@ Consumers can examine the various source provenance attestations to determine if
 
 ## Safe Expunging Process
 
-SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
-This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements.
+SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content,
+but MUST only allow these changes in order to meet legal or privacy compliance requirements.
+Content changed under this process includes changing files, history, references, or any other metadata stored by the SCS.
+
+### Warning
 
 Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact.
-In version control systems like Git, removal of a revision changes the object IDs of all subsequent revisions that were built on top of it.
-Although there is no "safe" way to do it, it sometimes necessary and there are steps you can take to mitigate the damage.
+For example, in VCSs like Git, removal of a revision changes the object IDs of all subsequent revisions that were built on top of it,
+this can break downstream consumers ability to refer to source they've already integrated into their products.
 
-When used as an attack, this is called “repo hijacking” (or “repo-jacking”) and is one of the primary threats source provenance attestations protect against.
-As such great care must be taken by SCSs when implementing these changes for legitimate purposes.
+It may be the case that the specific set of commits targeted by the takedown can be expunged in ways that do not impact revisions which can mitigate these problems.
+
+It is also the case that removing content from a repository won't necessarily remove it everywhere.
+The content may still exist in other copies of the repository, either in backups or on developer machines.
+
+### Process
+
+An SCS MUST document the Safe Expunging Process and describe how requests and actions are tracked and SHOULD log the fact that content was removed.
+Different organizations and tech stacks may have different approaches to the problem.
 
 SCSs SHOULD require multi-party approval for any changes made under this process.
 
-The Safe Expunging Process MUST be documented and describe how requests and actions are tracked and SHOULD log the fact that content was removed.
-
-On the git VCS, force pushes allow you to remove data from a branch.
-If a branch has been identified as consumable branch, force pushes to that branch must follow the safe expunging process.
-
-Different organizations and tech stacks may have different approaches to the problem.
-
-Safe Expunging Scenarios:
-
-### Legal Takedowns
-
-The only currently acceptable scenario to expunge data is in response to some legal or compliance request.
-
-The takedown request MUST be addressed by following an agreed-upon process.
-
-It may be the case that the specific set of commits targeted by the takedown can be expunged in ways that do not impact revisions.
-
 The application of the safe expunging process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
 and regulations which may require the change to be kept private to the extent possible.
+Organizations SHOULD prefer to make logs public if possible.
 
 ## Levels
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -56,7 +56,7 @@ Removing a revision from a repository is similar to deleting a package version f
 For example, in VCSs like Git, removal of a revision changes the object IDs of all subsequent revisions that were built on top of it,
 , breaking downstream consumers' ability to refer to source they've already integrated into their products.
 
-It may be the case that the specific set of changes targeted by the takedown can be expunged in ways that do not impact consumed revisions, which can mitigate these problems.
+It may be the case that the specific set of changes targeted by a legal takedown can be expunged in ways that do not impact consumed revisions, which can mitigate these problems.
 
 It is also the case that removing content from a repository won't necessarily remove it everywhere.
 The content may still exist in other copies of the repository, either in backups or on developer machines.
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -66,7 +66,7 @@ The content may still exist in other copies of the repository, either in backups
 An SCS MUST document the Safe Expunging Process and describe how requests and actions are tracked and SHOULD log the fact that content was removed.
 Different organizations and tech stacks may have different approaches to the problem.
 
-SCSs SHOULD have technical mechanisms in place which require multiple Trusted Persons to trigger any expunging (removals) made under this process.
+SCSs SHOULD have technical mechanisms in place which require an Administrator plus, at least, one additional 'trusted person' to trigger any expunging (removals) made under this process.
 
 The application of the safe expunging process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
 and regulations which may require the change to be kept private to the extent possible.
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -47,8 +47,12 @@ Consumers can examine the various source provenance attestations to determine if
 ## Safe Expunging Process
 
 SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
-This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements as well as administrative
-changes within a repository (see below for more information on the various scenarios).
+This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements.
+
+Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact. 
+In version control systems like git, removal of a revision changes the object id of all subsequent revisions that were built on top of it. 
+Although there is no "safe" way to do it, it sometimes necessary and there are steps you can take to mitigate the damage.
+
 When used as an attack, this is called “repo hijacking” (or “repo-jacking”) and is one of the primary threats source provenance attestations protect against.
 As such great care must be taken by SCSs when implementing these changes for legitimate purposes.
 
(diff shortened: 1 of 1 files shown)


TomHennen:
@adityasaky I'd also like your input here.

trishankatdatadog (approved):
Don't see anything controversial

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,46 @@ Consumers can examine the various source provenance attestations to determine if
...
-This includes changing files, history, or changing references in git and is used to accommodate legal or privacy compliance requirements.
+SCSs MAY allow _administrators_ to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
+This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements as well as administrative
+changes within a repository (see below for more information on the various scenarios).
 When used as an attack, this is called “repo hijacking” (or “repo-jacking”) and is one of the primary threats source provenance attestations protect against.
+As such great care must be taken by SCSs when implementing these changes for legitimate purposes.
+
+SCSs SHOULD require multi-party approval for any changes made under this process.
+
+The Safe Expunging Process MUST be documented and describe how requests and actions are tracked and SHOULD log the fact that content was removed.
```
adityasaky, line 57: Should they log _privately_ the content that was removed as well? is there an expectation of being able to _prove_ that only the content they were legally supposed to remove was removed, even if they can only do so for internal parties / auditors?

adityasaky: I guess this is hard to answer if the information that must be removed is something personal (say PII the person in question doesn't want a cloud SCP to hold on to even privately)...

adityasaky (commented):
I'm still a bit curious what this may end up looking like. In addition to the experiences @zachariahcox and others have had at GitHub, GitLab, etc., I wonder if something like this has ever come up within the LF for https://git.kernel.org/ or so? Is there a public guide for dealing with these scenarios that we can leverage for this?

docs/spec/draft/source-requirements.md
TomHennen: I think we're quite constrained here.  Different legal environments have different requirements.  If you think about [national security letters](https://en.wikipedia.org/wiki/National_security_letter) and their non-disclosure requirement, you could imagine a similar mechanism being used for a takedown request.  Of course, we need to think beyond just US law.  One could _imagine_ (I am not a lawyer!) [this Australian law](https://www.homeaffairs.gov.au/about-us/our-portfolios/national-security/lawful-access-telecommunications/assistance-and-access-overview) being used to require a change without logging it.

Additionally, requiring this logging mechanism would likely require a lot of new code and systems to be written, which would constrain implementers.  Maybe that's OK for SLSA L3, but it seems like a lot to ask for SLSA L2.  How readily do we want this to be adoptable?  How much do we want to delay the release and adoption of the source track while we get an answer to this question and implementations built?

adityasaky: Good points! I agree on not blocking too much on this and on not being very specific about how this is implemented as we can't foresee all scenarios and their variables.

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,46 @@ Consumers can examine the various source provenance attestations to determine if
...
 If a branch has been identified as consumable branch, force pushes to that branch must follow the safe expunging process.
 
-TODO: Determine how organizations can provide transparency around this process.
-At a minimum the organization would need to declare why data was removed from the branch.
-
-The goal of this sections is to document that this process is allowed.
 Different organizations and tech stacks may have different approaches to the problem.
 
-Scenarios that need to be addressed:
+Safe Expunging Scenarios:
```
adityasaky, line 64: Should we highlight this list isn't exhaustive?

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,46 @@ Consumers can examine the various source provenance attestations to determine if
...
+This process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
+and regulations which may require the change to be kept private to the extent possible.
+
 ### Commit metadata rewrites
 
 A team may decide that all reachable commits in the history of a revision need to follow a new metadata convention.
 In git VCS, compliance with this new policy will require history to be rewritten (commit metadata is included in the computation of the revision id).
 Policies in this category include things like commit signatures, author / committer formatting restrictions, closed-issue-linkage, etc.
 
+SCSs SHOULD make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
```
adityasaky, line 81: Thoughts on making this `SHOULD` a `MUST`? I understand why this can't be said for legal takedowns but this scenario (and others like it such as repo renames) seem to have no reason we can't require an announcement? If we can think of reasons, perhaps we list them as examples of when an announcement isn't made?

docs/spec/draft/source-requirements.md

docs/spec/draft/source-requirements.md
TomHennen: Sure.  I also changed SCSs to 'Organizations' since it's likely not the system that will do this but rather the people involved in running it.

adityasaky (approved):

docs/spec/draft/source-requirements.md
adityasaky: Thanks!

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,46 @@ Consumers can examine the various source provenance attestations to determine if
 
 ## Safe Expunging Process
 
-Administrators have the ability to expunge (remove) content from a repository and its change history without leaving a record of the removed content.
-This includes changing files, history, or changing references in git and is used to accommodate legal or privacy compliance requirements.
+SCSs MAY allow _administrators_ to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
```
zachariahcox, line 49:
```suggestion
SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
```

I'm hoping we can remove this as a general capability of repo admins one day. It's logically going to be "the org" (not just the admins) that make decisions about which data is inappropriate for the repo and must be removed, because the org defines the "intent" of the code.

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,46 @@ Consumers can examine the various source provenance attestations to determine if
...
 On the git VCS, force pushes allow you to remove data from a branch.
 If a branch has been identified as consumable branch, force pushes to that branch must follow the safe expunging process.
 
-TODO: Determine how organizations can provide transparency around this process.
-At a minimum the organization would need to declare why data was removed from the branch.
-
-The goal of this sections is to document that this process is allowed.
 Different organizations and tech stacks may have different approaches to the problem.
 
-Scenarios that need to be addressed:
```
zachariahcox, line 62: this was definitely "things to think about" more than scenarios we know we want to support.

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,46 @@ Consumers can examine the various source provenance attestations to determine if
...
+This process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
+and regulations which may require the change to be kept private to the extent possible.
+
 ### Commit metadata rewrites
 
 A team may decide that all reachable commits in the history of a revision need to follow a new metadata convention.
 In git VCS, compliance with this new policy will require history to be rewritten (commit metadata is included in the computation of the revision id).
 Policies in this category include things like commit signatures, author / committer formatting restrictions, closed-issue-linkage, etc.
 
+Organizations MUST make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
```
zachariahcox, line 81: I've been struggling with this topic recently. 

I'm certainly leaning towards a position that "we wish commit metadata were different" should never be a good enough reason to rewrite history. 
Commit shas are the foreign keys of package registries. Changing them causes unknowable damage to an ecosystem's ability to trust itself.

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,46 @@ Consumers can examine the various source provenance attestations to determine if
...
+
 ### Repository renames
 
 When a repo is transferred to a new organization ("donated"), or if a repo must be renamed or otherwise have its url changed within the same org, attestations for previous revisions of this repo will no longer be matched because the combination of the repository id and the revision id will have changed.
 
+Organizations MUST make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
```
zachariahcox, line 87: I think this would only be true for public oss repos. 

```suggestion
Organizations MUST inform all consumers of a repo of these changes when they occur.
Organizations SHOULD provide a means to resolve previous revision IDs to the new revision IDs.
```

TomHennen: Great point.

docs/spec/draft/source-requirements.md
TomHennen: agreed

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,46 @@ Consumers can examine the various source provenance attestations to determine if
...
+
 ### Commit metadata rewrites
 
 A team may decide that all reachable commits in the history of a revision need to follow a new metadata convention.
 In git VCS, compliance with this new policy will require history to be rewritten (commit metadata is included in the computation of the revision id).
 Policies in this category include things like commit signatures, author / committer formatting restrictions, closed-issue-linkage, etc.
 
+Organizations MUST make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
+
 ### Repository renames
```
zachariahcox, line 83: probably the best case scenario here is to just archive/lock the previous repo name so the attestations stay hosted and available.

docs/spec/draft/source-requirements.md
TomHennen: I can add some language to that effect.  "Orgs should strive to avoid rewriting commit metadata and only do so as a last resort" ?

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,46 @@ Consumers can examine the various source provenance attestations to determine if
...
-TODO: Determine how organizations can provide transparency around this process.
-At a minimum the organization would need to declare why data was removed from the branch.
-
-The goal of this sections is to document that this process is allowed.
 Different organizations and tech stacks may have different approaches to the problem.
 
-Scenarios that need to be addressed:
+Safe Expunging Scenarios (non-exhaustive):
 
 ### Legal Takedowns
```
zachariahcox, line 66: In git / merkle-tree-based-VCS, I think this is going to be the only legit reason to consider history rewrites due to the damage they cause to the trustworthiness of a repo. 

In these cases, we should probably build out a health mapping concept. 
These shas XYZ have been updated to ABC. 

We need to somehow say: provenance issued for XYZ is legitimate and still hosted but the revision is no longer available in the repo. 
It will depend on the situation whether packages built on XYZ are still legal.

docs/spec/draft/source-requirements.md
zachariahcox: I was considering just cutting this "commit metadata rewrites" section. Essentially saying "This process is built for legal challenges. Other uses are hard to justify."

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,47 @@ Consumers can examine the various source provenance attestations to determine if
 
 ## Safe Expunging Process
 
-Administrators have the ability to expunge (remove) content from a repository and its change history without leaving a record of the removed content.
-This includes changing files, history, or changing references in git and is used to accommodate legal or privacy compliance requirements.
+SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
```
zachariahcox, line 49: We could add a sentence here to help set the tone. 

```
Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact. 

In version control systems like git, removal of a revision changes the object id of all subsequent revisions that were built on top of it. 

Although there is no "safe" way to do it, it sometimes necessary and there are steps you can take to mitigate the damage.

File: docs/spec/draft/source-requirements.md (hunk 6 above)
Comment (adityasaky) [reply docs/spec/draft/source-requirements.md:81]: I like the idea of reducing the scope of use of safe expunging to legal challenges.

Comment (TomHennen) [reply docs/spec/draft/source-requirements.md:81]: That works for me.  Would I remove 'Repository renames' too?

Comment (zachariahcox) [reply docs/spec/draft/source-requirements.md:81]: good question -- maybe we could move that info into "how source provenance attestation subjects work?"

Essentially the take away should be "if you remove a sha or change the url of a repo, provenance attestations are voided. Also you may have broken the chain of custody for shipped packages."

**Review by marcelamelara — CHANGES_REQUESTED** [review]
Thanks for these revisions @TomHennen ! I agree with the other reviewers that the safe expunging process should be scoped to legal challenges.

File: docs/spec/draft/source-requirements.md (hunk 10 above)
Comment (TomHennen) [reply docs/spec/draft/source-requirements.md:49]: Done.

File: docs/spec/draft/source-requirements.md (hunk 5 above)
Comment (TomHennen) [reply docs/spec/draft/source-requirements.md:62]: I think we can resolve this since we've decided to only focus on legal requests for now.

File: docs/spec/draft/source-requirements.md (hunk 9 above)
Comment (TomHennen) [reply docs/spec/draft/source-requirements.md:70]: Removed metadata/history rewrites entirely, so I think this comment is obsolete.

File: docs/spec/draft/source-requirements.md (hunk 6 above)
Comment (TomHennen) [reply docs/spec/draft/source-requirements.md]: Done

File: docs/spec/draft/source-requirements.md [hunk 11]
```diff
@@ -46,37 +46,47 @@ Consumers can examine the various source provenance attestations to determine if
...
 Policies in this category include things like commit signatures, author / committer formatting restrictions, closed-issue-linkage, etc.
 
+Organizations MUST make the fact of these changes public when they occur (e.g. an issue, announcement, or other mechanism).
+
 ### Repository renames
 
 When a repo is transferred to a new organization ("donated"), or if a repo must be renamed or otherwise have its url changed within the same org, attestations for previous revisions of this repo will no longer be matched because the combination of the repository id and the revision id will have changed.
 
+Organizations MUST inform all consumers of a repo of these changes when they occur.
+Organizations SHOULD provide a means to resolve previous revision IDs to the new revision IDs.
```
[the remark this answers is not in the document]
TomHennen: Removed this section entirely.

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,47 @@ Consumers can examine the various source provenance attestations to determine if
 
 ## Safe Expunging Process
 
-Administrators have the ability to expunge (remove) content from a repository and its change history without leaving a record of the removed content.
-This includes changing files, history, or changing references in git and is used to accommodate legal or privacy compliance requirements.
+SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
+This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements as well as administrative
+changes within a repository (see below for more information on the various scenarios).
```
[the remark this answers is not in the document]
TomHennen: I think this is resolved now as we've removed this section?

docs/spec/draft/source-requirements.md
```diff
@@ -46,37 +46,47 @@ Consumers can examine the various source provenance attestations to determine if
...
-The goal of this sections is to document that this process is allowed.
 Different organizations and tech stacks may have different approaches to the problem.
 
-Scenarios that need to be addressed:
+Safe Expunging Scenarios (non-exhaustive):
 
 ### Legal Takedowns
 
-A DMCA takedown request will be addressed by following an agreed-upon process.
+A legal takedown request will be addressed by following an agreed-upon process.
```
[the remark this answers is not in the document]
TomHennen: I suspect it could be too restrictive.  We've seen some examples in National Security Letters [that forbid disclosure](https://en.wikipedia.org/wiki/National_security_letter#:~:text=NSLs%20typically%20contain%20a%20nondisclosure%20requirement%20forbidding%20the%20recipient%20of%20an%20NSL%20from%20disclosing%20the%20FBI%20had%20requested%20the%20information).  I could see something like that happening here too.

I figure laws are written by governments which have their own agendas and may not care what SLSA says.

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,37 @@ Consumers can examine the various source provenance attestations to determine if
 
 ## Safe Expunging Process
 
-Administrators have the ability to expunge (remove) content from a repository and its change history without leaving a record of the removed content.
-This includes changing files, history, or changing references in git and is used to accommodate legal or privacy compliance requirements.
+SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content.
+This includes changing files, history, or changing references in git and is used to accommodate legal/privacy compliance requirements.
+
+Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact.
+In version control systems like git, removal of a revision changes the object id of all subsequent revisions that were built on top of it.
```
adityasaky, line 53:
```suggestion
In version control systems like Git, removal of a revision changes the object IDs of all subsequent revisions that were built on top of it.
```

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,37 @@ Consumers can examine the various source provenance attestations to determine if
...
+Although there is no "safe" way to do it, it sometimes necessary and there are steps you can take to mitigate the damage.
```
adityasaky, line 54:
```suggestion
Although there is no "safe" way to do it, it is sometimes necessary and there are steps you can take to mitigate the damage.
```

adityasaky: I think it might flow better to swap this sentence with the previous one, especially since the "it" here is different from the "it" the last sentence ends with.

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,37 @@ Consumers can examine the various source provenance attestations to determine if
...
+
 When used as an attack, this is called “repo hijacking” (or “repo-jacking”) and is one of the primary threats source provenance attestations protect against.
```
adityasaky, line 56: Does this text still belong here? I always understood this to mean repo takeovers or name squatting if a popular repository is renamed, and I think we've taken out repo renames from this.

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,37 @@ Consumers can examine the various source provenance attestations to determine if
...
-TODO: Determine how organizations can provide transparency around this process.
-At a minimum the organization would need to declare why data was removed from the branch.
-
-The goal of this sections is to document that this process is allowed.
 Different organizations and tech stacks may have different approaches to the problem.
 
-Scenarios that need to be addressed:
+Safe Expunging Scenarios:
 
 ### Legal Takedowns
```
adityasaky, line 70: If we have only one scenario, I wonder about reframing the section to say it's for legal takedowns only. OTOH, if we foresee adding scenarios, that'd create more work then, so we can punt until we're ready to publish and see if we still only have one scenario?

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,37 @@ Consumers can examine the various source provenance attestations to determine if
...
-That process should be documented itself and followed.
-It may be the case that the specific set of commits targeted by the takedown can be expunged in ways that do not impact revisions.
-
-### Commit metadata rewrites
+The only currently acceptable scenario to expunge data is in response to some legal or compliance request.
 
-A team may decide that all reachable commits in the history of a revision need to follow a new metadata convention.
-In git VCS, compliance with this new policy will require history to be rewritten (commit metadata is included in the computation of the revision id).
-Policies in this category include things like commit signatures, author / committer formatting restrictions, closed-issue-linkage, etc.
+The takedown request MUST be addressed by following an agreed-upon process.
```
adityasaky, line 74: This agreed-upon process is the same as the documented safe expunging process mentioned above, right? If so, perhaps this could be combined or refer to it?

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,37 @@ Consumers can examine the various source provenance attestations to determine if
...
 
-### Repository renames
+It may be the case that the specific set of commits targeted by the takedown can be expunged in ways that do not impact revisions.
 
-When a repo is transferred to a new organization ("donated"), or if a repo must be renamed or otherwise have its url changed within the same org, attestations for previous revisions of this repo will no longer be matched because the combination of the repository id and the revision id will have changed.
+This process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
```
adityasaky, line 78:
```suggestion
The application of the safe expunging process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
```

Also, could this say something to the effect of "It must be as transparent / public as possible within the constraints of complying with local laws and regulations and protection of sensitive data (e.g. PII)." I think, as it's written, it's easy to default to making these actions fully private rather than a best effort at transparency?

docs/spec/draft/source-requirements.md
mlieberman85: I think we need to be clearer also with no "safe" way. I understand we're using quotation marks around safe to indicate irony but I think it is easy for it to get confused especially since above we call it safe expunging process.

Maybe switching to something like

> Although there is no completely safe way to do it

Or we if we think even that is confusing maybe we call it something other than safe expunging process. Given that this is part of a specification it will be potentially interpreted in ways we can't anticipate, so anything we can do to clarify would be great.

docs/spec/draft/source-requirements.md
TomHennen: Honestly my preference is to accept your suggestion but leave it otherwise unchanged.  Setting up all the mechanisms for this will be very difficult.  If we say "must be as transparent/public as possible" we _will_ get questions like "how do we quantify that?", "what counts as transparent?", etc...

I'd prefer to err on the side of being less prescriptive here so that folks can more easily adopt.

docs/spec/draft/source-requirements.md
TomHennen: Yeah, there's lots of duplication.  I've re-written a bunch and should have addressed this.

docs/spec/draft/source-requirements.md
TomHennen: I've reworked the whole thing and only cover legal takedowns.  There's now a section for 'Warning' where we tell folks the problems they're going to encounter and then a section for 'Process' where we talk a bit more about what to do.

docs/spec/draft/source-requirements.md
TomHennen: I did add something that says they SHOULD disclose if possible, which is hopefully easy enough?

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,31 @@ Consumers can examine the various source provenance attestations to determine if
...
 
-On the git VCS, force pushes allow you to remove data from a branch.
-If a branch has been identified as consumable branch, force pushes to that branch must follow the safe expunging process.
+### Warning
 
-TODO: Determine how organizations can provide transparency around this process.
-At a minimum the organization would need to declare why data was removed from the branch.
+Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact.
+For example, in VCSs like Git, removal of a revision changes the object IDs of all subsequent revisions that were built on top of it,
+this can break downstream consumers ability to refer to source they've already integrated into their products.
```
adityasaky, line 57:
```suggestion
and this can break downstream consumers ability to refer to source they've already integrated into their products.
```

nit

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,31 @@ Consumers can examine the various source provenance attestations to determine if
...
 
-The goal of this sections is to document that this process is allowed.
-Different organizations and tech stacks may have different approaches to the problem.
-
-Scenarios that need to be addressed:
+It may be the case that the specific set of commits targeted by the takedown can be expunged in ways that do not impact revisions which can mitigate these problems.
```
adityasaky, line 59:
```suggestion
It may be the case that the specific set of changes targeted by the takedown can be expunged in ways that do not impact consumed revisions, which can mitigate these problems.
```

nit: making language a bit more generic and highlighting that "revisions" is the consumed ones from the previous sentence.

adityasaky (approved):
Some minor language suggestions but I think this looks good!

docs/spec/draft/source-requirements.md
TomHennen: So I was really trying to explain why it's bad that the object ID changes.

So maybe something more like:
```suggestion
, breaking downstream consumers ability to refer to source they've already integrated into their products.
```

adityasaky: Looks good. Follow up nit I just noticed: `breaking downstream consumers' ability to`, with the possessive apostrophe.

docs/spec/draft/source-requirements.md
marcelamelara: @TomHennen Is there an issue tracking the repo renaming discussion?

TomHennen: No, I didn't create one.  It sounded as though we could drop it until we needed to resolve it.  Should I create one anyways?

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,31 @@ Consumers can examine the various source provenance attestations to determine if
...
-At a minimum the organization would need to declare why data was removed from the branch.
+Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact.
+For example, in VCSs like Git, removal of a revision changes the object IDs of all subsequent revisions that were built on top of it,
+this can break downstream consumers ability to refer to source they've already integrated into their products.
 
-The goal of this sections is to document that this process is allowed.
-Different organizations and tech stacks may have different approaches to the problem.
-
-Scenarios that need to be addressed:
+It may be the case that the specific set of changes targeted by the takedown can be expunged in ways that do not impact consumed revisions, which can mitigate these problems.
```
marcelamelara, line 59: Nit: I don't think we've used the term takedown anywhere before, so adding this for clarity

```suggestion
It may be the case that the specific set of changes targeted by a legal takedown can be expunged in ways that do not impact consumed revisions, which can mitigate these problems.
```

docs/spec/draft/source-requirements.md
```diff
@@ -46,36 +46,31 @@ Consumers can examine the various source provenance attestations to determine if
...
-### Commit metadata rewrites
-
-A team may decide that all reachable commits in the history of a revision need to follow a new metadata convention.
-In git VCS, compliance with this new policy will require history to be rewritten (commit metadata is included in the computation of the revision id).
-Policies in this category include things like commit signatures, author / committer formatting restrictions, closed-issue-linkage, etc.
+An SCS MUST document the Safe Expunging Process and describe how requests and actions are tracked and SHOULD log the fact that content was removed.
+Different organizations and tech stacks may have different approaches to the problem.
 
-### Repository renames
+SCSs SHOULD require multi-party approval for any changes made under this process.
```
marcelamelara, line 69: This requirement isn't entire clear. Who are these multiple parties? And how are they authorized to approve a legal takedown?

marcelamelara (commented):
I think this process is much clearer now, thanks @TomHennen ! I have one remaining question about the multi-party approval.

docs/spec/draft/source-requirements.md
TomHennen: Reworded this.  Hopefully the new words are better?

marcelamelara (approved):
Thank you for these final wording changes @TomHennen ! LGTM

zachariahcox (approved):
This looks good to me and I recommend merging it. 

We will need to come back at some point to review a real life expungement and see what can actually be accomplished to here. How smooth a process can we make it in git? 

I'm also not sure the scs can be on the hook for enforcing that the data removals were for legal reasons. 
Generally I think "the owner of the intellectual property in the repo (the root repo owner, in gh terms) can remove data. They should only do this for legal or privacy reasons due to the risk of severe reputational consequences (damage to artifact chain of custody)." Ie: it's indistinguishable from a repo hijack so you should have a good reason.

TomHennen:
> This looks good to me and I recommend merging it.

Thanks!

> 
> We will need to come back at some point to review a real life expungement and see what can actually be accomplished to here. How smooth a process can we make it in git?

Agreed.  We'll probably want similar feedback for the entire build track.  Once we're in the RC phase (hopefully soon!) we can try to elicit that sort of feedback?

> 
> I'm also not sure the scs can be on the hook for enforcing that the data removals were for legal reasons. Generally I think "the owner of the intellectual property in the repo (the root repo owner, in gh terms) can remove data. They should only do this for legal or privacy reasons due to the risk of severe reputational consequences (damage to artifact chain of custody)." Ie: it's indistinguishable from a repo hijack so you should have a good reason.

Filed #1222 to track.

PR #1204, opened 2024-10-16
content: draft: Clarify name of Source L3

### LINKED ISSUE #1112: copy: clarify names of source levels
related to: https://github.com/slsa-framework/slsa/pull/1097#discussion_r1718489268

## Level 2
my initial thoughts are that we're trying to get across the following concepts:

teams can have more than one branch
teams may need to indicate that consumers can / should / must ignore commits on users/* and only ship commits on /releases/* because branches have different security postures
only some branches have protected history, IE, we allow force push to user branches.
the logical VSA for this rule would need to verify that "the previous revision id is reachable from this new revision id" (IE: there was no potential for data loss due to force push or repo hijack)

https://github.com/slsa-framework/slsa/pull/1097#discussion_r1714156093

## Level 3

* https://github.com/slsa-framework/slsa/pull/1097#discussion_r1714081313

Image

TomHennen:
Updating name of Source Level 3 to make it more clear by removing the somewhat ambiguous 'Source Provenance' and including the language from PR #1143 instead.

fixes #1112

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -100,7 +100,7 @@ All organizations of any size producing software of any kind.
 Benefits:
 Allows source consumers to track changes to the software over time and attribute those changes to the people that made them.
 
-### Level 3: Source Provenance Attestations
+### Level 3: Authenticatable and auditable attestations
 
 Summary:
 A consumer can ask the SCS for everything it knows about a specific revision of a repository.


mlieberman85 (approved):
LGTM

docs/spec/draft/source-requirements.md
```diff
@@ -100,7 +100,7 @@ All organizations of any size producing software of any kind.
 Benefits:
 Allows source consumers to track changes to the software over time and attribute those changes to the people that made them.
 
-### Level 3: Source Provenance Attestations
+### Level 3: Authenticatable and auditable attestations
```
adityasaky, line 103:
```suggestion
### Level 3: Authenticatable and Auditable Attestations
```

nit: title casing

adityasaky (approved):

docs/spec/draft/source-requirements.md
```diff
@@ -100,7 +100,7 @@ All organizations of any size producing software of any kind.
 Benefits:
 Allows source consumers to track changes to the software over time and attribute those changes to the people that made them.
 
-### Level 3: Source Provenance Attestations
+### Level 3: Authenticatable and Auditable Attestations
```
zachariahcox, line 103: According to the summary below, I think we're saying level three is "produces an unverifiable info artifact that has been signed."

In many ways the attestation _itself_ will not be very auditable, it represents a summary of work that happens on the platform side from the perspective of one or more tools in the SCS.

zachariahcox: I guess I had missed this pr: https://github.com/slsa-framework/slsa/pull/1143/files

Generally, I would not say VSAs are "auditable," and these are pretty close to summary attestations mechanically. They are claims about what a closed system allegedly did.
They should be _signed_, which would make them _authentic_, so you can trust the data exactly as much as you trust the issuer.

zachariahcox: I have to say, I like reuse of "provenance" for this. It seems the same as build provenance in the sense of "we don't know what it means or if it's true, but we know who said it."

TomHennen: There's a lot going on in this comment thread :)

1. What name to use.  It sounds like your suggesting we replace 'Attestations' with 'Provenance' which is fine with me.
2. You seem to be suggesting these claims aren't as strong as we would like, looking at the content, yeah it doesn't seem that strong.  I think instead of just providing that information in "a documented and tamper-resistant format" we could require the SCS to store the information it uses to validate security settings a documented and tamper-resistant format and to do so _contemporaneously_ as the contents and security related settings of a repo are changed.  I think this would make it auditable and provide a stronger guarantee.
3. "I would say VSAs are not "auditable," and these are pretty close to summary attestations mechanically" -> I'm really confused by this... To me VSAs _are_ summary attestations (at least [in the source track](https://slsa.dev/spec/draft/source-requirements#summary-attestation:~:text=The%20source%20track%20issues%20summary%20attestations%20using%20Verification%20Summary%20Attestations%20(VSAs)%20as%20follows%3A)).  And yeah, I'd agree that what we're asking for here aren't VSAs, but rather [the provenance attestations](https://slsa.dev/spec/draft/source-requirements#provenance-attestations).

WDYT?

TomHennen: For (1) updated to 'Authenticatable and Auditable Provenance'
For (2) filed https://github.com/slsa-framework/slsa/issues/1216
For (3) agreed, and I think the change in (1) should clarify that.

TomHennen:
I think the change from 'attestations' to 'provenance' should be uncontroversial.  So I'm going to go ahead and merge.  If anyone feels otherwise please let me know!

PR #1205, opened 2024-10-16
content: draft: clarify "in the present context"

### LINKED ISSUE #1138: What does "exist in the present context" mean?
> What does "exist in the present context" mean?

_Originally posted by @marcelamelara in https://github.com/slsa-framework/slsa/pull/1094#discussion_r1722360178_

@zachariahcox
>  for git repos, a revision is just a tree id in a big database.
Preventing trees from existing can only happen at push receive time and causes a ton of developer friction.
> 
> Especially when teams rely on pull requests, the "reachable from this ref" security boundary is the most practical. It models whether a commit / tree id is included in a specific subset. That subset has certain properties. I think that's what we're saying by context here, but it might need some additional definition here!

TomHennen:
My read is that "the present context" is what branches or tags point at the given revision.  Making that explicit.

fixes #1138

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -196,7 +196,8 @@ Activities conducted on the SCS SHOULD be attributed to authenticated identities
 <td><td>✓<td>✓
 <tr id="source-attestations"><td>Source attestations<td>
 
-A source attestation contains information about how a specific revision was created and how it came to exist in its present context.
+A source attestation contains information about how a specific revision was created and how it came to exist in its present context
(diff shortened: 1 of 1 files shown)


TomHennen:
I'm taking a shot at resolving this issue based on the conversation we've had.

PTAL?  I can adjust if this isn't right.

zachariahcox (approved):

mlieberman85 (approved):
LGTM

marcelamelara (approved):
Thanks @TomHennen !

Issue #1129, opened 2024-09-09
Define where source attestations are stored


TomHennen:
We should provide guidance on where source attestations are stored.

We may not want to be too prescriptive but should provide an allowance for how a source control system (#1128) should do so.

marcelamelara:
I think this is a general question for SLSA. @paveliak and I were having this exact discussion about the build environment attestations today as well. We also don't want to be too prescriptive, but provide some practical recommendations.

TomHennen:
Perhaps we should require the Source Control System (or Build system in the case of the build environment attestations) to document where they're stored and then provide examples in the sample profiles (#1142)?

For `git` we might suggest "store in the VCS like so", for other systems maybe we suggest a "content addressable storage system by <repository identifier>/<revision id>".

TomHennen:
My inclination is to say something like

"Systems which implement SLSA SHOULD store attestations about attested artifacts adjacent to the artifacts themselves. This aligns the failure domain of attestation distribution with artifact distribution and it helps align read ACLs of attestations with the artifacts. This makes it easier for consumers to use attestations as needed.

In all cases Systems MUST document where consumers can find the relevant attestations for the artifacts produced by the system.

Systems MAY replace more detailed attestations (e.g. provenance) with Verification Summary Attestations (VSAs) as needed for ease-of-use, performance, confidentiality, or any other reason, unless forbidden by the relevant SLSA track."

WDYT @marcelamelara @adityasaky @zachariahcox ?

TomHennen:
@adityasaky I'd especially like any thoughts you have on where these can be stored for git revisions :)

zachariahcox:
This is an interesting question. 
It makes sense to say "if you can read the artifact you must be able to find and read the attestations"

Issuers can come from anywhere though and they could all store things in different places -- I'm not sure how far we can go with saying things should be adjacent? 

Should this just be "organizations must explain where to find 'org-issued provenance' for their artifacts"? 
That could easily be "in the same folder as the artifact" for some orgs, or it could be pgi sigstore

TomHennen:
"Issuers can come from anywhere though"

Oh that's interesting, are you thinking of third parties outside the organization responsible for the attested artifact or are you thinking multiple parties within the organization?

adityasaky:
[...]

PR #1217, opened 2024-10-23
content: draft: Improve strength of Source Level 3

### LINKED ISSUE #1216: Improve strength of Source Level 3
...
2. You seem to be suggesting these claims aren't as strong as we would like, looking at the content, yeah it doesn't seem that strong.  I think instead of just providing that information in "a documented and tamper-resistant format" we could require the SCS to store the information it uses to validate security settings a documented and tamper-resistant format and to do so _contemporaneously_ as the contents and security related settings of a repo are changed.  I think this would make it auditable and provide a stronger guarantee.
...

_Originally posted by @TomHennen in https://github.com/slsa-framework/slsa/pull/1204#discussion_r1809401329_

TomHennen:
Previously level 3 just required the provenance/attestations to be tamper-resistant but didn't require those attestations to be created at any particular time. If an SCS were to create these attestations on-demand it would leave revisions more susceptible to tampering within the SCS between the time of their production and the time of the request.

By creating the attestations contemporaneously it reduces the period of time during which a threat actor would be able to falsify this evidence.

Also changed 'Source Attestations' to 'Source Provenance' to be inline with #1204 where we call it 'Provenance' and not 'Attestations'.

fixes #1216

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -103,7 +103,7 @@ Allows source consumers to track changes to the software over time and attribute
 ### Level 3: Authenticatable and Auditable Provenance
 
 Summary:
-The SCS is able to provide credible, tamper-resistant, and contemporaneous evidence, in a documented format,
+The SCS is able to generate credible, tamper-resistant, and contemporaneous evidence, in a documented format,
 of how a specific revision was created to authorized users of the source repository.
 
 Intended for:
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -103,7 +103,7 @@ Allows source consumers to track changes to the software over time and attribute
 ### Level 3: Authenticatable and Auditable Provenance
 
 Summary:
-The SCS generates credible, tamper-resistant, and contemporaneous evidence of how a specific revision was created. 
+The SCS generates credible, tamper-resistant, and contemporaneous evidence of how a specific revision was created.
 It is provided to authorized users of the source repository in a documented format.
 of how a specific revision was created to authorized users of the source repository.
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -103,7 +103,7 @@ Allows source consumers to track changes to the software over time and attribute
 ### Level 3: Authenticatable and Auditable Provenance
 
 Summary:
-The SCS is able to generate credible, tamper-resistant, and contemporaneous evidence, in a documented format,
+The SCS generates credible, tamper-resistant, and contemporaneous evidence, in a documented format,
 of how a specific revision was created to authorized users of the source repository.
 
 Intended for:
@@ -215,8 +215,8 @@ Source Provenance are attestations that contain information about how a specific
 (e.g. the branches or tags that point, or pointed, at that revision).
 They are associated with the revision identifier delivered to consumers and are a statement of fact from the perspective of the SCS.
 
-At Source Level 3 Source Provenance MUST be created contemporaneously with revision creation such that they provide a credible, auditable, record
-of changes.
+At Source Level 3 Source Provenance MUST be created contemporaneously with the revision being made available such that they provide a credible,
+auditable, record of changes.
 
 If a consumer is authorized to access source on a particular branch, they MUST be able to fetch the source attestation documents for revisions in the history of that branch.
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -103,7 +103,8 @@ Allows source consumers to track changes to the software over time and attribute
 ### Level 3: Authenticatable and Auditable Provenance
 
 Summary:
-The SCS generates credible, tamper-resistant, and contemporaneous evidence, in a documented format,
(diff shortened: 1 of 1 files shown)


TomHennen:
@adityasaky I'd also like your thoughts here.  I think this is inline with how gittuf works anyways?

docs/spec/draft/source-requirements.md
```diff
@@ -208,12 +209,15 @@ These forms of identity are user-provided and not typically verified by the sour
...
-<tr id="source-attestations"><td>Source attestations<td>
+<tr id="source-provenance"><td>Source Provenance<td>
 
-A source attestation contains information about how a specific revision was created and how it came to exist in its present context
+Source Provenance are attestations that contain information about how a specific revision was created and how it came to exist in its present context
 (e.g. the branches or tags that point, or pointed, at that revision).
 They are associated with the revision identifier delivered to consumers and are a statement of fact from the perspective of the SCS.
 
+At Source Level 3 Source Provenance MUST be created contemporaneously with revision creation such that they provide a credible, auditable, record
+of changes.
```
arewm, line 219: What does this mean in practice? The provenance cannot be created at the time of a commit because that commit may be local. It would have to be done when the commit is pushed to the SCS.

Is there any way that this could be verified, or would it be wrapped up into verifying trust in the SCS identity?

TomHennen: > What does this mean in practice? The provenance cannot be created at the time of a commit because that commit may be local. It would have to be done when the commit is pushed to the SCS.

Good question. Perhaps "contemporaneously with the revision entering the SCS"

> 
> Is there any way that this could be verified, or would it be wrapped up into verifying trust in the SCS identity?

I'd imagine it would be when verifying trust in the identity.  I suppose it might be possible to use a time-stamp-authority, but I'm not sure we want to _require_ that?

arewm: Sounds good on both accounts.

I agree, I don't think that we could require the latter in the track itself. It just seemed like a gap that others might point out as well.

docs/spec/draft/source-requirements.md
```diff
@@ -103,14 +103,15 @@ Allows source consumers to track changes to the software over time and attribute
 ### Level 3: Authenticatable and Auditable Provenance
 
 Summary:
-A consumer can ask the SCS for everything it knows about a specific revision of a repository.
-The information is provided in a documented and tamper-resistant format.
+The SCS is able to provide credible, tamper-resistant, and contemporaneous evidence, in a documented format,
```
adityasaky, line 106:
```suggestion
The SCS is able to generate credible, tamper-resistant, and contemporaneous evidence, in a documented format,
```

"provide" to me places the responsibility of attestation storage on the SCS. While that may be the case, I think we need a more explicit treatment of required properties for attestation storage, perhaps similar to https://slsa.dev/spec/v1.0/distributing-provenance#where-attestations-are-published. Without handling that, we're simultaneously saying that these attestations keep the SCS (somewhat) honest but also placing the power in the SCS to equivocate the attestations provided.

TomHennen: Maybe this shouldn't be "is able to generate" but just "generates"

Since we're asking for it to be contemporaneous it's not optional.

> While that may be the case, I think we need a more explicit treatment of required properties for attestation storage

Tracked in https://github.com/slsa-framework/slsa/issues/1129

docs/spec/draft/source-requirements.md
TomHennen: IMO, if folks are using gittuf, then local copies are sort of included within the SCS?

But that's just nit-picking.

I do like the "contemporaneously with the revision being made available" language better.  It's easier to understand.

docs/spec/draft/source-requirements.md
adityasaky: "generates" sounds good!

docs/spec/draft/source-requirements.md
adityasaky: > IMO, if folks are using gittuf, then local copies are sort of included within the SCS?

Yep! Which is why I wasn't sure we wanted to say the provenance is generated when a revision enters the SCS, as me creating a local commit is within the SCS in the gittuf model, but meaningless until I publish the revision for others to consume.

adityasaky (approved):

zachariahcox (approved):

arewm (approved):
Thanks.

docs/spec/draft/source-requirements.md
```diff
@@ -103,14 +103,15 @@ Allows source consumers to track changes to the software over time and attribute
 ### Level 3: Authenticatable and Auditable Provenance
 
 Summary:
-A consumer can ask the SCS for everything it knows about a specific revision of a repository.
-The information is provided in a documented and tamper-resistant format.
+The SCS generates credible, tamper-resistant, and contemporaneous evidence, in a documented format,
```
zachariahcox, line 106: This is sort of a complicated sentence. I recommend something like: 
```suggestion
The SCS generates credible, tamper-resistant, and contemporaneous evidence of how a specific revision was created. 
It is provided to authorized users of the source repository in a documented format.
```

TomHennen: Yeah that's much better.  Thanks!

PR #1220, opened 2024-10-24
resourceUri SHOULD match the download URI

### LINKED ISSUE #1212: Clarify how end-users can know the expected value of resourceUri in a VSA
In #1191 @zachariahcox says

On
> > *Threat:* Replace the package with one built using an unofficial CI/CD pipeline
that does not build in the correct way.
> > 
> > *Mitigation:* Verifier requires provenance showing that the builder matched an
> This is a great place to expand. Thanks for starting this.
> 
> I think the mitigation for this is essentially:
> 
> * Have a way to know what "correct" looks like for this resourceUri.
>   
>   * If you have access to such a thing, you can verify the artifact yourself by inspecting the data in the provenance attestation.
>   * If you do not, you must rely on a VSA with access to definition of "correct"
>   * If you do not have either, you _cannot know_ whether the artifact is safe to use, even if it was SLSA Build L3
> 
> I think in the example below "build L3" is conflated with "built in the correct way" which is not necessarily the case.
https://github.com/slsa-framework/slsa/pull/1191/files#r1806779663

I think this really boils down to better define (or suggest) how the resourceUri should be set.

Internally we prefer that the resourceUri match (or be derivable from) the URI used to fetch the artifact being verified.  The consumer knows "I'm fetching binary X.  Let me make sure <trusted party> says the binary I got is approved to be X".

Reading the existing draft recommendations this way of using the `resourceUri` is [implicit](https://slsa.dev/spec/draft/verification_summary#:~:text=vsa.resourceUri%20%3D%3D%20resolvedDependencies.uri).  Making it explicit would be helpful.

TomHennen:
When verifying VSAs consumers are expected to match the resourceUri with the 'expected value' but the spec doesn't currently indicate how that expected value is to be determined.

In this change we suggest the resourceUri be set to the URI the consumer will fetch the artifact from. If it's set to something else the producer MUST tell the user how to determine the expected value.

fixes #1212

### DIFF
--- a/docs/spec/draft/verification_summary.md
+++ b/docs/spec/draft/verification_summary.md
@@ -155,6 +155,13 @@ of the other top-level fields, such as `subject`, see [Statement]._
 `resourceUri` _string ([ResourceURI]), required_
 
 > URI that identifies the resource associated with the artifact being verified.
+>
+> The `resourceUri` SHOULD be set to the URI the producer expects the consumer
+> to fetch the artifact being verified from. This enables the consumer to easily
+> determine the expected value when [verifying](#how-to-verify). If the
+> `resourceUri` is set to some other value, the producer MUST communicate the
+> expected value, or how to determine the expected value, to consumers through
+> out-of-band channel.
 
 <a id="policy"></a>
 `policy` _object ([ResourceDescriptor]), required_


TomHennen:
@adityasaky this was your suggestion, so I'd love your thoughts too.

docs/spec/draft/verification_summary.md
```diff
@@ -155,6 +155,13 @@ of the other top-level fields, such as `subject`, see [Statement]._
 `resourceUri` _string ([ResourceURI]), required_
 
 > URI that identifies the resource associated with the artifact being verified.
+>
+> The `resourceUri` SHOULD be set to the URI the producer expects the consumer
```
hepwori, line 159: Nit: Sentence reads a little tangled to me. Consider "The resourceURI SHOULD be set to the URI from which the producer expects the consumer to fetch the artifact for verification".

mlieberman85: +1

TomHennen: Done!

hepwori (approved):
lgtm

adityasaky (approved):
LGTM!

arewm (approved):

mlieberman85 (approved):
LGTM

trishankatdatadog (approved):

Issue #1181, opened 2024-10-09
TODO: Need mitigation description for "Dependency confusion" threat


lehors:
The 'Dependency Confusion' threat ([link](https://slsa.dev/spec/draft/threats#usage-threats:~:text=Threat%3A%20Register,of%20this%20section)) needs a mitigation section and perhaps examples.

TomHennen:
@meder would you be interested in sending a PR for this?  It's very much inline with the dependency track and [your recent blog post](https://slsa.dev/blog/2024/08/dep-confusion-and-typosquatting).

michaelwinser:
I would suggest that all builds should pull only from an internal artifact registry and that admission to that registry be explicitly managed and audited. Especially the namespace.

meder:
I suppose there are two ways to look at the mitigations on the threats page: using existing SLSA tracks vs best practices. I was leaning towards applying the existing SLSA tracks lens, given that I see explicit callouts to SLSA not addressing some threats. I think the idea is to give others a way to understand what SLSA does and doesn't address. Once the dependency track is a thing the page could be updated.

TomHennen:
> I suppose there are two ways to look at the mitigations on the threats page: using existing SLSA tracks vs best practices. I was leaning towards applying the existing SLSA tracks lens, given that I see explicit callouts to SLSA not addressing some threats. I think the idea is to give others a way to understand what SLSA does and doesn't address. Once the dependency track is a thing the page could be updated.

This is what I was thinking.  For now just say it's not addressed but link to Meder's recent blog post for ideas.  Then update once the dep track is done.

marcelamelara:
Resolved in #1226

PR #1226, opened 2024-10-29
content: Update mitigation section for the Dependency Confusion threat.

meder:
Documenting a SLSA-native and build trackccentric mitigation for Dependency Confusion attacks (#1181)

Would love to hear thoughts/opinions on the best way to reflect differing levels of adoption / maturity in native provenance verification across different ecosystems.

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -775,9 +775,18 @@ The consumer requests a package that it did not intend.
 on the victim's internal registry, and wait for a misconfigured victim to fetch
 from the public registry instead of the internal one.
 
-**TODO:** fill out the rest of this section
+*Mitigation:* The mitigation is for the software producer to build internal
+packages on a SLSA Level 2+ compliant build system and define expectations for
+build provenance. Expectations must be verified on installation of the internal
+packages. If a misconfigured victim attempts to install attacker's package with
+an internal name but from the public registry, then verification against
+expectations will fail.
+
+For more information see [Verifying artifacts](https://slsa.dev/spec/v1.1/verifying-artifacts)
+and [Defender's Perspective: Dependency Confusion and Typosquatting Attacks](https://slsa.dev/blog/2024/08/dep-confusion-and-typosquatting).
 
 </details>
+
 <details><summary>Typosquatting</summary>
 
 *Threat:* Register a package name that is similar looking to a popular package
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -782,8 +782,8 @@ packages. If a misconfigured victim attempts to install attacker's package with
 an internal name but from the public registry, then verification against
 expectations will fail.
 
-For more information see [Verifying artifacts](https://slsa.dev/spec/v1.1/verifying-artifacts)
-and [Defender's Perspective: Dependency Confusion and Typosquatting Attacks](https://slsa.dev/blog/2024/08/dep-confusion-and-typosquatting).
+For more information see [Verifying artifacts](/spec/v1.1/verifying-artifacts)
+and [Defender's Perspective: Dependency Confusion and Typosquatting Attacks](/blog/2024/08/dep-confusion-and-typosquatting).
 
 </details>
 


docs/spec/draft/threats.md
```diff
@@ -775,9 +775,18 @@ The consumer requests a package that it did not intend.
 on the victim's internal registry, and wait for a misconfigured victim to fetch
 from the public registry instead of the internal one.
 
-**TODO:** fill out the rest of this section
+*Mitigation:* The mitigation is for the software producer to build internal
+packages on a SLSA Level 2+ compliant build system and define expectations for
+build provenance. Expectations must be verified on installation of the internal
+packages. If a misconfigured victim attempts to install attacker's package with
```
arewm, line 781: Do we need to be more precise about what we mean by "define expectations for build provenance" here? It is not clear to me what expectations we are verifying. 

Also, if we are doing verifications on the provenance, that is after the build, right? If so, it doesn't really follow for me to verify on installation. Are you trying to say that we must verify that internal packages are installed/used and not some untrusted external one?

TomHennen: Checking 'expectations' is commonly used terminology on this page.  The other threats mention it without further explanation.  The explanation about how to do so is in https://slsa.dev/spec/v1.1/verifying-artifacts#step-2-check-expectations.

docs/spec/draft/threats.md
```diff
@@ -775,9 +775,18 @@ The consumer requests a package that it did not intend.
 on the victim's internal registry, and wait for a misconfigured victim to fetch
 from the public registry instead of the internal one.
 
-**TODO:** fill out the rest of this section
+*Mitigation:* The mitigation is for the software producer to build internal
```
TomHennen, line 778: optional: Would we prefer to say that SLSA doesn't currently address this threat but that a future tracks plans to (insert relevant links)?

One reason we might prefer not to address this threat as written is that not all methods of forming expectations will address it.  E.g. in 'verifying-artifacts' it talks about [forming expectations](https://slsa.dev/spec/v1.1/verifying-artifacts#forming-expectations) and the first thing listed is TOFU.  In the TOFU model this threat isn't addressed, we'd just wind up anchoring on the malicious package's provenance values.  'Defined in source' has a similar problems in that the package name is bound to a source repo, if the wrong name is requested then the malicious expectations will be pulled instead.

The only way of forming expectations that works here is "Defined by producer", but it's not clear how those expectations are communicated.  You could imagine a similar one called 'defined by organization' where the org sets a policy that says "unless you get an exception for a package it must come from builds A, B, C and source repos X, Y, Z".

adityasaky: Given that this threat is very specifically about a public package shadowing a name from the internal registry, it would still be possible to enforce expectations on the internal-only packages at least since the producer and consumer is the same entity, I think? In other words, set expectations for package `foo` (known internal) in `resolvedDependencies` but not package `bar` which is fine to pull in from the external registry.

Separately, this might need a higher level of completeness in build provenance than is currently typical. For example, looking at [NPM's use of provenance](https://search.sigstore.dev/?logIndex=139985224), `resolvedDependencies` doesn't record all the packages that are actually resolved but rather just the git revision. Should this be called out?

adityasaky: > Right, that's a better way of describing what I was trying to get at with expectations that are set by the organization.

My bad, I misread the last bit of that comment. It's the same idea. 😄 

> Not quite sure what you mean? That's all we need to address this threat as the source repo listed there is the 'main' one that drives the build itself (IIUC).

If the provenance only records a git revision, we don't have sufficient information to enforce the "defined by organization" list for internal dependencies as the packages fetched by the dependency resolver for the manifest at that git revision is unknown.

In other words, in addition to saying SLSA level 2+ builder, maybe this text needs to call out that `resolvedDependencies` needs to be more complete to enforce the expectations about the organization's internal packages.

docs/spec/draft/threats.md
meder: re: the last question: yes.

docs/spec/draft/threats.md
meder: Great discussion, thanks all.

Case 1 is indeed what I am describing with the caveat that the producer should be able to easily communicate/distribute the expectations to the consumers as the producer and the organization consuming dependencies are the same entity. The verification can then happen at different points in SDLC: it could happen at the time dependencies are ingested into the internal registry, it could happen on build/installation.

trishankatdatadog: Totally agree with Tom: dependency confusion is not within SLSA scope, but perhaps best within Securing Software Repos WG.

adityasaky: Ahh, yes, I was misunderstanding @meder's intent. I think we can close this thread.

adityasaky (approved):

docs/spec/draft/threats.md
```diff
@@ -775,9 +775,18 @@ The consumer requests a package that it did not intend.
 on the victim's internal registry, and wait for a misconfigured victim to fetch
```
zachariahcox, line 775: to be clear, is "fetch" the problem here or the fact that a bespoke tool needs to know that that a build should be failed if a package doesn't match its expected issuer and producer and you have to know ahead of time what's expected. 

potentially the mitigation for this is just a redirect to the "verifying-artifacts" file?

docs/spec/draft/threats.md
TomHennen: Just reopening because I think there's still some outstanding disagreement here.

Option 1: Leave this as is, even if it's hard to implement
Option 2: Say "SLSA doesn't currently handle this, but read this blog post, see this S2C2F docs, see the forthcoming dependency track".  <--- preferred by the folks at the SLSA spec meeting.

meder:
And here's a blogpost on the topic: [Defender's Perspective: Dependency Confusion and Typosquatting Attacks](https://slsa.dev/blog/2024/08/dep-confusion-and-typosquatting).

docs/spec/draft/threats.md
meder: Knowing ahead of time problem is solved by the fact that the organization is the producer of the internal artifacts  and thus are able to distribute expectations for provenance within the organization.

Custom or native tooling would work here, with native tooling being preferred due to the issue you highlight. Though given the page's context: threats that SLSA can solve/mitigate given current spec, I didn't really go into the implementation details.

meder:
Hi all, 

re: Dependency track: I think it can indeed make it easier to defend against dependency confusion and we can update this page with the new capabilities once that lands.

It might be worth agreeing on the purpose of the threats page. Given the current wording on the page I view its purpose as a place to highlight how current SLSA specification can help mitigate common supply chain threats **at a high level**. At its core dependency confusion is about the lack of verifiable provenance, hence the current wording suggested in this PR. Is this how others view the purpose of this page?

zachariahcox (approved):
this looks correct to me 👍

TomHennen (approved):

TomHennen:
@arewm, @trishankatdatadog any objections to merging this?  I think we have enough folks, and it's been sitting here for a  bit.

I'll probably merge by EoY unless you say otherwise. :)

TomHennen:
Er, I meant "end of week" not "end of year"...

We can always make further changes as needed.

Issue #1178, opened 2024-10-09
TODO: Needs mitigation description for "Software producer intentionally submits bad code" threat


lehors:

zachariahcox:
I think the changes needed are here: 
https://github.com/search?q=repo%3Aslsa-framework%2Fslsa%20Software%20producer%20intentionally&type=code

* https://github.com/slsa-framework/slsa/blob/a9ba55a455e1be6f93167d49e1e43217a0a3a51e/docs/spec/v1.1/about.md?plain=1#L133
* https://github.com/slsa-framework/slsa/blob/a9ba55a455e1be6f93167d49e1e43217a0a3a51e/docs/spec/v1.1/threats.md?plain=1#L84
* https://github.com/slsa-framework/slsa/blob/a9ba55a455e1be6f93167d49e1e43217a0a3a51e/docs/spec/v1.1/threats-overview.md?plain=1#L3

michaelwinser:
Mitigating against this threat and more generally across major organizational trust boundaries is not a well solved problem.

Question: is this a case of the "software producer intentionally submits bad code" or "someone inside the software producer organization intentionally submits bad code." 


The most formal approach is to trest all externally produced software as source code that is being imported into your internal source tree. 

1. All changes should be reviewed by two internal parties
2. Always build from source

In practice this is not always possible. Organizations can and do rely on 3rd party auditors. Organizations can and should also ask for evidence of secure practices. 

At the end of the day, if an organization cannot be trusted but its software is essential then sandboxing and runtime auditing woudl be required.

PR #1236, opened 2024-11-18
content: source track: update source threats for draft spec

zachariahcox:
fixes: https://github.com/slsa-framework/slsa/issues/1178

The threats.md file is intended to be a one-stop summary of all the threats mitigated by SLSA. 
From this page, we should plan to redirect to other parts of the documentation.

### DIFF
--- a/docs/spec/draft/future-directions.md
+++ b/docs/spec/draft/future-directions.md
@@ -59,8 +59,8 @@ the basis for a future Build Platform Operations track:
 
 The SLSA Source track will describe increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc).
 
-The Source track will be scoped to revisions of a single repository that is controlled by an organization.
-The organization determines the intent of the software in the repository, what Source level should apply to the repository, and administers technical controls to enforce that level.
+The Source track will be scoped to revisions of a single repository.
+The intent of each revision is determined by the [software producer](terminology.md#roles) who is also responsible for declaring which Source level should apply to each revision and administering technical controls to enforce that level.
 
-The primary purpose of the Source track will be to enable verification that the creation of a revision followed the expected process.
-Consumers will be able to examine source provenance attestations to determine if all sources used during the build meet their requirements.
+The primary purpose of the Source track will be to enable verification that the creation of a revision followed the producer's intended process.
+Consumers will be able to examine source provenance attestations to determine if a revision meet their requirements.
--- a/docs/spec/draft/future-directions.md
+++ b/docs/spec/draft/future-directions.md
@@ -60,7 +60,7 @@ the basis for a future Build Platform Operations track:
 The SLSA Source track will describe increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc).
 
 The Source track will be scoped to revisions of a single repository.
-The intent of each revision is determined by the [software producer](terminology.md#roles) who is also responsible for declaring which Source level should apply to each revision and administering technical controls to enforce that level.
+The intent of each revision is determined by the [software producer](terminology.md#roles) who is also responsible for declaring which Source level should apply and administering technical controls to enforce that level.
 
 The primary purpose of the Source track will be to enable verification that the creation of a revision followed the producer's intended process.
 Consumers will be able to examine source provenance attestations to determine if a revision meet their requirements.
--- a/docs/spec/draft/future-directions.md
+++ b/docs/spec/draft/future-directions.md
@@ -16,7 +16,7 @@ additional aspects of automatable supply chain security.
 ### Build L4
 
 A build L4 could include further hardening of the build platform and enabling
-corraboration of the provenance, for example by providing complete knowledge of
+corroboration of the provenance, for example by providing complete knowledge of
 the build inputs.
 
 The initial [draft version (v0.1)] of SLSA defined a "SLSA 4" that included the
@@ -54,3 +54,14 @@ the basis for a future Build Platform Operations track:
 </section>
 
 [draft version (v0.1)]: ../v0.1/requirements.md
+
+
+## Source Track
+
+The SLSA Source track will describe increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc).
+
+The Source track will be scoped to revisions of a single repository that is controlled by an organization.
+The organization determines the intent of the software in the repository, what Source level should apply to the repository, and administers technical controls to enforce that level.
+
+The primary purpose of the Source track will be to enable verification that the creation of a revision followed the expected process.
+Consumers will be able to examine source provenance attestations to determine if all sources used during the build meet their requirements.
--- a/docs/spec/draft/future-directions.md
+++ b/docs/spec/draft/future-directions.md
@@ -55,7 +55,6 @@ the basis for a future Build Platform Operations track:
 
 [draft version (v0.1)]: ../v0.1/requirements.md
 
-
 ## Source Track
 
 The SLSA Source track will describe increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc).
--- a/docs/spec/draft/threats-overview.md
+++ b/docs/spec/draft/threats-overview.md
@@ -23,10 +23,9 @@ availability. Integrity means protection against tampering or unauthorized
 modification at any stage of the software lifecycle. Within SLSA, we divide
 integrity into source integrity vs build integrity.
 
-**Source integrity:** Ensure that all changes to the source code reflect the
-intent of the software producer. Intent of an organization is difficult to
-define, so SLSA is expected to approximate this as approval from two authorized
-representatives.
+**Source integrity:** Ensure that source revisions contain only changes submitted by
+authorized contributors according to the process defined by the software producer and
+that source revisions are not modified as they pass between development stages.
 
 **Build integrity:** Ensure that the package is built from the correct,
 unmodified sources and dependencies according to the build recipe defined by the
--- a/docs/spec/draft/threats-overview.md
+++ b/docs/spec/draft/threats-overview.md
@@ -23,9 +23,7 @@ availability. Integrity means protection against tampering or unauthorized
 modification at any stage of the software lifecycle. Within SLSA, we divide
 integrity into source integrity vs build integrity.
 
-**Source integrity:** Ensure that source revisions contain only changes submitted by
-authorized contributors according to the process defined by the software producer and
-that source revisions are not modified as they pass between development stages.
+**Source integrity:** Ensure that the source revision represents the intent of the producer, that all expected processes were followed and that the revision was not modified after being accepted.
 
 **Build integrity:** Ensure that the package is built from the correct,
 unmodified sources and dependencies according to the build recipe defined by the
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -53,13 +53,11 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
 
 A source integrity threat is a potential for an adversary to introduce a change
 to the source code that does not reflect the intent of the software producer.
-This includes the threat of an authorized individual introducing an unauthorized
-change---in other words, an insider threat.
+This includes modification of the source data at rest as well as insider threats,
+when an authorized individual introduces an unauthorized change.
 
-SLSA v1.0 does not address source threats, but we anticipate doing so in a
-[future version](current-activities.md#source-track). In the meantime, the
-threats and potential mitigations listed here show how SLSA v1.0 can fit into a
-broader supply chain security program.
+SLSA does not yet address source threats, but we anticipate doing so in a
+[future version](future-directions#source-track).
 
 ### (A) Producer
 
@@ -67,70 +65,53 @@ The producer of the software intentionally produces code that harms the
 consumer, or the producer otherwise uses practices that are not deserving of the
 consumer's trust.
 
-Threats in this category likely *cannot* be mitigated through controls placed
-during the authoring/reviewing process, in contrast with (B).
+<details><summary>Software producer intentionally creates a malicious revision of the source</summary>
 
-<!--
-**TODO:** The difference between (A) and (B) is still a bit fuzzy, which would
-be nice to resolve. For example, compromised developer credentials - is that (A)
-or (B)?
--->
+*Threat:* A producer intentionally creates a malicious revision (or a VSA issuer intentionally creates a malicious attestation) with the intent of harming their consumers.
 
-<details><summary>Software producer intentionally submits bad code</summary>
+*Mitigation:*
+This kind of attack cannot be directly mitigated through SLSA controls.
+Consumers must establish some basis to trust the organizations from which they consume software.
+That basis may be:
+* The code is open source and has a sufficiently large user-base that malicious changes are likely to be detected.
+* The organization has sufficient legal or reputational incentives to dissuade it from making malicious changes.
 
-*Threat:* Software producer intentionally submits "bad" code, following all
-proper processes.
+Ultimately this is a judgement call with no straightforward answer.
 
-*Mitigation:* **TODO**
-
-*Example:* A popular extension author sells the rights to a new owner, who then
-modifies the code to secretly mine cryptocurrency at the users' expense. SLSA
-does not protect against this, though if the extension were open source, regular
-auditing may discourage this from happening.
+*Example:* A producer with an otherwise good reputation decides suddenly to produce a malicious artifact with the intent to harm their consumers.
 
 </details>
 
-<!--
-**TODO:** More producer threats? Perhaps the attack to xz where a malicious
-contributor gained enhanced privileges through social engineering?
--->
+### (B) Modifying the source
 
-### (B) Authoring & reviewing
+An adversary without any special administrator privileges attempts to introduce a change counter to the declared intent of the source by following the producer's official source control process.
 
-An adversary introduces a change through the official source control management
-interface without any special administrator privileges.
-
-Threats in this category *can* be mitigated by code review or some other
-controls during the authoring/reviewing process, at least in theory. Contrast
-this with (A), where such controls are likely ineffective.
+Threats in this category can be mitigated by following source control management best practices.
 
 #### (B1) Submit change without review
 
 <details><summary>Directly submit without review</summary>
 
-*Threat:* Submit bad code to the source repository without another person
-reviewing.
+*Threat:* Submit code to the source repository without another person reviewing.
 
-*Mitigation:* Source repository requires two-person approval for all changes.
+*Mitigation:* The producer requires approval of all changes before they are accepted.
 
-*Example:* Adversary directly pushes a change to a GitHub repo's `main` branch.
-Solution: Configure GitHub's "branch protection" feature to require pull request
-reviews on the `main` branch.
+*Example:* Adversary directly pushes a change to a git repo's `main` branch.
+Solution: The producer can configure branch protection rules on the `main` branch.
+A best practice is to require approval of any changes via a change management tool before they are accepted into the source.
 
 </details>
-<details><summary>Review own change through a sock puppet account</summary>
+<details><summary>Single actor controls multiple accounts</summary>
 
-*Threat:* Propose a change using one account and then approve it using another
-account.
+*Threat:* An actor is able to control multiple account and effectively approve their own code changes.
 
-*Mitigation:* Source repository requires approval from two different, trusted
-persons. If the proposer is trusted, only one approval is needed; otherwise two
-approvals are needed. The software producer maps accounts to trusted persons.
+*Mitigation:* The producer must ensure that no actor is able to control or influence multiple accounts with review privileges.
 
-*Example:* Adversary creates a pull request using a secondary account and then
-approves and merges the pull request using their primary account. Solution:
-Configure branch protection to require two approvals and ensure that all
-repository contributors and owners map to unique persons.
+*Example:* Adversary creates a pull request using a secondary account and approves it using their primary account.
+
+Solution: The producer must require strongly authenticated user accounts and ensure that all accounts map to unique persons.
+A common vector for this attack is to take over a robot account with the permission to contribute code.
+Control of the robot and an actors own legitimate account is enough to exploit this vulnerability.
 
 </details>
 <details><summary>Use a robot account to submit change</summary>
@@ -142,63 +123,63 @@ two-person review.
 robots.
 
 *Example:* A file within the source repository is automatically generated by a
-robot, which is allowed to submit without review. Adversary compromises the
-robot and submits a malicious change without review. Solution: Require human
-review for these changes.
-
-<!--
-> TODO([#196](https://github.com/slsa-framework/slsa/issues/196)) This solution
-> may not be practical. Should there be an exception for locked down robot
-> accounts?
--->
+robot, which is allowed to submit without review.
+Adversary compromises the robot and submits a malicious change.
+Solution: Require review for such changes.
 
 </details>
-<details><summary>Abuse review exceptions</summary>
+<details><summary>Abuse of rule exceptions</summary>
 
-*Threat:* Exploit a review exception to submit a bad change without review.
+*Threat:* Rule exceptions provide vector for abuse
 
-*Mitigation:* All changes require two-person review without exception.
+*Mitigation:* Remove rule exceptions.
 
-*Example:* Source repository requires two-person review on all changes except
-for "documentation changes," defined as only touching files ending with `.md` or
-`.html`. Adversary submits a malicious executable named `evil.md` without review
-using this exception, and then builds a malicious package containing this
-executable. This would pass the policy because the source repository is correct,
-and the source repository does require two-person review. Solution: Do not allow
-such exceptions.
-
-<!--
-> TODO This solution may not be practical in all circumstances. Are there any
-> valid exceptions? If so, how do we ensure they cannot be exploited?
--->
+*Example:* The intent of a producer is to require two-person review on "all changes except for documentation changes," defined as those only modifying `.md` files.
+Adversary submits a malicious executable named `evil.md` and a code review is not required due to the exception.
+Technically, the intent of the producer was followed and the produced malicious revision meets all defined policies.
+Solution: The producer adjusts the rules to prohibit such exceptions.
 
 </details>
 
-#### (B2) Evade code review requirements
+<details><summary>Highly-permissioned actor bypasses or disables controls</summary>
+
+*Threat:* Trusted actor with "admin" privileges in a repository submits code by disabling existing controls.
+
+*Mitigation:* All actors must be subject to same controls, whether or not they have
+administrator privileges.
+Changes to the controls themselves should require their own review process.
+
+*Example 1:* A GitHub repository-level admin pushes a change without review, even though GitHub branch protection is enabled.
+Solution: The producer can modify the rule to disallow bypass by administrators, or move the rule to an organization-level ruleset.
+
+*Example 2:* GitHub repository-level admin removes a branch requirement, pushes their change, then re-enables the requirement to cover their tracks.
+Solution: The producer can configure higher-permission-level rules (such as organization-level GitHub Rulesets) to prevent repository-level tampering.
+
+#### (B2) Evade change management process
 
 <details><summary>Modify code after review</summary>
 
 *Threat:* Modify the code after it has been reviewed but before submission.
 
-*Mitigation:* Source control platform invalidates approvals whenever the
-proposed change is modified.
+*Mitigation:* Source control platform invalidates approvals whenever the proposed change is modified.
 
 *Example:* Source repository requires two-person review on all changes.
-Adversary sends a "good" pull request to a peer, who approves it. Adversary then
-modifies it to contain "bad" code before submitting. Solution: Configure branch
-protection to dismiss stale approvals when new changes are pushed.
+Adversary sends an initial "good" pull request to a peer, who approves it.
+Adversary then modifies their proposal to contain "bad" code.
 
-> Note: This is not currently a SLSA requirement because the productivity hit is
-> considered too great to outweigh the security benefit. The cost of code review
-> is already too high for most projects, given current code review tooling, so
-> making code review even costlier would not further our goals. However, this
-> should be considered for future SLSA revisions once the state-of-the-art for
-> code review has improved and the cost can be minimized.
+Solution: Configure the code review rules to require review of the most recent revision before submission.
+Resetting or "dismissing" votes on a PR introduces substantial friction to the process.
+Depending on the security posture of the source, the producer has a few choices to deal with this situation.
+They may:
+
+-   Accept this risk. Code review is already expensive and the pros outweigh the cons here.
+-   Dismiss reviews when new changes are added. This is a common outcome when expert code review is required.
+-   Leave previous reviews intact, but require that "at least the last revision must be reviewed by someone."
 
 </details>
 <details><summary>Submit a change that is unreviewable</summary>
 
-*Threat:* Send a change that is meaningless for a human to review that looks
+*Threat:* Adversary crafts a change that is meaningless for a human to review that looks
 benign but is actually malicious.
 
 *Mitigation:* Code review system ensures that all reviews are informed and
@@ -206,8 +187,8 @@ meaningful.
 
 *Example:* A proposed change updates a file, but the reviewer is only presented
 with a diff of the cryptographic hash, not of the file contents. Thus, the
-reviewer does not have enough context to provide a meaningful review. Solution:
-the code review system should present the reviewer with a content diff or some
+reviewer does not have enough context to provide a meaningful review.
+Solution: the code review system should present the reviewer with a content diff or some
 other information to make an informed decision.
 
 </details>
@@ -220,44 +201,27 @@ different context.
 
 *Example:* MyPackage's source repository requires two-person review. Adversary
 forks the repo, submits a change in the fork with review from a colluding
-colleague (who is not trusted by MyPackage), then merges the change back into
-the upstream repo. Solution: The merge should still require review, even though
-the fork was reviewed.
+colleague (who is not trusted by MyPackage), then proposes the change to
+the upstream repo.
+Solution: The proposed change still requires two-person review in the upstream context even though it received two-person review in another context.
 
 </details>
-<details><summary>Compromise another account</summary>
-
-*Threat:* Compromise one or more trusted accounts and use those to submit and
-review own changes.
-
-*Mitigation:* Source control platform verifies two-factor authentication, which
-increases the difficulty of compromising accounts.
-
-*Example:* Trusted person uses a weak password on GitHub. Adversary guesses the
-weak password, logs in, and pushes changes to a GitHub repo. Solution: Configure
-GitHub organization to requires 2FA for all trusted persons. This would increase
-the difficulty of using the compromised password to log in to GitHub.
 
 </details>
-<details><summary>Hide bad change behind good one</summary>
+<details><summary>Commit graph attacks</summary>
 
-*Threat:* Request review for a series of two commits, X and Y, where X is bad
-and Y is good. Reviewer thinks they are approving only the final Y state whereas
-they are also implicitly approving X.
+*Threat:* Request review for a series of two commits, X and Y, where X is bad and Y is good.
+Reviewer thinks they are approving only the final Y state but they are also implicitly approving X.
 
-*Mitigation:* Only the version that is actually reviewed is the one that is
-approved. Any intermediate revisions don't count as being reviewed.
+*Mitigation:* The producer declares that only the final delta is considered approved.
+In this configuration, intermediate revisions are not considered to be approved and are not added to the protected context (e.g. the `main` branch).
+With git version control systems this is called a "squash" merge strategy.
 
-*Example:* Adversary sends a pull request containing malicious commit X and
-benign commit Y that undoes X. In the pull request UI, reviewer only reviews and
-approves "changes from all commits", which is a delta from HEAD to Y; they don't
-see X. Adversary then builds from the malicious revision X. Solution: Policy
-does not accept this because the version X is not considered reviewed.
+*Example:* Adversary sends a pull request containing malicious commit X and benign commit Y that undoes X.
+The produced diff of X + Y contains zero lines of changed code and the reviewer may not notice that X is malicious unless they review each commit in the request.
+If X is allowed to become reachable from the protected branch, the content may become available in secured environments such as developer machines.
 
-<!--
-> TODO This is implicit but not clearly spelled out in the requirements. We
-> should consider clarifying if there is confusion or incorrect implementations.
--->
+Solution: The code review tool does not merge contributor-created commits, and instead merges a single new commit representing only the reviewed "changes from all commits."
 
 </details>
 
@@ -267,8 +231,10 @@ does not accept this because the version X is not considered reviewed.
 
 *Threat:* Two trusted persons collude to author and approve a bad change.
 
-*Mitigation:* This threat is not currently addressed by SLSA. We use "two
-trusted persons" as a proxy for "intent of the software producer".
+*Mitigation:* The producer can arbitrarily increase friction of their policies to reduce risk, such as requiring additional, or more senior reviewers.
+The goal of policy here is to ensure that the approved changes match the intention of the producer for the source.
+Increasing the friction of the policies may make it harder to circumvent, but doing so has diminishing returns.
+Ultimately the producer will need to land upon a balanced risk profile that makes sense for their security posture.
 
 </details>
 <details><summary>Trick reviewer into approving bad code</summary>
@@ -293,29 +259,6 @@ stamping."
 An adversary introduces a change to the source control repository through an
 administrative interface, or through a compromise of the underlying
 infrastructure.
-
-<details><summary>Project owner bypasses or disables controls</summary>
-
-*Threat:* Trusted person with "admin" privileges in a repository submits "bad"
-code bypassing existing controls.
-
-*Mitigation:* All persons are subject to same controls, whether or not they have
-administrator privileges. Disabling the controls requires two-person review (and
-maybe notifies other trusted persons?)
-
-*Example 1:* GitHub project owner pushes a change without review, even though
-GitHub branch protection is enabled. Solution: Enable the "Include
-Administrators" option for the branch protection.
-
-*Example 2:* GitHub project owner disables "Include Administrators", pushes a
-change without review, then re-enables "Include Administrators". This currently
-has no solution on GitHub.
-
-<!--
-> TODO This is implicit but not clearly spelled out in the requirements. We
-> should consider clarifying since most if not all existing platforms do not
-> properly address this threat.
--->
 
 </details>
 <details><summary>Platform admin abuses privileges</summary>
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -74,7 +74,7 @@ This kind of attack cannot be directly mitigated through SLSA controls.
 Consumers must establish some basis to trust the organizations from which they consume software.
 That basis may be:
 
--   The code is open source and has a sufficiently large user-base that malicious changes are likely to be detected.
+-   The repo is open source with an active user-base. High numbers of engaged users may increase the likelihood that bad code is detected during code review and reduce the time-to-detection when bad code makes it in.
 -   The organization has sufficient legal or reputational incentives to dissuade it from making malicious changes.
 
 Ultimately this is a judgement call with no straightforward answer.
@@ -110,9 +110,10 @@ A best practice is to require approval of any changes via a change management to
 
 *Example:* Adversary creates a pull request using a secondary account and approves it using their primary account.
 
-Solution: The producer must require all accounts with 'write' and 'approval' permissions to be strongly authenticated and ensure they map to unique persons.
-A common vector for this attack is to take over a robot account with the permission to contribute code.
-Control of the robot and an actors own legitimate account is enough to exploit this vulnerability.
+Solution: The producer must track all actors who have both explicit review permissions and the independent ability to control a privileged bot.
+A common vector for this attack is to influence a robot account with the permission to review or contribute code.
+Control of the robot account and an actor's own personal account is enough to exploit this vulnerability.
+A common solution to this flow is to deny bot accounts from contributing or reviewing code, or to require more human reviews in those cases.
 
 </details>
 <details><summary>Use a robot account to submit change</summary>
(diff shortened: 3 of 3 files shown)


docs/spec/draft/threats.md
```diff
@@ -76,25 +76,20 @@ be nice to resolve. For example, compromised developer credentials - is that (A)
 or (B)?
 -->
 
-<details><summary>Software producer intentionally submits bad code</summary>
+<details><summary>The organization changes the intent of the repository and creates a malicious revision</summary>
```
TomHennen, line 79: What about this...

```suggestion
<details><summary>The organization intentionally creates a malicious revision</summary>
```

Rationale: the "changes the intent of the repository" may be a distraction.  Folks might think "this thing still does compression so the major intent of the repo is the same"

TomHennen (commented):
Just the one comment, otherwise this looks great!

docs/spec/draft/threats.md
```diff
@@ -76,25 +76,20 @@ be nice to resolve. For example, compromised developer credentials - is that (A)
 or (B)?
 -->
 
-<details><summary>Software producer intentionally submits bad code</summary>
+<details><summary>The organization intentionally creates a malicious revision</summary>
```
adityasaky, line 79: Is there a reason to switch the terminology to "organization" from "producer"? Organization seems narrower to me, and the figure also uses producer.

zachariahcox: that's a good question. 
My idea was to align with the "entity that determines the intent of the source," because "sudden change of intent" is really the attack vector here.

zachariahcox: I started a conversation in slack. Copied here: 

https://openssf.slack.com/archives/C03NUSAPKC6/p1732542950838849?thread_ts=1732542830.795929&cid=C03NUSAPKC6

The threats diagram uses the term "producer" in the source threats section, but is not one of the [defined terms for the proposed source track](https://github.com/slsa-framework/slsa/blob/4a5c7e682c304f73a9d2501d3ec0c819648c346f/docs/spec/draft/source-requirements.md#definitions)

At least to me, "producer" doesn't capture what we mean here.
"producer" seems like it must be akin to "organization" in our source track terminology, but there are other actor-based threats against the source that do not require a compromised organization (IE: "contributors" and "reviewers" also "produce" source.) 

The source track needs to talk about several kinds of actors with roles [described here](https://github.com/slsa-framework/slsa/blob/4a5c7e682c304f73a9d2501d3ec0c819648c346f/docs/spec/draft/source-requirements.md#source-roles). The only entity that could be described as the sole "producer" would be the org, which defines the intent of the software and the rules to enforce it.

arewm: A producer can be an individual or an organization. I think it makes sense to keep it general.

TomHennen (approved):

adityasaky (commented):
this isn't a complete review because I wasn't sure if more changes are coming, so I'm submitting this for visibility for now.

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,59 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
-be nice to resolve. For example, compromised developer credentials - is that (A)
-or (B)?
--->
+*Threat:* An producer intentionally creates a malicious revision with the intent of harming their consumers.
 
-<details><summary>Software producer intentionally submits bad code</summary>
+*Mitigation:*
+This kind of attack cannot be directly mitigated through SLSA controls.
+Tools like the [OSSF Scorecard](https://github.com/ossf/scorecard) can help to quantify the risk of consuming artifacts from specific organizations, but do not fully remove it.
+Trustworthiness scales with transparency, and consumers SHOULD push their vendors to follow transparency best-practices.
```
trishankatdatadog, line 72: "Transparency" is unfortunately an overloaded word in our world. WDYM specifically in this context?

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,59 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
 #### (B1) Submit change without review
 
 <details><summary>Directly submit without review</summary>
 
-*Threat:* Submit bad code to the source repository without another person
-reviewing.
+*Threat:* Submit code to the source repository without another person reviewing.
 
-*Mitigation:* Source repository requires two-person approval for all changes.
+*Mitigation:* The producer can require pre-approval for all changes.
```
trishankatdatadog, line 91: This new language is actually more confusing than the last one. What's a "pre-approval"?

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,59 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
+<details><summary>Single actor controls multiple accounts</summary>
 
-*Threat:* Propose a change using one account and then approve it using another
-account.
+*Threat:* An actor is able to control multiple account and effectively approve their own code changes.
 
-*Mitigation:* Source repository requires approval from two different, trusted
-persons. If the proposer is trusted, only one approval is needed; otherwise two
-approvals are needed. The software producer maps accounts to trusted persons.
+*Mitigation:* The produce must ensure that no actor is able to control or influence multiple accounts with review privileges.
```
trishankatdatadog, line 102:
```suggestion
*Mitigation:* The producer must ensure that no actor is able to control or influence multiple accounts with review privileges.
```

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,59 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
 
-*Example:* Adversary creates a pull request using a secondary account and then
-approves and merges the pull request using their primary account. Solution:
-Configure branch protection to require two approvals and ensure that all
-repository contributors and owners map to unique persons.
+*Example:* Adversary creates a pull request using a secondary account and approves it using their primary account.
+
+Solution: The producer must require strongly authenticated user accounts and ensure that all accounts map to unique persons.
+A common vector for this attack is to take over a robot account with the permission to contribute code.
+Control of the robot and an actors own legitimate account is enough to exploit this vulnerability.
```
trishankatdatadog, line 108: Confusing. The solution goes into another problem, but does not prescribe a solution for that. Best to separate the latter out IMHO.

docs/spec/draft/threats.md
```diff
@@ -142,72 +117,72 @@ two-person review.
...
+<details><summary>Abuse of rule exceptions</summary>
+
+*Threat:* Rule exceptions provide vector for abuse
+
+*Mitigation:* Remove rule exceptions.
+
+*Example:* The intent of a producer is to require two-person review on "all changes except for documentation changes," defined as those only modifying `.md` files.
+Adversary submits a malicious executable named `evil.md` and a code review is not required due to the exception.
+Technically, the intent of the producer was followed and the produced malicious revision meets all defined policies.
+Solution: Do not allow such exceptions.
```
trishankatdatadog, line 134: Who should not allow such exceptions? The SCS, the org, or someone else?

docs/spec/draft/threats.md
```diff
@@ -142,72 +117,72 @@ two-person review.
...
-<!--
-> TODO This solution may not be practical in all circumstances. Are there any
-> valid exceptions? If so, how do we ensure they cannot be exploited?
--->
+*Example 1:* A GitHub repository-level admin pushes a change without review, even though GitHub branch protection is enabled.
+Solution: The producer can modify the rule to disallow bypass by administrators, or move the rule to an organization-level ruleset.
 
-</details>
+*Example 2:* GitHub repository-level admin removes a branch requirement, pushes their change, then re-enables the requirement to cover their tracks.
+Solution: The producer can use higher-permission-level rulesets (such as organization-level) to prevent repository-level tampering.
```
trishankatdatadog, line 150: Have we defined "rulesets" in our terminology?

docs/spec/draft/threats.md
zachariahcox: oof, that's a good callout. 

I propose that I mean: 
* open source
* open build
* open policies
* published "definitions of correctness."

docs/spec/draft/threats.md
zachariahcox:
```suggestion
*Mitigation:* The producer requires approval of all changes before they are accepted.
```

zachariahcox: is that a clearer? pre-approval does sound like you're applying for a car loan lol

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,59 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
+*Threat:* Submit code to the source repository without another person reviewing.
 
-*Mitigation:* Source repository requires two-person approval for all changes.
+*Mitigation:* The producer can require pre-approval for all changes.
 
-*Example:* Adversary directly pushes a change to a GitHub repo's `main` branch.
-Solution: Configure GitHub's "branch protection" feature to require pull request
-reviews on the `main` branch.
+*Example:* Adversary directly pushes a change to a git repo's `main` branch.
+Solution: The producer can configure branch protection rules on the `main` branch.
```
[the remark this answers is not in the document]
zachariahcox: I think in many ways this section is duplicative of other sections. 

We could just say: 
```
*Threat*: source threats!
*Mitigation*: SLSA source track! <url>
```

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,60 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
+A best practice would be to required pre-approval of any changes via a change management tool (such a as GitHub pull request).
```
zachariahcox, line 96:
```suggestion
A best practice is to require approval of any changes via a change management tool before they are accepted into the source.
```

docs/spec/draft/threats.md
zachariahcox: I think we can just cut it. This is very duplicative of other sections in the source track.

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,60 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
 
-*Example:* Adversary creates a pull request using a secondary account and then
-approves and merges the pull request using their primary account. Solution:
-Configure branch protection to require two approvals and ensure that all
-repository contributors and owners map to unique persons.
+*Example:* Adversary creates a pull request using a secondary account and approves it using their primary account.
+
+Solution: The producer must require strongly authenticated user accounts and ensure that all accounts map to unique persons.
+A common vector for this attack is to take over a robot account with the permission to contribute code.
+Control of the robot and an actors own legitimate account is enough to exploit this vulnerability.
```
zachariahcox, line 109:
```suggestion
Solution: The producer must require strongly authenticated user accounts and ensure that all accounts map to unique persons. It must not be possible for one actor to exploit robot actors that possess review privileges. 
```

docs/spec/draft/threats.md
```diff
@@ -142,72 +118,72 @@ two-person review.
...
+<details><summary>Abuse of rule exceptions</summary>
+
+*Threat:* Rule exceptions provide vector for abuse
+
+*Mitigation:* Remove rule exceptions.
+
+*Example:* The intent of a producer is to require two-person review on "all changes except for documentation changes," defined as those only modifying `.md` files.
+Adversary submits a malicious executable named `evil.md` and a code review is not required due to the exception.
+Technically, the intent of the producer was followed and the produced malicious revision meets all defined policies.
+Solution: Do not allow such exceptions.
```
zachariahcox, line 135:
```suggestion
Solution: The producer adjusts the rules to prohibit such exceptions.
```

docs/spec/draft/threats.md
```diff
@@ -221,43 +197,25 @@ different context.
...
+Intermediate revisions don't count as being reviewed and are not added to the protected context (such as the `main` branch).
 
-*Example:* Adversary sends a pull request containing malicious commit X and
-benign commit Y that undoes X. In the pull request UI, reviewer only reviews and
-approves "changes from all commits", which is a delta from HEAD to Y; they don't
-see X. Adversary then builds from the malicious revision X. Solution: Policy
-does not accept this because the version X is not considered reviewed.
+*Example:* Adversary sends a pull request containing malicious commit X and benign commit Y that undoes X.
+The produced diff of X + Y contains zero lines of changed code and the reviewer may not notice that X is malicious unless they review each commit in the request.
+If X is allowed to become reachable from the protected branch, the content may become available in secured contexts, such as developer machines and vulnerable to exploits.
```
zachariahcox, line 216:
```suggestion
If X is allowed to become reachable from the protected branch, the content may become available in secured environments such as developer machines.
```

docs/spec/draft/threats.md
```diff
@@ -142,72 +118,72 @@ two-person review.
...
-<!--
-> TODO This solution may not be practical in all circumstances. Are there any
-> valid exceptions? If so, how do we ensure they cannot be exploited?
--->
+*Example 1:* A GitHub repository-level admin pushes a change without review, even though GitHub branch protection is enabled.
+Solution: The producer can modify the rule to disallow bypass by administrators, or move the rule to an organization-level ruleset.
 
-</details>
+*Example 2:* GitHub repository-level admin removes a branch requirement, pushes their change, then re-enables the requirement to cover their tracks.
+Solution: The producer can use higher-permission-level rulesets (such as organization-level) to prevent repository-level tampering.
```
zachariahcox, line 151:
```suggestion
Solution: The producer can configure higher-permission-level rules (such as organization-level GitHub Rulesets) to prevent repository-level tampering.
```

docs/spec/draft/threats.md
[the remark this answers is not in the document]
TomHennen: Note that this page discusses lots of threats to the build and doesn't just defer to the SLSA Build Track.

The general purpose of this page, AIUI, was to centralize all the various threats in one place and direct folks elsewhere as needed.

TomHennen:
If we merge this before 1.1 goes out will that make it harder to remove this from the 1.1 release?  Have we decided we want to include this in the 1.1 release (fine with me at this point)?

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,59 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
-*Threat:* Software producer intentionally submits "bad" code, following all
-proper processes.
-
-*Mitigation:* **TODO**
-
-*Example:* A popular extension author sells the rights to a new owner, who then
-modifies the code to secretly mine cryptocurrency at the users' expense. SLSA
-does not protect against this, though if the extension were open source, regular
-auditing may discourage this from happening.
+*Example:* The [libxz attack](https://access.redhat.com/security/cve/CVE-2024-3094) and the intentional sabotage of [colors.js and faker.js](https://fossa.com/blog/npm-packages-colors-faker-corrupted/) are clear examples of an organization with an otherwise good reputation intentionally producing malicious code.
```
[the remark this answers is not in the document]
zachariahcox: @TomHennen I think we missed the suggestion here ☝️

docs/spec/draft/threats.md
zachariahcox: @TomHennen I think SLSA tooling gives you a way to unilaterally improve trust when the source and builds are open source. This is possible independently of whether or not the producer has something to lose.

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,60 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
-Threats in this category likely *cannot* be mitigated through controls placed
-during the authoring/reviewing process, in contrast with (B).
+<details><summary>Software producer intentionally creates a malicious revision of the source</summary>
 
-<!--
-**TODO:** The difference between (A) and (B) is still a bit fuzzy, which would
-be nice to resolve. For example, compromised developer credentials - is that (A)
-or (B)?
--->
+*Threat:* A producer intentionally creates a malicious revision with the intent of harming their consumers.
```
zachariahcox, line 67: @TomHennen what if we update this to cover both: 

```suggestion
*Threat:* A producer intentionally creates a malicious revision (or a VSA issuer intentionally creates a malicious attestation) with the intent of harming their consumers.
```

The mitigation for both is a clearer -- you cannot.

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,60 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
 
-<details><summary>Software producer intentionally submits bad code</summary>
+*Mitigation:*
+This kind of attack cannot be directly mitigated through SLSA controls.
+Trustworthiness scales with transparency, and consumers SHOULD push their vendors to follow transparency best-practices.
+When transparency is not possible, consumers may choose not to consume the artifact, or may require additional evidence of correctness from a trusted third-party.
+Tools like the [OSSF Scorecard](https://github.com/ossf/scorecard) can help to quantify the risk of consuming artifacts from specific producers, but do not fully remove it.
+For example, a consumer may choose to only consume source artifacts from repositories that have a high score on the OSSF Scorecard.
```
zachariahcox, line 74:
```suggestion
Ultimately, consumers must decide which producers are trusted to produce artifacts and which issuers are trusted to produce VSAs. 
```

We could redirect from here over to "verifying platforms".

docs/spec/draft/threats.md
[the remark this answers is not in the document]
zachariahcox:
```suggestion
*Example:* A producer with an otherwise good reputation decides suddenly to produce a malicious artifact with the intent to harm their consumers. 
```
This is simpler. wdyt @TomHennen ?

docs/spec/draft/threats.md
zachariahcox: I left two suggestions. wdyt @trishankatdatadog @TomHennen ?

docs/spec/draft/threats.md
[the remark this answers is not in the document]
zachariahcox: Potentially then most of these examples should link to relevant sections of the relevant tracks. Maybe we'll take it as a todo.

docs/spec/draft/threats-overview.md
```diff
@@ -23,10 +23,9 @@ availability. Integrity means protection against tampering or unauthorized
 modification at any stage of the software lifecycle. Within SLSA, we divide
 integrity into source integrity vs build integrity.
 
-**Source integrity:** Ensure that all changes to the source code reflect the
-intent of the software producer. Intent of an organization is difficult to
-define, so SLSA is expected to approximate this as approval from two authorized
-representatives.
+**Source integrity:** Ensure that source revisions contain only changes submitted by
+authorized contributors according to the process defined by the software producer and
+that source revisions are not modified as they pass between development stages.
```
zachariahcox, line 28: this wording aligns better with the build integrity section below.

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,60 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
-does not protect against this, though if the extension were open source, regular
-auditing may discourage this from happening.
+*Example:* The [libxz attack](https://access.redhat.com/security/cve/CVE-2024-3094) and the intentional sabotage of [colors.js and faker.js](https://fossa.com/blog/npm-packages-colors-faker-corrupted/) are clear examples of an organization with an otherwise good reputation intentionally producing malicious code.
 
 </details>
 
-<!--
-**TODO:** More producer threats? Perhaps the attack to xz where a malicious
-contributor gained enhanced privileges through social engineering?
--->
```
zachariahcox, line 96: we could add another section here like 
```
*threat*: A member of the producer's organization acts against the intent of the producer.
```
We could say this is when "the producer UNintentially ships bad code."

docs/spec/draft/threats.md
```diff
@@ -221,43 +197,25 @@ different context.
...
-*Threat:* Request review for a series of two commits, X and Y, where X is bad
-and Y is good. Reviewer thinks they are approving only the final Y state whereas
-they are also implicitly approving X.
+*Threat:* Request review for a series of two commits, X and Y, where X is bad and Y is good.
+Reviewer thinks they are approving only the final Y state but they are also implicitly approving X.
 
-*Mitigation:* Only the version that is actually reviewed is the one that is
-approved. Any intermediate revisions don't count as being reviewed.
+*Mitigation:* The producer declares that only the final delta is considered approved.
+Intermediate revisions don't count as being reviewed and are not added to the protected context (such as the `main` branch).
```
zachariahcox, line 212:
```suggestion
In this configuration, intermediate revisions are not considered to be approved and are not added to the protected context (e.g. the `main` branch).
With git version control systems this is called a "squash" merge strategy.
```

docs/spec/draft/threats.md
```diff
@@ -267,8 +225,10 @@ does not accept this because the version X is not considered reviewed.
 
 *Threat:* Two trusted persons collude to author and approve a bad change.
 
-*Mitigation:* This threat is not currently addressed by SLSA. We use "two
-trusted persons" as a proxy for "intent of the software producer".
```
zachariahcox, line 271: I do not think we should use "two trusted persons" as a proxy for producer intent. We should probably say "the documented change process for the source" if we must, but I prefer just "the intent".

zachariahcox:
> If we merge this before 1.1 goes out will that make it harder to remove this from the 1.1 release? Have we decided we want to include this in the 1.1 release (fine with me at this point)?
> 
> @lehors thoughts?

@TomHennen! Potentially I am off base -- is 1.1 going to be different from the /draft folder? 
I think this kind of change should go out with the most recent version of the source track, whichever that is 👍 
It was not my intention to add items to the 1.1 release if that is different than the drafts folder.

lehors:
> > If we merge this before 1.1 goes out will that make it harder to remove this from the 1.1 release? Have we decided we want to include this in the 1.1 release (fine with me at this point)?
> > @lehors thoughts?
> 
> @TomHennen! Potentially I am off base -- is 1.1 going to be different from the /draft folder? I think this kind of change should go out with the most recent version of the source track, whichever that is 👍 It was not my intention to add items to the 1.1 release if that is different than the drafts folder. In the meantime I'll update the title to talk about drafts spec instead of 1.1

I'm glad @TomHennen remembers that we need to get 1.1 finalized. I was going to bring that up myself looking at this PR. Unfortunately, we don't have a very clean way of working on several versions in parallel.

The plan was to only work in the draft folder and when all 1.1 issues have been addressed to create a new folder for 1.1 selecting (manually) from the draft what is relevant. Of course this is really only practical if the differences are fairly coarse (like whole files to be left out). If the content of some files start changing in ways that are only relevant for the next version this operation becomes quite impractical.

If we want to go ahead and merge this, I will need to create the 1.1 folder before, and from then on every change made for 1.1 will have to be made in both folders (1.1 and draft). Not ideal but doable.

TomHennen:
> > > If we merge this before 1.1 goes out will that make it harder to remove this from the 1.1 release? Have we decided we want to include this in the 1.1 release (fine with me at this point)?
> > > @lehors thoughts?
> > 
> > 
> > @TomHennen! Potentially I am off base -- is 1.1 going to be different from the /draft folder? I think this kind of change should go out with the most recent version of the source track, whichever that is 👍 It was not my intention to add items to the 1.1 release if that is different than the drafts folder. In the meantime I'll update the title to talk about drafts spec instead of 1.1
> 
> I'm glad @TomHennen remembers that we need to get 1.1 finalized. I was going to bring that up myself looking at this PR. Unfortunately, we don't have a very clean way of working on several versions in parallel.
> 
> The plan was to only work in the draft folder and when all 1.1 issues have been addressed to create a new folder for 1.1 selecting (manually) from the draft what is relevant. Of course this is really only practical if the differences are fairly coarse (like whole files to be left out). If the content of some files start changing in ways that are only relevant for the next version this operation becomes quite impractical.
> 
> If we want to go ahead and merge this, I will need to create the 1.1 folder before, and from then on every change made for 1.1 will have to be made in both folders (1.1 and draft). Not ideal but doable.

I wonder if at this point we can just wait and include the source track (and the build track if it's ready).  We've waited this long, and there don't seem to be that many outstanding issues elsewhere.  By trying to separate them now we'd probably just cause ourselves more work for less gain. WDYT?

lehors:
> I wonder if at this point we can just wait and include the source track (and the build track if it's ready). We've waited this long, and there don't seem to be that many outstanding issues elsewhere. By trying to separate them now we'd probably just cause ourselves more work for less gain. WDYT?

You mean to skip 1.1? That would solve the problem of course. It's true that the more time goes by the less relevant it is.

TomHennen:
> > I wonder if at this point we can just wait and include the source track (and the build track if it's ready). We've waited this long, and there don't seem to be that many outstanding issues elsewhere. By trying to separate them now we'd probably just cause ourselves more work for less gain. WDYT?
> 
> You mean to skip 1.1? That would solve the problem of course. It's true that the more time goes by the less relevant it is.

Not so much "skip" as _include_ the source track in 1.1.  New tracks [are supposed to go](https://github.com/slsa-framework/slsa/blob/main/docs/spec-stages.md#:~:text=MINOR%20version%20when,Versioning%20Rules.) in minor releases

TomHennen:
We had a long, _long_, chat about tracks, threats pages, etc... at yesterday's meeting.  Do we still like this PR or do we want to simplify it some?  I have no opinion at this time.

docs/spec/draft/threats.md
trishankatdatadog: Sorry for late reply. Why do consumers even need to know which producers are trusted to produce artifacts (source code, I imagine, in this case)? Couldn't they just get away with the Source Track VSA from the issuer, and call it a day so long as its subjects/outputs (e.g., source code) matches the inputs to the next step (e.g., build)?

docs/spec/draft/threats.md
trishankatdatadog: Left a comment down there...

zachariahcox: good ideas. I'll incorporate these.

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,65 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
 
 A source integrity threat is a potential for an adversary to introduce a change
 to the source code that does not reflect the intent of the software producer.
-This includes the threat of an authorized individual introducing an unauthorized
-change---in other words, an insider threat.
+This includes modification of the source data at rest as well as insider threats,
+when an authorized individual introduces an unauthorized change.
 
-SLSA v1.0 does not address source threats, but we anticipate doing so in a
-[future version](current-activities.md#source-track). In the meantime, the
```
zachariahcox, line 60: I think this was a broken link? I have added a section to future-directions to target.

docs/spec/draft/future-directions.md
```diff
@@ -54,3 +54,14 @@ the basis for a future Build Platform Operations track:
 </section>
 
 [draft version (v0.1)]: ../v0.1/requirements.md
+
+
+## Source Track
```
zachariahcox, line 59: added the "objective" section from source-requirements.md, updated the verbs to be future tense. 
this is to be the target of future-direction links.

docs/spec/draft/threats.md
```diff
@@ -220,44 +201,27 @@ different context.
 
 *Example:* MyPackage's source repository requires two-person review. Adversary
 forks the repo, submits a change in the fork with review from a colluding
-colleague (who is not trusted by MyPackage), then merges the change back into
-the upstream repo. Solution: The merge should still require review, even though
-the fork was reviewed.
+colleague (who is not trusted by MyPackage), then proposes the change to
+the upstream repo.
+Solution: The proposed change still requires two-person review in the upstream context even though it received two-person review in another context.
```
zachariahcox, line 206: I updated the words to be a little more precise. 
In general, I'd expect reviews to be transitive for some situations. but yeah, if your org thinks this is a problem, then this is how you'd fix it.

zachariahcox:
Howdy friends! I have updated this PR to reference only 1.1 legal concepts. 
I also incorporated the most recent feedback from reviewers (thank you!)

There were a lot of structural changes to the repo and I ultimately rebased these smallish patches. 
Apologies if your comments fell off! I think I got to all of them but please let me know.

cc: @TomHennen @adityasaky @lehors @trishankatdatadog @arewm

docs/spec/draft/threats.md
TomHennen: agreed

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,66 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
-Threats in this category likely *cannot* be mitigated through controls placed
-during the authoring/reviewing process, in contrast with (B).
+<details><summary>Software producer intentionally creates a malicious revision of the source</summary>
 
-<!--
-**TODO:** The difference between (A) and (B) is still a bit fuzzy, which would
-be nice to resolve. For example, compromised developer credentials - is that (A)
-or (B)?
--->
+*Threat:* A producer intentionally creates a malicious revision (or a VSA issuer intentionally creates a malicious attestation) with the intent of harming their consumers.
```
TomHennen, line 70:
```suggestion
*Threat:* A producer intentionally creates a malicious revision with the intent of harming their consumers.
```

I think the VSA Issuer note should probably go in https://slsa.dev/spec/draft/threats#verification-threats (below) as it applies everywhere, and the 'producer' of the VSA isn't aligned with the producer being discussed here: https://slsa.dev/spec/draft/terminology#:~:text=Examples-,Producer,-A%20party%20who

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,66 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
-this with (A), where such controls are likely ineffective.
+Threats in this category can be mitigated by following source control management best practices.
 
 #### (B1) Submit change without review
 
 <details><summary>Directly submit without review</summary>
 
-*Threat:* Submit bad code to the source repository without another person
-reviewing.
+*Threat:* Submit code to the source repository without another person reviewing.
```
TomHennen, line 96: I think this leaves out that the code being submitted is malicious.

the vast majority of unreviewed changes aren't malicious.

Perhaps this can be rephrased as 

```suggestion
*Threat:* Malicious code submitted to the source repository.
```
And then the (partial) mitigation is "The producer requires approval of all changes before they are accepted".

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,66 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
+*Mitigation:* The producer must ensure that no actor is able to control or influence multiple accounts with review privileges.
 
-*Example:* Adversary creates a pull request using a secondary account and then
-approves and merges the pull request using their primary account. Solution:
-Configure branch protection to require two approvals and ensure that all
-repository contributors and owners map to unique persons.
+*Example:* Adversary creates a pull request using a secondary account and approves it using their primary account.
+
+Solution: The producer must require strongly authenticated user accounts and ensure that all accounts map to unique persons.
+A common vector for this attack is to take over a robot account with the permission to contribute code.
```
TomHennen, line 114: optional: Robot accounts are discussed below so this seems somewhat redundant.  Suggest removing.

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,66 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
-approvals are needed. The software producer maps accounts to trusted persons.
+*Mitigation:* The producer must ensure that no actor is able to control or influence multiple accounts with review privileges.
 
-*Example:* Adversary creates a pull request using a secondary account and then
-approves and merges the pull request using their primary account. Solution:
-Configure branch protection to require two approvals and ensure that all
-repository contributors and owners map to unique persons.
+*Example:* Adversary creates a pull request using a secondary account and approves it using their primary account.
+
+Solution: The producer must require strongly authenticated user accounts and ensure that all accounts map to unique persons.
```
TomHennen, line 113: Trying to scope this down some for folks.  There's no way an org can ensure all GitHub accounts map to unique persons.  They really just need it to be the case for people that can approve code.

```suggestion
Solution: The producer must require all accounts with 'write' and 'approval' permissions to be strongly authenticated and ensure they map to unique persons.
```

docs/spec/draft/threats.md
```diff
@@ -142,72 +124,72 @@ two-person review.
 robots.
 
 *Example:* A file within the source repository is automatically generated by a
-robot, which is allowed to submit without review. Adversary compromises the
-robot and submits a malicious change without review. Solution: Require human
-review for these changes.
+robot, which is allowed to submit without review.
+Adversary compromises the robot and submits a malicious change.
+Solution: Require review for such changes.
```
TomHennen, line 129: Optional: 

```suggestion
Solution: Require review for such changes or ensure all changes relevant to the robot's behavior require review.
```

E.g. imagine if the Robot is a GHA that generates code.  If the workflow that runs that code is reviewed, and the generation tool is reviewed, then it could be safe?

docs/spec/draft/threats.md
```diff
@@ -142,72 +124,72 @@ two-person review.
...
--->
+</details>
+<details><summary>Abuse of rule exceptions</summary>
+
+*Threat:* Rule exceptions provide vector for abuse
+
+*Mitigation:* Remove rule exceptions.
+
+*Example:* The intent of a producer is to require two-person review on "all changes except for documentation changes," defined as those only modifying `.md` files.
+Adversary submits a malicious executable named `evil.md` and a code review is not required due to the exception.
```
TomHennen, line 139: On first reading it wasn't clear to me what the actual problem was here.

```suggestion
Adversary submits a malicious executable masquerading as a documentation file, `help.md`. A code review is not required due to the exception, and a user (or another workflow) is then induced to _execute_ `help.md` and are compromised.
```

docs/spec/draft/threats-overview.md
TomHennen: I think this is 'fine', but we do lose the introduction of the concept of 'intent' (which we reference later) which isn't quite the same as 'process'.  I'd argue the process is one of the ways the producer tries to ensure their intent is tracked, but it's not really what the end user cares about.

I'd also probably argue that the build process is somewhat different and they don't _need_ to mirror each other.  To some extent the source is the closest approximation of the org's intent, and then we just rely on the build system to mechanically transform that source faithfully.

TomHennen (changes requested):
Looking good!

docs/spec/draft/threats.md
zachariahcox: That's fine with me too. 
Generally, it's hard to say whether a commit is malicious, and "exploitable" ends up being just as bad.

docs/spec/draft/threats.md
zachariahcox: That's probably fine. I _think_ these are different threat models, but not sure if the difference is worth calling out, or if these two blurbs are doing so. 
This one is about gaining access to two accounts with `write` and the robots one is about gaining access to an account with `bypass`. 

The mitigation for the first is _hard_ to execute and protect against. You'd have to somehow prevent account takeovers. 

The mitigation for the robots one is easier to execute (just run some kind of automation with a poisoned payload, usually) and protect against-- just do not allow any account to bypass code review requirements.

docs/spec/draft/threats.md
zachariahcox: (commented on the other thread)

docs/spec/draft/threats.md
TomHennen: Right.  I guess I'll leave it to you?

docs/spec/draft/threats.md
TomHennen: I suppose I think there are legitimate use cases for bypassing code review by robots, and I think it is possible to do safely?

I also don't think we have to address that here and now, especially if we don't have a fully fleshed out framework for doing so.

zachariahcox: I think eventually we'd need to say something like:
## autobypasses
for some security postures, this practice increases risk because due to threat of account takeover.
For others it's way to reduce risk because security patches are applied faster (or whatever the bot does)

zachariahcox:
I'm going to take another crack at the source threats overview. I mostly left it alone today from where it was in december.

docs/spec/draft/threats.md
```diff
@@ -110,7 +110,7 @@ A best practice is to require approval of any changes via a change management to
 
 *Example:* Adversary creates a pull request using a secondary account and approves it using their primary account.
 
-Solution: The producer must require strongly authenticated user accounts and ensure that all accounts map to unique persons.
+Solution: The producer must require all accounts with 'write' and 'approval' permissions to be strongly authenticated and ensure they map to unique persons.
```
adityasaky, line 113:
```suggestion
Solution: The producer must require all users trusted to write to the repository and approve changes to be strongly authenticated and ensure they map to unique persons.
```

I took a stab at generalizing this, WDYT?

TomHennen: To be honest, I think 'credentials' obscures the meaning.

'credentials' can be interpreted as a specific thing (e.g. password, cryptographic key, badge, etc).  I have _many_ of those things, and I use multiple of them when interacting with systems like GitHub.  Further those systems rarely restrict use based on the way a user authenticated.

adityasaky: That's a good point. Maybe:

```
The producer must require anyone trusted to write to the repository and approve changes to be strongly authenticated and ensure the credentials they use to authenticate themselves map to unique persons.
```

Perhaps an example as well after this?

TomHennen: 1. I think we need "write to the repository _or_ approve changes"
2. I think the use of credentials later in the sentence is still misplaced.

The ACLs systems implement are still based on accounts (users).  User auth to those accounts with their credentials, but the way you ensure they map to a single person is at the account level (since that's where the ACLs are).

Now I'm sure that in some systems account == credential, but those systems may have an even _harder_ (impossible) time enforcing this mapping.  What we really want is that no unique person can appear to be two or more people.

adityasaky: > I think we need "write to the repository or approve changes"

you're right.

> The ACLs systems implement are still based on accounts (users). User auth to those accounts with their credentials, but the way you ensure they map to a single person is at the account level (since that's where the ACLs are).

I agree with this where the point of enforcement is a hosted platform that is part of the SCS. I worry this doesn't apply when we're considering other types of systems such as gittuf, that is also listed in the source spec. I do see your point about the word "credentials" not applying well, though.

TomHennen: I guess when considering a system like gittuf the goal is still that "No unique person is able to take actions that would otherwise require two or more people."

I don't think talking about the credential actually helps in that case.  As written it's still possible to have two credentials that map to the same person (and in gittuf that probably has to happen in case folks lose a key, are using multiple hardware backed keys, etc?).  Alternatively the advise to implementors is "don't allow a unique person to have more than one credential in the ACL"

TomHennen: Oh, but also!  This is _not_ the source requirements doc, this is just _an example_ of a solution for how this can be addressed.

In the future this will link the source reqs doc.

(maybe that needs to be more clear)

adityasaky: >Oh, but also! This is not the source requirements doc, this is just an example of a solution for how this can be addressed.

Oh wow, I'd missed that when I paged things in for this PR!

> I guess when considering a system like gittuf the goal is still that "No unique person is able to take actions that would otherwise require two or more people."
>
> I don't think talking about the credential actually helps in that case. As written it's still possible to have two credentials that map to the same person (and in gittuf that probably has to happen in case folks lose a key, are using multiple hardware backed keys, etc?). Alternatively the advise to implementors is "don't allow a unique person to have more than one credential in the ACL"

You're right in that "credentials" doesn't necessarily address it for the gittuf case, which was also my concern for "account". But, given this isn't actually part of the source track, I think we can close this thread. :)

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,66 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
-approvals are needed. The software producer maps accounts to trusted persons.
+*Mitigation:* The producer must ensure that no actor is able to control or influence multiple accounts with review privileges.
 
-*Example:* Adversary creates a pull request using a secondary account and then
-approves and merges the pull request using their primary account. Solution:
-Configure branch protection to require two approvals and ensure that all
-repository contributors and owners map to unique persons.
+*Example:* Adversary creates a pull request using a secondary account and approves it using their primary account.
+
+Solution: The producer must require all accounts with 'write' and 'approval' permissions to be strongly authenticated and ensure they map to unique persons.
```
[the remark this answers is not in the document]
adityasaky: If two accounts under consideration must map to unique persons, don't we have also have the case that each person is mapped to at most one account?

That said, I think it is better to put the emphasis on the persons involved being distinct rather than the accounts, it would partially address the concern I had in the other thread I think.

adityasaky (approved):

docs/spec/draft/threats.md
```diff
@@ -220,44 +204,27 @@ different context.
...
-and Y is good. Reviewer thinks they are approving only the final Y state whereas
-they are also implicitly approving X.
+*Threat:* Request review for a series of two commits, X and Y, where X is bad and Y is good.
+Reviewer thinks they are approving only the final Y state but they are also implicitly approving X.
 
-*Mitigation:* Only the version that is actually reviewed is the one that is
-approved. Any intermediate revisions don't count as being reviewed.
+*Mitigation:* The producer declares that only the final delta is considered approved.
+In this configuration, intermediate revisions are not considered to be approved and are not added to the protected context (e.g. the `main` branch).
+With git version control systems this is called a "squash" merge strategy.
```
arewm, line 221: While it is likely harder to detect/enforce, another reasonable mitigation is just to review all commits. There could even be automation which flags content from earlier commits that have been removed/overwritten in order to try to detect this type of attack.

docs/spec/draft/threats.md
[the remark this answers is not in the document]
TomHennen: I think the trick is that it should still be possible for anonymous folks to send PRs.  _They_ don't need to be unique.  But you'd want to unique and trusted people to be involved.

I also wonder how much of this nuance is actually needed.  These are just example solutions, they're not really meant to be completely bulletproof (I don't think?).

Another alternative is we just punt and say "SLSA doesn't address this yet".  But that seems annoying. :p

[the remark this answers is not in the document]
arewm: If an anonymous person sends a PR then this can be protected against by _requiring_ 2-party reviews. I don't know how to word that though.

You are probably right that this much nuance isn't needed here. This might even be addressing a separate threat?

docs/spec/draft/future-directions.md
```diff
@@ -54,3 +54,13 @@ the basis for a future Build Platform Operations track:
 </section>
 
 [draft version (v0.1)]: ../v0.1/requirements.md
+
+## Source Track
+
+The SLSA Source track will describe increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc).
+
+The Source track will be scoped to revisions of a single repository that is controlled by an organization.
+The organization determines the intent of the software in the repository, what Source level should apply to the repository, and administers technical controls to enforce that level.
```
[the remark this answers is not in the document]
arewm: It is okay for it to not be _precise_ here, but I don't think it should be confusing. I interpret this blurb as an indication of the identities that might be involved in the source track. If that is the case, we should clarify that it relates to the organization hosting the version control system?

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,66 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
 
-<details><summary>Software producer intentionally submits bad code</summary>
+*Mitigation:*
+This kind of attack cannot be directly mitigated through SLSA controls.
+Consumers must establish some basis to trust the organizations from which they consume software.
+That basis may be:
 
-*Threat:* Software producer intentionally submits "bad" code, following all
-proper processes.
+-   The code is open source and has a sufficiently large user-base that malicious changes are likely to be detected.
```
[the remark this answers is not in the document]
zachariahcox: I guess both to some extent. 
Expert code review may be "required" to approve something 
into the main branch, but prs are often reviewed by non-authorized (but maybe informed) reviewers too. 
Those extra eyes help.

Maybe this should be a time-to-mitigation claim. 
`With more sets of eyes, the length of time that a vulnerability exists in the repo is reduced.`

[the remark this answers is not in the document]
zachariahcox: or maybe 

```The repo is open source with an active user-base. High numbers of engaged users increase the likelihood that bad code is detected during code review and reduce the time-to-detection when bad code makes it in.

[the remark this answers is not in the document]
Comment (arewm) [reply docs/spec/draft/threats.md:77]: I think commenting about a reduced time to detection feels right. We can also make an additional comment about an "expert" review as a potential additional mitigation.

How about 
```suggestion
-   The code is open source and has a large engaged user-base. Time to detect and remediate malicious changes to the code base are reduced.
-   The organization requires that all code is vetted by expert reviewers before revisions are accepted.
```

docs/spec/draft/threats.md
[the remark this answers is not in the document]
zachariahcox: > An account that creates a revision request doesn't necessarily need to have write permissions. 

Not to get tooooooo in the weeds, but in most implementations an open requires write permissions to created the proposed merge commits for CI.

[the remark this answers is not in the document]
zachariahcox: I'm not sure how to tighten this one up. 
I feel like the threat, mitigation and example are all good, but the solution needs work.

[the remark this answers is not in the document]
zachariahcox: I could just adjust it -- "the producer much prevent this kind of thing from happening somehow"

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,66 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
+*Mitigation:* The producer must ensure that no actor is able to control or influence multiple accounts with review privileges.
 
-*Example:* Adversary creates a pull request using a secondary account and then
-approves and merges the pull request using their primary account. Solution:
-Configure branch protection to require two approvals and ensure that all
-repository contributors and owners map to unique persons.
+*Example:* Adversary creates a pull request using a secondary account and approves it using their primary account.
+
+Solution: The producer must require all accounts with 'write' and 'approval' permissions to be strongly authenticated and ensure they map to unique persons.
+A common vector for this attack is to take over a robot account with the permission to contribute code.
```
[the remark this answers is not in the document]
zachariahcox: yeah, I don't like the word mapping here. let me try to remove it.

docs/spec/draft/threats.md
zachariahcox: that's technically possible, but I would say that getting that right requires multiple squash merge PRs -- you cannot really do it in one shot.

docs/spec/draft/threats.md
```diff
@@ -220,44 +205,25 @@ different context.
...
-increases the difficulty of compromising accounts.
-
-*Example:* Trusted person uses a weak password on GitHub. Adversary guesses the
-weak password, logs in, and pushes changes to a GitHub repo. Solution: Configure
-GitHub organization to requires 2FA for all trusted persons. This would increase
-the difficulty of using the compromised password to log in to GitHub.
 
 </details>
-<details><summary>Hide bad change behind good one</summary>
+<details><summary>Commit graph attacks</summary>
```
zachariahcox, line 215: @arewm @TomHennen 

I rewrote this section based on y'all's feedback. 

My goal was to remove the wonky git stuff about squashes and focus on reachable revisions following process.

mlieberman85: Oh, so one of the things I've done in a previous life was the attestation is actually an attestation of a particular ordering of the commits. so you are attesting to the change not each individual commit so if you roll back a commit it would no longer be valid since you're just attesting to HEAD. with the assumption you're just attesting to old HEAD -> PR/New HEAD

docs/spec/draft/threats.md
```diff
@@ -53,84 +53,67 @@ broadly adopted in an automated fashion, minimizing the chance of mistakes.
...
 
-<details><summary>Software producer intentionally submits bad code</summary>
+*Mitigation:*
+This kind of attack cannot be directly mitigated through SLSA controls.
+Consumers must establish some basis to trust the organizations from which they consume software.
+That basis may be:
 
-*Threat:* Software producer intentionally submits "bad" code, following all
-proper processes.
+-   The repo is open source with an active user-base. High numbers of engaged users may increase the likelihood that bad code is detected during code review and reduce the time-to-detection when bad code makes it in.
```
arewm, line 77: nit:
```suggestion
-   The repo is open source with an active user-base. High numbers of engaged users may increase the likelihood that bad code is detected during code review and reduce the time-to-detection when bad revisions are accepted.
```

arewm (approved):
Thanks! I left one nit comment outside of the review.

docs/spec/draft/threats.md
arewm: I know some folks that hold this to be a hotly debated and contentious topic. It feels right for it to be required at a higher SLSA level because of the higher protections that it guarantees. Some communities preferences may still prefer to not use this approach preventing the SLSA source level that can be achieved.

zachariahcox:
Ok! thanks everyone for your input and patience on this one!
I'm going to merge it -- and we'll focus on any followups in the next pr 👍

Issue #1184, opened 2024-10-09
TODO: Need mitigation description for "Use a compromised build tool" threat


lehors:

TomHennen:
I think a reasonable mitigation here is something like "Treat build tooling, including OS images, as any other software to be verified prior to use (as described in (G)).  This will allow the build platform to detect any modified binaries."

Once the attested build environments track is finalized we can update this mitigation to reference it specifically.  CC @marcelamelara

Separately, I'm wondering if this is a reason to move the mitigations from (G) (discussed in #1190, #1191) to (I) since 'usage' is more clearly aligned with this risk than with 'distribution channel'.  (Or maybe we can just tweak the mitigation language here to discuss).

marcelamelara:
@TomHennen Definitely happy to help with this as well!

PR #1250, opened 2024-12-03
content: draft: Moved example to a new threat.

### LINKED ISSUE #1223: Clarify 'Tamper with provenance or VSA' threat
```suggestion
*Threat:* Issue an attestation that purposefully misrepresents the subject.
```

_Originally posted by @zachariahcox in https://github.com/slsa-framework/slsa/pull/1191#discussion_r1816709266_

TomHennen:
The example was in the wrong place. It had been under a threat that was meant for attackers that modified signed attestations, yet it discussed replacing one valid attestation with another.

There wasn't an obivious existing threat to put this under so I created a new one.

fixes #1223

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -722,8 +722,7 @@ rejects because provenance is missing.
 
 *Threat:* Replace a package and its VSA with a malicious package and its valid VSA.
 
-*Mitigation*: Verifier checks that the `resourceUri` in the VSA matches the package
-they've requested not just the package they received.
+*Mitigation*: Consumer ensures that the VSA matches the package they've requested (not just the package they received) by following the [verification process](verification_summary#how-to-verify).
 
 *Example:* Adversary uploads a malicious package to `repo/evil-package`,
 getting a valid VSA for `repo/evil-package`. Adversary then replaces
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -718,6 +718,20 @@ on the package repository and deletes existing provenance. Solution: Verifier
 rejects because provenance is missing.
 
 </details>
+<details><summary>Replace package and VSA with another <span>(expectations)</span></summary>
+
+*Threat:* Replace a package and its VSA with a malicious package and its valid VSA.
+
+*Mitigation*: Verifier checks that the `resourceUri` in the VSA matches the package
+they've requested not just the package they received.
+
+*Example:* Adversary uploads a malicious package to `repo/evil-package`,
+getting a valid VSA for `repo/evil-package`. Adversary then replaces
+`repo/my-package` and its VSA with `repo/evil-package` and its VSA.
+Solution: Verifier rejects because the VSA `resourceUri` field lists
+`repo/evil-package` and not the expected `repo/my-package`.
+
+</details>
 <details><summary>Tamper with artifact after upload <span>(Build L1)</span></summary>
 
 *Threat:* Take a benign version of the package, modify it in some way, then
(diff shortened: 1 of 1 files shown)


docs/spec/draft/threats.md
```diff
@@ -717,6 +717,20 @@ corresponding signing key.
 on the package repository and deletes existing provenance. Solution: Verifier
 rejects because provenance is missing.
 
+</details>
+<details><summary>Replace package and VSA with another <span>(expectations)</span></summary>
+
+*Threat:* Replace a package and its VSA with a malicious package and its valid VSA.
+
+*Mitigation*: Verifier checks that the `resourceUri` in the VSA matches the package
+they've requested not just the package they received.
```
arewm, line 726: Are we assuming that the VSA producer is the same identity in both (and therefore the mitigation wouldn't just be to check the identity of the producer?

zachariahcox: I think the consumer has a list of data fields that must be true about their consumed artifacts. It's essentially the known-ahead-of-time-somehow "definition of correctness."

It would definitely include uri, issuer id, and maybe a bunch of domain-specific claims too

TomHennen: Yeah we already have a set of instructions for what to verify when checking the VSA: https://slsa.dev/spec/draft/verification_summary#how-to-verify

docs/spec/draft/threats.md
```diff
@@ -717,6 +717,20 @@ corresponding signing key.
...
+
+*Example:* Adversary uploads a malicious package to `repo/evil-package`,
+getting a valid VSA for `repo/evil-package`. Adversary then replaces
+`repo/my-package` and its VSA with `repo/evil-package` and its VSA.
+Solution: Verifier rejects because the VSA `resourceUri` field lists
+`repo/evil-package` and not the expected `repo/my-package`.
```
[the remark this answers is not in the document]
TomHennen: So right now this threats page is set up to enumerate lots of specific threats and then get into the details of how those specific threats can be mitigated.  Note how section 'd' covers each of the various expectations separately: https://slsa.dev/spec/draft/threats#d-external-build-parameters

What I hear you suggesting is that we change that model to something else?

FWIW there was a brief discussion [here](https://github.com/slsa-framework/slsa/issues/1184#:~:text=Separately%2C%20I%27m%20wondering%20if%20this%20is%20a%20reason%20to%20move%20the%20mitigations%20from%20(G)%20(discussed%20in%20%231190%2C%20%231191)%20to%20(I)%20since%20%27usage%27%20is%20more%20clearly%20aligned%20with%20this%20risk%20than%20with%20%27distribution%20channel%27.%20(Or%20maybe%20we%20can%20just%20tweak%20the%20mitigation%20language%20here%20to%20discuss).) about moving more things to 'Usage'.

docs/spec/draft/threats.md
zachariahcox: should it just be this then, to be more general? 
I don't have a very strong opinion here. 

```suggestion
*Mitigation*: Consumer ensures that the VSA matches the package they've requested (not just the package they received) by following the [verification process](verification_summary#how-to-verify).
```

zachariahcox (approved):

docs/spec/draft/threats.md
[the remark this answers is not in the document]
zachariahcox: > What I hear you suggesting is that we change that model to something else?

I bet this kind of detail is useful for search engine optimization too, so I'm generally okay with very specific examples. 

I don't think this section needs to be as thorough and complete as other sections of the spec. 
In these examples I'm fine saying "here's a good example of why a specific element is required in the VSA spec."

docs/spec/draft/threats.md
TomHennen: That's great. :)

docs/spec/draft/threats.md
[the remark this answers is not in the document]
TomHennen: Excellent, that seems to be WAI then.  Let me close this thread. (but please reopen if you have other thoughts)

PR #1251, opened 2024-12-03
content: draft: Add mitigation for compromised build tooling

### LINKED ISSUE #1184: TODO: Need mitigation description for "Use a compromised build tool" threat

TomHennen:
This threat can be mitigated in a number of ways, here I address it in the simplest one, verifying the tooling prior to use.  You can also imagine resolving it by recording the digests in the provenance, and propagating VSAs so that downstream verifiers can verify recursively, but that's pretty complicated.

You can also resolve this with the attested build environments track, but I don't think we should mention that here until it's finalized? Or maybe we can point to it now as 'coming soon'?

fixes #1184

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -866,14 +866,19 @@ libDep, resulting in MyPackage also having the security vulnerability.
 process, which alters the build process and injects unintended behavior into the
 output artifact.
 
+*Mitigation:* Treat build tooling, including OS images, as any other software
+to be verified prior to use (as described in (G)). This will allow the build
+platform to detect any modified binaries.
+
 *Example:* MyPackage is a tarball containing an ELF executable, created by
 running `/usr/bin/tar` during its build process. An adversary compromises the
 `tar` OS package such that `/usr/bin/tar` injects a backdoor into every ELF
 executable it writes. The next time MyPackage is built, the build picks up the
 vulnerable `tar` package, which injects the backdoor into the resulting
-MyPackage artifact.
-
-*Mitigation:* **TODO**
+MyPackage artifact.  Solution: The build platform verifies the disk image,
+or the individual components on the disk image, against the associated
+provenance or VSAs prior to running a build. The modified `/usr/bin/tar`
+will fail this verification.
 
 </details>
 


marcelamelara:
> You can also resolve this with the attested build environments track, but I don't think we should mention that here until it's finalized? Or maybe we can point to it now as 'coming soon'?

As a status update, I am still hoping to open a PR before EOY where we describe the detailed verification requirements for the BuildEnv track.

That said, these requirements are already laid out at a high level in `draft/attested-build-env-levels.md`. Not sure what level of detail makes more sense for the threats page. And #1244 may actually be the more pertinent PR to have merged in the context of the updates in this PR.

marcelamelara:
> And https://github.com/slsa-framework/slsa/pull/1244 may actually be the more pertinent PR to have merged in the context of the updates in this PR.

Looking at the current 1.1RC terminology page, the terminology page with the build environment model isn't included. So, I think the most we should do at this point is say that the "upcoming build environment track may provide additional assurances against this threat/"

docs/spec/draft/threats.md
```diff
@@ -866,14 +866,19 @@ libDep, resulting in MyPackage also having the security vulnerability.
...
 `tar` OS package such that `/usr/bin/tar` injects a backdoor into every ELF
 executable it writes. The next time MyPackage is built, the build picks up the
 vulnerable `tar` package, which injects the backdoor into the resulting
-MyPackage artifact.
-
-*Mitigation:* **TODO**
+MyPackage artifact.  Solution: The build platform verifies the disk image,
+or the individual components on the disk image, against the associated
+provenance or VSAs prior to running a build. The modified `/usr/bin/tar`
+will fail this verification.
```
marcelamelara, line 881: FWIW, this describes what the [BuildEnv track L1](https://slsa.dev/spec/draft/attested-build-env-levels#buildenv-l1-signed-build-image-provenance-exists) would have you do :) Maybe it is worth mentioning the BuildEnv track as upcoming here after all. You may also want to link to the [verifying platforms](https://slsa.dev/spec/v1.1/verifying-systems) page here as well.

docs/spec/draft/threats.md
```diff
@@ -866,14 +866,19 @@ libDep, resulting in MyPackage also having the security vulnerability.
 process, which alters the build process and injects unintended behavior into the
 output artifact.
 
+*Mitigation:* Treat build tooling, including OS images, as any other software
```
trishankatdatadog, line 869: What's not clear here is that this is, at best, a partial solution to the problem, which is fine. My point is that a VSA or a build provenance by itself does not[^1] necessarily guarantee that the originally built image does not contain a backdoor. The solution proposed here would catch any backdoor introduced _after_ that.

[^1]: Except if it goes thru a policy for [reproducible builds _and_ diverse double-compiling](https://research.swtch.com/nih).

arewm, line 869: We use artifact more than software in this page.

docs/spec/draft/threats.md
```diff
@@ -866,14 +866,19 @@ libDep, resulting in MyPackage also having the security vulnerability.
...
 running `/usr/bin/tar` during its build process. An adversary compromises the
 `tar` OS package such that `/usr/bin/tar` injects a backdoor into every ELF
 executable it writes. The next time MyPackage is built, the build picks up the
 vulnerable `tar` package, which injects the backdoor into the resulting
-MyPackage artifact.
-
-*Mitigation:* **TODO**
+MyPackage artifact.  Solution: The build platform verifies the disk image,
+or the individual components on the disk image, against the associated
+provenance or VSAs prior to running a build. The modified `/usr/bin/tar`
```
arewm, line 880: What do you mean here by verifying against the provenance or VSAs? Is this just verifying trust in the provenance/VSA and then ensuring that the artifact is the subject for the attestations?

docs/spec/draft/threats.md
arewm: Is the BuildEnv track intended to be verified at a lower level than the build platform itself or as part of the build platform (I suspect that the answer depends on the boundaries of the build platform).

This is also relevant to `[apply SLSA recursively]: verifying-artifacts.md#step-3-optional-check-dependencies-recursively` as mentioned earlier in the document.

TomHennen: I think we should definitely mention the future BuildEnv track, but I'm not quite sure what to link to yet?  Elsewhere when we talk about the source track (and other things) we tend to point to https://slsa.dev/spec/draft/future-directions#build-track.  WDYT about adding a blurb there about the BuildEnv track and then sending another PR to mention it here as needed?

arewm: add a future direction for the BuildEnv track and then add a xref in the threats?

TomHennen: > add a future direction for the BuildEnv track and then add a xref in the threats?

Right but in this PR or a future one?  I was lazily suggesting that @marcelamelara do so at her earliest convenience. :p

marcelamelara: I gotcha @TomHennen :) I hadn't realized the BuildEnv track wasn't mentioned in the future directions page. My thumbs up was meant to signify that I'd send the PR! I'll open an issue to track this change for the sake of completeness.

zachariahcox (approved):
this change looks good to me

docs/spec/draft/threats.md
TomHennen: I went ahead and added the reference to the track since you got that in before I got this in. :)

docs/spec/draft/threats.md
```diff
@@ -866,14 +866,23 @@ libDep, resulting in MyPackage also having the security vulnerability.
 process, which alters the build process and injects unintended behavior into the
 output artifact.
 
+*Mitigation:* This can be partially mitigated by treating build tooling,
+including OS images, as any other artifact to be verified prior to use.
+The threats described in this document apply recursively to build tooling
+as do the mitigations and examples.  A future
+[Build Environment track](future-directions#build-environment-track) may
+provide more comprehensive guidance on how to address this threat.
```
marcelamelara, line 874: Not necessarily a blocker, but I'm wondering if we should generalize and say something like "how to address _more specific vectors/aspects of_ this threat" just so we're extra clear that even the BuildEnv track may or may not fully address this threat.

marcelamelara (commented):
Thanks for the updates @TomHennen ! I'm happy to approve this once my comment has been resolved.

docs/spec/draft/threats.md
TomHennen: Done :)

docs/spec/draft/threats.md
```diff
@@ -866,14 +866,24 @@ libDep, resulting in MyPackage also having the security vulnerability.
 process, which alters the build process and injects unintended behavior into the
 output artifact.
 
+*Mitigation:* This can be partially mitigated by treating build tooling,
+including OS images, as any other artifact to be verified prior to use.
+The threats described in this document apply recursively to build tooling
+as do the mitigations and examples.  A future
+[Build Environment track](future-directions#build-environment-track) may
+provide more comprehensive guidance on how to address more specfiic
+aspects this threat.
```
marcelamelara, line 875: Nit
```suggestion
aspects of this threat.
```

marcelamelara (approved):
Thanks so much @TomHennen !

docs/spec/draft/threats.md
```diff
@@ -866,14 +866,24 @@ libDep, resulting in MyPackage also having the security vulnerability.
...
 executable it writes. The next time MyPackage is built, the build picks up the
 vulnerable `tar` package, which injects the backdoor into the resulting
-MyPackage artifact.
-
-*Mitigation:* **TODO**
+MyPackage artifact.  Solution: [apply SLSA recursively] to all build tools
+prior to the build.  The build platform verifies the disk image,
+or the individual components on the disk image, against the associated
+provenance or VSAs prior to running a build.  Depending on where the initial
+compromise took place the modified `/usr/bin/tar` will fail this verification.
```
trishankatdatadog, line 886:
```suggestion
compromise took place (i.e. before/during vs _after_ the build of the build tool itself), the modified `/usr/bin/tar` will fail this verification.
```

trishankatdatadog (approved):

arewm (approved):

PR #1252, opened 2024-12-03
clarify orgs are responsible for what gets expunged

### LINKED ISSUE #1222: Safe Expunging and 'legal' restrictions
I'm also not sure the scs can be on the hook for enforcing that the data removals were for legal reasons. 
Generally I think "the owner of the intellectual property in the repo (the root repo owner, in gh terms) can remove data. They should only do this for legal or privacy reasons due to the risk of severe reputational consequences (damage to artifact chain of custody)." Ie: it's indistinguishable from a repo hijack so you should have a good reason.

_Originally posted by @zachariahcox in https://github.com/slsa-framework/slsa/pull/1203#pullrequestreview-2393973461_

TomHennen:
The prior language made it sound like the SCS was responsible for only performing expunging for legal requests.  That's not reasonable from a technical perspective.  With this update we clarify that it's the _organization's_ responsibility to ensure this bar is met.

I had thought about adding a whole section of requirements just for the organization, but that seems out of scope for this change and it would be a  lot of work.  We can still do so if we want, but not here.

fixes #1222

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -47,7 +47,7 @@ Consumers can examine the various source provenance attestations to determine if
 ## Safe Expunging Process
 
 SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content,
-but MUST only allow these changes in order to meet legal or privacy compliance requirements.
+but the organization MUST only allow these changes in order to meet legal or privacy compliance requirements.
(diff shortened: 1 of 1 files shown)


trishankatdatadog (approved):

zachariahcox (approved):
100%, this makes sense. Thanks!
@TomHennen I think we should come through separately and talk about organization vs. "producer" terminology, based on our discussion from a couple weeks ago at the spec sync.

PR #1280, opened 2025-02-07
nonspec: Split Tracks out of the Draft specification into separate specs

lehors:
Turn the build, build environment, and source track sections into separate specifications.
Change the Draft specification to only hold general content and a highlevel description of the tracks with pointers to the individual specs.


lehors:
For the record: I think this works pretty well but I am struggling with the terminology section. Most of its current content is build related so I moved it into the Build track spec but I actually think that's too radical. The right thing might be to have a global terminology section that stays with the (main) draft spec and can be referenced from the different track specs. Then each track spec can have its own specialized terminology section that extends the main one with terms specific to that track.
At the same time, splitting the current terminology section into a general one and a build specific one requires careful surgery (for which I may not be the best person).

docs/_data/nav/main.yml
```diff
@@ -315,6 +286,48 @@
...
+
+  - title: Verifying artifacts
+    url: /build/v1.0/verifying-artifacts
+    description: Guidance for verifying software artifacts and their SLSA provenance, intended for platform implementers and software consumers
+
+  - title: Verifying build platforms
+    url: /build/v1.0/verifying-systems
+    description: Guidelines for securing SLSA Build L3+ builders, intended for platform implementers
+
+- title: SLSA Source WD
```
marcelamelara, line 316: Looking at the navigation menu, it's not immediately clear that this and the other tracks are "tracks". I'm wondering if these menu items could be renamed to something like "Track: Source WD" or simply "Source Track WD" to be very explicit. I also don't know if "WD" will be clear to everyone, but I do agree that we need something short.

lehors: I agree. Have a look at the latest update. Let me know what you think.
Thanks.

docs/spec/draft/threats-overview.md
```diff
@@ -74,7 +74,7 @@ For close source software SLSA does not provide any solutions for malicious prod
 <td>E
 <td>Build process
 <td><a href="https://www.crowdstrike.com/blog/sunspot-malware-technical-analysis/">SolarWinds</a>: Attacker compromised the build platform and installed an implant that injected malicious behavior during each build.
-<td>Higher SLSA levels require <a href="requirements#build-requirements">stronger security controls for the build platform</a>, making it more difficult to compromise and gain persistence.
+<td>Higher SLSA levels require <a href="../../build/v1.0/requirements">stronger security controls for the build platform</a>, making it more difficult to compromise and gain persistence.
```
paveliak, line 77: @lehors How would we manage threat modeling in this new design? With new tracks coming into SLSA spec I see people taking two approaches:
- `Source` track proposes to [amend existing document](https://github.com/slsa-framework/slsa/pull/1236)
- `BuildEnvironment` track proposes to [put threats into a separate document](https://github.com/slsa-framework/slsa/pull/1287)

Looking at this chage I think you adopt the first path with core specification defining the threats across all tracks. And then [BuildEnvironment threat PR](https://github.com/slsa-framework/slsa/pull/1287) needs to be updated moving all the changes into `threats.md`.

How does it sound?

docs/build/v1.0/terminology.md
```diff
@@ -0,0 +1,186 @@
...
+    source* and *build definition*.
+-   *Builder:* Usually means *build platform*, but might be used for *build
+    environment*, the user who invoked the build, or a build tool from
+    *dependencies*. To avoid confusion, we always use "build platform". The only
+    exception is in the [provenance](/provenance/v1), where `builder` is used as
+    a more concise field name.
+
+</details>
+
+### Build environment model
```
marcelamelara, line 92: This section should probably be moved to a separate terminology.md file under /build-env/draft/ because this content wasn't part of the build track 1.0

docs/spec/draft/levels.md
```diff
@@ -20,225 +21,46 @@ to recognize progress made in one aspect of security without blocking on an
...
 tracks without invalidating previous levels.
 
+## Build track levels
+
 | Track/Level | Requirements | Focus
 | ----------- | ------------ | -----
 | [Build L0]  | (none)       | (n/a)
 | [Build L1]  | Provenance showing how the package was built | Mistakes, documentation
 | [Build L2]  | Signed provenance, generated by a hosted build platform | Tampering after the build
 | [Build L3]  | Hardened build platform | Tampering during the build
```
marcelamelara, line 31: It looks like the links to the individual levels for this and the other tracks are missing. I think it may actually be ok to remove these links from this page since the actual track specs are referenced below each table.

lehors:
I'm closing this given the undesirable effect of this approach to have to manage two sets of version numbers and the new direction we are taking with the use of branches and subdirs for the tracks. There may still be some useful nuggets in there that we may want to leverage as we reorganize the tracks into subdirs but this PR clearly won't be merged as is.

PR #1283, opened 2025-02-11
fix(how-to): links to spec

scop:
HTML rendering does not appear to remove `.md` to "external" doc links.

### DIFF
--- a/docs/how-to/how-to-infra.md
+++ b/docs/how-to/how-to-infra.md
@@ -20,7 +20,7 @@ learn more about verifying your system for SLSA conformance, see
 provenance, see [Producing artifacts](/spec/v1.0/requirements). To learn more
 about the SLSA provenance format, see [Provenance](/provenance/v1).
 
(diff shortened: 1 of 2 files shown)


lehors:
Sorry, I beat you to it. PR #1282 addressed the problem. Note that this only occurs with links that have a fragment identifier. I don't know for sure but it looks to me like a bug in Jekyll.
Thanks though.

scop:
It does not seem to be limited to links with fragments. I checked every instance that I fixed in my version here before fixing them, and they were all broken similarly. The ones that seem to remain unaddressed are:

- The "SLSA Levels" link in https://slsa.dev/how-to/how-to-orgs#choosing-your-slsa-level
- The "Verifying Artifacts" link at end of https://slsa.dev/how-to/how-to-orgs#for-software-consumers

I've rebased & reworked my branch here to address those.

scop:
Hm, updating my branch did not seem to affect what is shown here, so I opened a new one with the remaining fixes: https://github.com/slsa-framework/slsa/pull/1286

PR #1286, opened 2025-02-11
fix(how-to-orgs): links to spec

scop:
HTML rendering does not appear to remove `.md` to "external" doc links.

Refs https://github.com/slsa-framework/slsa/pull/1283#issuecomment-2651904208


lehors (approved):
It's unclear to me why Jekyll chokes on those two links but thanks!

Issue #1267, opened 2025-01-03
Build Environment track threats


paveliak:
`Build` track defines requirements for the provenance that is generated for the build artifacts. Trustworthiness of the build provenance largely depends on the trustworthiness of the _environment_ the build runs in. Build track assumes full trust into the `Build Platform` providing no solutions to verify build environment integrity. `Build Environment` track intends to close this gap.

Build environment is bootstrapped from a `build image`, which is expected to be an artifact of a SLSA build pipeline. Build platform verifies image provenance upon starting up the environment and provides evidence to the tenant. 

Bootrapping the build environment is a complex process, especially at higher SLSA levels. [Build L3](https://slsa.dev/spec/v1.0/levels#build-l3-hardened-builds) usually requires significant changes to existing build platforms to maintain ephemeral build environments. It is not uncommon for the build platforms to rely on public cloud providers for managing compute resources that power build environments. This in turn might significantly increase attack surface because added build platform dependencies effectively become part of the `TCB`.

`Build Environment` track addresses TCB size concerns at `L2` and `L3` levels. L1 level assumes full trust into the Build Platform including underlying Compute Platform (eg. public cloud provider). L2 level adds capabilities for verifying Compute Platform. L3 level removes Compute Platform from TCB rooting the trust into the hardware.

# Threat model
This diagram outlines the lifetime of a build image between it being generated and used for creating a build environment. Image could be compromised at different times (components). Higher SLSA levels secure the build environment from a larger amount of threats.

```mermaid
flowchart LR
      BuildImage>Build Image] ---> |L1|BuildPlatform[[Build Platform]]
      BuildPlatform[[Build Platform]] ---> |L2|ComputeProvider[[Compute Provider]]
      ComputeProvider[[Compute Provider]] ---> |L3|BuildEnvironment[(Build Environment)]
```

`L1` level protects from threats that happened to the build image in between generating it and passing to the `Build Platform`. This covers cases of unauthorized modifications happening to the image as it is distributed (potentially via untrusted channels).

`L2` level protects from threats occurring at the `Build Platform` side as it interacts with the `Compute Provider` for actually creating a build environment. `Control plane` is the only `Build Platform` component that is considered trusted at `L2` as it performs remote attestation of the build environment.

[...]

PR #1287, opened 2025-02-12
content: Adding Build Environment Threats section

paveliak:
This PR integrates https://github.com/slsa-framework/slsa/issues/1267 into the Build Environment draft spec

### DIFF
--- a/docs/spec/draft/attested-build-env-levels.md
+++ b/docs/spec/draft/attested-build-env-levels.md
@@ -1,6 +1,7 @@
 ---
 title: Build Environment track
 description: This page gives an overview of the SLSA Build Environment track and its levels, describing their security objectives and general requirements.
+mermaid: true
 ---
 
 ## Rationale
@@ -96,7 +97,32 @@ A typical build environment will go through the following lifecycle:
 
 ### Build environment threats
 
-TODO
+SLSA [Build track] defines requirements for the provenance that is generated for the build artifacts. Trustworthiness of the build provenance largely depends on the trustworthiness of the [build environment] the build runs in. Build track assumes full trust into the [Build Platform] providing no solutions to verify integrity of the build environment. BuildEnv track intends to close this gap.
+
+Build environment is bootstrapped from a [build image], which is expected to be an artifact of a SLSA build pipeline. Build platform verifies image provenance upon starting up the environment and provides evidence to the tenant.
+
+Bootstrapping the build environment is a complex process, especially at higher SLSA levels. [Build L3] usually requires significant changes to existing build platforms to maintain ephemeral build environments. It is not uncommon for the build platforms to rely on public cloud providers for managing compute resources that power build environments. This in turn might significantly increase attack surface because added build platform dependencies effectively become part of the [TCB].
+
+BuildEnv track addresses TCB size concerns at [BuildEnv L2] and [BuildEnv L3] levels. [BuildEnv L1] level assumes full trust into the Build Platform including underlying [Compute Platform] (eg. public cloud provider). BuildEnv L2 level adds capabilities for verifying Compute Platform. BuildEnv L3 level removes Compute Platform from TCB rooting the trust into the hardware.
+
+This diagram outlines the lifetime of a build image between it being generated and used for creating a build environment. Build Image could be compromised at different times (components). Higher SLSA levels secure the build environment from a larger amount of threats.
+
+<div class="mermaid">
+flowchart LR    
+      BuildImage>Build Image] ---> |L1|BuildPlatform[[Build Platform]]
+      BuildPlatform[[Build Platform]] ---> |L2|ComputeProvider[[Compute Provider]]
+      ComputeProvider[[Compute Provider]] ---> |L3|BuildEnvironment[(Build Environment)]
+</div>
+
+[BuildEnv L1] protects from threats that happened to the build image in between generating it and passing to the Build Platform. This covers cases of unauthorized modifications happening to the image as it is distributed (potentially via untrusted channels).
+
+[BuildEnv L2] level protects from threats occurring at the Build Platform side as it interacts with the Compute Platform for actually creating a build environment. [Control plane] is the only Build Platform component that is considered trusted at L2 as it performs remote attestation of the build environment.
+
+[BuildEnv L3] level protects from threats coming from the Compute Platform. It requires build running in a trusted execution environment using technologies like [AMD SEV] and [Intel TDX]. Compute platform is considered untrusted at L3 with the trust rooted into hardware.
+
+NOTE: [Control Plane] is considered trusted at L2 and L3 as it performs remote attestation of the build environment. Build platforms MAY provide capabilities that let tenants perform remote attestation themselves. However, for the Control Plane to be considered untrusted it should have no back-door access to the build environment (e.g. via SSH). Besides, Control Plane provides input data to the build environment (i.e. build request message) and security risks associated with compromising inputs have to be considered as well.
+
+For the example threats refer to the [Build Threats] section
 
 ## BuildEnv levels
 
@@ -334,6 +360,7 @@ TODO
 
 <!-- Link definitions -->
 
+[Build track]: levels.md#build-track
 [Build L1]: levels.md#build-l1
 [Build L2]: levels.md#build-l2
 [Build L3]: levels.md#build-l3
@@ -341,25 +368,33 @@ TODO
 [BuildEnv L1]: #buildenv-l1
 [BuildEnv L2]: #buildenv-l2
 [BuildEnv L3]: #buildenv-l3
+[control plane]:  terminology.md#control-plane
+[dm-verity]: https://docs.kernel.org/admin-guide/device-mapper/verity.html
 [Release Attestation]: https://github.com/in-toto/attestation/blob/main/spec/predicates/release.md
 [SCAI]: https://github.com/in-toto/attestation/blob/main/spec/predicates/scai.md
 [Secure Boot]: https://wiki.debian.org/SecureBoot#What_is_UEFI_Secure_Boot.3F
 [SLSA Build Provenance]: provenance.md
 [TPM]: https://trustedcomputinggroup.org/resource/tpm-library-specification/
 [VSA]: verification_summary.md
-[build image]: #build-image
+[build image]: terminology.md#build-image
 [confidential computing]: https://confidentialcomputing.io/wp-content/uploads/sites/10/2023/03/Common-Terminology-for-Confidential-Computing.pdf
 [execution context]: terminology.md#build-environment
 [hosted]: requirements.md#isolation-strength
-[boot process]: #boot-process
-[build agent]: #build-agent
-[build image producer]: #build-image-producer
+[boot process]:  terminology.md#boot-process
+[build agent]: terminology.md#build-agent
+[build environment]: terminology.md#build-environment
+[build image producer]: terminology.md#build-image-producer
+[build platform]: terminology.md#platform
 [build platforms]: terminology.md#platform
-[compute platform]: #compute-platform
-[host interface]: #host-interface
-[measurement]: #measurement
+[compute platform]: terminology.md#compute-platform
+[host interface]: terminology.md#host-interface
+[measurement]: terminology.md#measurement
 [provenance]: terminology.md#provenance
-[quote]: #quote
-[reference values]: #reference-value
+[quote]: terminology.md#quote
+[reference values]: terminology.md#reference-value
 [several classes]: #build-environment-threats
 [vTPM]: https://trustedcomputinggroup.org/about/what-is-a-virtual-trusted-platform-module-vtpm/
+[AMD SEV]: https://www.amd.com/en/developer/sev.html
+[Intel TDX]: https://www.intel.com/content/www/us/en/developer/tools/trust-domain-extensions/overview.html
+[TCB]: https://csrc.nist.gov/glossary/term/trusted_computing_base
+[Build Threats]: threats.md#build-threats
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -416,10 +416,9 @@ the source repo does not match the expected value.
 ### (E) Build process
 
 An adversary introduces an unauthorized change to a build output through
-tampering of the build process; or introduces false information into the
-provenance.
+tampering of the build image or build process; or introduces false information into the provenance.
 
-These threats are directly addressed by the SLSA Build track.
+These threats are directly addressed by the SLSA Build and Build Environment tracks.
 
 <details><summary>Forge values of the provenance (other than output digest) <span>(Build L2+)</span></summary>
 
@@ -553,19 +552,71 @@ transitive closure of the inputs, and the cache entry is itself a SLSA Build L3
 build with its own provenance that corresponds to the key.
 
 </details>
-<details><summary>Compromise build platform admin <span>(verification)</span></summary>
+<details><summary>Compromise build platform admin<span>(BuildEnv L2+)</span></summary>
 
 *Threat:* An adversary gains admin permissions for the artifact's build platform.
 
-*Mitigation:* The build platform must have controls in place to prevent and
-detect abusive behavior from administrators (e.g. two-person approvals, audit
-logging).
+*Mitigation:* Remote (SSH) access is disabled in the build image. 
+Control Plane attests build environment before the job is started in it and 
+produces a [VSA][vsa] providing evidence that build environment has no outside-in access.
 
 *Example:* MyPackage is built on Awesome Builder. Awesome Builder allows
 engineers on-call to SSH into build machines to debug production issues. An
 adversary uses this access to modify a build in progress. Solution: Consumers
-do not accept provenance from the build platform unless they trust sufficient
-controls are in place to prevent abusing admin privileges.
+do not accept provenance from the build platform unless they can validate 
+build environment integrity.
+
+</details>
+<details><summary>Compromise build image<span>(BuildEnv L1+)</span></summary>
+
+*Threat:* An adversary injects malicious code into the build image at the generation time 
+or in transit.
+
+*Mitigation:* Build image comes with SLSA provenance. Control plane attests build environment 
+upon creation and at runtime and issues Build provenance only if environment passed attestation.
+
+*Example:* MyPackage is built on Awesome Builder. Awesome Builder uses VM images provided by a 
+Fancy Image partner. Adversary was able to install malicious tool into the image at the Fancy Image 
+side hence infiltrating all build environments that use that image.
+
+*Example 2:* MyPackage is built on Awesome Builder. An adversary gained access to the 
+build image storage and modified image at rest injecting a malicious tool into it. All build 
+environments created after that point were compromised.
+
+</details>
+<details><summary>Unexpected reuse of a build environment</summary>
+
+*Threat:* Adversary injects malicious code into the target build by being able to run random code in the same build environment
+
+_Mitigation_: Unique build identifier is included into the Build environment provenance (and [TPM] measurement) allowing [Control plane] to detect environment reuse. Needs [BuildEnv L2] level
+
+_Example_: Due to a bug in the build platform, the environment was used for running two or more jobs and effectively losing “ephemeral” property. Malicious actors could use this vulnerability to poison the build environments they should not have access to.
+
+</details>
+<details><summary>Compromise secure boot<span>(BuildEnv L3)</span></summary>
+
+*Threat:* An adversary having unauthorized access to the compute provider infrastructure was able 
+to deploy a bootkit and circumvent secure boot.
+
+*Mitigation:* Build platform relies on hardware-assisted mechanisms to attest integrity of the 
+build environment (e.g. hardware [TPM], [AMD SEV], [Intel TDX])
+
+*Example:* Awesome Builder uses Cloudy Sky compute provider for provisioning virtual machines. 
+An adversary got an unauthorized access to Cloudy Sky infrastructure and was able to deploy 
+malicious code into UEFI and vTPM and compromise virtual machine attestations.
+
+</details>
+<details><summary>Compromise compute provider<span>(BuildEnv L3)</span></summary>
+
+*Threat:* An adversary having unauthorized access to the host machine performs malicious change in the 
+build environment by using hypervisor capabilities to manipulate virtual machine state.
+
+*Mitigation:* Build platform uses hardware-assisted mechanisms to attest integrity of the build environment 
+(e.g. [AMD SEV] and [Intel TDX])
+
+*Example:* Awesome Builder uses Cloudy Sky compute provider for provisioning virtual machines. An adversary 
+got an unauthorized access to the host machines in Cloudy Sky and was able to hijack hypervisor and inject 
+malicious code into the virtual machine.
 
 </details>
 
@@ -1029,3 +1080,6 @@ Solution: Only accept cryptographic hashes with strong collision resistance.
 [supply chain threats]: threats-overview
 [vsa]: verification_summary
 [vsa_verification]: verification_summary#how-to-verify
+[TPM]: https://trustedcomputinggroup.org/resource/tpm-library-specification/
+[AMD SEV]: https://www.amd.com/en/developer/sev.html
+[Intel TDX]: https://www.intel.com/content/www/us/en/developer/tools/trust-domain-extensions/overview.html
--- a/docs/_includes/head.html
+++ b/docs/_includes/head.html
@@ -30,4 +30,7 @@
     {%- if page.noindex %}
     <meta name="robots" content="noindex">
     {% endif -%}
+    {%- if page.mermaid %}
+    {% include mermaid.html %}  
+    {% endif -%}
 </head>
--- a/docs/_includes/mermaid.html
+++ b/docs/_includes/mermaid.html
@@ -1,0 +1,11 @@
+<script type="module">
+    import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@11.4.1/+esm';
+    let config = { 
+        startOnLoad: true,
+        flowchart: { 
+            useMaxWidth: true, 
+            htmlLabels: true 
+        } 
+    };
+    mermaid.initialize(config);
+</script>


paveliak:
@marcelamelara Mermaid didn't work 😁 I need to convert diagram to an acceptable format. The text part is ready for review

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -47,7 +47,64 @@ environment, and the compute platform they used.
 
 ### Build environment threats
 
-TODO
+SLSA [Build track] defines requirements for the provenance that is generated for the build artifacts. Trustworthiness of the build provenance largely depends on the trustworthiness of the [build environment] the build runs in. Build track assumes full trust into the [Build Platform] providing no solutions to verify integrity of the build environment. BuildEnv track intends to close this gap.
+
+Build environment is bootstrapped from a [build image], which is expected to be an artifact of a SLSA build pipeline. Build platform verifies image provenance upon starting up the environment and provides evidence to the tenant.
+
+Bootrapping the build environment is a complex process, especially at higher SLSA levels. [Build L3] usually requires significant changes to existing build platforms to maintain ephemeral build environments. It is not uncommon for the build platforms to rely on public cloud providers for managing compute resources that power build environments. This in turn might significantly increase attack surface because added build platform dependencies effectively become part of the TCB.
```
kommendorkapten, line 54: Is `TCB` defined somewhere?

kommendorkapten: As this is the first time mentioning it (that I found), maybe spell it out first then put `TCB` in parenthesis.

kommendorkapten: Oh I see now that there is a link section below, `[TCB]` would then be preferred.

paveliak: Done. Added link to https://csrc.nist.gov/glossary/term/trusted_computing_base

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -47,7 +48,64 @@ environment, and the compute platform they used.
...
+
+What follows are the example threats.
+
+#### Invalid build image
+
+_Threat_: Wrong [build image] is used for the [build environment]. This could be a dev/test image or an older version of the image having security vulnerabilities.
+
+_Mitigation_: [Control Plane] verifies build image provenance upon creating build environment. Needs [BuildEnv L1] level
+
+_Example_: Malicious actor gained access to the build image supply chain and was ultimately able to configure the wrong image in the Build platform.
```
marcelamelara, line 84: For this and the other threats , I think we should flip the Mitigation and Example so the example follows the threat description.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -47,7 +48,64 @@ environment, and the compute platform they used.
...
+
+#### Reused build environment
+
+_Threat_: [Build environment] unexpectedly reused across two or more builds
+
+_Mitigation_: Unique build identifier is included into the Build environment provenance (and [TPM] measurement) allowing [Control plane] to detect environment reuse. Needs [BuildEnv L2] level
+
+_Example_: Due to a bug in the build platform, the environment was used for running two or more jobs and effectively losing “ephemeral” property. Malicious actors could use this vulnerability to poison the build environments they should not have access to.
+
+#### Continuous integrity of the build environment
```
marcelamelara, line 102: Two things: 1) "Continuous integrity" isn't really a threat, so we probably want to replace this with something like "runtime host system compromise". 
2) I think we need to define the scope of "integrity of the build environment" at L3 because hardware TPMs + Linux IMA can also be used to provide continuous integrity checking, but of the file system only, whereas TEEs can extend this to the memory.

marcelamelara (commented):
Thank you for this first draft @paveliak ! I've got a few comments.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -47,7 +48,64 @@ environment, and the compute platform they used.
...
+
+_Threat_: [Build environment] integrity was compromised due to unauthorized access on the [Compute platform] provider side.
+
+_Mitigation_: Trusted execution environment provided by hardware-assisted mechanisms like [AMD SEV] and [Intel TDX] secures access to the build environment state even in the event of a fully compromised compute provider. Requires [BuildEnv L3] level.
+
+_Example_: Malicious actor (potentially a rogue administrator) was able to retrieve encryption secrets from the build environment memory and modify contents of the root or temporary file system (used to store transient data including build artifacts).
```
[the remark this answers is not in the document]
marcelamelara: @bureado The example scenarios you provide are actually spot-on, we should add them to this section. Thank you!

I definitely agree, it's tricky to actually rely on IMA-like mechanisms to check "runtime integrity" for complex build environments (which most of them are). Even though there do exist HW TEEs that could support IMA (TDX comes to mind, there may be others), the non-determinism of the system still poses a challenge. This is the main reason why we've so far excluded runtime integrity from the BuildEnv track, and I still think we should. Now, if the build environment's FS, for example, was partitioned into read-only and RW segments or something like this, we could start considering an additional "runtime integrity" level in this track.

The other reason I think HW TEEs could be valuable for "integrity at runtime" is also their ability to detect modifications to CPU states and the TEE memory in use, in addition to their confidentiality properties. But at L3, I still think this particular property is nice-to-have.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -47,7 +48,64 @@ environment, and the compute platform they used.
...
+
+_Example_: Malicious actor gained unauthorized access to the build environment root file system and was able to modify it. For a build image that is served remotely via a network file system or a network block device this could happen if the remote server is compromised or data was compromised at rest.
+
+#### Reused build environment
+
+_Threat_: [Build environment] unexpectedly reused across two or more builds
+
+_Mitigation_: Unique build identifier is included into the Build environment provenance (and [TPM] measurement) allowing [Control plane] to detect environment reuse. Needs [BuildEnv L2] level
+
+_Example_: Due to a bug in the build platform, the environment was used for running two or more jobs and effectively losing “ephemeral” property. Malicious actors could use this vulnerability to poison the build environments they should not have access to.
```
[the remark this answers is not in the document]
marcelamelara: > This scenario appears to introduce a requirement that a control plane component (presumably in a higher privileged level than the build itself) is expected to compute a build identifier and measure it into a PCR... For example, if the TEE provides an instance identifier, you could use that. 

Yes, this needs to be clarified in this text. The actual [BuildEnv L2](https://slsa.dev/spec/draft/attested-build-env-levels#buildenv-l2-attested-build-environment-instantiation) requirements, as currently defined, actually do specify that it's the environment running with a TPM/in a TEE that is supposed to measure the instance ID into the PCR (or similar), not the control plane.

>While possible, this doesn't sound like the only (or even the most feasible) way to prevent environment reuse.

I think we actually need to clarify what we mean by "environment reuse." What we really mean is environment _instance_ reuse, in line with the non-interference requirements of Build L3, which seek to ensure that the same VM instance isn't reused by another tenant as a means to leak info about or influence another build. Perhaps we even need to make this distinction clearer in the Build track.

> it's perfectly fine to have multiple builds reuse the same build "environment" as long as the "environment" is pristine, that is, the runtime measurements match the boot time measurements.

This assumption is actually already built into the BuildEnv track. It's ok (and in fact, expected) to reuse the same "base" environment (i.e., base VM image, host machine). But what's maybe missing here is that we distinguish between the pre-build (running) environment and the actual "build at runtime". In the former, the CI platform has not handed off control of the build environment to the tenant, in the latter the tenant is executing their build and can essentially do whatever they want in the running VM. Because the instance ID is assigned _before_ the tenant gets the environment, we treat it as pre-build property. Does this make sense?

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
paveliak: > when I think of "runtime integrity" I'm thinking (...) IMA.

We can split vulnerabilities/threats for the build environment integrity into two buckets:
- "outside threats" - those that come from the compute platform that hosts the environment
- "inside threats" - those that come from the software running inside the build environment

This section (and I would say the whole Build environment track!) is scoped to the _outside threats_. Everything that happens inside the Build environment is considered "trusted" (for the scope of this track) and is passed on to the Build track (or a new future track).

Speaking about IMA it brings an interesting topic of dependencies across the tracks. Non-determinism of the builds could be addressed by the IMA policies tuned for the specific builds. In this model tenant (who schedules a build) is also a Build Image provider. By knowing all the details about the build process tenant may define strict IMA policies regarding what is allowed during the build process. These policies are passed to the Control Plane that attests Build Environment at the end of the build when it is requested to sign the build provenance.

The challenge with this model though is how much sense those polices make to the consumer of build artifacts? Obviously we can include those policies into the Build Environment provenance and we could also include the whole IMA log their too. That could be a lot of data with a low signal/noise ratio but that might be a common problem for all sorts of provenance.

docs/spec/draft/attested-build-env-levels.md
paveliak: I removed this threat as this topic require more thinking...

docs/spec/draft/threats.md
```diff
@@ -553,19 +552,71 @@ transitive closure of the inputs, and the cache entry is itself a SLSA Build L3
...
+
+_Mitigation_: Unique build identifier is included into the Build environment provenance (and [TPM] measurement) allowing control plane to detect environment reuse.
+
+_Example_: Due to a bug in the build platform, the environment was used for running two or more jobs and effectively losing “ephemeral” property. Malicious actors could use this vulnerability to poison the build environments they should not have access to.
+
+</details>
+<details><summary>Compromise secure boot<span>(BuildEnv L3)</span></summary>
+
+*Threat:* An adversary having unauthorized access to the compute provider infrastructure was able 
+to deploy a bootkit and circumvent secure boot.
```
marcelamelara, line 599: This is about the VM's secure boot, not the host's, right? We probably want to clarify that here.

marcelamelara (commented):
@paveliak Thanks for the updates to this. Excited to see the threat model come together!

TomHennen:
@deeglaze any feedback on this?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -96,7 +97,32 @@ A typical build environment will go through the following lifecycle:
...
+      BuildImage>Build Image] ---> |L1|BuildPlatform[[Build Platform]]
+      BuildPlatform[[Build Platform]] ---> |L2|ComputeProvider[[Compute Provider]]
+      ComputeProvider[[Compute Provider]] ---> |L3|BuildEnvironment[(Build Environment)]
+</div>
+
+[BuildEnv L1] protects from threats that happened to the build image in between generating it and passing to the Build Platform. This covers cases of unauthorized modifications happening to the image as it is distributed (potentially via untrusted channels).
+
+[BuildEnv L2] level protects from threats occurring at the Build Platform side as it interacts with the Compute Platform for actually creating a build environment. [Control plane] is the only Build Platform component that is considered trusted at L2 as it performs remote attestation of the build environment.
+
+[BuildEnv L3] level protects from threats coming from the Compute Platform. It requires build running in a trusted execution environment using technologies like [AMD SEV] and [Intel TDX]. Compute platform is considered untrusted at L3 with the trust rooted into hardware.
```
deeglaze, line 121:
```suggestion
[BuildEnv L3] protects from threats coming from the Compute Platform.
It requires build running in a trusted execution environment using technologies like [AMD SEV] and [Intel TDX].
The compute platform (i.e., privileged remote operators and host system software) is considered semi-trusted at L3.
Side channels can still be utilized to extract information when analyzing information like memory access patterns, and these cannot be ruled out without further integrity protections and assurances on the system software and human access controls.
The [SEV-SNP/TDX threat model](https://www.kernel.org/doc/Documentation/security/snp-tdx-threat-model.rst) describes this level of trust reduction through the use of memory encryption, integrity protection, and remote attestation with cryptographic keys that are certified by the hardware manufacturer.
TEE technologies are not infallible, so physical human access to hardware is still a risk that is accepted at L3.
```
AMD SEV does not have a particularly compelling integrity story. It does not remove trust from the hypervisor the way SEV-SNP and TDX do. Do you mean SEV-SNP?

I don't think we can say L3 makes the host completely untrusted either.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -96,7 +97,32 @@ A typical build environment will go through the following lifecycle:
...
+
+NOTE: [Control Plane] is considered trusted at L2 and L3 as it performs remote attestation of the build environment. Build platforms MAY provide capabilities that let tenants perform remote attestation themselves. However, for the Control Plane to be considered untrusted it should have no back-door access to the build environment (e.g. via SSH). Besides, Control Plane provides input data to the build environment (i.e. build request message) and security risks associated with compromising inputs have to be considered as well.
```
deeglaze, line 123:
```suggestion
NOTE: [Control Plane] is considered trusted at L2 and L3 because it _verifies_ the remote attestation of the build environment.
Build platforms MAY provide capabilities that let tenants perform remote attestation themselves.
However, for the Control Plane to be considered untrusted it should have no back-door access to the build environment (e.g. via SSH).
Besides, Control Plane provides input data to the build environment (i.e., build request message) and security risks associated with compromising inputs have to be considered as well.
```
I'm not very convinced by this. It seems to assume a particular build platform architecture I'm not familiar with.

docs/spec/draft/threats.md
```diff
@@ -553,19 +552,71 @@ transitive closure of the inputs, and the cache entry is itself a SLSA Build L3
...
+*Example 2:* MyPackage is built on Awesome Builder. An adversary gained access to the 
+build image storage and modified image at rest injecting a malicious tool into it. All build 
+environments created after that point were compromised.
+
+</details>
+<details><summary>Unexpected reuse of a build environment<span>(BuildEnv L2+)</span></summary>
+
+*Threat:* Adversary injects malicious code into the target build by being able to run random code in the same build environment
+
+_Mitigation_: Unique build identifier is included into the Build environment provenance (and [TPM] measurement) allowing control plane to detect environment reuse.
```
deeglaze, line 591:
```suggestion
*Mitigation:* Unique build identifier is included into the Build environment provenance (and [TPM] measurement) allowing control plane to detect environment reuse.
```
Use consistent typesetting.

docs/spec/draft/threats.md
```diff
@@ -553,19 +552,71 @@ transitive closure of the inputs, and the cache entry is itself a SLSA Build L3
...
+
+_Example_: Due to a bug in the build platform, the environment was used for running two or more jobs and effectively losing “ephemeral” property. Malicious actors could use this vulnerability to poison the build environments they should not have access to.
```
deeglaze, line 593:
```suggestion
*Example:* Due to a bug in the build platform, the environment was used for running two or more jobs and effectively losing “ephemeral” property. Malicious actors could use this vulnerability to poison the build environments they should not have access to.
```
Consistent typesetting.

deeglaze (changes requested):

deeglaze:
> @deeglaze any feedback on this?

Thanks for the ping.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -96,7 +97,32 @@ A typical build environment will go through the following lifecycle:
 
 ### Build environment threats
 
-TODO
+SLSA [Build track] defines requirements for the provenance that is generated for the build artifacts. Trustworthiness of the build provenance largely depends on the trustworthiness of the [build environment] the build runs in. Build track assumes full trust into the [Build Platform] providing no solutions to verify integrity of the build environment. BuildEnv track intends to close this gap.
+
+Build environment is bootstrapped from a [build image], which is expected to be an artifact of a SLSA build pipeline. Build platform verifies image provenance upon starting up the environment and provides evidence to the tenant.
```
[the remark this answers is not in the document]
marcelamelara: Agreed, we should lay out the benefits of providing evidence for security properties that are in place. We should also note that what is meant here is actually SLSA Provenance specifically, and that at higher levels of the track several other security properties are checked.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -96,7 +97,32 @@ A typical build environment will go through the following lifecycle:
...
+This diagram outlines the lifetime of a build image between it being generated and used for creating a build environment. Build Image could be compromised at different times (components). Higher SLSA levels secure the build environment from a larger amount of threats.
+
+<div class="mermaid">
+flowchart LR    
+      BuildImage>Build Image] ---> |L1|BuildPlatform[[Build Platform]]
+      BuildPlatform[[Build Platform]] ---> |L2|ComputeProvider[[Compute Provider]]
+      ComputeProvider[[Compute Provider]] ---> |L3|BuildEnvironment[(Build Environment)]
+</div>
+
+[BuildEnv L1] protects from threats that happened to the build image in between generating it and passing to the Build Platform. This covers cases of unauthorized modifications happening to the image as it is distributed (potentially via untrusted channels).
```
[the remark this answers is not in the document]
marcelamelara: I'd replace "transmission" with "distribution" since that's the term we use throughout the spec more commonly.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
paveliak: >  image security properties including provenance

What would be the reasoning for having image security properties that are not part of the provenance? Are they being passed on the side in addition to the provenance?

I can imagine that _environment_ has additional security properties beyond those that come with the provenance from the build image provider and so when build platform attests build environment it considers a combination of both. Is it what you had in mind?

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -96,7 +97,32 @@ A typical build environment will go through the following lifecycle:
...
+
+Bootstrapping the build environment is a complex process, especially at higher SLSA levels. [Build L3] usually requires significant changes to existing build platforms to maintain ephemeral build environments. It is not uncommon for the build platforms to rely on public cloud providers for managing compute resources that power build environments. This in turn might significantly increase attack surface because added build platform dependencies effectively become part of the [TCB].
+
+BuildEnv track addresses TCB size concerns at [BuildEnv L2] and [BuildEnv L3] levels. [BuildEnv L1] level assumes full trust into the Build Platform including underlying [Compute Platform] (eg. public cloud provider). BuildEnv L2 level adds capabilities for verifying Compute Platform. BuildEnv L3 level removes Compute Platform from TCB rooting the trust into the hardware.
+
+This diagram outlines the lifetime of a build image between it being generated and used for creating a build environment. Build Image could be compromised at different times (components). Higher SLSA levels secure the build environment from a larger amount of threats.
```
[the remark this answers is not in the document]
paveliak: I meant to say that there is an inherent lifetime of an image between generating it and creating a build environment with it. And along this time line image is being passed across different parties and so there is both time and space scale at play. I will refine wording

docs/spec/draft/threats.md
```diff
@@ -553,19 +552,71 @@ transitive closure of the inputs, and the cache entry is itself a SLSA Build L3
...
+<details><summary>Compromise build platform admin<span>(BuildEnv L2+)</span></summary>
 
 *Threat:* An adversary gains admin permissions for the artifact's build platform.
 
-*Mitigation:* The build platform must have controls in place to prevent and
-detect abusive behavior from administrators (e.g. two-person approvals, audit
-logging).
+*Mitigation:* Remote (SSH) access is disabled in the build image. 
+Control Plane attests build environment before the job is started in it and 
+produces a [VSA][vsa] providing evidence that build environment has no outside-in access.
```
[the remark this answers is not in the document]
paveliak: It is the responsibility of the build image provider to disable remote access. And tenant should either trust build image provider or be the image provider themselves. Once control plane attested build environment we can be confident that remote access is disabled because we know what's in image (e.g. SSH server is not running).

There is a separate problem, whether we actually trust what's in image or it may have vulnerabilities/backdoors but we either have to trust platform or we have to trust image. Perhaps we should be more specific about the context of "trust" in these discussions and then:
- We trust Build Image Provider to disable all "public" remote access components and services in the image (like SSH) while assuming non zero possibility of having vulnerabilities in the image that could be exploited to gain remote access
- We trust Build Platform to have a firewall solution that blocks unexpected network interactions with the image (given the policies defined by tenant)
- We accept risks that malicious actor could gain admin access to the Platform, disable firewall and leverage those vulnerabilities in the image to gain remote access

I am not sure though if these details are worth putting here given the level of details that other threats provide. What do you think?

docs/spec/draft/threats.md
```diff
@@ -553,19 +552,71 @@ transitive closure of the inputs, and the cache entry is itself a SLSA Build L3
...
+side hence infiltrating all build environments that use that image.
+
+*Example 2:* MyPackage is built on Awesome Builder. An adversary gained access to the 
+build image storage and modified image at rest injecting a malicious tool into it. All build 
+environments created after that point were compromised.
+
+</details>
+<details><summary>Unexpected reuse of a build environment<span>(BuildEnv L2+)</span></summary>
+
+*Threat:* Adversary injects malicious code into the target build by being able to run random code in the same build environment
```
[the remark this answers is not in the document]
paveliak: That's a risk too but it could be easily mitigated by deleting secrets at the end of the build. And besides for the open source projects this is not a risk at all, unless they publish artifacts to the registry and malicious code gets access to  those credentials effectively being able to overwrite data in the registry. However quite often those credentials are revoked at the end of the job so even if next job gains access to them they are no longer valid.

A larger security risk comes from being able to access _following_ jobs. Then bad actor can plant an agent that could mangle subsequent builds injecting backdoors (aka Solarwind attack) and it can exfiltrate secrets that are active.

Perhaps it depends on the specific context and target of the attack which case is actually the worst one. Both are bad.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -96,7 +97,32 @@ A typical build environment will go through the following lifecycle:
 
 ### Build environment threats
 
-TODO
+SLSA [Build track] defines requirements for the provenance that is generated for the build artifacts. Trustworthiness of the build provenance largely depends on the trustworthiness of the [build environment] the build runs in. Build track assumes full trust into the [Build Platform] providing no solutions to verify integrity of the build environment. BuildEnv track intends to close this gap.
```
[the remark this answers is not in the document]
paveliak: Will do. Thanks for explaining the sentence per line formatting. I was not aware about it.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -96,7 +97,32 @@ A typical build environment will go through the following lifecycle:
...
+
+Build environment is bootstrapped from a [build image], which is expected to be an artifact of a SLSA build pipeline. Build platform verifies image provenance upon starting up the environment and provides evidence to the tenant.
+
+Bootstrapping the build environment is a complex process, especially at higher SLSA levels. [Build L3] usually requires significant changes to existing build platforms to maintain ephemeral build environments. It is not uncommon for the build platforms to rely on public cloud providers for managing compute resources that power build environments. This in turn might significantly increase attack surface because added build platform dependencies effectively become part of the [TCB].
```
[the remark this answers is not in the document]
paveliak: > why and how the attack surface is altered in a public cloud setting

It mostly comes from the shire size and obscurity of cloud providers. When Build Platform uses specific feature of a cloud provider (e.g. virtual machine IaaS) there is no way to trust a specific part of a cloud provider infra because internal interactions are hidden from the consumer. Disks for those VMs probably come from a blob storage over a network. All of that is managed by its own control plane which uses some kind of IAM to access it. It snowballs pretty fast. Ultimately tenants have to trust the whole cloud provider which is huge.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
deeglaze: Provenance says nothing about quality assurance endorsements for example.

docs/spec/draft/attested-build-env-levels.md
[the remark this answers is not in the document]
paveliak: I think it's both. It definitely alters attack surface due to the shared responsibility model. But what I wanted to emphasize here is that it _expands_ attack surface because cloud providers are massive companies with complex infrastructure that provide limited (if any) abilities to scope the context of the trust and continuously validate it over time when using their services.

If I have a box under my desk I might not be good with installing all the security updates onto it but my hardware setup is minimalistic and transparent (to me). If I move this box into the cloud then it will be patched on time but on the downside there is a massive amount of complex hardware and software and people that make it possible to run my single VM (and myriad of other things). And even if I had a complete visibility into their infrastructure (which I have not) I won't be able to wrap my head around their security posture. And no single person would probably be able to.

Perhaps what I need to add here is literally this?
```
Cloud providers are massive companies with complex infrastructure that provide limited (if any) abilities to scope the context of the trust and continuously validate it over time when using their services.
```

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -96,7 +97,32 @@ A typical build environment will go through the following lifecycle:
...
+<div class="mermaid">
+flowchart LR    
+      BuildImage>Build Image] ---> |L1|BuildPlatform[[Build Platform]]
+      BuildPlatform[[Build Platform]] ---> |L2|ComputeProvider[[Compute Provider]]
+      ComputeProvider[[Compute Provider]] ---> |L3|BuildEnvironment[(Build Environment)]
+</div>
+
+[BuildEnv L1] protects from threats that happened to the build image in between generating it and passing to the Build Platform. This covers cases of unauthorized modifications happening to the image as it is distributed (potentially via untrusted channels).
+
+[BuildEnv L2] level protects from threats occurring at the Build Platform side as it interacts with the Compute Platform for actually creating a build environment. [Control plane] is the only Build Platform component that is considered trusted at L2 as it performs remote attestation of the build environment.
```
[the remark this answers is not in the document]
paveliak: @deeglaze An example threat could be a bad actor gaining admin access to the Build platform. Then they can do various things ranging from bootstrapping a different image (from what tenant expects) to installing additional components into the environments before handing it over to the tenant for the CI job execution.

docs/spec/draft/attested-build-env-levels.md
paveliak: Yes, I used AMD SEV as an umbrella acronym for all generations including SEV-SNP. The link brings to the AMD page that covers them all but I can update it for better visibility

docs/spec/draft/attested-build-env-levels.md
paveliak: @deeglaze which part exactly? Control plane attesting build environment?

We had discussions about explicitly defining attestation service that could be external to the build platform (and which obviously has to be trusted). If we do it then the only other piece that we need to consider are the interfaces between the Control Plane and the Build Environment (i.e. inputs for the Build process).

Would it make things clearer here if we decoupled attestation service from the control plane?

docs/spec/draft/attested-build-env-levels.md
paveliak: > The compute platform (i.e., privileged remote operators and host system software) is considered semi-trusted

@deeglaze How can we be more specific here? CoCo claims to isolate VM from the host interface and "semi-trusted" wording may not be inspiring enough for the people to invest into L3. I am OK with trusting physical access but trusting remote access may contradict CoCo claims. Would it be accurate to say that we need to trust remote privileged operator to the extent of that operator being able to perform side channel attacks to compromise TEE? Or there is more to that?

deeglaze: It's not just side channels that are a problem if you're running a typical guest stack, since there are still virtual firmwares and device drivers that aren't fully hardened to not trust the VMM. I don't think there are many guest environments that have both reached that level of hardening and also have a rich enough environment to run a full build service.

docs/spec/draft/attested-build-env-levels.md
deeglaze: I think that yes the two aspects should be decoupled. There's some actor that will verify the remote attestation of the build environment to establish the L3 claim of the build environment. That actor need not be part of the build process. That L3 quality could be assigned well after build by a relying party that is checking the attestation measurements after the fact. Operationally that's a rather expensive choice to make, but there are certainly folks that want to have the opportunity to check these qualities themselves without an architecture that hides raw measurements.

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -132,15 +132,13 @@ flowchart LR
...
-It requires build running in a trusted execution environment using technologies like [AMD SEV] and [Intel TDX]. 
-The compute platform (i.e., privileged remote operators and host system software) is considered semi-trusted at L3. 
-Side channels can still be utilized to extract information when analyzing information like memory access patterns, and these cannot be ruled out without further integrity protections and assurances on the system software and human access controls. 
-The [SEV-SNP/TDX threat model](https://www.kernel.org/doc/Documentation/security/snp-tdx-threat-model.rst) describes this level of trust reduction through the use of memory encryption, integrity protection, and remote attestation with cryptographic keys that are certified by the hardware manufacturer. 
-TEE technologies are not infallible, so physical human access to hardware is still a risk that is accepted at L3. 
+[BuildEnv L3] protects from threats coming from the Compute Platform by adopting hardware root of trust.
+Boot time integrity is verifiable using the cryptographic evidences signed by a key unique to each hardware device (i.e. CPU chip).
```
deeglaze, line 140:
```suggestion
Boot time integrity is verifiable using the cryptographic evidences signed by a key unique to each hardware device (e.g., CPU chip). 
```

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -132,15 +132,13 @@ flowchart LR
...
-The compute platform (i.e., privileged remote operators and host system software) is considered semi-trusted at L3. 
-Side channels can still be utilized to extract information when analyzing information like memory access patterns, and these cannot be ruled out without further integrity protections and assurances on the system software and human access controls. 
-The [SEV-SNP/TDX threat model](https://www.kernel.org/doc/Documentation/security/snp-tdx-threat-model.rst) describes this level of trust reduction through the use of memory encryption, integrity protection, and remote attestation with cryptographic keys that are certified by the hardware manufacturer. 
-TEE technologies are not infallible, so physical human access to hardware is still a risk that is accepted at L3. 
+[BuildEnv L3] protects from threats coming from the Compute Platform by adopting hardware root of trust.
+Boot time integrity is verifiable using the cryptographic evidences signed by a key unique to each hardware device (i.e. CPU chip). 
+Practically, achieving L3 might require build running in a trusted execution environment using technologies like [AMD SEV-SNP] and [Intel TDX].
```
deeglaze, line 141:
```suggestion
Practically, achieving L3 might require build running in a trusted execution environment using technologies like [AMD SEV-SNP] and [Intel TDX] without dependence on a host-provided virtual TPM device.
```

marcelamelara: Yes, and we should clarify that the VM secure boot requirements from L2 remain nonetheless.

marcelamelara (changes requested):
Thanks for the updates @paveliak ! I mostly have clarifying changes at this point, so I think this is almost ready.

marcelamelara:
@mssun Would be great to get your feedback on this PR!

docs/spec/draft/attested-build-env-levels.md
```diff
@@ -132,13 +132,13 @@ flowchart LR
...
+The Compute platform is fully trusted at this level as it provides a virtual TPM device that performs boot measurements. 
 [Control plane] is the only Build Platform component that is considered trusted at L2 as it performs remote attestation of the build environment.
 
 [BuildEnv L3] protects from threats coming from the Compute Platform by adopting hardware root of trust.
-Boot time integrity is verifiable using the cryptographic evidences signed by a key unique to each hardware device (i.e. CPU chip). 
-Practically, achieving L3 might require build running in a trusted execution environment using technologies like [AMD SEV-SNP] and [Intel TDX]. 
+Boot time integrity is verifiable using the cryptographic evidences signed by a key unique to each hardware device (e.g., CPU chip). 
+Practically, achieving L3 requires Compute Providers supporting remote attestation of the hypervisor stack or having the build running in a trusted execution environment using technologies like [AMD SEV-SNP] and [Intel TDX] without dependence on a host-provided virtual TPM device.
```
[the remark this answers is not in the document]
paveliak: > Who is the remote party that seeks to attest the hypervisor stack

@marcelamelara that would be Control Plane. Next paragraph mentions that `Control Plane is considered trusted at L2 and L3 because it _verifies_ the remote attestation of the build environment`

Issue #1213, opened 2024-10-21
Rephrase "The update did not match the code submitted to GitHub"?


TomHennen:
>The update did not match the code submitted to GitHub

this phrase is used a few times here, but I'm not sure what it means. 
I think it has to mean basically "use of compromised dependency."
IE, the revision consumed was not the one provided in the build inputs, but I think that's not very clear from this sentence. 

I recommend replacing all usage with: "The update used unauthorized build inputs."

_Originally posted by @zachariahcox in https://github.com/slsa-framework/slsa/pull/1209#discussion_r1806499521_

TomHennen:
From what I can tell this phrase is only used once in "Known example" text for "Use compromised dependency".  So it's referring to a specific event (the [event-stream](https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/) attack).

In that attack the idea is that the maintainer had a package that _purported_ to be from GitHub repo X, but uploaded a package that wasn't from repo X.  Since there wasn't any SLSA verification in place, I don't think it's correct to say the update used unauthorized build inputs.

Perhaps "The updated binary was not built from the purported source code"?

zachariahcox:
update: we have a few usages ([query](https://github.com/search?q=repo%3Aslsa-framework%2Fslsa+%22The+update+did+not+match+the+code%22&type=code)), only two of which need to be addressed: 

- [ ] https://github.com/slsa-framework/slsa/blob/f28c3deb3da02017430ebd5dbc510ef92c3509a2/docs/spec/draft/threats-overview.md?plain=1#L98
- [ ] https://github.com/slsa-framework/slsa/blob/f28c3deb3da02017430ebd5dbc510ef92c3509a2/docs/spec/v1.1/threats-overview.md?plain=1#L70

PR #1292, opened 2025-03-17
content: simplify description of event-stream threat

zachariahcox:
fixes https://github.com/slsa-framework/slsa/issues/1213

Proposing we remove the detail relating to a mismatch between the uploaded package and its purported source revisions. 
That is _definitely bad_ but the primary threat here is that the producer changed their intent. 

This patch only covers v1.1 and vdraft.

### DIFF
--- a/docs/spec/draft/threats-overview.md
+++ b/docs/spec/draft/threats-overview.md
@@ -98,7 +98,7 @@ For close source software SLSA does not provide any solutions for malicious prod
 <tr>
 <td>N/A
 <td>Dependency threats (i.e. A-H, recursively)
-<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior.
+<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the binary version of the dependency to add malicious behavior without updating the corresponding source code.
 <td>Applying SLSA recursively to all dependencies would prevent this particular vector, because the provenance would indicate that it either wasn't built from a proper builder or that the source did not come from GitHub.
 </table>
 
--- a/docs/spec/draft/threats-overview.md
+++ b/docs/spec/draft/threats-overview.md
@@ -98,8 +98,8 @@ For close source software SLSA does not provide any solutions for malicious prod
 <tr>
 <td>N/A
 <td>Dependency threats (i.e. A-H, recursively)
-<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the binary version of the dependency to add malicious behavior without updating the corresponding source code.
-<td>Applying SLSA recursively to all dependencies would prevent this particular vector, because the provenance would indicate that it either wasn't built from a proper builder or that the source did not come from GitHub.
+<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker controls an innocuous dependency and publishes a malicious binary version without a corresponding update to the source code.
+<td>Applying SLSA recursively to all dependencies would prevent this particular vector, because the provenance would indicate that it either wasn't built from a proper builder or that the binary did not match the source.
 </table>
 
 <table>
--- a/docs/spec/draft/threats-overview.md
+++ b/docs/spec/draft/threats-overview.md
@@ -98,7 +98,7 @@ For close source software SLSA does not provide any solutions for malicious prod
 <tr>
 <td>N/A
 <td>Dependency threats (i.e. A-H, recursively)
-<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior. The update did not match the code submitted to GitHub (i.e. attack F).
+<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior.
 <td>Applying SLSA recursively to all dependencies would prevent this particular vector, because the provenance would indicate that it either wasn't built from a proper builder or that the source did not come from GitHub.
 </table>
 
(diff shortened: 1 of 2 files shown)


docs/spec/draft/threats-overview.md
```diff
@@ -98,7 +98,7 @@ For close source software SLSA does not provide any solutions for malicious prod
 <tr>
 <td>N/A
 <td>Dependency threats (i.e. A-H, recursively)
-<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior. The update did not match the code submitted to GitHub (i.e. attack F).
+<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior.
```
TomHennen, line 101: I don't think this change properly motivates the "How SLSA could help".  Applying SLSA recursively wouldn't necessarily help if the malicious change was submitted to GitHub.

Maybe...

```suggestion
<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the binary version of the dependency to add malicious behavior without updating the corresponding source code.
```

docs/spec/v1.1/threats-overview.md
```diff
@@ -67,7 +67,7 @@ Many recent high-profile attacks were consequences of supply chain integrity vul
 <tr>
 <td>D
 <td>Use compromised dependency (i.e. A-H, recursively)
-<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior. The update did not match the code submitted to GitHub (i.e. attack F).
+<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior.
```
TomHennen, line 70:
```suggestion
<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the binary version of the dependency to add malicious behavior without updating the corresponding source code.
```

TomHennen (changes requested):

docs/spec/draft/threats-overview.md
zachariahcox: Oo, I like that take. IE: SLSA doesn't actually help if your trusted publisher decides to attack you, but it can make forensics easier.

I guess we're saying that high-level SLSA Provenance techniques could make it clear that the package was built from different source?
The verification step would be: does the attested provenance agree with claimed dependency graph?

zachariahcox: either way, this change seems better than what was there before, so accepted!

TomHennen (approved):
LGTM!

docs/spec/draft/threats-overview.md
```diff
@@ -98,7 +98,7 @@ For close source software SLSA does not provide any solutions for malicious prod
 <tr>
 <td>N/A
 <td>Dependency threats (i.e. A-H, recursively)
-<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the dependency to add malicious behavior. The update did not match the code submitted to GitHub (i.e. attack F).
+<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker added an innocuous dependency and then later updated the binary version of the dependency to add malicious behavior without updating the corresponding source code.
```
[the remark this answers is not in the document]
TomHennen: It's like this:

1. Attacker has control of innocuous npm library 'foo', with the source stored in GitHub.com/attacker/foo
2. The latest version source version of foo is v1.0.0, which matches exactly the version published to npm that's also labeled v1.0.0.
3. Attacker proposes that innocent library 'bar' take a dependency on 'foo'.
4. 'bar' takes the dependency on foo. It solves some problem they have and seems fine.
5. Attacker updates GitHub.com/attacker/foo to add some basic, innocuous, fix, updates the GitHub tag to v1.0.1.
6. Attacker creates a malicious 'binary' version of foo and uploads it directly to npm as v1.0.1.
7. Dependabot (or the attacker) proposes that 'bar' update its dependency to v1.0.1
8. When bar is built it takes the malicious binary version of foo from npm.

So, I guess, if I'm reading you correctly, it means it uploads the malicious version to the package hosting location.

Any suggestions on a better way to express that?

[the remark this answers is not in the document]
arewm: So this is an xz-style attack in a way? That flow makes sense. It sounded like the content was text was implying that v1.0.0 was re-uploaded with a malicious binary version. 

I think it will be challenging to clarify this without adding too many words. Is it critical to say 
> Attacker added an innocuous dependency and then later updated

Maybe some wording like this?
```suggestion
<td><a href="https://web.archive.org/web/20210909051737/https://schneider.dev/blog/event-stream-vulnerability-explained/">event-stream</a>: Attacker controls an innocuous dependency and publishes a malicious binary version without a corresponding update to the source code.
```

It doesn't matter how the attacker controls the dependency nor does it necessarily matter who depends on this artifact. The key part of this is the ability for an attacker to insert itself into a dependency's supply chain to break the direct link from source to binary.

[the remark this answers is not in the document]
TomHennen: > It sounded like the content was text was implying that v1.0.0 was re-uploaded with a malicious binary version.

I think that could be too.  Probably shouldn't be excluded.  I left it out of my example because some package managers don't allow that (but some do!).

I like the suggestion!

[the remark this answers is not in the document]
marcelamelara: +1 To @arewm's suggestion. I was also having a hard time clearly seeing the connection between the original text and the actual event-stream attack.

docs/spec/draft/threats-overview.md
```diff
@@ -98,7 +98,7 @@ For close source software SLSA does not provide any solutions for malicious prod
...
 <td>Applying SLSA recursively to all dependencies would prevent this particular vector, because the provenance would indicate that it either wasn't built from a proper builder or that the source did not come from GitHub.
```
marcelamelara, line 102: Given the updated text per [Andrew's suggestion](https://github.com/slsa-framework/slsa/pull/1292/files#r2007672262), I think we may also want to clarify this explanation.
```suggestion
<td>Applying SLSA recursively to all dependencies would prevent this particular vector, because the provenance would indicate that it either wasn't built from a proper builder or that the binary did not match the source on GitHub.
```

TomHennen: yes, important distinction!

TomHennen:
LGTM, thanks!

marcelamelara (approved):
LGTM thank you!

arewm (approved):

Issue #1303, opened 2025-03-26
Build threats - "Poison the build cache" doesn't have a 'solution'


TomHennen:
The threat lists a mitigation, and has an example, but the example doesn't include a solution or a conclusion of any sort.

https://deploy-preview-1298--slsa.netlify.app/spec/v1.1-rc2/threats#c-source-code-management:~:text=Poison%20the%20build%20cache

TomHennen:
I think the 'example' might also be a bit wonky, or doesn't have enough data.

If the cache is keyed by the digest of the source file, what are the contents of that file?  How does the attacker poison it?

Maybe the idea is that _other_ inputs besides the source file went into that content and that wasn't taken into consideration when constructing the cache.  If so we should say that.

Alternatively we _remove_ a mention of the key and just say the solution is to use separate caches.

TomHennen:
Going to keep it simple and just say builders should always use a separate caches for each build.

I suppose an outstanding question might be "what good is the cache then?"

IDK.  Not sure we'll resolve it here, but maybe @MarkLodato has some thoughts?

MarkLodato:
I agree that the current wording is poor. Sorry this post is so long - I didn't have time to make it shorter 😄.

For a long writeup of build cache poisoning, see https://adnanthekhan.com/2024/05/06/the-monsters-in-your-build-cache-github-actions-cache-poisoning/. This is specifically about GitHub's [actions/cache](https://github.com/actions/cache) but the same concepts apply to other build caches.

I suppose there are two classes of vulnerabilities, **both** of which most build caches exhibit:

#### Issue 1: Hash Collisions

**Problem:** The cache key does not fully and securely cover the transitive closure of all inputs. For example, suppose you want to cache a single `.cc` --> `.o` compilation. If you just cache by the hash of the source `.cc` file but *not* the compiler command line or the hash of the compiler, then two different builds with different command lines or compiler versions will use the same hash key.

While this is more of a correctness issue, it is still a vulnerability that can be actively exploited in some situations. The biggest issue is persistence, which is what [cacheract](https://github.com/adnaneKhan/cacheract) shows. Once someone compromises a single build, they can gain persistence and stay basically forever.

[...]

PR #1310, opened 2025-03-26
content: draft: address issues in threats.md

### LINKED ISSUE #1302: Build threats - "Compromise project owner" doesn't have a 'solution'
The threat lists a mitigation, and has an example, but the example doesn't include a solution or a conclusion of any sort.

https://deploy-preview-1298--slsa.netlify.app/spec/v1.1-rc2/threats#c-source-code-management:~:text=Compromise%20project%20owner

That should be addressed.

### LINKED ISSUE #1306: "Delete the code" threat has a solution that doesn't exist in 1.1
"Solution: Verifier rejects the package because it lacks a positive attestation showing that some system, such as GitHub, ensured retention and availability of the source code."

This solution isn't part of SLSA 1.1

https://deploy-preview-1298--slsa.netlify.app/spec/v1.1-rc2/threats#availability-threats:~:text=Solution%3A%20Verifier%20rejects%20the%20package%20because%20it%20lacks%20a%20positive%20attestation%20showing%20that%20some%20system%2C%20such%20as%20GitHub%2C%20ensured%20retention%20and%20availability%20of%20the%20source%20code.

We should update.

### LINKED ISSUE #1307: "Forge change metadata" likely belongs in "Source threats" not "Verification threats"
The text doesn't directly discuss verification or expectations so it seems misplaced.

https://deploy-preview-1298--slsa.netlify.app/spec/v1.1-rc2/threats#availability-threats:~:text=Forge%20change%20metadata

### LINKED ISSUE #1308: Example in "Exploit cryptographic hash collisions" is source specific
The text is generally fine, but it's talking about hash collisions in source files, which aren't part of how SLSA verifies artifacts.  Elsewhere when we talk about verification we're talking about verification of artifacts to their provenance and that isn't referenced here.

We should update to talk about binary files and how a collision impacts provenance verification.

TomHennen:
This change addresses multiple issues in threats.md.

* Don't let an attacker debug an in progress build. Fixes #1302 
* Use a separate cache for each build. refs #1303
  * Leaves this issue open for further work.
* Don't reference source attestations in threats yet. Fixes #1306 
* Move "forge change metadata" to "Source Threats". Fixes #1307 
* Use an executable as a hash collision example instead of source file. Fixes #1308

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -987,13 +987,12 @@ two-party review.
 *Threat:* Exploit a cryptographic hash collision weakness to bypass one of the
 other controls.
 
-*Mitigation:* Require cryptographically secure hash functions for commit
-IDs and provenance subjects, such as SHA-256.
+*Mitigation:* Choose secure algorithms when using cryptographic digests, such
+as SHA-256.
 
-*Examples:* Attacker crafts a malicious executable with the same MD5 hash as a
-target benign executable. Attacker replaces the benign executable with
-the malicious executable. Solution: Only accept cryptographic hashes with strong
-collision resistance.
+*Examples:* Attacker crafts a malicious file with the same MD5 hash as a target
+benign file. Attacker replaces the benign file with the malicious file.
+Solution: Only accept cryptographic hashes with strong collision resistance.
 
 </details>
 
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -263,13 +263,7 @@ stamping."
 *Threat:* Forge the change metadata to alter attribution, timestamp, or
 discoverability of a change.
 
-*Mitigation:* Source control platform strongly authenticates actor identity,
-timestamp, and parent revisions.
-
-*Example:* Adversary submits a git commit with a falsified author and timestamp,
-and then rewrites history with a non-fast-forward update to make it appear to
-have been made long ago. Solution: Consumer detects this by seeing that such
-changes are not strongly authenticated and thus not trustworthy.
+*Mitigation:* This threat is not currently addressed by SLSA.
 
 </details>
 
@@ -916,7 +910,7 @@ machine when Dep is loaded at runtime. An end user installs MyPackage, which in
 turn installs the compromised version of Dep. When the user runs MyPackage, it
 loads and executes the malicious code from Dep.
 
-*Mitigation:* N/A - This threat is not currently addressed by SLSA. SLSA's
+*Mitigation:* N/A - SLSA's
 threat model does not explicitly model runtime dependencies. Instead, each
 runtime dependency is considered a distinct artifact with its own threats.
 
@@ -939,15 +933,7 @@ SLSA v1.0 does not address availability threats, though future versions might.
 revision or cause it to get garbage collected, preventing anyone from inspecting
 the code.
 
-*Mitigation:* Some system retains the revision and its version control history,
-making it available for inspection indefinitely. Users cannot delete the
-revision except as part of a transparent legal or privacy process.
-
-*Example:* An adversary submits malicious code to the MyPackage GitHub repo,
-builds from that revision, then does a force push to erase that revision from
-history or deletes the entire repo. This would make the revision
-unavailable for inspection. Solution: Verifier rejects packages built from
-source repos that do not have strong retention guarantees.
+*Mitigation:* This threat is not currently addressed by SLSA.
 
 </details>
 <details><summary>A dependency becomes temporarily or permanently unavailable to the build process</summary>
@@ -963,14 +949,14 @@ impact of this threat.
 
 *Threat:* The package registry stops serving the artifact.
 
-*Mitigation:* N/A - This threat is not currently addressed by SLSA.
+*Mitigation:* This threat is not currently addressed by SLSA.
 
 </details>
 <details><summary>De-list provenance</summary>
 
 *Threat:* The package registry stops serving the provenance.
 
-*Mitigation:* N/A - This threat is not currently addressed by SLSA.
+*Mitigation:* This threat is not currently addressed by SLSA.
 
 </details>
 
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -472,11 +472,12 @@ forged digest is irrelevant.
 build process or provenance generation.
 
 *Example:* MyPackage is built on Awesome Builder under the project "mypackage".
-Adversary is an administrator of the "mypackage" project. Awesome Builder allows
-administrators to debug build machines via SSH. An adversary uses this feature
-to alter a build in progress. Solution: Build L3 requires provenance to be
-resistant to forgery by tenants, so the attacker would be forbidden from
-altering a build in progress.
+Adversary is an owner of the "mypackage" project. Awesome Builder allows
+owners to debug the build environment via SSH. An adversary uses this feature
+to alter a build in progress. Solution: Build L3 requires complete provenance, the
+attackers access and/or actions within the SSH would be enumerated within the
+external parameters. The updated external parameters will not match the declared
+expectations causing verification to fail.
 
 </details>
 <details><summary>Compromise other build <span>(Build L3)</span></summary>
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -463,7 +463,9 @@ build process or provenance generation.
 *Example:* MyPackage is built on Awesome Builder under the project "mypackage".
 Adversary is an administrator of the "mypackage" project. Awesome Builder allows
 administrators to debug build machines via SSH. An adversary uses this feature
-to alter a build in progress.
+to alter a build in progress. Solution: Build L3 requires provenance to be
+resistant to forgery by tenants, so the attacker would be forbidden from
+altering a build in progress.
 
 </details>
 <details><summary>Compromise other build <span>(Build L3)</span></summary>
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -517,6 +517,7 @@ such cache poisoning attacks.
 the source file. Adversary runs a malicious build that creates a "poisoned"
 cache entry with a falsified key, meaning that the value wasn't really produced
 from that source. A subsequent build then picks up that poisoned cache entry.
+Solution: Builder uses a separate cache for each build.
 
 </details>
 <details><summary>Compromise build platform admin <span>(verification)</span></summary>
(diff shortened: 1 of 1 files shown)


docs/spec/draft/threats.md
```diff
@@ -256,6 +256,23 @@ stamping."
...
 </details>
 
+#### (B4) Render logs ineffective
+
+<details><summary>Forge change metadata</summary>
+
+*Threat:* Forge the change metadata to alter attribution, timestamp, or
+discoverability of a change.
+
+*Mitigation:* Source control platform strongly authenticates actor identity,
```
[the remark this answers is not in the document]
zachariahcox: I recommend dropping this item, saying that slsa doesn't support until a future track, or mentioning that consumers have to have their own way to meaningfully identify contributors?

docs/spec/draft/threats.md
```diff
@@ -515,6 +534,7 @@ such cache poisoning attacks.
 the source file. Adversary runs a malicious build that creates a "poisoned"
 cache entry with a falsified key, meaning that the value wasn't really produced
 from that source. A subsequent build then picks up that poisoned cache entry.
+Solution: Builder uses a separate cache for each build.
```
zachariahcox, line 537:
```suggestion
Solution: Builder.id uses a separate cache for each build.
```
Right? there's no way for actions as a product to use a separate cache for all builds, but you might want something like that for specific configurations.

zachariahcox: potentially in this case, the solution is: "for critical builds, do not use a cache."

docs/spec/draft/threats.md
```diff
@@ -926,9 +946,8 @@ revision except as part of a transparent legal or privacy process.
 *Example:* An adversary submits malicious code to the MyPackage GitHub repo,
 builds from that revision, then does a force push to erase that revision from
 history (or requests that GitHub delete the repo.) This would make the revision
```
zachariahcox, line 948:
```suggestion
history deletes the repo. This would make the revision
```

docs/spec/draft/threats.md
```diff
@@ -926,9 +946,8 @@ revision except as part of a transparent legal or privacy process.
...
-unavailable for inspection. Solution: Verifier rejects the package because it
-lacks a positive attestation showing that some system, such as GitHub, ensured
-retention and availability of the source code.
+unavailable for inspection. Solution: Verifier rejects packages built from
+source repos that do not have strong retention guarantees.
```
zachariahcox, line 950: who makes the guarantee in this case?

we could also say "verifier rejects the package because it lacks positive attestations that consumed source is still available."

docs/spec/draft/threats.md
```diff
@@ -999,10 +1004,9 @@ other controls.
 *Mitigation:* Require cryptographically secure hash functions for commit
 checksums and provenance subjects, such as SHA-256.
```
zachariahcox, line 1005:
```suggestion
IDs and provenance subjects, such as SHA-256.
```

docs/spec/draft/threats.md
TomHennen: I'm trying to get away from talking about attestations in the source threats section because those attestations don't exist yet (they're part of the source track).

> "who makes the guarantee in this case?"

This would be the source control platform.  I don't know how many actually do this.

I'd also be happy saying SLSA doesn't address this threat yet, which would be simpler.

Note that I don't think _anything_ we're planning in the source track would prevent a repo owner from deleting that repo.  Whenever we address these things we'd probably want to provide separate examples/mitigations.

docs/spec/draft/threats.md
TomHennen: Builder.id isn't a component in the [Build model](https://slsa.dev/spec/draft/terminology#build-model), but it does [_refer to_](https://slsa.dev/spec/v1.0/provenance#builder) the 'Build Platform'.  In this doc we're always using 'Builder' as a shorthand for 'Build Platform'.

In the case of GitHub Actions, I don't think we'd consider GitHub Actions the "build platform" instead it's a component of that platform.  Much like for an [SCS](https://slsa.dev/spec/draft/source-requirements#definitions:~:text=Source%20Control%20System%20(SCS)) the system is a collection of tools and services that people build on top of to meet the requirements.  For GHA and SLSA builds it's really the specific workflow that brings everything together and I think _it_ can decide how caches are or aren't used?  So I think we're fine?

> potentially in this case, the solution is: "for critical builds, do not use a cache."

I'd rather not start talking about 'critical builds'.  Everywhere else in this page we just talk about 'builds'

docs/spec/draft/threats.md
```diff
@@ -999,10 +1004,9 @@ other controls.
...
 
-*Examples:* Construct a benign file and a malicious file with the same SHA-1
-hash. Get the benign file reviewed and then submit the malicious file.
-Alternatively, get the benign file reviewed and submitted and then build from
-the malicious file. Solution: Only accept cryptographic hashes with strong
+*Examples:* Attacker crafts a malicious executable with the same MD5 hash as a
```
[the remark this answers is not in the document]
TomHennen: I was just trying to make the example as meaningful as possible.  I can replace executable with file though. That's fine.

Generally I think commit IDs might be misplaced here.  Why are they special?  Even once we have the source track we'll still really just care about what's in the provenance.

Maybe we just drop 'commit IDs' and only refer to provenance subjects?

Another potential issue (that I'm not sure I want to get into) is that sometimes there are [use cases](https://github.com/in-toto/attestation/blob/main/spec/v1/digest_set.md#use-cases-for-non-cryptographic-immutable-digests) for non-cryptographic identifiers in digests and this would seem to preclude that (but not really since this isn't the requirements).  If we wanted to clarify _that_ we'd have to say something like "*Mitigation:* Choose **secure** algorithms when using cryptographic digests"

docs/spec/draft/threats.md
adityasaky: > I'd also be happy saying SLSA doesn't address this threat yet, which would be simpler.

I support this approach.

docs/spec/draft/threats.md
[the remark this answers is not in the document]
adityasaky: > Maybe we just drop 'commit IDs' and only refer to provenance subjects?

I'd drop commit IDs from here. I know this is "only" the threats page and not a requirement for a track, but I wonder if some folks will get confused and think that SLSA requires them to start using git with SHA-256 identifiers.

> Generally I think commit IDs might be misplaced here. Why are they special? Even once we have the source track we'll still really just care about what's in the provenance.

I think the source provenance or VSA subject might still refer to a revision using `gitCommit` or so as the digest type, which ends up being SHA-1? (At least that's what we're doing for the VSA PoC in gittuf)

docs/spec/draft/threats.md
TomHennen: Done.

docs/spec/draft/threats.md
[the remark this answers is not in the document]
TomHennen: Done.  Now just saying SLSA doesn't address this yet.

docs/spec/draft/threats.md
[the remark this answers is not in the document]
TomHennen: Yes it does, that's a good point!

docs/spec/draft/threats.md
```diff
@@ -256,6 +256,17 @@ stamping."
 
 </details>
 
+#### (B4) Render logs ineffective
+
+<details><summary>Forge change metadata</summary>
```
arewm, line 261: Can we change the heading to be something more representative of the summary? I assume that this is logs for incident response, etc. but to me, simple statements of logs are often build logs.

docs/spec/draft/threats.md
```diff
@@ -463,7 +474,9 @@ build process or provenance generation.
 *Example:* MyPackage is built on Awesome Builder under the project "mypackage".
 Adversary is an administrator of the "mypackage" project. Awesome Builder allows
 administrators to debug build machines via SSH. An adversary uses this feature
-to alter a build in progress.
+to alter a build in progress. Solution: Build L3 requires provenance to be
+resistant to forgery by tenants, so the attacker would be forbidden from
+altering a build in progress.
```
arewm, line 479: Altering a build is different from forging provenance. This is also not an action by a tenant but an admin. I initially thought that this related to isolation, but that is more scoped for preventing parallel builds from affecting each other.

I feel like this falls more into the "future" operations track. How can a build platform be run securely so that admin users have the least possible impact.

docs/spec/draft/threats.md
TomHennen: Do you have a suggestion?

docs/spec/draft/threats.md
TomHennen: So, now that you call this out I think the example differs from the 'threat' and 'mitigation'.

The threat and mitigation are talking about an attacker gaining ownership of a _project_.  The example then changes terminology to talk about 'administrator' access which makes us think it's talking about administration of the entire build platform, but it is really just supposed to be talking about _the project itself_.

Re "forging", the need for the provenance to be complete comes from the high level "Provenance is Unforgeable" requirement.  Let me replace that with something more specific.

So I think we really want something like

> *Example:* MyPackage is built on Awesome Builder under the project "mypackage".
Adversary is an _owner_ of the "mypackage" project. Awesome Builder allows
_owners_ to debug the _build environment_ via SSH. An adversary uses this feature
to alter a build in progress. Solution: Build L3 requires _complete_ provenance, the
attackers access and/or actions within the SSH would be enumerated within the
external parameters.  The updated external parameters will not match the declared
expectations causing verification to fail.


I'd agree that any attacks by the Build Platform administrator would definitely fall under some new tracks (like Build Environment).

SG?

docs/spec/draft/threats.md
arewm:
```
#### (B4) Render commit metadata
```
?

docs/spec/draft/threats.md
TomHennen: Fixed.  Thanks for the catch.

docs/spec/draft/threats.md
TomHennen: How about "Render change metadata ineffective"

"commit" -> "change" to stay consistent (and somewhat platform agnostic)

TomHennen: Done.  Let me know if you don't like the new thing. :)

arewm (approved):
Thanks for this!

TomHennen:
Leaving the cache poisoning issue open to address Mark's comment on the #1303 in a separate PR.

zachariahcox (approved):
Woohoo!

PR #1311, opened 2025-03-26
content: draft: Update 'unreviewable change' to reference a JPEG

### LINKED ISSUE #1300: Example in "Submit a change that is unreviewable" could be better.
It currently says "A proposed change updates a file, but the reviewer is only presented with a diff of the cryptographic hash, not of the file contents. Thus, the reviewer does not have enough context to provide a meaningful review. Solution: the code review system should present the reviewer with a content diff or some other information to make an informed decision."

I don't think this is sufficiently motivated.  _Why_ is the tool only showing the cryptographic hash?

Instead I'd like to update this with something like 

"A proposed change updates a JPEG file, but the reviewer is only presented with a diff of the binary file contents.  The reviewer is unable to parse the contents themselves so they do not have enough context to provide a meaningful review. Solution: the code review system should present the reviewer with a rendering of the image and the embedded metadata, allowing them to make an informed decision."

TomHennen:
This provides a more clear example of how a change might not be reviewable.

This use case is more common and seems like it is easier to understand vs why
a code review system would just display the cryptographic hash of a file.

fixes #1300

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -197,8 +197,9 @@ message, but the reviewer is only presented with a diff of the binary
 file contents. The reviewer is unable to parse the contents themselves
 so they do not have enough context to provide a meaningful review.
 Solution: the code review system should present the reviewer with a
-rendering of the image and the embedded metadata, allowing them to make
-an informed decision.
+rendering of the image and the [embedded
+metadata](https://en.wikipedia.org/wiki/Exif), allowing them to make an
+informed decision.
 
 </details>
 <details><summary>Copy a reviewed change to another context</summary>
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -192,11 +192,13 @@ benign but is actually malicious.
 *Mitigation:* Code review system ensures that all reviews are informed and
 meaningful.
 
-*Example:* A proposed change updates a file, but the reviewer is only presented
-with a diff of the cryptographic hash, not of the file contents. Thus, the
-reviewer does not have enough context to provide a meaningful review.
-Solution: the code review system should present the reviewer with a content diff or some
-other information to make an informed decision.
+*Example:* A proposed change updates a JPEG file to include a malicious
+message, but the reviewer is only presented with a diff of the binary
+file contents. The reviewer is unable to parse the contents themselves
+so they do not have enough context to provide a meaningful review.
+Solution: the code review system should present the reviewer with a
+rendering of the image and the embedded metadata, allowing them to make
+an informed decision.
 
 </details>
 <details><summary>Copy a reviewed change to another context</summary>
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -190,7 +190,8 @@ They may:
 benign but is actually malicious.
 
 *Mitigation:* Code review system ensures that all reviews are informed and
-meaningful.
+meaningful to the extent possible. For example the system could show
+& resolve symlinks, render images, or verify & display provenance.
 
 *Example:* A proposed change updates a JPEG file to include a malicious
 message, but the reviewer is only presented with a diff of the binary


arewm (approved):

docs/spec/draft/threats.md
```diff
@@ -192,11 +192,13 @@ benign but is actually malicious.
...
-with a diff of the cryptographic hash, not of the file contents. Thus, the
-reviewer does not have enough context to provide a meaningful review.
-Solution: the code review system should present the reviewer with a content diff or some
-other information to make an informed decision.
+*Example:* A proposed change updates a JPEG file to include a malicious
+message, but the reviewer is only presented with a diff of the binary
+file contents. The reviewer is unable to parse the contents themselves
+so they do not have enough context to provide a meaningful review.
+Solution: the code review system should present the reviewer with a
+rendering of the image and the embedded metadata, allowing them to make
```
adityasaky, line 200: I don't mean to nitpick but I'm not sure what "embedded metadata" would entail here.

adityasaky (approved):
My question aside, I think this is a good change!

docs/spec/draft/threats.md
TomHennen: https://en.wikipedia.org/wiki/Exif.  Maybe I should link to it.

TomHennen: Done

adityasaky: Thanks!

marcelamelara (approved):
LGTM thanks!

docs/spec/draft/threats.md
```diff
@@ -192,11 +192,14 @@ benign but is actually malicious.
...
-with a diff of the cryptographic hash, not of the file contents. Thus, the
-reviewer does not have enough context to provide a meaningful review.
-Solution: the code review system should present the reviewer with a content diff or some
-other information to make an informed decision.
+*Example:* A proposed change updates a JPEG file to include a malicious
+message, but the reviewer is only presented with a diff of the binary
+file contents. The reviewer is unable to parse the contents themselves
+so they do not have enough context to provide a meaningful review.
+Solution: the code review system should present the reviewer with a
+rendering of the image and the [embedded
```
trishankatdatadog, line 200: This is going to be tricky if we think about all sorts of binaries, no? For example, how would the opaque XZ Utils [backdoor](https://www.akamai.com/blog/security-research/critical-linux-backdoor-xz-utils-discovered-what-to-know) (a shared object hidden inside a test file) have been caught here? Also, rendering images themselves is a tricky business which could open the renderer to RCEs. I think so long as we note these practical limitations, we'll be OK.

TomHennen: I can add something to that effect if you'd like.

In the end this is just an example meant to get people thinking along the right lines and thinking about what's better/worse.  I don't think anyone would expect a user to parse a jpeg themselves looking for all these things, but the tool could render it for them.  For binary files that are the result of a build a solution could be 1) verify and show the user the provenance for the binary, and then they can decide if that seems right. or 2) (better) have a policy for that path that checks the provenance itself and says "hey the policy verification passed, this is fine, click for more details if you want".

TomHennen: Added some details to 'mitigation'.

arewm: The `xz` attack certainly had some opaque data present, but one of the key pieces of that attack is that the artifact was produced in someone's local environment that had a modified version of the `build-to-host.m4` file. This is the file that started the chain of events leading to the exploit. The artifact was neither built in an isolated environment nor did it have any sort of provenance.

SLSA is not the solution for every attack. If someone is able to create a JPEG that contains malicious payload while seeming completely innocuous to reviewers, there isn't _much_ that can be done. Even though I am not sure where this "malicious message" would be included in the JPEG, the point of this is that systems of review should give the participating parties as much actionable information as reasonably possible to enable informed decisions about the proposed revisions.

trishankatdatadog (approved):

lehors (approved):

PR #1312, opened 2025-03-27
content: draft: Clarify cache poisoning example

### LINKED ISSUE #1303: Build threats - "Poison the build cache" doesn't have a 'solution'

TomHennen:
Provides more detailed examples of cache poisoning leveraging @MarkLodato's helpful comment in #1303.

Also link to a detailed real-world example.

Fixes #1303.

### DIFF
--- a/docs/spec/draft/threats.md
+++ b/docs/spec/draft/threats.md
@@ -525,27 +525,27 @@ key.
 benign build process ([example][build-cache-poisoning-example]).
 
 *Mitigation:* Build caches must be [isolated][isolated] between builds to prevent
-such cache poisoning attacks.
+such cache poisoning attacks. In particular, the cache SHOULD be keyed by the
+transitive closure of all inputs to the cached artifact, and the cache must
+either be only writable by the trusted control plane or have SLSA Build L3
+provenance for each cache entry.
 
 *Example 1:* The cache key does not fully cover the transitive closure of all
 inputs and instead only uses the digest of the source file itself. Adversary runs
 a build over `auth.cc` with command line flags to gcc that define a marco
 replacing `CheckAuth(ctx)` with `true`. When subsequent builds build `auth.cc`
 they will get the attacker's poisoned instance that does not call `CheckAuth`.
-Solution: SLSA Build Level 3 requires build cache's to be isolated and a properly
-isolated build cache digest must cover the *transitive closure* of all inputs to
-the cached artifact.
+Solution: Build cache is keyed by digest of auth.cc, command line, and digest of
+compiler so changing the command line flags results in a different cache entry.
 
 *Example 2:* The tenant controlled build process has full write access to the
 cache. Adversary observes a legitimate build of `auth.cc` which covers the
 transitive closure of all inputs and notes the digest used for caching. The
 adversary builds a malicious version of `auth.o` and directly writes it to the
 build cache using the observed digest. Subsequent legitimate builds will use
-the malicious version of `auth.o`. Solution: SLSA Build Level 3 requires build
-caches to be isolated. A properly isolated build cache can be guaranteed to
-be created by the trusted control plane either by restricting write access to
-the cache or ensuring the provenance of the digest used to key the cache
-is not directly influenced by the tenants.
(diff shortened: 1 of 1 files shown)


docs/spec/draft/threats.md
```diff
@@ -522,16 +522,30 @@ key.
...
+*Example 2:* The tenant controlled build process has full write access to the
+cache. Adversary observes a legitimate build of `auth.cc` which covers the
+transitive closure of all inputs and notes the digest used for caching. The
+adversary builds a malicious version of `auth.o` and directly writes it to the
+build cache using the observed digest. Subsequent legitimate builds will use
+the malicious version of `auth.o`. Solution: SLSA Build Level 3 requires build
+caches to be isolated and a properly isolated build cache only allows the
+trusted control plane to have write access to the cache and ensures the digest
+used to key the cache comes from the build inputs and is not directly influenced
+by the tenants.
```
arewm, line 548: It isn't strictly necessary for the cache to only be writeable by the build platform if we can depend on the provenance of the cache. While I don't think that the spec is currently written that way today, we can have confidence in the cache if we can have confidence in the fact that the cache was produced by the build platform. This can take the form of ensuring that only the build platform has the ability to write to the cache or by treating the cache as a byproduct which has its own provenance. In this latter model, it would effectively be recursive SLSA (so we might defer it for now?). If the cache has provenance, ideally we would be verifying it regardless of whether the builder is the only one with write access.

This is the direction of a potential change...
```suggestion
the malicious version of `auth.o`. Solution: SLSA Build Level 3 requires build
caches to be isolated. A properly isolated build cache can be guaranteed to 
be created by the trusted control plane either by restricting write access to
the cache and ensures the provenance of the digest used to key the cache
is not directly influenced by the tenants.
```

TomHennen: "and ensures..." -> "or by ensuring..."

Earlier it says "either" but we don't have an 'or', just an 'and'.

docs/spec/draft/threats.md
```diff
@@ -522,16 +522,30 @@ key.
...
-the source file. Adversary runs a malicious build that creates a "poisoned"
-cache entry with a falsified key, meaning that the value wasn't really produced
-from that source. A subsequent build then picks up that poisoned cache entry.
-Solution: Builder uses a separate cache for each build.
+*Example 1:* The cache key does not fully cover the transitive closure of all
+inputs and instead only uses the digest of the source file itself. Adversary runs
+a build over `auth.cc` with command line flags to gcc that define a marco
+replacing `CheckAuth(ctx)` with `true`. When subsequent builds build `auth.cc`
+they will get the attacker's poisoned instance that does not call `CheckAuth`.
+Solution: SLSA Build Level 3 requires build cache's to be isolated and a properly
```
MarkLodato, line 535: optional suggestion: you might want to consider moving these two "Solution:" bits up to the "Mitigation" line above to explain what specifically we mean when we say "isolated", then here in these examples explain how the solution protects against this specific attack. So maybe:

*Mitigation:* Build caches ... attacks. In particular, the cache SHOULD be keyed by the transitive closure of all inputs to the cached artifact, and the cache must either be only writable by the trusted control plane or have SLSA Build L3 provenance for each cache entry.  (Or something like that - see @arewm's suggestion.)

*Example 1:* ... Solution: Build cache is keyed by digest of {digest of `auth.cc`, command line, digest of compiler}, so changing the command line or compiler results in a different cache entry.

*Example 2:* ... Solution: Each cache entry is keyed by the transitive closure of the inputs, and the cache entry is itself a SLSA Build L3 build with its own provenance that corresponds to the key.

(Perhaps needs better wording.)

MarkLodato (approved):
Overall LGTM. Just one suggestion. Thanks for writing this up!

arewm (approved):

PR #1298, opened 2025-03-25
content: V1.1 RC2

lehors:
Add SLSA v1.1 Release Candidate 2 (RC2).
Rename SLSA v1.1 to v1.1 RC1 (the way it should have been in the first place!) and label it as retired.
Redirect v1.1 to v1.1-rc2.
Update current activities.

Note to reviewers: Besides checking the preview, you might want to checkout this PR and do a diff between docs/spec/v1.1-rc1 and docs/spec/v1.1-rc2 to see what has changed between the 2. RC2 shouldn't include any of the future stuff (i.e., source track, build-env). Otherwise, looking at the files changed on GitHub will show all the RC2 as new and be of little value for review.

PR #1315, opened 2025-03-28
content: draft: build L3 requires complete external parameters, not provenance

arewm:
It was raised in the review of 1.1-RC2 that build L3 does not require provenance to be complete. Instead, it only requires the external parameters to be complete. With this requirement, it is still sufficient to mitigate an attack by SSH on the artifact as the verifier would be able to ensure that the external parameters meet expectations.

ref: https://github.com/slsa-framework/slsa/pull/1298#discussion_r2018863889

Signed-off-by: arewm <arewm@users.noreply.github.com>


lehors (approved):
LGTM. Nice catch!

TomHennen (approved):

TomHennen:
Thanks Andrew!

adityasaky (approved):

PR #1318, opened 2025-04-04
fix: Fix link to current activities

skade:
The "current activities" link in the "project status" part of the main page goes to the wrong target. Make it go to the right one.


lehors (approved):
Thanks!

PR #1319, opened 2025-04-07
fix: 1.1-rc2: address minor typo

zachariahcox:
addressing minor spelling errors.


TomHennen:
Per https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md I think this is considered a 'fix' which has reduced review requirements (1) and waiting period (0).

TomHennen (approved):

PR #1320, opened 2025-04-07
fix: 1.1 rc2 run grammar check and fix typos. 

zachariahcox:
This PR should introduce no change to the meaning of any sentence.


arewm (approved):
Awesome, thanks! It looks like we need a grammar linter. :)

lehors (approved):
Thanks for the careful nit picking! This should clearly be labelled as a fix rather than a content change though.

zachariahcox:
> This should clearly be labelled as a fix rather than a content change though.
Aha! I didn't know about that prefix. Thanks!

Issue #1134, opened 2024-09-19
How to handle submodules in the source track?


TomHennen:
Should we address submodules and other 'indirection' elsewhere in the spec or is it better to keep it aligned with the attestations themselves?

_Originally posted by @TomHennen in https://github.com/slsa-framework/slsa/pull/1094#discussion_r1672743142_

zachariahcox:
Some thoughts on this topic. 

Submodules are just files. In that sense, nothing special is needed. 

However, if you direct a tool to act on the content of those files (EG: `git clone --recurse-submodules`), you will end up with other repos in your repo. 
In this sense, submodule objects are similar to dependency lock files. 

Not all submodules are required (it might not be necessary to clone every submodule) and there are other ways to get the nested repo pattern needed by a project (EG: just clone them directly from a `makefile`).

TomHennen:
I'd agree, so long as we require that the builder include each submodule repo as a separate 'resolvedDependency' so that it's easy for package verifiers to know that all these other repos were used.

Would that be reasonable?

TomHennen:
Where should we document this?  A 'faq' section?

TomHennen:
This should be fixed by #1324

PR #1324, opened 2025-04-18
editorial: Describe a potential use case of source `dependencyLevels`

arewm:
The current wording of the `dependencyLevels` VSA property for the source track just states that dependencies are not common. While this may be true, git submodules are not too uncommon, so we should call these out specifically as an example.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -415,7 +415,7 @@ SCSs that do not use cryptographic digests MUST define a canonical type that is 
 E.g. `git+https://github.com/foo/hello-world`.
 5.  `verifiedLevels` MUST include the SLSA source track level the SCS asserts the revision meets. One of `SLSA_SOURCE_LEVEL_0`, `SLSA_SOURCE_LEVEL_1`, `SLSA_SOURCE_LEVEL_2`, `SLSA_SOURCE_LEVEL_3`.
 MAY include additional properties as asserted by the SCS.  The SCS MUST include _only_ the highest SLSA source level met by the revision.
-6.  `dependencyLevels` MAY be empty as source revisions are typically terminal nodes in a supply chain.
+6.  `dependencyLevels` MAY be empty as source revisions are typically terminal nodes in a supply chain. This COULD be used to indicate the source level of any git submodules present in the revision.
 
 The SCS MAY issue these attestations based on its understanding of the underlying system (e.g. based on design docs, security reviews, etc...),
 but at SLSA Source Level 3 MUST use tamper-proof [provenance attestations](#provenance-attestations) appropriate to their SCS when making the assessment.


docs/spec/draft/source-requirements.md
```diff
@@ -327,7 +327,7 @@ SCSs that do not use cryptographic digests MUST define a canonical type that is
 E.g. `git+https://github.com/foo/hello-world`.
 5.  `verifiedLevels` MUST include the SLSA source track level the verifier asserts the revision meets. One of `SLSA_SOURCE_LEVEL_0`, `SLSA_SOURCE_LEVEL_1`, `SLSA_SOURCE_LEVEL_2`, `SLSA_SOURCE_LEVEL_3`.
 MAY include additional properties as asserted by the verifier.  The verifier MUST include _only_ the highest SLSA source level met by the revision.
-6.  `dependencyLevels` MAY be empty as source revisions are typically terminal nodes in a supply chain.
+6.  `dependencyLevels` MAY be empty as source revisions are typically terminal nodes in a supply chain. For example, this COULD be used to indicate the source level of any git submodules present in the revision.
```
TomHennen, line 330: This is an interesting idea.

Let's discuss what this would look like from an e2e perpsective.

For build track VSAs the way we'd implement is something like:

The build provenance lists 5 binary dependencies, they each had their own VSA.  When the verification of the build provenance happens the verifier would populate dependencyLevels based on those binaries existing in the build provenance & their VSAs.

When dealing with source, my expectation for build provenance is that it would list each source repo fetched (including submodules!) in `resolvedDependencies`.  AIUI it's the git client that fetches these submodules anyways.  Do we like that properly (it seems nice to me) or would we want to modify it?

If we do want to handle this in the source track (where source VSAs would cover their submodules) how would we verify it?  I'm thinking we'd want to update the source provenance to list those submodules.  Should we do that?  IDK!

TomHennen: Also, maybe it's possible it would make sense for other 'linking' strategies that might not be as obvious as the ones git employs.  E.g. in Git maybe we'd be happy to just have them expressed in the build provenance and leave it out of scope for the source VSA, but maybe there are other source control systems where that wouldn't be the case.

arewm: (maybe I should have created an issue first instead of jumping to a PR)
> AIUI it's the git client that fetches these submodules anyways. Do we like that properly (it seems nice to me) or would we want to modify it?

That is fair. It is possible to clone a repository without recursion. It may therefore be possible to configure a build that doesn't perform this recursion. Should the presence of git submodules in provenance's `resolvedDependencies` be called out for the build track?

That being said, if it still seems reasonable to include a source dependency relationship for submodules. Just like artifacts can indicate dependencies, it is possible to perform builds without all of those dependencies present. The set of source dependencies doesn't have the be a subset of the dependencies for producing artifacts.

Does this question revolve around what the dependency track reporting might shake out to be?

zachariahcox:
It seems to me that submodules are the same as a lock file. 
If the build track describes those as dependencies, then we should probably consider doing the same here too.

docs/spec/draft/source-requirements.md
zachariahcox, line 330: I guess my vote would be that we leave the spec content similar to [other provenance formats](https://github.com/slsa-framework/slsa/blob/8c24af63cc1bc878bf8d350db1da8bf76386d8f2/docs/spec/v1.1-rc2/verification_summary.md?plain=1#L97C3-L101C5) unless we have a reason to be different here. 

The lock file analogy seems solid. If you're using submodules to describe your dependency graph, it does make sense to talk about the levels. 
I don't have a lot of experience working with submodules in high security builds. All else equal, rather than submodules it seems better to explicitly describe the repos you need and scope down the read token permissions only to those repos at the required revisions.

TomHennen: Sorry for taking so long to get back on this.  Let's leave it in.  Making one minor suggestion to remove 'For example' since the example is actually counter to text (e.g. it won't be empty in the example).

```suggestion
6.  `dependencyLevels` MAY be empty as source revisions are typically terminal nodes in a supply chain. This COULD be used to indicate the source level of any git submodules present in the revision.
```

TomHennen (approved):
FWIW I've come around on this.  Seems fine to include.  Just a minor change requested (but we should probably resolve other questions too).

zachariahcox (approved):

adityasaky (approved):
Looks good to me, but with @TomHennen's proposal to drop "for example".

TomHennen:
FYI I think this fixes #1134 

Just need to drop 'for example' and then rebase?

docs/spec/draft/source-requirements.md
arewm: I just pushed this change.

arewm:
The requested changes have been made. I think this PR is ready to be merged?

TomHennen (approved):

TomHennen:
The linter errors are fixed in main and are unrelated to this content.

TomHennen:
We have all the approvers needed and this PR has had the required amount of time for content.

PR #1325, opened 2025-04-18
content: SLSA v1.1 publication

lehors:
This PR publishes v1.1 RC2 as v1.1, the new 'Approved specification'. It marks v1.0 as 'Retired' and includes an update of the current activities.
This also hides v1.1 RC1 and RC2 without removing them (they can still be accessed directly, but don't appear in the navigation bar).


arewm (approved):
🚀

TomHennen (approved):
Woo!

adityasaky (approved):

PR #1328, opened 2025-04-20
nonspec: tooling for migration to a late-branching strategy, standardized devcontainers

zachariahcox:
fixes: https://github.com/slsa-framework/slsa/issues/742

* add devcontainer
* migrate tools to tools folder
* propose plan for late-branching refactor
* add workflow file to call and record history of that refactor (not sure best way to preserve those logs forever
* exploration into how netlify preview builds could be supported. (for draft updates, it's easy, for release updates, what should happen?) Either way, all those workflows are disabled. 

[Copilot is generating a summary...]

PR #1329, opened 2025-04-20
nonspec: new website build strategy

zachariahcox:
super draft. 

depends on https://github.com/slsa-framework/slsa/pull/1328


lehors:
Thanks for doing this. I like it. Just some thoughts that came up as I went through this a first time:

What's the reason to change from docs/ to www/ ? It's no big deal but this creating a lot of noise in this PR.

We'll also need to document what steps need to be taken when releasing a new spec (something that's missing today!) One will still need to update files like _data/versions.yml _data/nav/* _redirects among others. Experience with the release of v1.1 just showed that there are quite few bits and pieces all over the place that need updating. We shouldn't need to find those the hard way every time.

We should aim to eventually have combine-versions.sh pull in the list of specs to deploy from one of the config files - I think _data/versions.yml.

Just to be clear: I'm not asking for any of the above to be part of this PR! :-)

PR #1331, opened 2025-04-21
nonspec: Update home page

lehors:
Removes the project status section which is outdated and no longer needed since we have "current activities" in the navigation bar.
Also adds some config and redirect to make the homepage more resilient to the publication of new spec versions.

### DIFF
--- a/docs/_redirects
+++ b/docs/_redirects
@@ -57,9 +57,10 @@
 
 /spec                   /spec/v1.0              302
 /spec/faq               /spec/v1.0/faq          302
-/spec/v1/*              /spec/v1.0/:splat       302
+/spec/v1/*              /spec/v1.1/:splat       302
 /spec/v1.1/*            /spec/v1.1-rc2/:splat   302
 /spec/current-activities /current-activities    301
+/spec/latest            /spec/v1.1/             302
 
 # Note: Versions prior to v1.0 stay in /verification_summary.
 /verification_summary           /spec/v1.0/verification_summary     302  # floating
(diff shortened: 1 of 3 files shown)


docs/_redirects
```diff
@@ -57,9 +57,10 @@
 
 /spec                   /spec/v1.0              302
 /spec/faq               /spec/v1.0/faq          302
-/spec/v1/*              /spec/v1.0/:splat       302
+/spec/v1/*              /spec/v1.1/:splat       302
 /spec/v1.1/*            /spec/v1.1-rc2/:splat   302
 /spec/current-activities /current-activities    301
+/spec/latest            /spec/v1.1/             302
```
TomHennen, line 63: Do we need/want a latest splat?  IDK

e.g. for spec/latest/foo?

docs/index.md
TomHennen, line 1: There are some other direct links to 1.0 things here too.  Should we fix those now?

TomHennen (changes requested):

docs/_redirects
arewm: I think that if we are only going to have one redirect for `latest`, it should be a splat.

docs/index.md
lehors: done

lehors:
I see @arewm submitted a concurrent PR but I suggest we merge this ASAP to address the most obvious problems until we figure the right way to go with all the redirects proposed in PR #1332.

arewm (approved):

PR #1336, opened 2025-04-23
content: draft: Source VSAs shouldn't be produced for all revisions.

TomHennen:
Currently the langauge makes it sound like they're produced for all revisions, but we really only need (or want!) them to be produced for ones the SCS wants to make claims about.

This change makes that more clear and indicates how consumers should think about revisions without VSAs.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -173,12 +173,16 @@ See also [Use cases for non-cryptographic, immutable, digests](https://github.co
 <td>✓<td>✓<td>✓
 <tr id="source-summary"><td>Source Summary Attestations<td>
 
-The SCS MUST generate [summary attestations](#summary-attestation) to enable users to determine the source level of a given revision.
+The SCS MUST generate [source summary attestations](#summary-attestation) to
+indicate the SLSA Source Level of a given revision.
 
-If a consumer is authorized to access source on a particular branch, they MUST be able to fetch the summary attestations for revisions in the history of that branch.
+If a consumer is authorized to access a revision, they MUST be able to fetch the
+corresponding source summary attestations.
 
-It is possible that an SCS can make no claims about a particular revision.
-For example, this would happen if the revision was created on another SCS, or if the revision was not the result of an accepted change management process.
+It is possible that an SCS makes no claims about a particular revision.
+In these cases the SCS MAY NOT generate a summary attestation. Consumers MUST
+interpret the absence of a source summary attestation as though the revision
+is SLSA Source Level 0.
 
 <td>✓<td>✓<td>✓
 <tr id="branches"><td>Branches<td>
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -180,7 +180,8 @@ The summary attestation "interprets" what the data in the source provenance atte
 If a consumer is authorized to access a revision, they MUST be able to fetch the
 corresponding source summary attestations.
 
-If the SCS DOES NOT generate a summary attestation for a revision, the revision MUST be verified as having SLSA Source Level 0.
+If the SCS DOES NOT generate a summary attestation for a revision, the revision
+cannot be verified and thus has Source Level 0.
 
 <td>✓<td>✓<td>✓
 <tr id="branches"><td>Branches<td>
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -174,7 +174,7 @@ See also [Use cases for non-cryptographic, immutable, digests](https://github.co
 <tr id="source-summary"><td>Source Summary Attestations<td>
 
 The SCS MUST generate a [source summary attestation](#summary-attestation) to
-indicate the SLSA Source Level of revisions it wishes to make claims about.
+indicate the SLSA Source Level of any revision at Level 1 or above.
 
 If a consumer is authorized to access a revision, they MUST be able to fetch the
 corresponding source summary attestations.
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -180,10 +180,7 @@ The summary attestation "interprets" what the data in the source provenance atte
 If a consumer is authorized to access a revision, they MUST be able to fetch the
 corresponding source summary attestations.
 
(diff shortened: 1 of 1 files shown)


arewm (approved):

docs/spec/draft/source-requirements.md
```diff
@@ -173,12 +173,16 @@ See also [Use cases for non-cryptographic, immutable, digests](https://github.co
...
+The SCS MUST generate [source summary attestations](#summary-attestation) to
+indicate the SLSA Source Level of a given revision.
 
-If a consumer is authorized to access source on a particular branch, they MUST be able to fetch the summary attestations for revisions in the history of that branch.
+If a consumer is authorized to access a revision, they MUST be able to fetch the
+corresponding source summary attestations.
 
-It is possible that an SCS can make no claims about a particular revision.
-For example, this would happen if the revision was created on another SCS, or if the revision was not the result of an accepted change management process.
+It is possible that an SCS makes no claims about a particular revision.
```
zachariahcox, line 182: it seems like line 176 is saying they must, but here they don't have to. 
I think it's that they MAY.

docs/spec/draft/source-requirements.md
```diff
@@ -173,12 +173,16 @@ See also [Use cases for non-cryptographic, immutable, digests](https://github.co
 <td>✓<td>✓<td>✓
 <tr id="source-summary"><td>Source Summary Attestations<td>
 
-The SCS MUST generate [summary attestations](#summary-attestation) to enable users to determine the source level of a given revision.
+The SCS MUST generate [source summary attestations](#summary-attestation) to
+indicate the SLSA Source Level of a given revision.
```
zachariahcox, line 177: I suspect users will get a little confused between the two types of attestations. 

```suggestion
The SCS MAY generate a [source summary attestation](#summary-attestation) to
indicate the SLSA Source Level of a revision.
The summary attestation "interprets" what the data in the source provenance attestation means.
```

docs/spec/draft/source-requirements.md
```diff
@@ -173,12 +173,16 @@ See also [Use cases for non-cryptographic, immutable, digests](https://github.co
...
-If a consumer is authorized to access source on a particular branch, they MUST be able to fetch the summary attestations for revisions in the history of that branch.
+If a consumer is authorized to access a revision, they MUST be able to fetch the
+corresponding source summary attestations.
 
-It is possible that an SCS can make no claims about a particular revision.
-For example, this would happen if the revision was created on another SCS, or if the revision was not the result of an accepted change management process.
+It is possible that an SCS makes no claims about a particular revision.
+In these cases the SCS MAY NOT generate a summary attestation. Consumers MUST
+interpret the absence of a source summary attestation as though the revision
+is SLSA Source Level 0.
```
zachariahcox, line 185:
```suggestion
If the SCS DOES NOT generate a summary attestation for a revision, the revision MUST be verified as having SLSA Source Level 0.
```

or maybe "cannot be verified and thus has source lvl 0?"

zachariahcox (approved):
I left some comments, also ok as-is.

docs/spec/draft/source-requirements.md
TomHennen: I'll take your suggestion and tweak, thanks.

docs/spec/draft/source-requirements.md
```diff
@@ -173,12 +173,15 @@ See also [Use cases for non-cryptographic, immutable, digests](https://github.co
 <td>✓<td>✓<td>✓
 <tr id="source-summary"><td>Source Summary Attestations<td>
 
-The SCS MUST generate [summary attestations](#summary-attestation) to enable users to determine the source level of a given revision.
+The SCS MAY generate a [source summary attestation](#summary-attestation) to
+indicate the SLSA Source Level of a revision.
+The summary attestation "interprets" what the data in the source provenance attestation means.
```
arewm, line 178: What do you mean by this? Any interpretations are also only possible if the source provenance attestation exists.

TomHennen: I realized I hadn't actually _pushed_ the changes I made, which should address this... Sorry!

PTAL at the new lanaguge?

arewm: It is odd that the summary attestation is required but we state that the SCS MAY create it. I get what you are saying, it just doesn't sit right.

Should we say something like
```
The summary attestation "interprets" what the data in the source provenance attestation means if it exists for the same commit.
```

TomHennen: Oh the MAY stuff is the old language.  If you check out the latest revision it says

"The SCS MUST generate a [source summary attestation](#summary-attestation) to
indicate the SLSA Source Level of revisions it wishes to make claims about."

and then later

"When source provenance is available the SCS MAY use it to generate the
source summary attestation."

arewm (approved):

docs/spec/draft/source-requirements.md
arewm: Interesting. Not sure how I got stuck between revisions. I like the latest!

zachariahcox (approved):

docs/spec/draft/source-requirements.md
```diff
@@ -173,12 +173,17 @@ See also [Use cases for non-cryptographic, immutable, digests](https://github.co
 <td>✓<td>✓<td>✓
 <tr id="source-summary"><td>Source Summary Attestations<td>
 
-The SCS MUST generate [summary attestations](#summary-attestation) to enable users to determine the source level of a given revision.
+The SCS MUST generate a [source summary attestation](#summary-attestation) to
+indicate the SLSA Source Level of revisions it wishes to make claims about.
```
zachariahcox, line 177:
```suggestion
The SCS MAY generate a [source summary attestation](#summary-attestation) to
indicate the SLSA Source Level of a revision.
```
Does "must if you wish" mean "may"?

zachariahcox:
approving, with a minor comment for thoughts.

docs/spec/draft/source-requirements.md
TomHennen: "If you want to claim a revision has a SLSA level you MUST create a VSA for it"

TomHennen: Removed the wishes bit.  I think this covers it. :)

PR #1337, opened 2025-04-23
content: draft: Simplify source organization requirements

arewm:
Similar to the build track, the requirements for organizations in the source track are to choose a source control system which is capable of achieving the desired levels.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -118,11 +118,12 @@ Many examples in this document use the [git version control system](https://git-
 
 #### Choose an appropriate source control system
 
-An organization producing source revisions MUST select a source control system
-capable of reaching their desired SLSA Source Level.
+An organization producing source revisions MUST select a SCS capable of reaching
+their desired SLSA Source Level.
 
 For example, if an organization wishes to produce revisions at Source Level 3,
-they MUST choose a source control system of producing Source Level 3 attestations.
+they MUST choose a source control system capable of producing Source Level 3
+attestations.
 
 #### Choose an appropriate change management process
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -132,7 +132,7 @@ the same source control system which manages the revisions.
 
 ### Source Control System
 
-#### Generate Provenance
+#### Revision management
 
 <table>
 <tr><th>Requirement<th>Description<th>L1<th>L2<th>L3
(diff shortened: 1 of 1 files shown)


docs/spec/draft/source-requirements.md
```diff
@@ -112,41 +112,27 @@ within the SCS's storage systems.
...
-They may be self-hosted or hosted in the cloud using vendors like GitLab, GitHub, Bitbucket, etc.
+#### Choose an appropriate source control system
 
-When self-hosting a solution, local, unauthenticated storage is not acceptable.
+An organization producing source revisions MUST select a source control system
+capable of reaching their desired SLSA Source Level.
 
-Branch protection is not required, nor are there any other constraints on the configuration of the tools.
+For example, if an organization wishes to produce revisions at Source Level 3,
+they MUST choose a source control system of producing Source Level 3 attestations.
```
TomHennen, line 125:
```suggestion
they MUST choose a source control system capable of producing Source Level 3 attestations.
```

docs/spec/draft/source-requirements.md
```diff
@@ -112,41 +112,27 @@ within the SCS's storage systems.
...
-
-The organization MUST document how the detailed [provenance attestations](#provenance-attestations)
-are distributed for relevant source repositories.
-<td><td><td>✓
-
-</table>
+### Source Control System
 
-### Source Control System Requirements
+#### Generate Provenance
```
TomHennen, line 135: I don't think this is the right heading?

If the intent is to match the build track ("Provenance generation") the problem is that it doesn't really include the IAM or other stuff we have here.

So maybe "Revision Management" instead?

```suggestion
#### Revision Management
```

Alternatively, just remove the heading?

TomHennen (approved):

docs/spec/draft/source-requirements.md
arewm: I'll change it to revision management for now. I added an additional heading here because there are two relevant sets of requirements for a SCS. We can revisit whether there should be two headings in the future.

zachariahcox (approved):
nice!

docs/spec/draft/source-requirements.md
```diff
@@ -112,41 +112,27 @@ within the SCS's storage systems.
...
-<tr id="use-modern-tools"><td>Use modern tools<td>
+[Organization]: #organization
 
-The organization MUST manage the source using tools specifically designed to manage source code.
-Tools like git, Perforce, Subversion are great examples.
-They may be self-hosted or hosted in the cloud using vendors like GitLab, GitHub, Bitbucket, etc.
+#### Choose an appropriate source control system
 
-When self-hosting a solution, local, unauthenticated storage is not acceptable.
+An organization producing source revisions MUST select a source control system
```
zachariahcox, line 121: minor, should we use the acronym here? 
```suggestion
An organization producing source revisions MUST select an SCS
```

PR #1344, opened 2025-04-23
content: draft: Rename 'Branches' to 'Protected Branches'

TomHennen:
And clarify they're protected by the org's change mangagment process.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -181,17 +181,18 @@ It is possible that an SCS can make no claims about a particular revision.
 For example, this would happen if the revision was created on another SCS, or if the revision was not the result of an accepted change management process.
 
 <td>✓<td>✓<td>✓
-<tr id="branches"><td>Branches<td>
+<tr id="branches"><td>Protected Branches<td>
 
-If the SCS supports multiple branches, the organization MUST indicate which branches are intended for consumption.
-This may be implied or explicit.
+If the SCS supports branches, the organization MUST indicate which branches are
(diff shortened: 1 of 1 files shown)


arewm (approved):

zachariahcox (approved):

docs/spec/draft/source-requirements.md
```diff
@@ -181,17 +181,18 @@ It is possible that an SCS can make no claims about a particular revision.
 For example, this would happen if the revision was created on another SCS, or if the revision was not the result of an accepted change management process.
 
 <td>✓<td>✓<td>✓
-<tr id="branches"><td>Branches<td>
+<tr id="branches"><td>Protected Branches<td>
 
-If the SCS supports multiple branches, the organization MUST indicate which branches are intended for consumption.
-This may be implied or explicit.
+If the SCS supports branches, the organization MUST indicate which branches are
+intended for consumption. This may be implied or explicit.
```
zachariahcox, line 187: I don't think you should have branches intended for consumption that are not protected. 
This is not explicitly called out though

TomHennen: Discussed in person and the thought is that we can merge and revisit this language later.

TomHennen:
After discussion, we just want to merge this with 'Branch continuity'.

We can then rename it to 'protected branch'.

TomHennen:
> After discussion, we just want to merge this with 'Branch continuity'.
> 
> We can then rename it to 'protected branch'.

Nevermind, it didn't work out.

Went with something simpler.

PR #1346, opened 2025-04-24
fix: add missing verifying source page to TOC

arewm:


lehors:
You need to add it to main.yml instead. I should verify it and remove it if that's correct but I don't think draft.yml is ever used anymore.

arewm:
Thanks, @lehors !

PR #1349, opened 2025-04-24
content: draft: Describe the need for continuity in the source track

arewm:
Since revisions are proposed relative to their parent. Since any revision can potentially have at least a partial dependency on its parent, we can only ensure that controls are effectively protecting the current revision by checking that they have been continuously enabled since at least the previous commit.

This is a concept that will be relevant to multiple controls within the source track.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -191,7 +191,7 @@ For systems like GitHub or GitLab, this can be accomplished by enabling branch p
 
 It MUST NOT be possible to delete the entire repository (including all branches) and replace it with different source.
 
-These branch protection enforcements MUST be verified for [continuity](#revision-continuity).
+These branch protection enforcements MUST be verified for [continuity](#revision-control-continuity).
 
 <td><td>✓<td>✓
 <tr id="tag-hygiene"><td>Tag Hygiene<td>
@@ -258,7 +258,7 @@ or if the revision was not the result of an accepted change management process.
 
 The SCS MUST provide a mechanism for organizations to enforce additional
 technical controls which govern changes to a [branch](#definitions). These
-technical controls MAY be verified for [continuity](#revision-continuity).
+technical controls MAY be verified for [continuity](#revision-control-continuity).
 
 For example, this could be accomplished by:
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -191,16 +191,7 @@ For systems like GitHub or GitLab, this can be accomplished by enabling branch p
 
 It MUST NOT be possible to delete the entire repository (including all branches) and replace it with different source.
 
-Continuity exceptions are allowed via the [safe expunging process](#safe-expunging-process).
-
-<td><td>✓<td>✓
-<tr id="continuous-controls"><td>Continuous Controls<td>
-
-The SCS MUST provide a mechanism for organizations to indicate which controls
-are enforced for every revision on a protected branch since that protection was
-enabled.
-
-Continuity exceptions are allowed via the [safe expunging process](#safe-expunging-process).
+These branch protection enforcements MUST be verified for [continuity](#revision-continuity).
 
 <td><td>✓<td>✓
 <tr id="tag-hygiene"><td>Tag Hygiene<td>
@@ -266,9 +257,8 @@ or if the revision was not the result of an accepted change management process.
 <tr id="change-management-process"><td>Enforced change management process<td>
 
 The SCS MUST provide a mechanism for organizations to enforce additional
-technical controls which govern changes to a [branch](#definitions). The
-continuity of these controls MAY be enforced by the
-[continuous controls](#continuous-controls) requirement.
+technical controls which govern changes to a [branch](#definitions). These
+technical controls MAY be verified for [continuity](#revision-continuity).
 
 For example, this could be accomplished by:
 
@@ -279,6 +269,22 @@ For example, this could be accomplished by:
 
 <td><td><td>✓
 </table>
+
+#### Revision continuity
+
+Revisions are created by applying a specific code change (a "diff" in git) on
+top of an earlier revision. Since a single revision does not provide sufficient
+context to assess the source code controls on its own, an SCS MUST provide a
+means to ensure continuity of controls from one revision to another. Continuity
+MAY be accomplished by an SCS ensuring that a control has not lapsed between when
+a revision and its parent are created.
+
+While continuity MUST be established and tracked from a specific point in time,
+an SCS MAY choose not to expose that to consumers in attestations it produces.
+If there is a lapse in continuity for a specific control, that continuity MUST
+be re-established from a new point in time.
+
+Continuity exceptions are allowed via the [safe expunging process](#safe-expunging-process).
 
 ### Provide a change management tool
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -279,10 +279,13 @@ means to ensure continuity of controls from one revision to another. Continuity
 MAY be accomplished by an SCS ensuring that a control has not lapsed between when
 a revision and its parent are created.
 
-While continuity MUST be established and tracked from a specific point in time,
-an SCS MAY choose not to expose that to consumers in attestations it produces.
-If there is a lapse in continuity for a specific control, that continuity MUST
-be re-established from a new point in time.
+Continuity MUST be established and tracked from a specific point in time. If
+there is a lapse in continuity for a specific control, that continuity MUST be
+re-established from a new point in time.
+
+An SCS SHOULD expose the time a control was established in the source provenance
+and other attestations it produces (with the exception of the source summary
+attestation).
 
 Continuity exceptions are allowed via the [safe expunging process](#safe-expunging-process).
 
@@ -434,3 +437,4 @@ Example source provenance attestations:
   phishing resistant.
 -   Protect against authentication token theft by forbidding bearer tokens
   (e.g. PATs).
+-   Including length of continuity in the VSAs
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -194,6 +194,15 @@ It MUST NOT be possible to delete the entire repository (including all branches)
 Continuity exceptions are allowed via the [safe expunging process](#safe-expunging-process).
 
 <td><td>✓<td>✓
+<tr id="continuous-controls"><td>Continuous Controls<td>
+
+The SCS MUST provide a mechanism for organizations to indicate which controls
+are enforced for every revision on a protected branch since that protection was
+enabled.
+
+Continuity exceptions are allowed via the [safe expunging process](#safe-expunging-process).
+
+<td><td>✓<td>✓
 <tr id="tag-hygiene"><td>Tag Hygiene<td>
 
 If the SCS supports tags (or other non-branch tracks), additional care must be
@@ -257,7 +266,9 @@ or if the revision was not the result of an accepted change management process.
 <tr id="change-management-process"><td>Enforced change management process<td>
 
 The SCS MUST provide a mechanism for organizations to enforce additional
-technical controls which govern changes to a [branch](#definitions).
+technical controls which govern changes to a [branch](#definitions). The
+continuity of these controls MAY be enforced by the
+[continuous controls](#continuous-controls) requirement.
 
 For example, this could be accomplished by:
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -48,7 +48,7 @@ Consumers can examine the various source provenance attestations to determine if
 
 When onboarding a branch to the SLSA Source Track or increasing the level of
 that branch, organizations are making claims about how the branch is managed
-from that time or revision forward.
+from that time or revision forward. This establishes [continuity](#revision-control-continuity).
 
 No claims are made for prior revisions.
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -276,7 +276,7 @@ Revisions are created by applying a specific code change (a "diff" in git) on
 top of an earlier revision. Since a single revision does not provide sufficient
 context to assess the source code controls on its own, an SCS MUST provide a
 means to ensure continuity of controls from one revision to another. Continuity
-MAY be accomplished by an SCS ensuring that a control has not lapsed between when
+MAY be accomplished by an SCS showing that a control has been enforced when
 a revision and its parent are created.
 
 Continuity MUST be established and tracked from a specific point in time. If
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -270,7 +270,7 @@ For example, this could be accomplished by:
 <td><td><td>✓
 </table>
 
-#### Revision continuity
+#### Revision control continuity
 
 Revisions are created by applying a specific code change (a "diff" in git) on
 top of an earlier revision. Since a single revision does not provide sufficient


TomHennen:
After our discussion at lunch I think @arewm is planning to make some updates?  If so I'll hold off on reviewing.

arewm:
@TomHennen, feel free to review this PR now.

docs/spec/draft/source-requirements.md
```diff
@@ -256,6 +257,22 @@ For example, this could be accomplished by:
 <td><td><td>✓
 </table>
 
+#### Revision continuity
```
TomHennen, line 260:
```suggestion
#### Revision control continuity
```

?

docs/spec/draft/source-requirements.md
```diff
@@ -256,6 +257,22 @@ For example, this could be accomplished by:
...
+top of an earlier revision. Since a single revision does not provide sufficient
+context to assess the source code controls on its own, an SCS MUST provide a
+means to ensure continuity of controls from one revision to another. Continuity
+MAY be accomplished by an SCS ensuring that a control has not lapsed between when
+a revision and its parent are created.
+
+While continuity MUST be established and tracked from a specific point in time,
+an SCS MAY choose not to expose that to consumers in attestations it produces.
+If there is a lapse in continuity for a specific control, that continuity MUST
+be re-established from a new point in time.
```
TomHennen, line 272:
```suggestion
Continuity MUST be established and tracked from a specific point in time.
If there is a lapse in continuity for a specific control, that continuity MUST
be re-established from a new point in time.

An SCS SHOULD expose the time a control was established in the source provenance and other attestations it produces (with the exception of the source summary attestation).
```

Pulling out exposing the time in the attestations to a separate line to aid clarity.  Also specifying that exposing the information in VSAs isn't required.

I wonder if we should add something like "include property age in VSAs" to "Future considerations" ?

TomHennen (changes requested):

docs/spec/draft/source-requirements.md
arewm: I don't have an issue with that other than it is longer. It is certainly more precise as we want to establish continuity in the various controls around new revisions.

docs/spec/draft/source-requirements.md
arewm: I'll make this change locally so I can fix the line endings. I think we did mention including that in future considerations previously. I don't know how it will fit within the VSA spec but it doesn't hurt to include it for the future.

TomHennen: If we were to do this we might change `verifiedLevels` from a list to a map.  Which would be a breaking change.  One could imagine something like 

```
verifiedLevels: {
  "SLSA_SOURCE_LEVEL_3": "2025-01-02",
  "TWO_PARTY_REVIEW": "2024-02-03"
}
```

docs/spec/draft/source-requirements.md
TomHennen: I think my trouble with just 'revision continuity' is that it sounds a lot like 'branch continuity' and that seems confusing, especially since we're talking about controls.

The prior 'control continuity' we'd discussed would be OK too?  Not sure if that's as nice though.

docs/spec/draft/source-requirements.md
arewm: Ah, sure. If we are going to introduce a breaking change, we could just break it all.

I wonder if we should re-evaluate moving the VSA spec to in-toto. This would enable us to have more flexibility with the it without requiring the pinning to the _one_ SLSA version.

docs/spec/draft/source-requirements.md
arewm: I prefer `Revision control continuity` to `Control continuity`. The previous subsection is `Revision management` so I wanted to keep a parallel construction there.

docs/spec/draft/source-requirements.md
```diff
@@ -178,7 +178,7 @@ For systems like GitHub or GitLab, this can be accomplished by enabling branch p
 
 It MUST NOT be possible to delete the entire repository (including all branches) and replace it with different source.
 
-Continuity exceptions are allowed via the [safe expunging process](#safe-expunging-process).
+These branch protection enforcements MUST be verified for [continuity](#revision-continuity).
```
TomHennen, line 181:
```suggestion
These branch protection enforcements MUST be verified for [continuity](#revision-control-continuity).
```

TomHennen (approved):
Just one tweak to fix a broken link.

docs/spec/draft/source-requirements.md
arewm: @TomHennen , is there anything remaining from this thread or can I resolve it?

TomHennen: yeah we can resolve

docs/spec/draft/source-requirements.md
```diff
@@ -269,6 +270,25 @@ For example, this could be accomplished by:
 <td><td><td>✓
 </table>
 
+#### Revision control continuity
+
+Revisions are created by applying a specific code change (a "diff" in git) on
+top of an earlier revision. Since a single revision does not provide sufficient
+context to assess the source code controls on its own, an SCS MUST provide a
+means to ensure continuity of controls from one revision to another. Continuity
+MAY be accomplished by an SCS ensuring that a control has not lapsed between when
```
zachariahcox, line 279: "Has not lapsed" is hard to prove. 
We might want: "show that it was enforced."

This will also help when you need to go vsa commits from the past that predate the slsa source track but which were retroactively compliant.

docs/spec/draft/source-requirements.md
```diff
@@ -269,6 +270,25 @@ For example, this could be accomplished by:
 <td><td><td>✓
 </table>
 
+#### Revision control continuity
+
+Revisions are created by applying a specific code change (a "diff" in git) on
+top of an earlier revision. Since a single revision does not provide sufficient
+context to assess the source code controls on its own, an SCS MUST provide a
```
zachariahcox, line 277: > Since a single revision does not provide sufficient
context to assess the source code controls on its own

Doesn't it have to? The source track is about provenance on the correct application of change sets. In that sense it does need to assess controls on its own, right?

It's also true that in any repo with old commits, some blobs in the revision predate the application of slsa controls. Those are kind of mysterious. 

If you want to be super safe, you could bootstrap a repo with a line by line review of the whole thing and issue a vsa of all content. 

A more typical option might be to say, we require linear history and all commits reached on the main branch are "newer than the branch's previous revision" and were therefore subject to all the modern controls too. 

The "old blob reachability" problem should probably only apply to the bootstrap phase (where we think you need a blanket attestation) or if you allow multiple parents on the merge commit (which maybe we just shouldn't recommend for higher levels).

docs/spec/draft/source-requirements.md
arewm: That's fair. How about:
```suggestion
MAY be accomplished by an SCS showing that a control has been enforced when
```

In context with the surrounding lines,

> means to ensure continuity of controls from one revision to another. Continuity
MAY be accomplished by an SCS showing that a control has been enforced when
a revision and its parent are created.

arewm:
@zachariahcox , this PR has stalled. Are there things that we can improve with it? Or are there follow-up clarifications that can capture in an issue to improve before we release the source track?

docs/spec/draft/source-requirements.md
adityasaky: I'm not sure this should be an SCS requirement. If we take a step back and consider a parallel for the build track, this seems akin to saying the build system is responsible for ensuring it doesn't drop below a certain level between different build executions. That's the responsibility of the organization that oversees the system out of band, and the consumers of build provenance verify that their expectations (perhaps established also out of band in conjunction with the organization) are met.

Similarly, I think the continuity of controls is something the organization should oversee. I do see why this is more a concern in the source track; revisions build on prior revisions which build on prior revisions. However, I think this shouldn't be an SCS requirement that the **consumer relies on being enforced** (in the process undermining source provenance as a way of at least partially keeping the SCS honest). Instead, I propose we highlight control / level continuity as something the consumer must establish for the verification pieces of the source track. For example:
1. Organization communicates protected branch `main`
2. Organization communicates `main` is at source level 3 from revision `abcdef01`
3. Consumer sets expectations that any revision on `main` after `abcdef01` must be at least level 3
4. Successive revisions are verified after `abcdef01` using corresponding source provenance attestations by the consumer to ensure level 3 requirements are met rather than trusting the SCS to have enforced level 3 requirements continuously

Scenarios where a revision falls below a level ought to be reconciled out of band by between the consumer and the organization. This is probably, as you suggest, by defining a new starting revision for the control.

Please let me know if I've misunderstood this entirely, thanks!

adityasaky: Now, an organization may choose to internally enforce continuity for their own convenience via an SCS that has such a feature, but I think that's different from how the consumer looks at a revision and the controls applied to said revision.

TomHennen: I wonder if the changes in #1362 might help resolve this. It provides a bit more detail on these org-provided technical controls, how they get communicated to others, and _who_ is making those claims.

The SCS should still be able to enforce that the control the org identifies as `USER_SOURCE_FOO` has been applied for each commit on a branch, but it doesn't necessarily have insight into how that control works or how strong it is (e.g. can everyone just bypass it).

Maybe the question would be "well what's the point then?"

If nothing else, orgs can make mistakes in their technical control implementations and they want to be able to identify those things.  They want to be able to tell the story to themselves and their auditors.

I'd also push back on the idea that the consumer should establish continuity themselves in all cases.  That's a lot of work!  There might be 3 revisions to check but there might also be 10,000.  It's one of the reasons the source track currently produces a VSA (it makes verification easier) and we also acknowledge that implementations will differ and have different provenance formats and requirements (which will make it harder for consumers to implement recursive provenance verification).  Consumers can still do the verification themselves, but they shouldn't have to.

docs/spec/draft/source-requirements.md
```diff
@@ -48,7 +48,7 @@ Consumers can examine the various source provenance attestations to determine if
 
 When onboarding a branch to the SLSA Source Track or increasing the level of
 that branch, organizations are making claims about how the branch is managed
-from that time or revision forward.
+from that time or revision forward. This establishes [continuity](#revision-control-continuity).
```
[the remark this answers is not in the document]
zachariahcox: branch continuity ensures that there's a stable history for a revision at the tip of a branch.

"claims continuity" should logically mean something similar: there is a minimum subset of claims that can be made about all commits in the history of a branch. 
There is a claims "discontinuity" whenever the SCS did not make that subset of claims about a commit in the lineage. 

In practice, I think claims continuity will be a lot easier to maintain over the long haul.

docs/spec/draft/source-requirements.md
zachariahcox: > Consumers can still do the verification themselves, but they shouldn't have to.

I agree. I think there are still multiple parties here. 
While not all them will generate docs for sigstore, there are at least 3 parties:
1. source prov issuer
1. change management tool claims issuer
1. VSA summarizer issuer

Consumers should _mostly_ lean on VSAs. 
VSA issuers will need to require claims continuity for new revisions on branches. 

The source prov issuer will need to summarize claims from the change tooling. 
Change management tools are going to be pretty dumb -- they are usually just going to have lists of facts.

I think in this context, it's the source provenance claims that need to remain stable for the VSA issuer to be happy.

docs/spec/draft/source-requirements.md
```diff
@@ -256,6 +257,25 @@ For example, this could be accomplished by:
 <td><td><td>✓
 </table>
 
+#### Revision control continuity
+
+Revisions are created by applying a specific code change (a "diff" in git) on
+top of an earlier revision. Since a single revision does not provide sufficient
+context to assess the source code controls on its own, an SCS MUST provide a
+means to ensure continuity of controls from one revision to another. Continuity
```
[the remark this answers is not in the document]
zachariahcox: I wasn't thinking that the claims were related to git, but that you could learn all the relevant claims by examining: the source prov attestation and (if you really want to), the lineage, fetching each of their source prov attestations as far back as you want. 

The lack of provenance / VSA is the same as the lack of claims, which is the same as discontinuity of any particular claim. 

In my example above (and extremely relevant for larger dev teams), all rulesets could have been modified between revisions B and A but that's ok as long as controls X and Y are continuously enforced on all new revisions.

docs/spec/draft/source-requirements.md
```diff
@@ -269,6 +270,25 @@ For example, this could be accomplished by:
...
+MAY be accomplished by an SCS showing that a control has been enforced when
+a revision and its parent are created.
+
+Continuity MUST be established and tracked from a specific point in time. If
+there is a lapse in continuity for a specific control, that continuity MUST be
+re-established from a new point in time.
+
+An SCS SHOULD expose the time a control was established in the source provenance
+and other attestations it produces (with the exception of the source summary
+attestation).
```
zachariahcox, line 288:
```suggestion
A branch has "control continuity" if contiguous revisions were accepted 
under the same set of controls. If a new revision is accepted under a 
different set of controls, it does not have control-continuity with its predecessor. 
In this example, the sequence of revisions E through A has control continuity for control x. 
Control y is only continuous from commits C through A due to a lapse at D.

main
|
A -- controls: x, y
|
B -- controls: x, y
|
C -- controls: x, y
|
D -- controls: x
|
E -- controls: x, y
```

zachariahcox: leaving here for comments -- this would be a shift towards a mode that is less approximate.
We can definitely talk about control continuity based on timestamps, but especially for enterprise- and org-level policies might be targeted based on non-public information, or repository properties, or other means of dynamically finding places they should apply. 

I'm not sure how we would deal with those cases if we put too much on policy time-since-last-mutation

docs/spec/draft/source-requirements.md
adityasaky: I think I'm convinced after the SLSA meeting this week. Let me try and summarize my understanding:
a) SCS provides mechanisms to enforce various controls, including organization defined ones
b) SCS has ways of ensuring continuity of these controls that the organization can rely on
c) The VSA will include the controls that were enforced, and also state that they were enforced continuously (I'm a little confused on whether this is implicit from a control being listed or explicit)
d) The source provenance will _not_ list controls but rather contain more "raw" facts about what happened in producing a revision, and thus at higher levels a consumer can use the source provenance(s) if desired to verify one or more controls listed in the VSA + their continuity

I think the last point is what I feel most strongly about (which I also mentioned on the call); I think we want to ensure that as much as possible, whatever we put in the provenance is "raw" information that can be used as evidence to verify the claims in a VSA (even if this is difficult!), both the application of a control (e.g., two party review) and the continuity of that control (e.g., has every revision since X had two party review).

Does this sound right? :)

adityasaky: If we are agreed, I think the change still necessary here is further down:

> An SCS SHOULD expose the time a control was established in the source provenance
and other attestations it produces (with the exception of the source summary
attestation).

TomHennen: I'd tweak this somewhat and maybe replace 'controls' with 'properties'.  [Over here](https://github.com/slsa-framework/slsa/blob/main/docs/spec/draft/source-requirements.md#:~:text=%E2%9C%93-,Enforced%20change%20management%20process,-The%20SCS%20MUST) it's the control that gets enforced and is turned into a [property](https://github.com/slsa-framework/slsa/blob/main/docs/spec/draft/source-requirements.md#additional-properties) embedded within the VSA.  The property doesn't have any of the raw data associated with it.

Meanwhile, in the source PoC the provenance does list something it calls [controls](https://github.com/slsa-framework/slsa-source-poc/blob/main/DESIGN.md#:~:text=main%22%2C-,%22controls%22%3A%20%5B,-%7B%0A%20%20%20%20%20%20%20%20%22name%22) which contain _a bit_ more information and let us know if that property has been continuously enforced on the referenced branch or not.

There could certainly be additional information that we want to store (you could imagine storing the entire REST response from the GH API for the 'CONTINUITY_ENFORCED' control in another attestation for example).  Also note, just because language is hard the 'CONTINUITY_ENFORCED' property there refers to _branch_ continuity.  So that might need to be tightened up.

docs/spec/draft/source-requirements.md
```diff
@@ -191,7 +191,7 @@ For systems like GitHub or GitLab, this can be accomplished by enabling branch p
 
 It MUST NOT be possible to delete the entire repository (including all branches) and replace it with different source.
 
-Continuity exceptions are allowed via the [safe expunging process](#safe-expunging-process).
+These branch protection enforcements MUST be verified for [continuity](#revision-control-continuity).
```
zachariahcox, line 194: I think this section is only talking about "no force pushes" which is sort of different than control continuity as I understand it.

arewm:
Closing in favor of #1381.

PR #1351, opened 2025-04-24
content: draft: Add onboarding blurb, clarify 'branch' definition.

TomHennen:
The content of a revision is built over time. The controls in place today may be more strict than the controls in place yesterday.

Users of existing repositories will need to be able to onboard those repositories with some understanding of
what claims are being made about prior revisions.

This is a first attempt at handling that issue.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -26,7 +26,9 @@ Consumers can examine the various source provenance attestations to determine if
 | Source Control System (SCS) | A suite of tools and services (self-hosted or SaaS) relied upon by the organization to produce new revisions of the source. The role of the SCS may be fulfilled by a single service (e.g., GitHub / GitLab) or rely on a combination of services (e.g., GitLab with Gerrit code reviews, GitHub with OpenSSF Scorecard, etc).
 | Source Provenance | Information about how a revision came to exist, where it was hosted, when it was generated, what process was used, who the contributors were, and what parent revisions it was based on.
 | Repository / Repo | A uniquely identifiable instance of a VCS. The repository controls access to the Source in the VCS. The objective of a repository is to reflect the intent of the organization that controls it.
-| Branch | A named pointer to a revision. Branches may be modified by authorized actors. Branches may have different security requirements.
+| Branch | A named pointer to a revision where the revision pointed to changes
+over time. Branches may be modified by authorized actors. Different branches may
+have different security requirements.
 | Change | A set of modifications to the source in a specific context. A change can be proposed and reviewed before being accepted.
 | Change History | A record of the history of revisions that preceded a specific revision.
 | Push / upload / publish | When an actor authenticates to a Repository to add or modify content. Typically makes a new revision reachable from a branch.
@@ -43,6 +45,14 @@ Consumers can examine the various source provenance attestations to determine if
 | Proposer | An actor that proposes (or uploads) a particular change to the source.
 | Reviewer / Voter / Approver | An actor that reviews (or votes on) a particular change to the source.
 | Merger | An actor that applies a change to the source. This actor may be the proposer.
+
+## Onboarding
+
+When onboarding a branch to the SLSA Source Track or increasing the level of
+that branch, organizations are making claims about how the branch is managed
+from that time or revision forward.
+
+No claims are made for prior revisions.
 
 ## Safe Expunging Process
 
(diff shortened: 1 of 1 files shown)


docs/spec/draft/source-requirements.md
```diff
@@ -26,7 +26,7 @@ Consumers can examine the various source provenance attestations to determine if
 | Source Control System (SCS) | A suite of tools and services (self-hosted or SaaS) relied upon by the organization to produce new revisions of the source. The role of the SCS may be fulfilled by a single service (e.g., GitHub / GitLab) or rely on a combination of services (e.g., GitLab with Gerrit code reviews, GitHub with OpenSSF Scorecard, etc).
 | Source Provenance | Information about how a revision came to exist, where it was hosted, when it was generated, what process was used, who the contributors were, and what parent revisions it was based on.
 | Repository / Repo | A uniquely identifiable instance of a VCS. The repository controls access to the Source in the VCS. The objective of a repository is to reflect the intent of the organization that controls it.
-| Branch | A named pointer to a revision. Branches may be modified by authorized actors. Branches may have different security requirements.
+| Branch | A named pointer to a revision where the revision pointed to changes over time. Branches may be modified by authorized actors. Different branches may have different security requirements.
```
zachariahcox, line 29:
```suggestion
| Branch | A named pointer to a revision. Branches may be modified by authorized actors. Different branches may have different security requirements.
```
This section looks like it's missing part of a commit?

zachariahcox: Also do we want to define a word for "branches committed to slsa source track" ?

docs/spec/draft/source-requirements.md
```diff
@@ -44,6 +44,14 @@ Consumers can examine the various source provenance attestations to determine if
 | Reviewer / Voter / Approver | An actor that reviews (or votes on) a particular change to the source.
 | Merger | An actor that applies a change to the source. This actor may be the proposer.
 
+## Onboarding
+
+When onboarding a branch to the SLSA Source Track or increasing the level of
+that branch, organizations are making claims about how the branch is managed
+from that time or revision forward.
```
arewm, line 51: I make a comment about this when I am trying to centralize continuity in #1349 as well.

docs/spec/draft/source-requirements.md
TomHennen: Oh I actually wanted to include the "where the revision pointed to changes over time" so maybe it's just bad grammar.

I wanted to have some sense that the branch pointer changes (it doesn't always point to the same thing).  Maybe it was fine as is?

docs/spec/draft/source-requirements.md
TomHennen: Oh, so I wonder if we can just drop this change if we take your PR?

arewm: Does my change sufficiently describe onboarding? I don't use that word specifically.

docs/spec/draft/source-requirements.md
zachariahcox: ah! i see!

maybe, a movable pointer to a revision?

docs/spec/draft/source-requirements.md
TomHennen: Oh ok.  I wasn't sure if you were suggesting this was redundant.  Perhaps you're saying it can be referenced by the continuity work?

docs/spec/draft/source-requirements.md
TomHennen: Yes I like that better.  I tweaked it a bit more "A named, _movable_, pointer" and "Branches may be modified _to point to different revisions_ by authorized actors"

docs/spec/draft/source-requirements.md
arewm: I was suggesting that they might be redundant. We don't have to assess redundancy immediately though.

TomHennen: Ok.  Let's merge then and see when we do a full review how it reads?

arewm: Sounds good.

arewm: I suspect that onboarding will be good here but that we will add one more xref to state that onboarding is establishing the start of continuity.

arewm (approved):

docs/spec/draft/source-requirements.md
TomHennen: So I think the idea is:

1. Merge this.
2. Add an xref
?

zachariahcox (approved):
Looks good to me.

PR #1359, opened 2025-04-29
fix: remove reference to previous version being level 0.1

zachariahcox:
This pull request removes outdated explanatory notes from the `levels.md` documentation file to streamline and clarify the content.

I think we should remove this reference to 0.1 being the "previous version." 
From context it may be that "previous version" is currently encoded somewhere to always be 0.1. 
Edit: found it and deleted it.

For reference, I was poking around here: https://slsa.dev/spec/v1.1/levels

It has been a full release, so I think it is also fine to remove the entire note.


arewm (commented):
Do you want to remove this or to change the previous version to be 1.0? We are not removing the link but the target of the link ... unless nothing was using this target.

arewm (approved):
I don't _necessarily_ have a problem with this change. Another option could be to modify the removed block to say 

> Note: A [previous version] of the specification

The question that comes to mind is "how long do we need to start calling out differences between v0.1/2 and the current version?" This was a big change... but do SLSA consumers need to be reminded of them?

The message about a future source track is still relevant.

I ended up approving it because this content is still visible in the 1.0 spec pages.

lehors:
> Another option could be to modify the removed block to say
> 
> > Note: A [previous version] of the specification

I had the same idea and think that would be a better way to address the issue on the mere basis that we're now modifying a published document which should be left alone as much as possible, although obviously we're not talking about a significant change.

lehors:
> I think those are per file and that appeared to be the only use.

For sure.

TomHennen:
How are we feeling about this change?  Is it ready to merge or is there a bit more work to do?

arewm:
Since we are modifying a published version, can we keep the reference in but not use the previous version? I think that would just be a one-word change to

```
> Note: A [previous version] of the specification used a single unnamed track,
```

zachariahcox:
@arewm 's idea is way better. Let's do that!

docs/spec/v1.1/levels.md
```diff
@@ -27,13 +27,6 @@ tracks without invalidating previous levels.
 | [Build L2]  | Signed provenance, generated by a hosted build platform | Tampering after the build
 | [Build L3]  | Hardened build platform | Tampering during the build
 
-<!-- For comparison: a future Build L4's focus might be reproducibility or
-hermeticity or completeness of provenance -->
-
-> Note: The [previous version] of the specification used a single unnamed track,
-> SLSA 1–4. For version 1.0, the Source aspects were removed to focus on the
-> Build track. A Source track may be added in [future versions].
```
zachariahcox, line 35: so, we could just remove the "The" and replace with "A" here. 

That would still leave a kind of non-sequitur comment about the differences between 0.1 and 1.0.

PR #1362, opened 2025-05-02
content: draft: Expand 'organization requirements'

### LINKED ISSUE #1361: Better organize & motivate source track requirements
I was chatting with @MarkLodato today about the source track requirements and one thing that came up was how it was somewhat confusing (and maybe non-optimal) to have the ['Enforced change management process'](https://slsa.dev/spec/draft/source-requirements#:~:text=Enforced%20change%20management%20process) talk about just providing the _capability_ that orgs can enforce additional requirements.

I think part of the problem generally is that we might have crammed too much stuff into the SCS requirements table and/or didn't talk enough about what producers need to do with the capabilities of the SCS.

Suggestion:

1. Have the 'Enforced change management process' requirement read a bit more like the ['Isolated'](https://slsa.dev/spec/v1.1/requirements#isolation-strength:~:text=%E2%9C%93-,Isolated,-The%20build%20platform) requirement in that the SCS MUST ensure organization defined technical controls are enforced for changes to protected branches.
1. Expand 'producer' requirements to include requirements to specify which branches are protected at L2+ (moving it out of being defined in the SCS requirements)
1. Under 'producer' define technical controls that consumers can form expectations about (somewhat mirroring ['Follow a consistent build process'](https://slsa.dev/spec/v1.1/requirements#follow-a-consistent-build-process) which can leverage whatever the 'Enforced change management process' becomes under the SCS requirements.    Alternatively have the producer make explicit claims about what consumers can expect (e.g. by using ['policy'](https://github.com/slsa-framework/slsa-source-poc/blob/main/DESIGN.md#policy) which the SCS must enforce.

TomHennen:
Require orgs to:

1. Specify which branches and tags are protected (L2+)
2. Specify what additional controls they want users to expect (L3+)

_Allow_ orgs to define properties to be included in the VSA to communicate which controls they've enabled.

Updated the 'Enforced change management process' to be require the SCS enforce the technical controls from #2.

Moved "safe expunging process" into this table (it was an org requirement anyways). Only requiring at L2+ because that's when branch continuity comes into play anyways (prior to that folks can technically make whatever changes they want to a repo).

fixes #1361

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -179,24 +179,19 @@ Organizations SHOULD prefer to make logs public if possible.
 <tr id="specify-control-expectations"><td>Specify control expectations<td>
 
 The organization MUST specify what technical controls consumers can expect to be
-enforced for revisions in each branch.
+enforced for revisions in each branch using the
+[Enforced change management process](#enforced-change-management-process).
 
 For example, an organization may wish consumers to form an expectation that
 revisions on 'main' require unit tests to have passed prior to merge.  The
 organization would then configure the SCS to enforce this requirement and
-embed the relevant metadata in the [source summary](#summary-attestation) and/or
-[source provenance](#provenance-attestations) attestations for all affected
-revisions.  Consumers of those revisions would be able to determine if these
-expectations have been met via those attestations.
+embed the `USER_SOURCE_UNIT_TESTED` tag in the
+[source summary](#summary-attestation) and have the SCS store a corresponding
+[test result attestation].  Consumers of those revisions would be able to
+determine if these expectations have been met by looking for the
+`USER_SOURCE_UNIT_TESTED` tag in the VSA.
 
-Organizations SHOULD NOT specify technical controls they do not wish consumers
-to form expectations on.
-
-For example, an organization may be conducting a trial of a new control and
-they are not yet sure if they will keep it, or it may have implemented a
-proprietary control it does not want to publicize. In these cases the
-organization would not configure the SCS to embed metadata about this control
-into the attestations.
+[test result attestation]: https://github.com/in-toto/attestation/blob/main/spec/predicates/test-result.md
 
 <td><td><td>✓<td>✓
 
@@ -317,7 +312,7 @@ For example, this would happen if the revision was created on another SCS,
 or if the revision was not the result of an accepted change management process.
 
 <td><td><td>✓<td>✓
-<tr id="change-management-process"><td>Enforced change management process<td>
+<tr id="enforced-change-management-process"><td>Enforced change management process<td>
 
 The SCS MUST
 
@@ -379,7 +374,7 @@ Examples:
 
 ### Provide a change management tool
 
-The change management tool MUST be able to authoritatively state that each new revision reachable from the protected branch represents only the changes managed via the [process](#change-management-process).
+The change management tool MUST be able to authoritatively state that each new revision reachable from the protected branch represents only the changes managed via the [process](#enforced-change-management-process).
 
 <table>
 <tr><th>Requirement<th>Description<th>L1<th>L2<th>L3<th>L4
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -319,12 +319,24 @@ or if the revision was not the result of an accepted change management process.
 <td><td><td>✓<td>✓
 <tr id="change-management-process"><td>Enforced change management process<td>
 
-The SCS MUST ensure organization-defined technical controls are enforced for
-changes made to protected branches.  It MUST also embed relevant metadata about
-these controls in the [source summary](#summary-attestation) and/or
-[source provenance](#provenance-attestations) attestations
+The SCS MUST
 
-For example, this could be accomplished by:
+-   Ensure organization-defined technical controls are enforced for changes made
+   to protected branches.
+-   Allow organizations to specify
+   [additional tags](#additional-tags) to be included in the
+   [source summary](#summary-attestation) when the corresponding controls are
+enforced.
+-   Allow organizations to distribute additional attestations related to their
+   technical controls to consumers authorized to access the corresponding source
+   revision.
+
+The SCS MUST NOT allow organization specified tags to begin with any value other
+than `USER_SOURCE_` or `INTERNAL_USER_` unless the SCS endorses the veracity of
+any corresponding claims.
+
+Enforcement of the organization defined technical controls could be accomplished
+by, for example:
 
 -   The configuration of branch protection rules (e.g.[GitHub](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets), [GitLab](https://docs.gitlab.com/ee/user/project/repository/branches/protected.html)) which require additional checks to 'pass'
     (e.g. unit tests, linters), or
@@ -427,6 +439,8 @@ MAY include additional properties as asserted by the SCS.  The SCS MUST include 
 
 The SCS MAY issue these attestations based on its understanding of the underlying system (e.g. based on design docs, security reviews, etc...),
 but at SLSA Source Level 3 MUST use tamper-proof [provenance attestations](#provenance-attestations) appropriate to their SCS when making the assessment.
+
+#### Additional tags
 
 The SLSA source track MAY create additional tags to include in `verifiedLevels` which attest
 to other properties of a revision (e.g. if it was code reviewed).  All SLSA source tags will start with `SLSA_SOURCE_`.  Consumers MAY assume all SLSA source tags are meant
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -185,11 +185,12 @@ enforced for revisions in each branch using the
 For example, an organization may claim that revisions on `main` passed unit
 tests before being accepted.  The organization could then configure the SCS to
 enforce this requirement and store corresponding [test result attestations] for
-all affected revisions.  They may then embed the `USER_SOURCE_UNIT_TESTED` tag
-in the [source summary attestations](#summary-attestation). Consumers would then
-expect that future revisions on `main` have been united tested and determine if
-that expectation has been met by looking for the `USER_SOURCE_UNIT_TESTED` tag
-in the VSAs and, if desired, consult the [test result attestations] as well.
+all affected revisions.  They may then embed the `USER_SOURCE_UNIT_TESTED`
+property in the [source summary attestations](#summary-attestation). Consumers
+would then expect that future revisions on `main` have been united tested and
+determine if that expectation has been met by looking for the
+`USER_SOURCE_UNIT_TESTED` property in the VSAs and, if desired, consult the
+[test result attestations] as well.
 
 [test result attestations]: https://github.com/in-toto/attestation/blob/main/spec/predicates/test-result.md
 
@@ -319,16 +320,16 @@ The SCS MUST
 -   Ensure organization-defined technical controls are enforced for changes made
    to protected branches.
 -   Allow organizations to specify
-   [additional tags](#additional-tags) to be included in the
+   [additional properties](#additional-properties) to be included in the
    [source summary](#summary-attestation) when the corresponding controls are
 enforced.
 -   Allow organizations to distribute additional attestations related to their
    technical controls to consumers authorized to access the corresponding source
    revision.
 
-The SCS MUST NOT allow organization specified tags to begin with any value other
-than `USER_SOURCE_` or `INTERNAL_USER_` unless the SCS endorses the veracity of
-any corresponding claims.
+The SCS MUST NOT allow organization specified properties to begin with any value
+other than `USER_SOURCE_` or `INTERNAL_USER_` unless the SCS endorses the
+veracity of any corresponding claims.
 
 Enforcement of the organization-defined technical controls could be accomplished
 by, for example:
@@ -435,22 +436,24 @@ MAY include additional properties as asserted by the SCS.  The SCS MUST include 
 The SCS MAY issue these attestations based on its understanding of the underlying system (e.g. based on design docs, security reviews, etc...),
 but at SLSA Source Level 3 MUST use tamper-proof [provenance attestations](#provenance-attestations) appropriate to their SCS when making the assessment.
 
-#### Additional tags
+#### Additional properties
 
-The SLSA source track MAY create additional tags to include in `verifiedLevels` which attest
-to other properties of a revision (e.g. if it was code reviewed).  All SLSA source tags will start with `SLSA_SOURCE_`.  Consumers MAY assume all SLSA source tags are meant
+The SLSA source track MAY create additional properties to include in
+`verifiedLevels` which attest to other claims concerning a revision (e.g. if it
+was code reviewed). All SLSA source properties will start with `SLSA_SOURCE_`.
+Consumers MAY assume all SLSA source properties are meant
 
-The SCS MAY embed user-provided tags within `verifiedLevels` corresponding to
+The SCS MAY embed user-provided properties within `verifiedLevels` corresponding to
 technical controls enforced by the SCS if they are prefixed with:
 
--   `USER_SOURCE_` to indicate a tag that is meant for consumption by external
-   consumers.
--   `INTERNAL_USER_` to indicate a tag that is not meant for consumption by
+-   `USER_SOURCE_` to indicate a property that is meant for consumption by
+   external consumers.
+-   `INTERNAL_USER_` to indicate a property that is not meant for consumption by
    external consumers.
 
-The meaning of the tags is left entirely to the organization. Inclusion of user
-provided tags within `verifiedLevels` SHOULD NOT be considered an endorsement of
-the veracity of the organization defined claim by the SCS.
+The meaning of the properties is left entirely to the organization. Inclusion of
+user provided properties within `verifiedLevels` SHOULD NOT be considered an
+endorsement of the veracity of the organization defined property by the SCS.
 
 #### Populating source_refs
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -182,15 +182,14 @@ The organization MUST specify what technical controls consumers can expect to be
 enforced for revisions in each branch using the
 [Enforced change management process](#enforced-change-management-process).
 
-For example, an organization may claim that revisions on `main` passed unit tests before being accepted.  The organization could then
-configure the SCS to enforce this requirement and store
-corresponding [test result attestations] for all affected revisions. 
-They may then embed the `USER_SOURCE_UNIT_TESTED` tag in the
-[source summary attestations](#summary-attestation).
-Consumers would be able to establish the expectation that revisions on `main`
-have been united tested and determine if that expectation has been met by
-looking for the `USER_SOURCE_UNIT_TESTED` tag in the VSA and, if desired,
-consult the [test result attestations] as well.
+For example, an organization may claim that revisions on `main` passed unit
+tests before being accepted.  The organization could then configure the SCS to
+enforce this requirement and store corresponding [test result attestations] for
+all affected revisions.  They may then embed the `USER_SOURCE_UNIT_TESTED` tag
+in the [source summary attestations](#summary-attestation). Consumers would then
+expect that future revisions on `main` have been united tested and determine if
+that expectation has been met by looking for the `USER_SOURCE_UNIT_TESTED` tag
+in the VSAs and, if desired, consult the [test result attestations] as well.
 
 [test result attestations]: https://github.com/in-toto/attestation/blob/main/spec/predicates/test-result.md
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -179,8 +179,8 @@ Organizations SHOULD prefer to make logs public if possible.
 
 <tr id="specify-control-expectations"><td>Specify control expectations<td>
 
-The organization MUST specify what technical controls consumers can expect
-revisions on a branch to have met.
+The organization MUST specify what technical controls consumers can expect to be
+enforced for revisions in each branch.
 
 For example, an organization may wish consumers to form an expectation that
 revisions on 'main' require unit tests to have passed prior to merge.  The
@@ -189,6 +189,15 @@ embed the relevant metadata in the [source summary](#summary-attestation) and/or
 [source provenance](#provenance-attestations) attestations for all affected
 revisions.  Consumers of those revisions would be able to determine if these
 expectations have been met via those attestations.
+
+Organizations SHOULD NOT specify technical controls they do not wish consumers
+to form expectations on.
+
+For example, an organization may be conducting a trial of a new control and
+they are not yet sure if they will keep it, or it may have implemented a
+proprietary control it does not want to publicize. In these cases the
+organization would not configure the SCS to embed metadata about this control
+into the attestations.
 
 <td><td><td>✓<td>✓
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -52,33 +52,6 @@ from that time or revision forward.
 
 No claims are made for prior revisions.
 
-## Safe Expunging Process
-
-SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content,
-but the organization MUST only allow these changes in order to meet legal or privacy compliance requirements.
-Content changed under this process includes changing files, history, references, or any other metadata stored by the SCS.
-
-### Warning
-
-Removing a revision from a repository is similar to deleting a package version from a registry: it's almost impossible to estimate the amount of downstream supply chain impact.
-For example, in VCSs like Git, removal of a revision changes the object IDs of all subsequent revisions that were built on top of it, breaking downstream consumers' ability to refer to source they've already integrated into their products.
-
-It may be the case that the specific set of changes targeted by a legal takedown can be expunged in ways that do not impact consumed revisions, which can mitigate these problems.
-
-It is also the case that removing content from a repository won't necessarily remove it everywhere.
-The content may still exist in other copies of the repository, either in backups or on developer machines.
-
-### Process
-
-An organization MUST document the Safe Expunging Process and describe how requests and actions are tracked and SHOULD log the fact that content was removed.
-Different organizations and tech stacks may have different approaches to the problem.
-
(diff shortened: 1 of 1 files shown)


arewm (approved):
This looks good. I was too optimistic for keeping the organization requirements to be table-less. This breakout makes sense.

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +124,74 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+<td>✓<td>✓<td>✓<td>✓
+
+<tr id="choose-process"><td>Choose an appropriate change management process<td>
+
+An organization producing source revisions MUST implement a change management
+process to ensure changes to source matches the organization's intent.
+
+<td><td>✓<td>✓<td>✓
+
+<tr id="specify-protection"><td>Specify which branches and tags are protected<td>
```
zachariahcox, line 136: we technically don't define "tag" in the definitions section. 
Generally, we mean it to be a named source context. Just like our definition of branch, it is a named pointer to a revision. 
We could talk about it separately from the perspective of git, but maybe not strictly relevant. 
```suggestion
<tr id="specify-protection"><td>Specify which branches are protected<td>
```

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +124,74 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+
+The organization MUST indicated which branches and tags it protects with Source
+Level 2+ controls.  Only those indicated branches and tags will be eligible for
+Source Level 2+.
```
zachariahcox, line 140:
```suggestion
Level 2+ controls.
```

I'm guessing we should drop this -- elsewhere I think we say anything with a VSA could have a SLSA level.

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +124,74 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+
+For example, if an organization has branches 'main' and 'experimental' and it
+intends for 'main' to be protected then it MUST indicate to the SCS that 'main'
+should be protected. From that point forward revisions on 'main' will be
+eligible for Source Level 2+ while revisions made solely on 'experimental' will
+not.
```
zachariahcox, line 147: I guess we need the explicit declaration of branch intent to help consumers "know where to start".

IE, I don't know that you'd be able to find a VSA before a sha, and the sha comes from specific branch pointers typically. You'd probably read what branch you're "supposed to use" then wire up your build pipeline to consume it. 

Then, in the normal verification flows, you'd start with the sha-pointed-to-by-the-ref-or-lockfile and check for VSA, without any further strict requirement on branch setup.

docs/spec/draft/source-requirements.md
```diff
@@ -271,8 +310,10 @@ or if the revision was not the result of an accepted change management process.
 <td><td><td>✓<td>✓
 <tr id="change-management-process"><td>Enforced change management process<td>
 
-The SCS MUST provide a mechanism for organizations to enforce additional
-technical controls which govern changes to a [branch](#definitions).
+The SCS MUST ensure organization defined technical controls are enforced for
+changes made to protected branches.  It MUST also embed relevant metadata about
+these controls in the [source summary](#summary-attestation) and/or
+[source provenance](#provenance-attestations) attestations
```
zachariahcox, line 316:
```suggestion
The SCS MUST ensure organization-defined technical controls are enforced for
changes made to protected branches.  It MUST also embed relevant metadata about
these controls in the [source summary](#summary-attestation) and/or
[source provenance](#provenance-attestations) attestations
```

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +124,74 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+The application of the safe expunging process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
+and regulations which may require the change to be kept private to the extent possible.
+Organizations SHOULD prefer to make logs public if possible.
+
+<td><td>✓<td>✓<td>✓
+
+<tr id="specify-control-expectations"><td>Specify control expectations<td>
+
+The organization MUST specify what technical controls consumers can expect
+revisions on a branch to have met.
```
zachariahcox, line 182:
```suggestion
The organization MUST specify what technical controls are enforced on revisions in each branch.
```

minor nit: "to have met" is hard to parse?

docs/spec/draft/source-requirements.md
TomHennen: Hmm, so this is a preview of what I'd like to improve about how we currently specify tag requirements.

E.g. [here](https://slsa.dev/spec/draft/source-requirements#future-considerations:~:text=If%20a%20tag%20is%20used%20to%20identify%20a%20specific%20commit%20to%20external%20systems%2C%20it%20MUST%20NOT%20be%20possible%20to%20move%20or%20delete%20those%20tags.) we say "If a tag is used to identify a specific commit to external systems, it MUST NOT be possible to move or delete those tags" but it's not clear how the SCS can _know_ if a tag is used to specify a commit to an external system.

What I'm trying to do here is put that requirement on the organization.  E.g. they have to indicate which tags get this protection.  So, since we do talk about tags and how they need to meet different requirements depending on how they're used it seems relevant to include.

I can definitely add a tag definition.

WDYT?

docs/spec/draft/source-requirements.md
TomHennen: Yeah, I wasn't trying to say they couldn't get a SLSA level at all, they just can't get Level 2+.

But what I think you're really saying is that it's redundant?

docs/spec/draft/source-requirements.md
TomHennen: I think I agree, but I'm not sure. :)

Are you looking for any changes?  I think this is getting at what you've described.

docs/spec/draft/source-requirements.md
TomHennen: Hmm, so what I was thinking is something like:

An org enforces controls A, B, C on a repo.  They're still not sure about C, or it's private or something.   So 1. they don't want to publicize C, and 2. they don't want users forming expectations (#1365) that C will always be there.

So while the two versions (mine and your suggestion) might seem the same, the key difference is that when I read the proposed version it requires _all_ controls to be specified, not just the ones "consumers can expect".

Maybe  "The organization MUST specify what technical controls consumers can expect to be enforced for revisions in each branch."  ?

TomHennen: and then maybe I need to add an example about controls that don't need to be specified?

docs/spec/draft/source-requirements.md
TomHennen: Added a definition for 'tag' and tweaked the 'branch' definition. PTAL

docs/spec/draft/source-requirements.md
TomHennen: Gave it a shot. PTAL?

docs/spec/draft/source-requirements.md
zachariahcox: Yes, and also that current or previous branch reachability is not required for any VSA.
If your process requires reachability, cool, but it's not a required part of the spec.

docs/spec/draft/source-requirements.md
```diff
@@ -26,7 +26,8 @@ Consumers can examine the various source provenance attestations to determine if
...
 | Source Provenance | Information about how a revision came to exist, where it was hosted, when it was generated, what process was used, who the contributors were, and what parent revisions it was based on.
 | Repository / Repo | A uniquely identifiable instance of a VCS. The repository controls access to the Source in the VCS. The objective of a repository is to reflect the intent of the organization that controls it.
-| Branch | A named, moveable, pointer to a revision. Branches may be modified to point to different revisions by authorized actors. Different branches may have different security requirements.
+| Branch | A named, moveable, pointer to a revision that tracks development in the named context over time. Branches may be modified to point to different revisions by authorized actors. Different branches may have different security requirements.
+| Tag | A named pointer to a revision that does not typically move. Tags may be created (and sometimes modified) by authorized actors. Different tags may be used for different purposes (e.g. denote a specific version of the software, or the 'latest' release).
```
zachariahcox, line 30: Does any VCS have this concept besides git?

TomHennen: The tag concept?  Yes.

* Subversion: https://tortoisesvn.net/docs/release/TortoiseSVN_en/tsvn-dug-branchtag.html
* Perforce: https://help.perforce.com/helix-core/server-apps/p4guide/current/Content/P4Guide/labels.revision.html
* CVS: https://www.gnu.org/software/trans-coord/manual/cvs/html_node/Tags.html#:~:text=There%20is%20seldom%20reason%20to%20tag%20a%20file%20in%20isolation.%20A%20more%20common%20use%20is%20to%20tag%20all%20the%20files%20that%20constitute%20a%20module%20with%20the%20same%20tag%20at%20strategic%20points%20in%20the%20development%20life%2Dcycle%2C%20such%20as%20when%20a%20release%20is%20made.
* Mercurial: https://wiki.mercurial-scm.org/Tag

zachariahcox: What if we just updated the branch definition to say "on git, when the spec says 'branch' we mean all refs, including refs/tags"

TomHennen: I guess because we'd discussed how people use tags differently from branches and were hoping to apply different requirements (e.g. tag hygiene) to tags but not branches.

We also don't expect folks to have continuity for tags.

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +125,83 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+<tr id="choose-process"><td>Choose an appropriate change management process<td>
+
+An organization producing source revisions MUST implement a change management
+process to ensure changes to source matches the organization's intent.
+
+<td><td>✓<td>✓<td>✓
+
+<tr id="specify-protection"><td>Specify which branches and tags are protected<td>
+
+The organization MUST indicated which branches and tags it protects with Source
```
zachariahcox, line 139:
```suggestion
The organization MUST indicate which branches and tags it protects with Source
```

docs/spec/draft/source-requirements.md
zachariahcox: That's a neat list!
My question was not very clear -- in all those VCS products, the producer is responsible for guaranteeing or not what labels/tags mean vs branches. 
Functionally, I think all products allow writers to change them. This is quite different from revision ids which are not mutable, even by producers/writers. 

What if we just said:
```suggestion
| Tag | A named pointer to a revision that does not typically move. Similar to branches, tags may be modified by authorized actors. Tags are often used by producers to indicate a more permanent name for a revision.
```

TomHennen: Oh yes I like this. :)

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +125,83 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+The organization MUST specify what technical controls consumers can expect to be
+enforced for revisions in each branch.
+
+For example, an organization may wish consumers to form an expectation that
+revisions on 'main' require unit tests to have passed prior to merge.  The
+organization would then configure the SCS to enforce this requirement and
+embed the relevant metadata in the [source summary](#summary-attestation) and/or
+[source provenance](#provenance-attestations) attestations for all affected
+revisions.  Consumers of those revisions would be able to determine if these
+expectations have been met via those attestations.
```
zachariahcox, line 191: minor. The "to form an expectation" verbiage is a little wonky. 
I think the producer is making a promise, and that promise should be verified by consumers. 

```suggestion
For example, an organization may claim that revisions on 'main' require unit 
tests to have passed prior to merge.  The organization would then configure 
the SCS to enforce this requirement and record evidence of the enforcement in 
the [source summary](#summary-attestation) and/or
[source provenance](#provenance-attestations) attestation for all affected
revisions.  Consumers of those revisions can verify that the promises were 
met via the evidence in those attestations.
```

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +125,83 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+
+Organizations SHOULD NOT specify technical controls they do not wish consumers
+to form expectations on.
```
zachariahcox, line 194: What's the intent of this one? 
I'm thinking "expectations" is just "what claims you should plan to check for in your verifications", but that's sort of different than what might be provided as evidence on an attestation. 
EG, there may be extra claims that apply only to machine-produced code that don't apply to humans. Those claims may or may not exist in an attestation -- they cannot be 100% required to exist on any attestation, but they are an important part of the guarantees offered by the producer, so they should be sort of "expected" in a sense.

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +125,83 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+
+For example, an organization may be conducting a trial of a new control and
+they are not yet sure if they will keep it, or it may have implemented a
+proprietary control it does not want to publicize. In these cases the
+organization would not configure the SCS to embed metadata about this control
+into the attestations.
```
zachariahcox, line 200: one of my expectations is that the source provenance will largely be authored by authorities. 
I would not expect the PR product to allow configuration of which claims it emits for example -- it should emit whatever it thinks is relevant about the operation from its own perspective. 
For example, if a control were enabled, but an authorized user bypassed it, it should not be configurable whether or not to emit that evidence in a provenance attestation -- the PR tool's reputation is on the line. 

I do agree that producers will need to ultimately declare that a subset of their enabled technical controls are guaranteed as evidence of SLSA level. Other control information may be included as optional fun facts.

zachariahcox (commented):
Thanks Tom! I left some non-blocking notes and comments. Please take or leave any of them as you see fit.

docs/spec/draft/source-requirements.md
adityasaky: I tend to agree; I think consumers should form expectations based on what an organization explicitly communicates rather than what the organization may be producing above and beyond. FWIW, a trial of a control (from the example further down) would likely also include checking that the metadata can be embedded and used within the SCS? Not allowing such a scenario may lead to partially untested claims being rolled out when the organization switches on a new control.

docs/spec/draft/source-requirements.md
```diff
@@ -271,8 +320,10 @@ or if the revision was not the result of an accepted change management process.
 <td><td><td>✓<td>✓
 <tr id="change-management-process"><td>Enforced change management process<td>
 
-The SCS MUST provide a mechanism for organizations to enforce additional
-technical controls which govern changes to a [branch](#definitions).
+The SCS MUST ensure organization-defined technical controls are enforced for
+changes made to protected branches.  It MUST also embed relevant metadata about
+these controls in the [source summary](#summary-attestation) and/or
+[source provenance](#provenance-attestations) attestations
```
adityasaky, line 326: I think this is just an "and" rather than "and/or" as at these levels the SCS is required to produce both summary and provenance attestations.

adityasaky: Ah, I think I misunderstood the "and/or". It's not about embedding the attestation itself but embedding the control in the attestation. In that sense, I wonder if we want the controls that the provenance is used to verify should be included in the provenance itself? Maybe it only belongs in the summary attestation?

docs/spec/draft/source-requirements.md
TomHennen: Ok, this is good feedback.  I've addressed it by:

1. Actually defining tag prefixes in the VSA that are reserved for users wishing to record claims in the VSA and how they can distinguish between stuff that's just for their own internal use vs stuff they actually want folks to set expectations for.
2. Allowing orgs to distribute additional attestations with supporting evidence if they want.

This comment thread is now homeless since the relevant text is gone.  I'm going to preemptively resolve but _please_ take a look at the updated language. :)

docs/spec/draft/source-requirements.md
TomHennen: Yes agreed.  I've done a lot of rewriting to this section to take this into account.  PTAL?

This thread is now outdated so I'm going to resolve. :)

docs/spec/draft/source-requirements.md
TomHennen: You're right.  I reworked this a lot with the other changes, hopefully it's better.

I do want to keep some of the "expectation" language as it relates to the change in #1365 which mirrors what we're doing for the build track.

docs/spec/draft/source-requirements.md
TomHennen: I think this does the trick (and this thread is outdated).  So closing it.

docs/spec/draft/source-requirements.md
TomHennen: Yes.  I think this is now more clear with the specific tags to use (and hopefully in combination with #1365).

docs/spec/draft/source-requirements.md
TomHennen: I'll take two 👍  and one "no blocking notes" comment from @zachariahcox as an indication I can close this. :)

adityasaky: in future, would it help if I left an agreeing comment rather than a reaction?

docs/spec/draft/source-requirements.md
```diff
@@ -371,15 +430,29 @@ SCSs that do not use cryptographic digests MUST define a canonical type that is
...
 
-Verifiers MAY issue these attestations based on their understanding of the underlying system (e.g. based on design docs, security reviews, etc...),
+The SCS MAY issue these attestations based on its understanding of the underlying system (e.g. based on design docs, security reviews, etc...),
 but at SLSA Source Level 3 MUST use tamper-proof [provenance attestations](#provenance-attestations) appropriate to their SCS when making the assessment.
 
+#### Additional tags
+
 The SLSA source track MAY create additional tags to include in `verifiedLevels` which attest
-to other properties of a revision (e.g. if it was code reviewed).  All SLSA source tags will start with `SLSA_SOURCE_`.
+to other properties of a revision (e.g. if it was code reviewed).  All SLSA source tags will start with `SLSA_SOURCE_`.  Consumers MAY assume all SLSA source tags are meant
```
adityasaky, line 443: this sentence trails off. Is the idea that any tag that starts with `SLSA_SOURCE_` can be used by the consumers in verifying VSA expectations?

docs/spec/draft/source-requirements.md
```diff
@@ -371,15 +430,29 @@ SCSs that do not use cryptographic digests MUST define a canonical type that is
...
+
+The SCS MAY embed user-provided tags within `verifiedLevels` corresponding to
```
adityasaky, line 445: Non-blocking suggestion, maybe we can discuss this after this PR is merged.

Do we want to support additional tags? Maybe the consumer can put `USER_SOURCE_` tags in another attestation (e.g., SCAI) that they then work out how to use in conjunction with the VSA? That avoids even accidental cases where `USER_SOURCE_` tags are ascribed to the organization by some party that's several layers removed from the spec for this and doesn't understand the intent.

I have a similar opinion for `INTERNAL_USER_`. I'm guessing this is to prevent consumers relying on a claim the organization doesn't (yet?) want to provide. However, there will still need to be some out of band exchange of expectations between the org and consumer (e.g., for the `SLSA_SOURCE_` tags), so perhaps this is additional complexity we can do without? Does this address any other scenario that I've missed?

docs/spec/draft/source-requirements.md
TomHennen: Ah, I can clean that up.  The idea is that all `SLSA_SOURCE_` tags are 'endorsed' by the SCS itself and should be well defined by SLSA.

docs/spec/draft/source-requirements.md
```diff
@@ -269,12 +314,26 @@ For example, this would happen if the revision was created on another SCS,
...
+   revision.
 
-The SCS MUST provide a mechanism for organizations to enforce additional
-technical controls which govern changes to a [branch](#definitions).
+The SCS MUST NOT allow organization specified tags to begin with any value other
+than `USER_SOURCE_` or `INTERNAL_USER_` unless the SCS endorses the veracity of
+any corresponding claims.
 
-For example, this could be accomplished by:
+Enforcement of the organization defined technical controls could be accomplished
```
zachariahcox, line 335:
```suggestion
Enforcement of the organization-defined technical controls could be accomplished
```

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +125,79 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+The organization MUST specify what technical controls consumers can expect to be
+enforced for revisions in each branch using the
+[Enforced change management process](#enforced-change-management-process).
+
+For example, an organization may wish claim that revisions on `main` require
+unit tests to have passed prior to merge.  The organization could then
+configure the SCS to enforce this requirement and embed the
+`USER_SOURCE_UNIT_TESTED` tag in the
+[source summary attestations](#summary-attestation) and have the SCS store
+corresponding [test result attestations] for all affected revisions.
```
zachariahcox, line 190:
```suggestion
configure the SCS to enforce this requirement and store
corresponding [test result attestations] for all affected revisions. 
They may then embed the `USER_SOURCE_UNIT_TESTED` tag in the
[source summary attestations](#summary-attestation).
```

docs/spec/draft/source-requirements.md
```diff
@@ -148,11 +125,79 @@ For example, if an organization wishes to produce revisions at Source Level 3,
...
+Consumers would be able to establish the expectation that revisions on `main`
+have been united tested and determine if that expectation has been met by
+looking for the `USER_SOURCE_UNIT_TESTED` tag in the VSA and, if desired,
+consult the [test result attestations] as well.
```
zachariahcox, line 194:
```suggestion
Consumers would be able to determine whether the control was enforced by
looking for the `USER_SOURCE_UNIT_TESTED` tag in the VSA and, if desired,
consulting the [test result attestations] as well.
```

docs/spec/draft/source-requirements.md
```diff
@@ -371,15 +430,29 @@ SCSs that do not use cryptographic digests MUST define a canonical type that is
...
+technical controls enforced by the SCS if they are prefixed with:
+
+-   `USER_SOURCE_` to indicate a tag that is meant for consumption by external
+   consumers.
+-   `INTERNAL_USER_` to indicate a tag that is not meant for consumption by
+   external consumers.
+
+The meaning of the tags is left entirely to the organization. Inclusion of user
+provided tags within `verifiedLevels` SHOULD NOT be considered an endorsement of
+the veracity of the organization defined claim by the SCS.
```
zachariahcox, line 455:
```suggestion
The meaning of the tags is left entirely to the organization. Inclusion of user
provided tags within `verifiedLevels` SHOULD NOT be considered an endorsement of
the veracity of the organization defined claim by the SCS.
```

The term "user" here is confusing me. 
Do we mean "producer" in this case? I don't think we mean developer or PR author or something, so it's not the same as "user provided timestamps".

These are claims introduced (probably) by the organization based on controls they've specified. Shouldn't they be just as trustworthy as any other claims based on tamper-resistant evidence?

zachariahcox (approved):
Generally, looks good. Thanks!!

I left one comment at the bottom about "user tags" (which we probably want to rename now that we have "tags" as a defined term?).

I still think "claims" might be the right mental model. 
"This issuer makes these claims, take 'em or leave 'em," etc.

docs/spec/draft/source-requirements.md
TomHennen: God catch, yes "org" or "producer" is better than "user" in this case.

Re-trustworthiness:

I'm making a distinction between trust in the SCS and trust in the Organization which can be different (it is for the source-poc).

docs/spec/draft/source-requirements.md
TomHennen: I've clarified the language some but I think we need to keep the expectation setting part.

It's not just about how they check if a control was implemented but also about how they know if they _should_ check.

So it goes something like:

1. I have no expectations.
2. I see a VSA that says "this thing on main is unit tested"
3. I now expect that future things on main are unit tested
4. I look for the corresponding claim for future revisions.

TomHennen:
> I left one comment at the bottom about "user tags" (which we probably want to rename now that we have "tags" as a defined term?).

Yeah good catch.  Renamed to 'properties' (it's what we call them already) where those properties represent claims.  That language might use some better cleaning later...

docs/spec/draft/source-requirements.md
TomHennen: Updated the language.  Based on other comments I think we're agreed about how trustworthy these are ("take it or leave it").

If not please reopen?

docs/spec/draft/source-requirements.md
TomHennen: I actually went ahead and removed this.  We're talking about additional properties here so there's no reason to talk about `SLSA_SOURCE_` stuff.  I've also gone ahead and indicated which prefixes MUST be used.

docs/spec/draft/source-requirements.md
TomHennen: Filed https://github.com/slsa-framework/slsa/issues/1369

docs/spec/draft/source-requirements.md
```diff
@@ -371,15 +429,33 @@ SCSs that do not use cryptographic digests MUST define a canonical type that is
...
+was code reviewed).
+
+The SCS MAY embed organization-provided properties within `verifiedLevels`
+corresponding to technical controls enforced by the SCS. If such properties are
+provided they MUST be prefixed with `ORG_SOURCE_` to distinguish them from other
+properties the SCS may wish use.
+
+-   `ORG_SOURCE_` to indicate a property that is meant for consumption by
+   external consumers.
+-   `ORG_SOURCE_INTERNAL_` to indicate a property that is not meant for
```
zachariahcox, line 452: It might be implied here that this distinction is only used in VSAs. The source prov would probably just write lists of data -- it doesn't necessarily know which are important.

zachariahcox: one interesting case to consider might be the same commit reached from two protected branches, both with different required rules.

TomHennen: > It might be implied here that this distinction is only used in VSAs. The source prov would probably just write lists of data -- it doesn't necessarily know which are important.

Yup, this is under the 'Summary attestation' section so it's just VSAs.

> one interesting case to consider might be the same commit reached from two protected branches, both with different required rules.

Can we leave this undefined and up to the SCS itself for now?

adityasaky (approved):
LGTM! I think we can merge this and work on #1369 and #1372 separately.

TomHennen:
Thanks for the reviews everyone.  I'm going to go ahead and merge (draft content) we can always make more changes later.

Issue #1352, opened 2025-04-24
Merge "Provide a change management tool" with the other requirements


TomHennen:
It's not really separate from the source control system itself.  Having it off by itself is somewhat confusing.

Also, merging with the other table will make it somewhat easier when talking about two-party review #1350.

arewm:
I just left a comment in 1350

> Should we include this type of verbiage elsewhere in the spec too (i.e. Enforced change management process)? Or should we say this as part of the Organization requirements?
>
> Is there an Organization requirement to establish change management processes to ensure consistency with their intent?

It should be the responsibility of the SCS to provide a change management tool but shouldn't it also be the responsibility of the organization to use that change management tool to formalize the organization's intent?

TomHennen:
> I just left a comment in 1350
> 
> > Should we include this type of verbiage elsewhere in the spec too (i.e. Enforced change management process)? Or should we say this as part of the Organization requirements?
> > Is there an Organization requirement to establish change management processes to ensure consistency with their intent?
> 
> It should be the responsibility of the SCS to provide a change management tool but shouldn't it also be the responsibility of the organization to use that change management tool to formalize the organization's intent?

Maybe https://github.com/slsa-framework/slsa/pull/1362 helps there?

Separately, I'm wondering what the [context](https://slsa.dev/spec/draft/source-requirements#:~:text=L4-,Context,-The%20change%20management) requirement really gets us.  Is it redundant with the others?  Can we drop it?

TomHennen:
We [discussed this](https://docs.google.com/document/d/1PwhekVB1iDpcgCQRNVN_aesoVdOiTruoebCs896aGxw/edit?tab=t.0#heading=h.kzk34jq57wgq) in today's spec meeting and the general consensus was that while timestamps are certainly useful in things like incident response, the source track might not be the right place to address that.  E.g. maybe things like enabling IR should go in the proposed [Platform operations track](https://slsa.dev/spec/v1.1/future-directions#platform-operations-track).

So we decided it would be simplest to remove the Verified Timestamp requirement entirely.

PR #1363, opened 2025-05-02
content: draft: merge timestamp req with source prov

TomHennen:
Merge the "Verified Timestamps" requirement with the source provenance requirement.  I think that's really where we need it?

refs #1352

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -261,6 +261,9 @@ At Source Level 3 Source Provenance MUST be created contemporaneously with the
 branch being updated to use that revision such that they provide a credible,
 auditable, record of changes.
 
+The Source Provenance MUST record a trustworthy timestamp indicating when the
+update occurred.  User-provided timestamps MUST NOT be used for this timestamp.
+
 If a consumer is authorized to access, they MUST be able to fetch the source
 provenance documents for relevant revisions.
 
@@ -331,12 +334,6 @@ The change management tool MUST record the "target" context for the change propo
 For example, for the git version control system, the change management tool MUST record the branch name that was updated.
 
 Branches may have differing security postures, and a change can be approved for one context while being unapproved for another.
-
-<td><td><td>✓<td>✓
-<tr id="verified-timestamps"><td>Verified Timestamps<td>
-
-The change management tool MUST record timestamps for all contributions and review-related activities.
-User-provided timestamps MUST NOT be used.
 
 <td><td><td>✓<td>✓
 </table>
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -260,7 +260,7 @@ from the perspective of the SCS.
 At Source Level 3 Source Provenance MUST be created contemporaneously with the
 branch being updated to use that revision such that they provide a credible,
 auditable, record of changes. The Source Provenance MUST record the time the
-update occurred. The user MUST NOT be able to falsify this timestamp.
+update occurred. The user MUST NOT be able to falsify this time.
 
 If a consumer is authorized to access, they MUST be able to fetch the source
 provenance documents for relevant revisions.
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -259,10 +259,8 @@ from the perspective of the SCS.
 
 At Source Level 3 Source Provenance MUST be created contemporaneously with the
 branch being updated to use that revision such that they provide a credible,
-auditable, record of changes.
-
-The Source Provenance MUST record a trustworthy timestamp indicating when the
-update occurred.  User-provided timestamps MUST NOT be used for this timestamp.
+auditable, record of changes. The Source Provenance MUST record the time the
+update occurred. User-provided timestamps MUST NOT be used for this timestamp.
 
 If a consumer is authorized to access, they MUST be able to fetch the source
 provenance documents for relevant revisions.
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -260,7 +260,7 @@ from the perspective of the SCS.
 At Source Level 3 Source Provenance MUST be created contemporaneously with the
 branch being updated to use that revision such that they provide a credible,
 auditable, record of changes. The Source Provenance MUST record the time the
-update occurred. User-provided timestamps MUST NOT be used for this timestamp.
+update occurred. The user MUST NOT be able to falsify this timestamp.
 
 If a consumer is authorized to access, they MUST be able to fetch the source
 provenance documents for relevant revisions.


TomHennen:
Note, I didn't move the context requirement because, [maybe we'll drop it](https://github.com/slsa-framework/slsa/issues/1352#issuecomment-2847749340)?

docs/spec/draft/source-requirements.md
```diff
@@ -261,6 +261,9 @@ At Source Level 3 Source Provenance MUST be created contemporaneously with the
 branch being updated to use that revision such that they provide a credible,
 auditable, record of changes.
 
+The Source Provenance MUST record a trustworthy timestamp indicating when the
+update occurred.  User-provided timestamps MUST NOT be used for this timestamp.
```
arewm, line 265: This is moved from the "change management tool" section into the provenance section. I think that it is important to have this clarification here, but do we also need it elsewhere related to the change management process?

TomHennen: Right I suppose the question is "where do we want them recorded?"

The current wording is fairly vague.  Do we just want them stored in logs the SCS maintains somewhere?  That could be OK?  For the GitHub source-poc story we might want to find some documentation (or state assumptions) about where the timestamps in the various APIs come from.

adityasaky: Couple questions :)

What counts as a trustworthy timestamp? And why do we require source provenance to include one (i.e., what properties are we looking for)? For example, if this is to ensure the provenance was produced at about the same time as the revision itself, maybe we should say that instead and let the implementation work out how to do it?

TomHennen: Good feedback. 

Since we have a 'contemporaneous' requirement this is meant to ensure we know when the update occurred.

1. Moved this up to the prior block to make that more clear.
2. Simplified the text about what this is for "record the time the update occurred"
3. Clarified that the user shouldn't be able to _falsify_ the timestamp instead of talking about where the time came from.

arewm (approved):

docs/spec/draft/source-requirements.md
TomHennen: I think we are still leaving it to the SCS to enforce the contemporaneous aspect.  This just records _when_ that happened (maybe this is nuanced distinction).

I do think knowing the time the update occurred is a helpful security property.  It allows correlation of updates to branches with other event logs which can be pretty helpful in incident response.

> In gittuf, the source provenance (such as a code review attestation) is required to be created ahead of the consumable revision being created.

I think you might be taking a broad view of source provenance when perhaps a more narrow view would help.  I'd say, what if a code review attestation isn't meant to meet the 'source provenance' attestation and instead you use another existing or new attestation to cover when the branch itself is updated to point at the new revision?

> We don't use time at all because of how spoofable it can be, instead we rely on this relativity for contemporaneous approvals.

It would be possible to use a TSA to meet this requirement, right?  Maybe that's impossible for gittuf?

Either way, this sounds like a discussion that is better handled in an issue?

zachariahcox (changes requested):
The change management tool needs to generate its own timestamps. 
It's also important to call out that are multiple activities to track, not just revision creation. 
At least the following are important: 
* push receive
* review receive
* commit create


I'm ok with adding more SCS-generated timestamp requirements, but we probably shouldn't remove the one for PR tools.

docs/spec/draft/source-requirements.md
```diff
@@ -332,12 +333,6 @@ For example, for the git version control system, the change management tool MUST
 
 Branches may have differing security postures, and a change can be approved for one context while being unapproved for another.
 
-<td><td><td>✓<td>✓
-<tr id="verified-timestamps"><td>Verified Timestamps<td>
```
zachariahcox, line 336: we could change the name here to "server-side-generated timestamps" if that's better? 
I think we softened it to this to make the spec less hosted-vendor-centric. 

Ultimately, the PR tool needs to issue an attestation based on information about which it is authoritative. This can never include user-generated timestamps.

david-a-wheeler:
I'm okay with "The Source Provenance MUST record the time the update occurred" for *source* provenance, but I'm increasingly dubious about "The user MUST NOT be able to falsify this time."

I see a lot of challenges in making such a bold claim:

* If you're using git, and create a commit, that time is part of the commit. I expect any attempt to "change" that time will cause failure of the cryptographic hash.
* The servers are not necessarily trustworthy. *Usually* they're well-synched to an NTP server that's got it right, but usually hides many sins :-). And are we assuming it's impossible for an attacker to break into (or control) a server repo? I wouldn't make that presumption, it's better to have multiple copies and data sources to cross-compare if there's a question.

docs/spec/draft/source-requirements.md
adityasaky: (This response is a bit rushed, sorry!)

> I think we are still leaving it to the SCS to enforce the contemporaneous aspect. This just records when that happened (maybe this is nuanced distinction).

Yes! However, "when" something happened can be established relatively or absolutely. If I'm not mistaken, we actually only care about relative guarantees here in the spec, in that the provenance was issued contemporaneously to the revision. We _may_ choose to use absolute timestamps to show that, but it's not the only way the SCS has.

> I do think knowing the time the update occurred is a helpful security property. It allows correlation of updates to branches with other event logs which can be pretty helpful in incident response.

I agree with this. I only think that the "time in source provenance" should be a SHOULD or MAY, not a MUST. That way, there's flexibility in including a trustworthy source of time for these correlation benefits that aren't necessarily part of the source track. OTOH, if an organization uses gittuf to meet SLSA source track, they get the relative ordering offered by gittuf to ensure that attestations that count as source provenance are contemporaneous of some revision.

> I think you might be taking a broad view of source provenance when perhaps a more narrow view would help. I'd say, what if a code review attestation isn't meant to meet the 'source provenance' attestation and instead you use another existing or new attestation to cover when the branch itself is updated to point at the new revision?

I actually think we're saying very similar things? :) I think the hypothetical you suggest here highlights that different SCSs can have very different approaches to how the contemporaneous property is guaranteed, and the requirement on the SCS from the source track ought to only be about ensuring the contemporaneous property. So, gittuf accomplishes it by using code review attestations as part of the source provenance. Another SCS may create a new attestation and include a server-based timestamp! This is why I think the spec should say "Source provenance MAY include time that cannot be falsified by the user to ensure the provenance was contemporaneous with the revision" or something similar.

> It would be possible to use a TSA to meet this requirement, right? Maybe that's impossible for gittuf?

It's not impossible, but it is very limiting, especially since the motivating factor for using a TSA is addressed without timestamps already. More generally, I'm not sure the specification should set hard, possibly expensive requirements (such as a trustworthy timestamp / TSA) if the property we're trying to meet doesn't actually need said requirements.

docs/spec/draft/source-requirements.md
adityasaky: We define SCS broadly enough that there may not be a server to generate timestamps!

If the SCS has a trustworthy server to provide timestamp, then this could work just fine. But since our aim is to ensure provenance was generated contemporaneously with a revision, and we can achieve this without timestamps, I suggest we mention timestamps as _one possible solution_, and not a hard requirement.

TomHennen: Thinking of @david-a-wheeler's comment [here](https://github.com/slsa-framework/slsa/pull/1363#issuecomment-2867089162)

Maybe we can say that the provenance signed/attestor believes this to be the current time.

For a centralized SCS like we have with the source-poc we don't want user provided timestamps (there are better sources).  In things like gittuf the signers are going to be individuals & the gittuf software.

Maybe it's not even worth mentioning?

As mentioned above I do think time is an important thing to capture.  As David mentions we might not actually need to be unforgeable.

zachariahcox: I agree with @david-a-wheeler -- the timestamp itself is not as relevant. 
I think the key is that issuer (the change management process tooling) must use timestamps from "a trusted source." 
For git, we need to specifically highlight that commit metadata is not a trusted source by default, and orgs need to do a bunch more work to be compliant if they want to rely on git commit timestamps.

docs/spec/draft/source-requirements.md
TomHennen: > I agree with this. I only think that the "time in source provenance" should be a SHOULD or MAY, not a MUST. That way, there's flexibility in including a trustworthy source of time for these correlation benefits that aren't necessarily part of the source track. OTOH, if an organization uses gittuf to meet SLSA source track, they get the relative ordering offered by gittuf to ensure that attestations that count as source provenance are contemporaneous of some revision.

When I was referring to correlation with other logs I meant logs that might be outside the scope of an SCS (e.g. networking logs, EDR logs, etc).

In incident response, for example, it's [extremely useful to have timestamps](https://library.mosse-institute.com/articles/2022/06/understanding-linux-timestamps-for-dfir/understanding-linux-timestamps-for-dfir.html#:~:text=One%20of%20the%20most%20important%20pieces%20of%20data%20that%20can%20be%20collected%20during%20a%20digital%20forensics%20and%20incident%20response%20investigation%20is%20timestamps.%20Timestamps%20can%20be%20used%20to%20help%20determine%20when%20files%20were%20created%2C%20accessed%2C%20and%20modified%20and%20can%20also%20provide%20insight%20into%20the%20behavior%20of%20an%20attacker).

The relative ordering in gittuf doesn't help there.

As I understand it your major concern is that it's hard to get "trusted" time in a distributed system.  Is that right?  If that requirement is removed I'm not sure I see any blockers?

Is there something I'm missing?

TomHennen:
We can abandon this PR since it was resolved another way in #1370

PR #1366, opened 2025-05-05
fix: Company name nitpick in sprint blog by-line

simonbaird:
Use the official correct way to spell "Red Hat".

(See https://www.redhat.com/en/about/brand/standards/naming-and-trademarks about half way down under the "Referring to Red Hat" heading.)


arewm (approved):
Thanks for fixing this. I missed it on my review!

PR #1373, opened 2025-05-16
content: draft: source track: cleanup objective section

zachariahcox:
This is a strict copy edit pass. 
There were a few typos, and an attempted simplification of the objective.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -6,8 +6,8 @@ The primary purpose of the SLSA Source track is to provide evidence that a repos
 It describes increasing levels of trustworthiness and completeness of a revision's <dfn>provenance</dfn>.
 In this track, provenance describes how a source revision came to exist.
 
-The Source track is scoped to single revisions of repositories managed by single organizations.
-The organization determines the intent of the software in the repository, the expected process for the creation of new revisions, and administers technical controls to enforce the process.
+Source provenance applies to a single revision of a repository, not the entire repository.
+The expected process for creating a new revision is determined solely by that repository's owner (the organization) who also determines the intent of the software in the repository and administers technical controls to enforce the process.
 
 Consumers can review source provenance attestations to verify whether a particular revision meets their standards.
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -365,12 +365,11 @@ be reviewed as well.
 **[Context-specific approvals]** Approvals are for a specific context, such as a
 repo + branch in git. Moving fully reviewed content from one context to another
 still requires review. The exact definition of “context” depends on the project,
-and this does not preclude well-understood automatic merges, such as cutting a 
-release branch.
+and this does not preclude well-understood automatic merges, such as cutting a release branch.
 
 **[Trusted Robot Contributions]** An organization MAY choose to grant a Trusted
 Robot a perpetual exception to a policy (e.g. a bot may be able to merge a change
-that has not been reviewed by two parties). 
+that has not been reviewed by two parties).
 
 Examples:
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -9,7 +9,7 @@ In this track, provenance describes how a source revision came to exist.
 Source provenance applies to a single revision of a repository, not the entire repository.
 The expected process for creating a new revision is determined solely by that repository's owner (the organization) who also determines the intent of the software in the repository and administers technical controls to enforce the process.
 
-Consumers can review source provenance attestations to verify whether a particular revision meets their standards.
+Consumers can review attestations to verify whether a particular revision meets their standards.
 
 ## Definitions
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -2,7 +2,7 @@
 
 ## Objective
 
-The primary purpose of the SLSA Source track is to provide evidence that a repository revision was created following the expected process.
+The primary purpose of the SLSA Source track is to provide producers and consumers with increasing levels of trust in the source code they produce and consume.
 It describes increasing levels of trustworthiness and completeness of a revision's <dfn>provenance</dfn>.
 In this track, provenance describes how a source revision came to exist.
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -1,19 +1,15 @@
 # SLSA Source Track
 
-## Outstanding TODOs
-
-Open issues are tracked with the [source-track](https://github.com/slsa-framework/slsa/issues?q=is%3Aissue+is%3Aopen+label%3Asource-track) label in the [slsa-framework/slsa](https://github.com/slsa-framework/slsa) repository.
-Source Track issues are triaged on the [SLSA Source Track](https://github.com/orgs/slsa-framework/projects/5) project board.
-
 ## Objective
 
-The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc...).
+The primary purpose of the SLSA Source track is to provide evidence that a repository revision was created following the expected process.
+It describes increasing levels of trustworthiness and completeness of a revision's <dfn>provenance</dfn>. 
+In this track, provenance describes how a source revision came to exist.
 
-The Source track is scoped to revisions of a single repository that is controlled by an organization.
-That organization determines the intent of the software in the repository, what Source level should apply to the repository and administers technical controls to enforce that level.
+The Source track is scoped to single revisions of repositories managed by single organizations. 
+The organization determines the intent of the software in the repository, the exected process for the creation of new revisions, and administers technical controls to enforce the process.
 
-The primary purpose of the Source track is to enable verification that the creation of a revision followed the expected process.
-Consumers can examine the various source provenance attestations to determine if all sources used during the build meet their requirements.
+Consumers can review source provenance attestations to verify whether a particular revision meets their standards.
 
 ## Definitions
 
@@ -22,16 +18,16 @@ Consumers can examine the various source provenance attestations to determine if
 | Source | An identifiable set of text and binary files and associated metadata. Source is regularly used as input to a build system (see [SLSA Build Track](requirements.md)).
 | Organization | A set of people who collectively create the Source. Examples of organizations include open-source projects, a company, or a team within a company. The organization defines the goals and methods of the source.
 | Version Control System (VCS)| Software for tracking and managing changes to source. Git and Subversion are examples of version control systems.
-| Revision | A specific state of the source with an identifier provided by the version control system. As an example, you can identify a git revision by its tree hash.
-| Source Control System (SCS) | A suite of tools and services (self-hosted or SaaS) relied upon by the organization to produce new revisions of the source. The role of the SCS may be fulfilled by a single service (e.g., GitHub / GitLab) or rely on a combination of services (e.g., GitLab with Gerrit code reviews, GitHub with OpenSSF Scorecard, etc).
-| Source Provenance | Information about how a revision came to exist, where it was hosted, when it was generated, what process was used, who the contributors were, and what parent revisions it was based on.
+| Revision | A specific state of the source with an identifier provided by the version control system. As an example, you can identify a git revision by its commit object ID.
+| Source Control System (SCS) | A suite of tools and services (self-hosted or SaaS) relied upon by the organization to produce new revisions of the source. The role of the SCS may be fulfilled by a single service (e.g., GitHub / GitLab) or a combination of services (e.g., GitLab with Gerrit code reviews, GitHub with OpenSSF Scorecard, etc).
+| Source Provenance | Information about how a revision came to exist, where it was hosted, when it was generated, what process was used, who the contributors were, and which parent revisions preceded it.
 | Repository / Repo | A uniquely identifiable instance of a VCS. The repository controls access to the Source in the VCS. The objective of a repository is to reflect the intent of the organization that controls it.
 | Branch | A named, moveable, pointer to a revision that tracks development in the named context over time. Branches may be modified to point to different revisions by authorized actors. Different branches may have different security requirements.
 | Tag | A named pointer to a revision that does not typically move. Similar to branches, tags may be modified by authorized actors. Tags are often used by producers to indicate a more permanent name for a revision.
 | Change | A set of modifications to the source in a specific context. A change can be proposed and reviewed before being accepted.
 | Change History | A record of the history of revisions that preceded a specific revision.
-| Push / upload / publish | When an actor authenticates to a Repository to add or modify content. Typically makes a new revision reachable from a branch.
-| Review / approve / vote | When an actor authenticates to a change review tool to comment upon, endorse, or reject a source change proposal.
+| Push / upload / publish | When an actor adds a new revision to the repository. Typically also modifies a branch to point to the new revision.
+| Review / approve / vote | When an actor uses a change review tool to comment upon, endorse, or reject a source change proposal.
 
 ## Source Roles
 
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -27,7 +27,7 @@ Consumers can review source provenance attestations to verify whether a particul
 | Change | A set of modifications to the source in a specific context. A change can be proposed and reviewed before being accepted.
 | Change History | A record of the history of revisions that preceded a specific revision.
 | Push / upload / publish | When an actor adds or modifies the Source, Branches or Tags in the repository.
-| Review / approve / vote | When an actor uses a change review tool to comment upon, endorse, or reject a source change proposal.
+| Review / approve / vote | When an actor uses a change management tool to comment upon, endorse, or reject a source change proposal.
 
 ## Source Roles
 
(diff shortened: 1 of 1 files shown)


docs/spec/draft/source-requirements.md
```diff
@@ -1,19 +1,15 @@
 # SLSA Source Track
 
-## Outstanding TODOs
```
zachariahcox, line 3: removed, none of the other sections are using this technique now.

docs/spec/draft/source-requirements.md
```diff
@@ -1,19 +1,15 @@
...
-
-Open issues are tracked with the [source-track](https://github.com/slsa-framework/slsa/issues?q=is%3Aissue+is%3Aopen+label%3Asource-track) label in the [slsa-framework/slsa](https://github.com/slsa-framework/slsa) repository.
-Source Track issues are triaged on the [SLSA Source Track](https://github.com/orgs/slsa-framework/projects/5) project board.
-
 ## Objective
 
-The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc...).
+The primary purpose of the SLSA Source track is to provide evidence that a repository revision was created following the expected process.
+It describes increasing levels of trustworthiness and completeness of a revision's <dfn>provenance</dfn>. 
+In this track, provenance describes how a source revision came to exist.
```
zachariahcox, line 7: I like the phrasing used on the build-env track: https://github.com/slsa-framework/slsa/blob/f671d6ce4f286557dd79c93ed4b5ecd07e1f6641/docs/spec/draft/attested-build-env-levels.md?plain=1#L28C19-L29C24

docs/spec/draft/source-requirements.md
```diff
@@ -1,19 +1,15 @@
...
 
-The Source track is scoped to revisions of a single repository that is controlled by an organization.
-That organization determines the intent of the software in the repository, what Source level should apply to the repository and administers technical controls to enforce that level.
+The Source track is scoped to single revisions of repositories managed by single organizations.
```
zachariahcox, line 9: I tried to simplify it, but it might need more. 

I think the key ideas here are: 
1. provenance is issued on a per-revision basis (not per branch or per repo)
2. "correctness" / expectations are set by the owning organization
3. (optional I guess) there must be exactly one organization in charge of a repo? I'm not sure if we need this requirement anymore. I think we just need "it's a revision of a repo where some org is responsible for it"

docs/spec/draft/source-requirements.md
```diff
@@ -1,19 +1,15 @@
...
 
-The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc...).
+The primary purpose of the SLSA Source track is to provide evidence that a repository revision was created following the expected process.
+It describes increasing levels of trustworthiness and completeness of a revision's <dfn>provenance</dfn>. 
+In this track, provenance describes how a source revision came to exist.
 
-The Source track is scoped to revisions of a single repository that is controlled by an organization.
-That organization determines the intent of the software in the repository, what Source level should apply to the repository and administers technical controls to enforce that level.
+The Source track is scoped to single revisions of repositories managed by single organizations. 
+The organization determines the intent of the software in the repository, the exected process for the creation of new revisions, and administers technical controls to enforce the process.
```
zachariahcox, line 10: edited this a bit. 

According to the lead in
1. provenance is a bunch of facts. 
1. Orgs determine what needs to happen, and they pick certain source track techniques to improve the trustworthiness of their process claims.
1. If they have certain sets of controls and techniques, they are eligible for SLSA source summary VSAs.

docs/spec/draft/source-requirements.md
```diff
@@ -1,19 +1,15 @@
...
 
-The primary purpose of the Source track is to enable verification that the creation of a revision followed the expected process.
-Consumers can examine the various source provenance attestations to determine if all sources used during the build meet their requirements.
+Consumers can review source provenance attestations to verify whether a particular revision meets their standards.
```
zachariahcox, line 12: this might not be enough, and we might still need the summary concept here.

docs/spec/draft/source-requirements.md
```diff
@@ -22,16 +18,16 @@ Consumers can examine the various source provenance attestations to determine if
...
 | Branch | A named, moveable, pointer to a revision that tracks development in the named context over time. Branches may be modified to point to different revisions by authorized actors. Different branches may have different security requirements.
 | Tag | A named pointer to a revision that does not typically move. Similar to branches, tags may be modified by authorized actors. Tags are often used by producers to indicate a more permanent name for a revision.
 | Change | A set of modifications to the source in a specific context. A change can be proposed and reviewed before being accepted.
 | Change History | A record of the history of revisions that preceded a specific revision.
-| Push / upload / publish | When an actor authenticates to a Repository to add or modify content. Typically makes a new revision reachable from a branch.
-| Review / approve / vote | When an actor authenticates to a change review tool to comment upon, endorse, or reject a source change proposal.
+| Push / upload / publish | When an actor adds a new revision to the repository. Typically also modifies a branch to point to the new revision.
```
zachariahcox, line 29: removed the reiteration of authentication. 
the verbs here should be "add" and "review" not "authenticates" I think.

docs/spec/draft/source-requirements.md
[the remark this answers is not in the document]
zachariahcox: Sigh. That's a good point Mr. Copilot.

Actually, maybe we should avoid the "is scoped to" wording. 

I think we really mean "the spec allows for writing provenance about _revisions_, not whole repos."

docs/spec/draft/source-requirements.md
```diff
@@ -1,19 +1,15 @@
...
 
-The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc...).
+The primary purpose of the SLSA Source track is to provide evidence that a repository revision was created following the expected process.
+It describes increasing levels of trustworthiness and completeness of a revision's <dfn>provenance</dfn>.
+In this track, provenance describes how a source revision came to exist.
 
-The Source track is scoped to revisions of a single repository that is controlled by an organization.
-That organization determines the intent of the software in the repository, what Source level should apply to the repository and administers technical controls to enforce that level.
+The Source track is scoped to single revisions of repositories managed by single organizations.
+The organization determines the intent of the software in the repository, the expected process for the creation of new revisions, and administers technical controls to enforce the process.
```
zachariahcox, line 10:
```suggestion
Source provenance applies to a single revision of a repository, not the entire repository.
The expected process for creating a new revision is determined solely by that repository's owner (the organization) who also determines the intent of the software in the repository and administers technical controls to enforce the process.
```

docs/spec/draft/source-requirements.md
```diff
@@ -22,16 +18,16 @@ Consumers can examine the various source provenance attestations to determine if
...
 | Branch | A named, moveable, pointer to a revision that tracks development in the named context over time. Branches may be modified to point to different revisions by authorized actors. Different branches may have different security requirements.
 | Tag | A named pointer to a revision that does not typically move. Similar to branches, tags may be modified by authorized actors. Tags are often used by producers to indicate a more permanent name for a revision.
 | Change | A set of modifications to the source in a specific context. A change can be proposed and reviewed before being accepted.
 | Change History | A record of the history of revisions that preceded a specific revision.
-| Push / upload / publish | When an actor authenticates to a Repository to add or modify content. Typically makes a new revision reachable from a branch.
-| Review / approve / vote | When an actor authenticates to a change review tool to comment upon, endorse, or reject a source change proposal.
+| Push / upload / publish | When an actor adds or modifies the Source, Branches or Tags in the repository.
```
zachariahcox, line 29: avoid use of new term "content"

docs/spec/draft/source-requirements.md
```diff
@@ -1,19 +1,15 @@
...
 
-## Outstanding TODOs
-
-Open issues are tracked with the [source-track](https://github.com/slsa-framework/slsa/issues?q=is%3Aissue+is%3Aopen+label%3Asource-track) label in the [slsa-framework/slsa](https://github.com/slsa-framework/slsa) repository.
-Source Track issues are triaged on the [SLSA Source Track](https://github.com/orgs/slsa-framework/projects/5) project board.
-
 ## Objective
 
-The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc...).
+The primary purpose of the SLSA Source track is to provide evidence that a repository revision was created following the expected process.
```
TomHennen, line 5: I think we should probably try to highlight the trustworthiness first (that's the objective), meanwhile the (a?) mechanism we use is by ensuring the revisions followed the expected process.

docs/spec/draft/source-requirements.md
TomHennen: Hmm, considering we only require the document 'provenance' at level 3 I'm somewhat concerned that we're overloading terms.

_Here_ we're talking about provenance generally, _there_ we're talking about a specific document.

We can probably address this by:

1. Being very careful when we talk about the provenance document(s) to always talk about "provenance _attestations_"
2. We can use a different work for the document entirely
3. We can drop the general purpose use of the word 'provenance' (I don't think it's used that broadly here?)
4. Other?

We can also probably track this as a separate issue and not block this PR on its resolution.

docs/spec/draft/source-requirements.md
TomHennen: "but consumers who issue VSAs do consume source provenance."

I've been considering the entity that issues VSAs as being part of the SCS and not a separate consumer (which might actually explain some of the disagreements we've had :p ).  ([reference](https://slsa.dev/spec/draft/source-requirements#definitions:~:text=%E2%9C%93-,Source%20Summary%20Attestations,-The%20SCS%20MUST))

I bet we need a diagram or something that calls out the various parts of an SCS to make this more clear.

docs/spec/draft/source-requirements.md
```diff
@@ -22,16 +18,16 @@ Consumers can examine the various source provenance attestations to determine if
...
 | Tag | A named pointer to a revision that does not typically move. Similar to branches, tags may be modified by authorized actors. Tags are often used by producers to indicate a more permanent name for a revision.
 | Change | A set of modifications to the source in a specific context. A change can be proposed and reviewed before being accepted.
 | Change History | A record of the history of revisions that preceded a specific revision.
-| Push / upload / publish | When an actor authenticates to a Repository to add or modify content. Typically makes a new revision reachable from a branch.
-| Review / approve / vote | When an actor authenticates to a change review tool to comment upon, endorse, or reject a source change proposal.
+| Push / upload / publish | When an actor adds or modifies the Source, Branches or Tags in the repository.
+| Review / approve / vote | When an actor uses a change review tool to comment upon, endorse, or reject a source change proposal.
```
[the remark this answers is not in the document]
TomHennen: yes I think that's a good change, when I see "review" I think specifically about the thing that people use to view the diffs and approve.

TomHennen (changes requested):
These changes generally LGTM, a few comments though.

docs/spec/draft/source-requirements.md
zachariahcox: That makes sense -- "trustworthy evidence" maybe?

TomHennen: What about

```suggestion
The primary purpose of the SLSA Source track is to provide consumers with increasing levels of trust in the source code they consume.
```

We can get into revisions, processes, and evidence later.

docs/spec/draft/source-requirements.md
zachariahcox: Yes. I think it may be normal that parts the SCS are owned by different teams / orgs / producers. This implies that all together the combo of tools represent "the scs" used by "the producer" for "a repo" but those tools may have internal dependencies on each other that matter.

docs/spec/draft/source-requirements.md
```diff
@@ -22,16 +18,16 @@ Consumers can examine the various source provenance attestations to determine if
...
 | Tag | A named pointer to a revision that does not typically move. Similar to branches, tags may be modified by authorized actors. Tags are often used by producers to indicate a more permanent name for a revision.
 | Change | A set of modifications to the source in a specific context. A change can be proposed and reviewed before being accepted.
 | Change History | A record of the history of revisions that preceded a specific revision.
-| Push / upload / publish | When an actor authenticates to a Repository to add or modify content. Typically makes a new revision reachable from a branch.
-| Review / approve / vote | When an actor authenticates to a change review tool to comment upon, endorse, or reject a source change proposal.
+| Push / upload / publish | When an actor adds or modifies the Source, Branches or Tags in the repository.
+| Review / approve / vote | When an actor uses a change management tool to comment upon, endorse, or reject a source change proposal.
```
zachariahcox, line 30: @TomHennen I went ahead and made this change 👍

docs/spec/draft/source-requirements.md
```diff
@@ -363,19 +359,18 @@ The following combinations are acceptable:
...
+submitted. I.e. if additional changes are made during the review process, those changes MUST
 be reviewed as well.
 
 **[Context-specific approvals]** Approvals are for a specific context, such as a
 repo + branch in git. Moving fully reviewed content from one context to another
-still requires review. (Exact definition of “context” depends on the project,
-and this does not preclude well-understood automatic or reviewless merges, such
-as cutting a release branch.)
+still requires review. The exact definition of “context” depends on the project,
+and this does not preclude well-understood automatic merges, such as cutting a
```
zachariahcox, line 368: removing the word review-less -- there may be many policies that are bypassed automatically in this way.

docs/spec/draft/source-requirements.md
```diff
@@ -363,19 +359,18 @@ The following combinations are acceptable:
...
-Robot to author and submit changes to source code without 2-party approval if
-the Robot’s identity is specifically allowed to bypass two-party review for the
-protected branch.
+still requires review. The exact definition of “context” depends on the project,
+and this does not preclude well-understood automatic merges, such as cutting a 
+release branch.
+
+**[Trusted Robot Contributions]** An organization MAY choose to grant a Trusted
+Robot a perpetual exception to a policy (e.g. a bot may be able to merge a change
+that has not been reviewed by two parties).
```
zachariahcox, line 373: solidly into scope creep here, lol.
In the previous wording, it wasn't clear if a bot could bypass other policies too. 
We should probably add a rider about "please document any actors with perpetual exceptions to stuff"

docs/spec/draft/source-requirements.md
adityasaky: > this might not be enough. end-users should probably use the summary attestations, but consumers who issue VSAs do consume source provenance.

I think the sentence in the PR now is good. While most consumers _should_ use VSAs, the ability to independently use source provenance at higher levels should be available to all consumers, not just those producing VSAs.

> I bet we need a diagram or something that calls out the various parts of an SCS to make this more clear.

big +1! filed https://github.com/slsa-framework/slsa/issues/1377 to track this.

docs/spec/draft/source-requirements.md
adityasaky: I think the rider makes a lot of sense, I would even say it should be a MUST for the organization.

Do you want to add it to this PR or would it make sense to update this independently? I lean towards the latter since it doesn't seem like a copy edit anymore.

adityasaky (approved):
I think this LGTM!

docs/spec/draft/source-requirements.md
TomHennen: > While most consumers should use VSAs, the ability to independently use source provenance at higher levels should be available to all consumers, not just those producing VSAs.

Yup, didn't mean to exclude that and the spec does require provenance be made available to them.  Do you think that needs to be called out in the objective? 

Maybe in the objective we just say something like

```suggestion
Consumers can review attestations to verify whether a particular revision meets their standards.
```

docs/spec/draft/source-requirements.md
TomHennen: Once we start talking about org defined controls it's going to become very difficult to document all that stuff.

We're currently not even requiring them to document how the controls are implemented in the first place: https://slsa.dev/spec/draft/source-requirements#:~:text=Specify%20control%20expectations

With each new 'MUST' requirement I ask myself "how would an implementation prove that this was done".  The more requirements that harder implementations become.

I'd suggest we leave things as is and perhaps as something to pursue in a later revision.

docs/spec/draft/source-requirements.md
TomHennen: @zachariahcox do you think this thread has been resolved in some way?  It still looks the same to me.

docs/spec/draft/source-requirements.md
```diff
@@ -1,19 +1,15 @@
...
-
-Open issues are tracked with the [source-track](https://github.com/slsa-framework/slsa/issues?q=is%3Aissue+is%3Aopen+label%3Asource-track) label in the [slsa-framework/slsa](https://github.com/slsa-framework/slsa) repository.
-Source Track issues are triaged on the [SLSA Source Track](https://github.com/orgs/slsa-framework/projects/5) project board.
-
 ## Objective
 
-The SLSA source track describes increasing levels of trustworthiness and completeness in a repository revision's provenance (e.g. how it was generated, who the contributors were, etc...).
+The primary purpose of the SLSA Source track is to provide producers and consumers with increasing levels of trust in the source code they produce and consume.
+It describes increasing levels of trustworthiness and completeness of a revision's <dfn>provenance</dfn>.
+In this track, provenance describes how a source revision came to exist.
```
TomHennen, line 7:
```suggestion
It describes increasing levels of trustworthiness and completeness of how a source revision was produced.
```

docs/spec/draft/source-requirements.md
```diff
@@ -1,19 +1,15 @@
...
 
-The Source track is scoped to revisions of a single repository that is controlled by an organization.
-That organization determines the intent of the software in the repository, what Source level should apply to the repository and administers technical controls to enforce that level.
+Source provenance applies to a single revision of a repository, not the entire repository.
```
zachariahcox, line 9:
```suggestion
Source provenance attestations apply to a single revision of a repository, not the entire repository.
```

TomHennen (approved):

PR #1394, opened 2025-06-04
content: Rework tag hygiene to be protected tags

adityasaky:
Rename tag hygiene to be protected tags and indicate that the organization must be able to define what tags are protected.


zachariahcox (approved):
lgtm

TomHennen (approved):

PR #1398, opened 2025-06-05
content: draft: Move Common Source Controls to Requirements

### LINKED ISSUE #1396: Move "Common Source Controls and their Applications" someplace more appropriate
The https://slsa.dev/spec/draft/verifying-source#common-source-controls-and-their-applications probably isn't in the right spot.

We had initially thought they could go in Verified Properties ([reference](https://github.com/slsa-framework/slsa/issues/1355)), but that
doesn't seem quite right since we don't want to specify any requirements for those properties now.

They seem more like the sorts of things organizations might want to enable as a part of their change management process.

Perhaps we can move them to source-requirements.md, perhaps in an appendix?

TomHennen:
Moving "Common Source Controls and their Applications" from verifying-source.md to source-requirements.md.

In the process I rejiggered the introduction to focus on the control aspect and how they can be validated with some of the information in the VSA or provenance (based on the enforced change management process stuff).

I also cut the two party review control because that's now a part of the levels themselves so it shouldn't need mentioning.

It might be worth linking to the verifying-source page at the start??

fixes #1396


TomHennen:
@zachariahcox I'd be curious for your thoughts on the question of if these should really be properties or not.  My intuition is that we don't quite know enough (or have the time right now?) to properly define 'verified properties' for them (https://github.com/slsa-framework/slsa/pull/1397) but maybe I'm being conservative.

Perhaps it would be good enough to just say "the org asserts this and here's the standard way to represent that in a VSA".  I can really go either way here.

PR #1399, opened 2025-06-08
impl: tools: add devcontainers and migration script

zachariahcox:
This PR is a paired down version of https://github.com/slsa-framework/slsa/pull/1328 focused on adding the devcontainer. 

I will also leave the late-branch script in -- I do think we should run it after the release of 1.2.

Here's an example branch produced by this kind of tool. 
* https://github.com/zachariahcox/slsa/tree/test/releases/v1.0


tools/migrate-to-late-branch.sh
```diff
@@ -0,0 +1,67 @@
...
+for version in "${VERSIONS[@]}"; do
+  # --- Create a new branch for this version ---
+  BRANCH="$REF_SPEC/$version"
+  git checkout -b "$BRANCH"
+
+  # --- move the version folder to the root of the spec directory ---
+  mv "$SPECDIR/$version" "spec"
+
+  # --- remove docs folder ---
+  rm -rf docs
```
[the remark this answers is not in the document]
zachariahcox: removing the docs folder is desirable.

tools/migrate-to-late-branch.sh
```diff
@@ -0,0 +1,75 @@
...
+done
+echo "No existing version refs found. Proceeding with migration..."
+
+for version in "${VERSIONS[@]}"; do
+  # --- Create a new branch for this version ---
+  BRANCH="$REF_SPEC/$version"
+  git checkout -b "$BRANCH"
+
+  # --- move the version folder to the root of the spec directory ---
+  mv "$SPECDIR/$version" "spec"
```
TomHennen, line 57: Would a `git mv` preserve history?  maybe it's preserved anyways?

.devcontainer/devcontainer.json
TomHennen, line 1: Would it be worth adding some note about how to use this to docs somewhere?  Maybe https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md#submitting-changes

tools/lint.sh
TomHennen, line 1: I think we'll have to update docs that refer to this command.  Like https://github.com/slsa-framework/slsa/blob/main/CONTRIBUTING.md#submitting-changes

TomHennen (changes requested):
Thanks!  It would be good to get reviews from some other folks too because I'm not confident enough in my git or bash scripting.

tools/migrate-to-late-branch.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
```
zachariahcox, line 1: @TomHennen here's a python version if we like that better?

```python
#!/usr/bin/env python3

"""
migrate-to-late-branch.py

Purpose: Migrate SLSA repository to a late-branch model.

Before: All versions are in subfolders in docs/spec/<version>
After:  Each version lives in its own branch under releases/<version>
        with content directly under /spec/

For each version subfolder:
1. Create and checkout a new branch named releases/<version>
2. Move the version content from docs/spec/<version> to spec/
3. Remove the docs/ directory
4. Commit and push the branch
"""

import argparse
import os
import subprocess
import sys
import re
from datetime import datetime
from typing import List, Optional


# --- Configuration constants ---
SPECDIR = "docs/spec"
REF_SPEC = "releases"


# --- Helper functions ---
def log(message: str) -> None:
    """Print a timestamped log message."""
    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
    print(f"[{timestamp}] {message}")


def error(message: str) -> None:
    """Print an error message and exit."""
    log(f"ERROR: {message}")
    sys.exit(1)


def confirm(message: str) -> bool:
    """Prompt the user for confirmation."""
    response = input(f"{message} [y/N] ").strip().lower()
    return response in ["y", "yes"]


def run_command(command: List[str], dry_run: bool = False, check: bool = True) -> Optional[str]:
    """
    Run a shell command.
    
    Args:
        command: The command to run as a list of strings
        dry_run: If True, print the command but don't execute it
        check: If True, raise an exception if the command fails
        
    Returns:
        The command output as a string, or None if dry_run is True
    """
    cmd_str = " ".join(command)
    if dry_run:
        log(f"[DRY RUN] Would execute: {cmd_str}")
        return None
    
    log(f"Executing: {cmd_str}")
    try:
        result = subprocess.run(
            command,
            check=check,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            text=True
        )
        return result.stdout.strip()
    except subprocess.CalledProcessError as e:
        error(f"Command failed: {cmd_str}\n{e.stderr}")
        return None


def get_repo_root() -> str:
    """Get the absolute path to the Git repository root."""
    result = run_command(["git", "rev-parse", "--show-toplevel"], check=True)
    if not result:
        error("Could not determine repository root.")
    return str(result)


def get_current_branch() -> str:
    """Get the name of the current Git branch."""
    result = run_command(["git", "rev-parse", "--abbrev-ref", "HEAD"], check=True)
    if not result:
        error("Could not determine current branch.")
    return str(result)


def has_uncommitted_changes() -> bool:
    """Check if there are uncommitted changes in the repository."""
    result = run_command(["git", "status", "--porcelain"])
    return bool(result)


def branch_exists(branch_name: str) -> bool:
    """Check if a branch exists locally or remotely."""
    # Check if branch exists locally
    local_result = subprocess.run(
        ["git", "show-ref", "--verify", "--quiet", f"refs/heads/{branch_name}"],
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE
    )
    
    # Check if branch exists remotely
    remote_result = subprocess.run(
        ["git", "show-ref", "--verify", "--quiet", f"refs/remotes/origin/{branch_name}"],
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE
    )
    
    return local_result.returncode == 0 or remote_result.returncode == 0


def get_version_dirs() -> List[str]:
    """Get all release version directories from the spec directory."""
    if not os.path.isdir(SPECDIR):
        error(f"{SPECDIR} does not exist.")
        
    # List all items in the directory
    items = os.listdir(SPECDIR)
    
    # Filter out README files, hidden files, and draft directories
    versions = [item for item in items if os.path.isdir(os.path.join(SPECDIR, item)) 
                and not re.match(r'^(\.|README|draft$)', item)]
    
    return versions


def migrate_version(version: str, dry_run: bool = False) -> None:
    """
    Migrate a specific version to its own branch.
    
    Args:
        version: The version to migrate
        dry_run: If True, print commands but don't execute them
    """
    branch = f"{REF_SPEC}/{version}"
    log(f"Processing version: {version} -> branch: {branch}")
    
    # Create a new branch for this version
    run_command(["git", "checkout", "-b", branch], dry_run=dry_run)
    
    # Move the version folder to the root of the spec directory
    log(f"Moving {SPECDIR}/{version} to spec/")
    run_command(["git", "mv", f"{SPECDIR}/{version}", "spec"], dry_run=dry_run)
    
    # Remove docs folder
    log("Removing docs/ directory")
    run_command(["git", "rm", "-rf", "docs"], dry_run=dry_run)
    
    # Commit the changes
    log("Committing changes")
    run_command([
        "git", "commit", "-m", f"Migrate {version} to its own branch ({branch})"], 
        dry_run=dry_run
        )
    
    # Push the branch
    log("Pushing branch to origin")
    run_command(["git", "push", "-u", "origin", branch], dry_run=dry_run)


def main() -> None:
    """Main function to execute the migration script."""
    # Parse command line arguments
    parser = argparse.ArgumentParser(
        description="Migrate SLSA repository to a late-branch model"
    )
    parser.add_argument(
        "--dry-run",
        action="store_true",
        help="Show what would be done without making changes"
    )
    parser.add_argument(
        "--force",
        action="store_true",
        help="Skip confirmation prompts"
    )
    args = parser.parse_args()
    
    # Check for uncommitted changes
    if has_uncommitted_changes():
        error("You have uncommitted changes. Please commit or stash them first.")
    
    # Ensure script is run from the repo root
    repo_root = get_repo_root()
    os.chdir(repo_root)
    
    # Save the current branch to return to later
    orig_branch = get_current_branch()
    log(f"Current branch: {orig_branch}")
    
    # Find all subfolders in docs/spec
    log("Finding version directories...")
    versions = get_version_dirs()
    log(f"Found versions: {', '.join(versions)}")
    
    # Check if any of the version refs already exist
    log("Checking for existing version refs...")
    existing_branches = []
    for version in versions:
        branch = f"{REF_SPEC}/{version}"
        if branch_exists(branch):
            existing_branches.append(branch)
    
    if existing_branches:
        error(f"Branches already exist: {', '.join(existing_branches)}. Please delete them first.")
    
    log("No existing version refs found. Proceeding with migration...")
    
    # Confirm the operation
    if not args.dry_run and not args.force:
        if not confirm(f"This script will create {len(versions)} new branches. Continue?"):
            log("Operation cancelled by user.")
            sys.exit(0)
    
    # Process each version
    try:
        for version in versions:
            migrate_version(version, dry_run=args.dry_run)
            
            # Return to the original branch after processing each version
            if not args.dry_run:
                run_command(["git", "checkout", orig_branch])
                
        # Summary of updated branches
        log("Operation complete!")
        if args.dry_run:
            log("[DRY RUN] Would have created these branches:")
        else:
            log("All branches created:")
            
        for version in versions:
            log(f"  - {REF_SPEC}/{version}")
            
    except Exception as e:
        error(f"An error occurred: {str(e)}")
        # Try to return to the original branch
        if not args.dry_run:
            try:
                run_command(["git", "checkout", orig_branch], check=False)
            except:
                pass
        sys.exit(1)


if __name__ == "__main__":
    main()

File: tools/migrate-to-late-branch.py [hunk 4]
```diff
@@ -0,0 +1,266 @@
+#!/usr/bin/env python3
+
+"""
+migrate-to-late-branch.py
```
zachariahcox, line 4: We probably don't need to ship both of these? 
I prefer python I guess (I can use a debugger), but bash is probably more traditional here.

TomHennen (approved):

zachariahcox:
@TomHennen @arewm sorry for the runaround -- I found a _substantially_ better way to build this devcontainer using the prebuilt ruby image. It cut build times for me to ~4 minutes from 15 or so on a 2core machine.

TomHennen (approved):

PR #1400, opened 2025-06-08
nonspec: remove controls folder

zachariahcox:
As far as I can tell, this content is old and not referenced by the site anymore. 
Generally, content these days should be versions in one of the spec folders. 

Let's cut it.


TomHennen (approved):

lehors (approved):
LGTM

PR #1401, opened 2025-06-09
fix: remove left-over line

lehors:
As I was reading through the source track draft I noticed an extra line that was left-over from some previous editing. This deletes it.


arewm (approved):
Thanks for catching this!

PR #1411, opened 2025-06-09
editorial: draft: Fix formatting in "Enforced Change Management Process"

### LINKED ISSUE #1410: Example in "Enforced Change Management Process" looks funny
Image

We can probably fix this by removing the bullets.  The "some other mechanism as enforced by the [Change management tool](https://deploy-preview-1409--slsa.netlify.app/spec/draft/source-requirements#change-management-tool-requirements)." text is outdated and the link is broken so that point can likely just be dropped?

TomHennen:
fixes #1410

* Example doesn't use bullets anymore. (should fix rendering issue)
* Example dropped reference to "some other mechanism as enforced by the Change management tool."
  * The referenced tool doesn't exist anymore.
  * The example doesn't need to be exhaustive, so it should be fine to remove.
* Added "The SCS MUST prevent organization-specified" to the other bulleted list.
  * Makes it more cohesive with the other requirements
  * Makes it more clear the example applies to the list, not just the "MUST prevent ..." language.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -267,18 +267,17 @@ enforced.
 -   Allow organizations to distribute additional attestations related to their
    technical controls to consumers authorized to access the corresponding source
    revision.
+-   Prevent organization-specified properties from beginning with any value
+   other than `ORG_SOURCE_` unless the SCS endorses the veracity of the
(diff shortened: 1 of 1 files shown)


adityasaky (approved):

lehors (approved):
LGTM

PR #1414, opened 2025-06-10
content: draft: Add landing pages, define 'tracks' page, levels->build-track-basics

### LINKED ISSUE #1402: Make room for new tracks with renaming/retitling/landing pages
The currently published spec only has one track and it shows in some of the naming.

* "requirements" - when talking about build track requirements
* "provenance" - when talking about build provenance
* "verifying-systems" - when talking about verifying builders

We should resolve some of these things before publishing 1.2

Note that the *provenance* file probably needs to stay where it is because the attestation predicate type points to it: "https://slsa.dev/provenance/v1" and we really don't want to change that.

TomHennen:
* Added landing page for provenance and requirements disambiguation.
* requirements.md isn't linked in the nav because we don't really want people to go there.
* removed requirements redirect so folks go to the landing page.
* Added 'tracks.md' which provides an overview of each track taking existing language from other pages.
* Moved levels.md to build-track-basics.md (linked from tracks.md) scoped down to just build stuff.
    * Now when talking about levels I either link to build-track-basics if the link was to build levels specifically, or about if its about the levels concept.

fixes #1402

### DIFF
--- a/docs/_data/nav/main.yml
+++ b/docs/_data/nav/main.yml
@@ -301,7 +301,7 @@
       description: Guidelines for verifying source provenance
 
     - title: Verified Properties
-      url: /spec/draft/verified-properties.md
+      url: /spec/draft/verified-properties
       description: SLSA allows a common way to express verified properties that may not fit within a SLSA track.
 
   - title: Attestation formats
--- a/docs/_data/nav/draft.yml
+++ b/docs/_data/nav/draft.yml
@@ -98,7 +98,7 @@
     description: Guidelines for verifying source provenance
 
   - title: Verified Properties
-    url: /spec/draft/verified-properties.md
+    url: /spec/draft/verified-properties
     description: SLSA allows a common way to express verified properties that may not fit within a SLSA track.
 
 - title: Attestation formats


zachariahcox:
not sure if related, but I get a 404 on https://deploy-preview-1414--slsa.netlify.app/spec/draft/verified-properties.md

docs/_data/nav/main.yml
```diff
@@ -260,6 +260,14 @@
...
       description: Terminology and model used by SLSA
 
+    - title: Tracks
+      url: /spec/draft/tracks
+      description: Provides an overview of each track and links to more specific information.
+
+    - title: Provenance
+      url: /spec/draft/provenance
+      description: Provides a description of the concept of provenance and links to the various tracks specific definitions.
+
```
zachariahcox, line 270: we can cut the security levels section below

TomHennen:
> should we add the build env track too? https://deploy-preview-1414--slsa.netlify.app/spec/draft/attested-build-env-levels

I was waiting since we won't want to include that in the 1.2 RC1 release?  If we put it in here we'd have to remove it when we cut the RC.

docs/_data/nav/main.yml
TomHennen: Done

TomHennen:
> not sure if related, but I get a 404 on https://deploy-preview-1414--slsa.netlify.app/spec/draft/verified-properties.md

Not related, but fixed.

zachariahcox (approved):
LGTM!

marcelamelara:
> should we add the build env track too? https://deploy-preview-1414--slsa.netlify.app/spec/draft/attested-build-env-levels

I'd appreciate it if we did add build env at this point, since the basics have been defined for a while.

TomHennen:
> > should we add the build env track too? https://deploy-preview-1414--slsa.netlify.app/spec/draft/attested-build-env-levels
> 
> I'd appreciate it if we did add build env at this point, since the basics have been defined for a while.

Ok, so then an interesting question is, do we:

1. Include it here, but remove when publishing RC1 (unless you want the build env track to go out in RC1)
2. Include it here but ensure it's noted as _draft_ and keep a persistent link to the draft instance.  That makes it easier for people to find when we publish RC1, doesn't require changes when publishing RC1, and still makes it clear it's draft.

WDYT?

marcelamelara (approved):
LGTM thanks! I'd like to add the BuildEnv track to the new page in time for the RC, but since there are a couple of other smaller changes I'd like to make as part of that (e.g., renaming the nav link to the track basics), I'll send a separate PR for that as soon as this gets merged.

marcelamelara:
Ok, so in the time I was writing up one reply, 2 more came in...

I was going to say that I can mark the track as draft on the current page, but _now_ it sounds like maybe let's not add it since we only want to release non-draft content in the RC. Fair enough. I will still send a PR with some BuildEnv track related changes I think ought to happen prior to RC1.

TomHennen:
> > Ok, so in the time I was writing up one reply, 2 more came in...
> > I was going to say that I can mark the track as draft on the current page, but _now_ it sounds like maybe let's not add it since we only want to release non-draft content in the RC. Fair enough. I will still send a PR with some BuildEnv track related changes I think ought to happen prior to RC1.
> 
> No, again, the BuildEnv track should be added to the draft. The draft is not meant to be exactly what's going to be in RC1. As part of putting together RC1 we can decide what ought to be in it and what should be left out.

Basically, @marcelamelara can go ahead and send the PR adding it to the tracks page and we can remove from the RC if needed?

PR #1416, opened 2025-06-12
content: draft: 'Context' -> 'History'

### LINKED ISSUE #1407: Can we further clarify 'Continuity' now that 'change management process' has moved?
We are moving this here as discussed. Should we modify the text for the next requirement from 
> Continuity MUST also apply to any [change management processes](#enforced-change-management-process)
which are enforced on the protected branches.

i.e. can we simplify continuity as it relates to _all_ controls that are in place now and for higher levels?

_Originally posted by @arewm in https://github.com/slsa-framework/slsa/pull/1395#discussion_r2136159509_

TomHennen:
Reworded the 'Context' requirement to a 'History' requirement
taking some content from the 'Continuity' requirement to do so.

The continuity requirement now focuses on the need to maintain
continuity of controls (both history and enforced change
management process) and leaves much of the rest of the text
to 'history'.

Moved 'protected branches' ahead of  'history' since history
only applies to protected branches (so seems like a better
ordering).

fixes #1407

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -290,9 +290,9 @@ policies.
 In a source control system, each new revision is built on top of prior
 revisions. Controls (e.g. [history](#history) or
 [enforced change management process](#enforced-change-management-process)) are
-only  effective if they are used continuously from one revision to another. If
+only effective if they are used continuously from one revision to another. If
 a control is disabled for the introduction of a new revision and then re-enabled
-it difficult to reason about the effectiveness of the control. 'Continuity' is
+it is difficult to reason about the effectiveness of the control. 'Continuity' is
 the concept of ensuring controls are enforced continuously from the time they
 were introduced, leading to a higher degree of trust in the revisions produced
 after their introduction.


docs/spec/draft/source-requirements.md
```diff
@@ -233,26 +234,46 @@ the SCS issued [source provenance](#source-provenance) when making the issuing
...
 revision in that context.
+
 > For example, for the git version control system, the
-SCS MUST record the branch name that was updated, its new revision and its previous revision.
+SCS MUST record the branch name that was updated, its new revision and its
+previous revision.
 
-<td><td>✓<td>✓<td>✓
-<tr id="branches"><td>Protected Branches<td>
+The SCS MUST prevent tampering with the history of revisions on protected
```
TomHennen, line 267: Note: if we wanted to simplify things further, I think we could basically _just_ include this requirement here.

E.g. explain what history is and require that it's retained.

Do we really need to explain that the change itself needs to be recorded etc (along with the explanatory text)?  It could be constructed as both redundant and hard to understand.

docs/spec/draft/source-requirements.md
```diff
@@ -233,26 +234,46 @@ the SCS issued [source provenance](#source-provenance) when making the issuing
...
+`refs/heads/releases/*`, but not `refs/heads/playground/*`.
+
+<td><td>✓<td>✓<td>✓
+<tr id="history"><td>History<td>
 
-The SCS MUST record the specific code change (a "diff" in git) or instructions
-to recreate it. In git, this is typically defined to be three revision IDs: the tip
-of the "topic" branch, the tip of the target branch, and closest shared ancestor
-between the two (such as determined by `git-merge-base`).
+Revisions are created by applying specific code changes (a "diffs" in git) on
```
arewm, line 248: Is it a "diffs" and not a "diff"?

docs/spec/draft/source-requirements.md
```diff
@@ -233,26 +234,46 @@ the SCS issued [source provenance](#source-provenance) when making the issuing
...
+top of earlier revisions of a branch. This sequence of changes, the revisions
+they produced, and how they were introduced into a branch constitute the history
+of a revision.
+
+The SCS MUST record the specific code change (a "diff" in git) responsible for
+creating a new revision or instructions to recreate it.
```
arewm, line 254: Is this too much of a duplication of the above?

docs/spec/draft/source-requirements.md
```diff
@@ -233,26 +234,46 @@ the SCS issued [source provenance](#source-provenance) when making the issuing
...
+tip of the "topic" branch, the tip of the target branch, and closest shared
+ancestor between the two (such as determined by `git-merge-base`).
 
 The SCS MUST record the "target" context for the change and the previous
 revision in that context.
+
 > For example, for the git version control system, the
-SCS MUST record the branch name that was updated, its new revision and its previous revision.
+SCS MUST record the branch name that was updated, its new revision and its
+previous revision.
```
arewm, line 265: The rationale behind having both of these "SCS MUST" statements isn't clear to me. maybe it is clear to others.

docs/spec/draft/source-requirements.md
arewm: I think that is what my comments above effectively state. For _me_, they don't provide specific value. Maybe they would for others... or maybe it could just be an implementation detail for an SCS.

docs/spec/draft/source-requirements.md
TomHennen: I figured plural since `revisions` was plural.  E.g. each revision is created by applying a single diff, but multiple revisions -> multiple diffs?

docs/spec/draft/source-requirements.md
TomHennen: Great, I'd be really happy to make this simplification.  Let's see what @zachariahcox thinks since I think he liked this other language.

docs/spec/draft/source-requirements.md
adityasaky: Perhaps singular removes confusion?

A revision is created by applying specific code changes (a "diff" in git) on top of an earlier revision of a branch. The sequence of changes, prior revisions produced, and how they were introduced into a branch constitute the history of the revision.

docs/spec/draft/source-requirements.md
adityasaky: I think it's worth being explicit about this; earlier we're more describing how revisions are created but not what the SCS must record for our purposes.

docs/spec/draft/source-requirements.md
```diff
@@ -233,26 +234,46 @@ the SCS issued [source provenance](#source-provenance) when making the issuing
...
-between the two (such as determined by `git-merge-base`).
+Revisions are created by applying specific code changes (a "diffs" in git) on
+top of earlier revisions of a branch. This sequence of changes, the revisions
+they produced, and how they were introduced into a branch constitute the history
+of a revision.
+
+The SCS MUST record the specific code change (a "diff" in git) responsible for
+creating a new revision or instructions to recreate it.
+
+> For example, In git, this is typically defined to be three revision IDs: the
```
adityasaky, line 256:
```suggestion
> For example, in git, this is typically defined to be three revision IDs: the
```

docs/spec/draft/source-requirements.md
```diff
@@ -233,26 +234,46 @@ the SCS issued [source provenance](#source-provenance) when making the issuing
...
+tip of the "topic" branch, the tip of the target branch, and closest shared
+ancestor between the two (such as determined by `git-merge-base`).
 
 The SCS MUST record the "target" context for the change and the previous
 revision in that context.
+
 > For example, for the git version control system, the
-SCS MUST record the branch name that was updated, its new revision and its previous revision.
+SCS MUST record the branch name that was updated, its new revision and its
```
adityasaky, line 264:
```suggestion
SCS MUST record the branch name that was updated, its new revision, and its
```

docs/spec/draft/source-requirements.md
adityasaky: I like the explicitness of what's currently there, fwiw.

docs/spec/draft/source-requirements.md
TomHennen: For reference I think this proposed language is both explicit, more readable, more compact, and more comprehensive? https://github.com/slsa-framework/slsa/pull/1416#discussion_r2145246118

docs/spec/draft/source-requirements.md
adityasaky: I think this is fine!

docs/spec/draft/source-requirements.md
TomHennen: Fixed.

adityasaky (approved):

docs/spec/draft/source-requirements.md
```diff
@@ -254,6 +242,22 @@ should be protected by SLSA Source Level 2+ requirements.
...
+top of earlier revisions of a branch. This sequence of changes, the revisions
+they produced, and how they were introduced into a branch constitute the history
+of a revision.
+
+The SCS MUST record the sequence of changes, the revisions they created,
+the actors that introduced them and the context they were introduced into.
+
+The SCS MUST prevent tampering with these records on protected branches.
+
+> For example, systems like GitHub or GitLab, this can be accomplished by
```
marcelamelara, line 258: typo
```suggestion
> For example, in systems like GitHub or GitLab, this can be accomplished by
```

docs/spec/draft/source-requirements.md
```diff
@@ -283,24 +287,12 @@ policies.
...
-The SCS MUST prevent the entire repository (including all branches) from being
-deleted and replaced by different source.
-
-Continuity MUST be established and tracked from a specific revision. If there is
-a lapse in continuity for a specific control, that continuity MUST be
-re-established from a new revision.
-
-Continuity MUST also apply to any [change management processes](#enforced-change-management-process)
-which are enforced on the protected branches.
+On [protected branches](#branches) continuity for [history](#history) and
```
marcelamelara, line 290: Not sure if this should be addressed in this PR, but as someone with less context around the source track, I find the term "continuity" to be a bit ambiguous. To me, this requirement seems to be more about the linearity of history.

marcelamelara (approved):
Thanks! I have only a couple of smaller comments, and one that I think is up for longer discussion potentially. So, I agree that this PR is basically ready to be merged for now.

docs/spec/draft/source-requirements.md
TomHennen: let me see if I can address...

TomHennen: How's this?

docs/spec/draft/source-requirements.md
```diff
@@ -283,24 +287,22 @@ policies.
...
-Continuity MUST be established and tracked from a specific revision. If there is
-a lapse in continuity for a specific control, that continuity MUST be
-re-established from a new revision.
-
-Continuity MUST also apply to any [change management processes](#enforced-change-management-process)
-which are enforced on the protected branches.
+In a source control system, each new revisions is built on top of prior
+revisions. Controls (e.g. [history](#history) or
+[enforced change management process](#enforced-change-management-process)) are
+only  effective if they are used continuously from one revision to another. If
```
adityasaky, line 293:
```suggestion
only effective if they are used continuously from one revision to another. If
```

docs/spec/draft/source-requirements.md
```diff
@@ -283,24 +287,22 @@ policies.
...
+a control is disabled for the introduction of a new revision and then re-enabled
+it difficult to reason about the effectiveness of the control. 'Continuity' is
```
adityasaky, line 295:
```suggestion
it is difficult to reason about the effectiveness of the control. 'Continuity' is
```

Issue #1369, opened 2025-05-10
Where to put additional claims?


TomHennen:
Non-blocking suggestion, maybe we can discuss this after this PR is merged.

Do we really want to support additional tags? Maybe the consumer can put `USER_SOURCE_` tags in another attestation (e.g., SCAI) that they then work out how to use in conjunction with the VSA? That avoids even accidental cases where `USER_SOURCE_` tags are ascribed to the organization by some party that's several layers removed from the spec for this and doesn't understand the intent.

I have a similar opinion for `INTERNAL_USER_`. I'm guessing this is to prevent consumers relying on a claim the organization doesn't (yet?) want to provide. However, there will still need to be some out of band exchange of expectations between the org and consumer (e.g., for the `SLSA_SOURCE_` tags), so perhaps this is additional complexity we can do without? Does this address any other scenario that I've missed?

_Originally posted by @adityasaky in https://github.com/slsa-framework/slsa/pull/1362#discussion_r2082328428_

TomHennen:
> Do we really want to support additional tags?

I think it's actually quite important to put them in the VSA if we want to keep the model that the user can easily verify a revision using only the VSA without additional evidence.  If we put these simple properties in yet another attestation there's no longer a simple verification story and the spec will be much more complex.

Note that as written the PR (https://github.com/slsa-framework/slsa/pull/1362) does allow/encourage additional evidence to be stored as well.

> However, there will still need to be some out of band exchange of expectations between the org and consumer (e.g., for the SLSA_SOURCE_ tags), so perhaps this is additional complexity we can do without?

This should be addressed in the "Specify control expectations" section where a TOFU model is presented (also discussed [here](https://slsa.dev/spec/draft/verifying-source#forming-expectations) when we talk about forming expectations generally).

I'm sure that language can use work.

What I was going for with the two different prefixes is that it makes the story of how those expectations are formed easier and allows for inband expectation setting.  That prevents each SCS from having to define a different mechanism and consumers needing to understand that mechanism.

TomHennen:
On second reading I think what you're pointing out is that there can be confusion between orgs.

Org A might issue a VSA with "ORG_SOURCE_TESTED", and then maybe they include (via submodules #1324) repos from Org B where they issued a VSA for the revisions they control that also say "ORG_SOURCE_TESTED" which gets included in the original VSA under `dependencyLevels`.  Then consumers can't distinguish them or who issued them.

I think we have a couple options:

[...]

PR #1417, opened 2025-06-12
content: draft: Update VSA spec to handle SLSA levels beyond the build track

TomHennen:
This basically follows the same strategy as before but provides a formula
instead of a full enumeration and links to verified-properties.md for
additional SLSA_ values.

I've decided _not_ to include a prohibition on `ORG_` prefixed policies because
the SlsaResult documentation already tells users (i.e. VSA issuers) that they
MAY use custom values as long as they don't start with `SLSA_`. `ORG_`
prefixed values, as they're defined by the organization, _are_ custom values
so it seems we're already aligned with the spec.  We're also not trying to
establish (within SLSA) any specific meaning for these ORG properties.
That's left to the org itself to document.

fixes https://github.com/slsa-framework/slsa/issues/1369

### DIFF
--- a/docs/spec/draft/verification_summary.md
+++ b/docs/spec/draft/verification_summary.md
@@ -373,13 +373,18 @@ verifiers they add to their roots of trust.
 </div>
 
 The result of evaluating an artifact (or set of artifacts) against SLSA.
-SHOULD be one of these values:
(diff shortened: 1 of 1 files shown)


adityasaky (approved):

marcelamelara (approved):
LGTM thanks!

PR #1422, opened 2025-06-13
content: draft: Add "source" to provenance attestations title in source-requirements

adityasaky:
This may be a bit of a nit but I think it's good for consistency and avoids confusion with non source provenance attestations in contexts outside of the track.


TomHennen (approved):

TomHennen:
I think this is sufficiently uncontroversial that we can merge with just one approval.

PR #1434, opened 2025-06-17
style: Add basic anchor links to source requirements

### LINKED ISSUE #1433: Make it easier to link to requirement rows
On the source track, one pain point I've hit is that I cannot link to specific requirements:
IE, I would like to hyperlink here to facilitate other discussions
image

_Originally posted by @zachariahcox in https://github.com/slsa-framework/slsa/issues/1428#issuecomment-2977720073_

TomHennen:
Adds anchor links to each row in the source requirements.

I'm not yet sure how well this is going to look.  Might need further tweaking.

fixes #1433

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -161,7 +161,7 @@ Many examples in this document use the [git version control system](https://git-
 <table>
 <tr><th>Requirement<th>Description<th>L1<th>L2<th>L3<th>L4
 
-<tr id="choose-scs"><td>Choose an appropriate source control system<td>
+<tr id="choose-scs"><td>Choose an appropriate source control system<a href="#choose-scs">🔗</a><td>
 
 An organization producing source revisions MUST select a SCS capable of reaching
 their desired SLSA Source Level.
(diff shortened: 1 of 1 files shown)


lehors:
This isn't bad actually but having to add these links everywhere doesn't seem desirable and only adding those few seems pretty arbitrary.
I'm actually used to dig into the HTML to find those anchors (in general, not just in the case of SLSA) but I agree that this isn't very user-friendly and something like what's done by GitHub would be much nicer. Maybe we should investigate a bit what a general solution would take.

TomHennen:
> This isn't bad actually but having to add these links everywhere doesn't seem desirable and only adding those few seems pretty arbitrary. I'm actually used to digging into the HTML to find those anchors (in general, not just in the case of SLSA) but I agree that this isn't very user-friendly and something like what's done by GitHub would be much nicer. Maybe we should investigate a bit what a general solution would take.

I think the big question is if we want this to block 1.2 RC1 (I think probably not).

I did look into this some and Jekyll, liquid, etc didn't seem to have a great general purpose solution.  Maybe there's some dope JavaScript library that can handle it.

zachariahcox (approved):
🚢 it

TomHennen:
I'll go ahead and merge now so we can include in RC1.  If nothing else it will make it easier for folks to file issues about things they don't like.  We can always remove later once we figure something else out.

I don't get the impression that @lehors objects to merging this now as much as he'd eventually like to see something better (and I agree this isn't great long term).

Filed https://github.com/slsa-framework/slsa/issues/1436 to track a longer term solution.

PR #1428, opened 2025-06-15
Create SLSA 1.2-rc1

TomHennen:
Creates SLSA 1.2-rc1

* I copied docs/spec/draft to docs/spec/v1.2-rc1
* I removed references to the build environment track from within v1.2-rc1
* I then followed [these instructions](https://github.com/slsa-framework/slsa/blob/main/docs/README.md#publishing-a-new-version-of-the-slsa-specification) to update nav, etc.

fixes #1419

PR #1439, opened 2025-06-18
editorial: draft: small changes to address comments on the 1.2-rc1 PR

TomHennen:
This change address @arewm's comments on #1428.

They're all fairly minor wording tweaks.

If these are acceptable I'll merge here and then update #1428.


arewm (approved):
Thanks!

Issue #1376, opened 2025-05-19
Remove references to 'people' in Source Requirements


TomHennen:
Per [guiding principles](https://slsa.dev/spec/draft/principles#support-anonymous-and-pseudonymous-contributions) we want to support anonymous contributions.

Given that, language like [attribute those changes to the **people** that made them](https://slsa.dev/spec/draft/source-requirements#:~:text=to%20the%20people%20that%20made%20them) seems to exceed the bounds of what we're looking for.

Consider replacing 'people' with 'actors' and conduct a review of the rest of the text to see if this sort of thing exists anywhere else.

tombedfordgit:
@TomHennen you can assign this to me, i'm happy to find/replace across the spec/site

TomHennen:
> [@TomHennen](https://github.com/TomHennen) you can assign this to me, i'm happy to find/replace across the spec/site

Done, thank you!

tombedfordgit:
These are the pages in which the terms 'people', 'person' or 'persons' are used:

- https://slsa.dev/spec/draft/principles
- https://slsa.dev/spec/v1.1/threats-overview
- https://slsa.dev/spec/v1.1/use-cases
- https://slsa.dev/spec/v1.1/principles
- https://slsa.dev/spec/v1.1/terminology
- https://slsa.dev/spec/v1.1/requirements
- https://slsa.dev/spec/v1.1/verifying-systems
- https://slsa.dev/spec/v1.1/threats
- https://slsa.dev/spec/v1.1/provenance

I'll review them individually, as there are phrases where substituting the words with 'actor' would unlikely make sense.

zachariahcox:
This has been mitigated in the draft folder by other PRs 👍

PR #1448, opened 2025-06-23
editorial: Update requirements.md to align verbiage

tombedfordgit:
Per https://github.com/slsa-framework/slsa/issues/1376, aligning verbiage to represent the fact that non-human actors can interface with a platform.

### DIFF
--- a/docs/spec/v1.1/requirements.md
+++ b/docs/spec/v1.1/requirements.md
@@ -125,7 +125,7 @@ distributing provenance.
 
 A package's <dfn>build platform</dfn> is the infrastructure used to transform the
 software from source to package. This includes the transitive closure of all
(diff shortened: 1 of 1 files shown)


zachariahcox (approved):
thanks!

lehors (approved):
LGTM

lehors (changes requested):
Sorry, I take it back. I don't think we change v1.1. Instead this change should be made to draft so that it gets fixed in v1.2-rc2 and beyond.

zachariahcox:
>  I don't think we change v1.1

OOF I missed that too. 
Maybe we should add some extra rules if the PR is trying to change a published spec version.

@tombedfordgit I think you just need to migrate this change to the draft folder 👍

PR #1449, opened 2025-06-23
editorial: Update threats.md to align verbiage

tombedfordgit:
Per https://github.com/slsa-framework/slsa/issues/1376, aligning verbiage to represent the fact that non-human actors can interface with a system/can be considered 'trusted'.


lehors (approved):
LGTM

lehors (changes requested):
Sorry, I take it back. I don't think we change v1.1. Instead this change should be made to draft so that it gets fixed in v1.2-rc2 and beyond.

tombedfordgit:
Closing this, and will make edits to draft-only

PR #1453, opened 2025-06-26
content: draft: adding depndency-track to nav bar (main.yml)

tombedfordgit:
adding Dependency track to nav bar, under working draft


TomHennen:
Changed title to `style` as this is just a layout change and doesn't change the meaning of the spec.

It's also draft and seems fine so I'll go ahead and merge.

PR #1458, opened 2025-07-12
blog: SLSA end-to-end

TomHennen:
A blog post version of [the SLSA e2e](https://docs.google.com/document/d/1Wk_5qKWoQZ3qSZhGXdKjDBK3GW41LoZ1dS2DXOibuLI/edit?resourcekey=0-ygY03McfHKFObtQzjNKsQQ&tab=t.0) proposal as [discussed at the spec meeting Monday](https://docs.google.com/document/d/1PwhekVB1iDpcgCQRNVN_aesoVdOiTruoebCs896aGxw/edit?tab=t.0#bookmark=id.yu3a5qraobex).


docs/_posts/2025-07-16-slsa-e2e.md
```diff
@@ -0,0 +1,202 @@
...
+([SLSA](slsa.dev)) framework. The goal is to create a comprehensive
+demonstration of how SLSA can be used to secure the software supply chain, from
+source code to end-user consumption. These implementations will serve as a
+reference for the community, showcasing best practices and providing a clear
+adoption path for organizations looking to improve their software supply chain
+security.
+
+We’re looking for examples that document the entire end-to-end story and include
+working implementations that people can try for themselves. We can then share
+these examples by hosting them on the SLSA blog or by linking to them elsewhere.
```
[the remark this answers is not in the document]
arewm: What would you anticipate organizations without an open source implementation would be able to demonstrate?

arewm (commented):
Thanks for converting this from the google document before I had a chance to.

docs/_posts/2025-07-16-slsa-e2e.md
```diff
@@ -0,0 +1,202 @@
...
+-   **Generate an SBOM:** (Optional) Generate a Software Bill of Materials (SBOM)
+  in a standard format (e.g., SPDX, CycloneDX) that lists all the components and
+  dependencies of the built artifact.
+-   **Collect source VSAs for all used source dependencies:** For each dependency,
+  retrieve and store its source VSA to verify its origin and integrity.
+    -   **Optionally (additionally) verify VSA during build and fail build if
+    absent/doesn’t meet requirements:** Implement a build-time check to
+    [verify the validity of each dependency's source VSA](/spec/v1.2-rc1/verifying-source#how-to-verify-slsa-a-source-revision)
+    and fail the build if a VSA is missing or does not meet the defined security
+    policy.
```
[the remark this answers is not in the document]
arewm: Ah, of course. The use of the word `dependency` threw me off. We don't use the word dependency within the source track with this meaning. I think the sources would be represented as `resolvedDependencies` in the provenance. 

Does this make it any clearer or worse?

```suggestion
-   **Collect source VSAs for all used source resolvedDendencies:** For each source,
  retrieve and store its VSA to verify its origin and integrity.
    -   **Optionally (additionally) verify VSA during build and fail build if
    absent/doesn’t meet requirements:** Implement a build-time check to
    [verify the validity of each dependency's source VSA](/spec/v1.2-rc1/verifying-source#how-to-verify-slsa-a-source-revision)
    and fail the build if a VSA is missing or does not meet the defined security
    policy.
```

docs/_posts/2025-07-16-slsa-e2e.md
```diff
@@ -0,0 +1,202 @@
...
+        policy.
+    -   **Check SBOM:** Check that an SBOM from the expected tooling exists.
+    -   **Check vulnerability attestation and triage:** Verify that all identified
+    vulnerabilities have been triaged and that the remediation status is
+    acceptable according to the security policy.
+-   **Issue VSA including:**
+    -   **SLSA build level:** The VSA should include the SLSA build level that the
+    artifact has achieved.
+    -   **Minimum SLSA Source Level of all sources:** The VSA should include the
+    minimum SLSA source level of all the source code used in the build.
```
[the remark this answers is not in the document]
arewm: I feel like this might be better to leave out for now? We can create an issue for it to discuss it further as well.

If you have a dependency on a Python wheel, this statement is not saying anything about the source levels. If you consume an sdist instead, however, this requirement would come into effect.

Maybe that is intended? Should the source track provide a way to distinguish between these sources?

docs/_posts/2025-07-16-slsa-e2e.md
```diff
@@ -0,0 +1,202 @@
...
+-   **Support for attestations from different producers:** The implementation may
+  be able to consume and verify attestations from different producers, such as
+  different builders, source control systems, and SBOM generators.
+
+## Submission
+
+Submissions should document:
+
+-   **How they work:** Explains the end-to-end workflow of the implementation.
+-   \*\*How they leverage existing standards: \*\*Note which standards or existing
```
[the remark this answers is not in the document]
arewm: You marked the conversation as resolved with a suggestion that hasn't been applied. :)

docs/_posts/2025-07-16-slsa-e2e.md
```diff
@@ -0,0 +1,202 @@
...
+-   **Gaps and future areas of improvement:** A document that identifies the
+  current limitations of the implementation and suggests areas for future
+  improvement.
+-   **A working example:** A complete, working example of the implementation.
+
+To submit your example end-to-end implementations please either send a PR
+creating the blog post yourself
+([example](https://github.com/slsa-framework/slsa/pull/1435)) or
+[create an issue](https://github.com/slsa-framework/slsa/issues) with the title
+“SLSA e2e: …”..
```
arewm, line 200:
```suggestion
To submit your example end-to-end implementations please either send a PR
creating the blog post yourself
([example](https://github.com/slsa-framework/slsa/pull/1435)) or
[create an issue](https://github.com/slsa-framework/slsa/issues) with the title
“SLSA e2e: …”. If your example is too large for either of these formats,
you may put the remaining walkthrough content in an external resource
(i.e. a git repository) and link to it.
```

docs/_posts/2025-07-16-slsa-e2e.md
```diff
@@ -0,0 +1,202 @@
...
+  policy for advanced users.
+-   **Integration with other security tools**
+-   **Support for additional attestation formats**
+-   **Support for attestations from different producers:** The implementation may
+  be able to consume and verify attestations from different producers, such as
+  different builders, source control systems, and SBOM generators.
+
+## Submission
+
+Submissions should document:
```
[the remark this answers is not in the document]
arewm: Added a suggestion below.

trishankatdatadog:
Thanks! Will review this PR today...

docs/_posts/2025-07-16-slsa-e2e.md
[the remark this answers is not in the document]
arewm: I don't think we came to a resolution. I don't think we have precisely defined language to differentiate between the sources code for an artifact's build vs its dependencies.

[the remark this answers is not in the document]
arewm: How about something like this? It isn't resolved... if want to show how you capture it, then great.
```suggestion
    -   **Minimum SLSA Source Level of all sources:** The VSA should include the
    minimum SLSA source level of all the source code used in the build. This can
    optionally include any dependencies if they are built from source.
```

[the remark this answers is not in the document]
TomHennen: Works for me! Thanks!

arewm (approved):

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,207 @@
...
+  This can link to existing documentation of any technology used.
+-   **The policy management process for the software producer:** A document that
+  explains how to create, manage, and enforce security policies.
+-   **Gaps and future areas of improvement:** A document that identifies the
+  current limitations of the implementation and suggests areas for future
+  improvement.
+-   **A working example:** A complete, working example of the implementation.
+
+To submit your example end-to-end implementations please either send a PR
+creating the blog post yourself
```
puerco, line 198: Not really important but a nit to provide a little bit more clarity:

Until I read all the way to the end I it became clear that the examples are to be published as blog posts in this line which just takes it as a given. Perhaps this should be clarified in the intro paragraphs. Coming from an implementer's point of view I was wondering about the submission medium and shape all the way until the end.

puerco (approved):
Looks good to me, just a small suggestion:

docs/_posts/2025-07-21-slsa-e2e.md
TomHennen: Thanks, that's a good point.  Updated the objective section and clarified.

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,205 @@
...
+-   **Generate Build Provenance:** Generate SLSA build Level 3-compliant build
+  provenance, which includes information about the build environment, the
+  specific build steps, and the inputs and outputs of the build process.
+-   **Generate Build Environment attestations**: (Optional) Generate SLSA BuildEnv
+  L2-compliant BuildEnv attestations, which includes integrity information about
+  the build environment at the start of the build.
+-   **Generate an SBOM:** (Optional) Generate a Software Bill of Materials (SBOM)
+  in a standard format (e.g., SPDX, CycloneDX) that lists all the components and
+  dependencies of the built artifact.
+-   **Collect source VSAs for all used source repos (as listed in resolvedDendencies):** For each source,
```
zachariahcox, line 74: It seems like we should verify VSAs for all dependencies, not just source, right?

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,205 @@
...
+Implementations should:
+
+-   **Protect a source repo with the SLSA Source Track:** Implement SLSA Source
+  Level 3 for source control, including measures like branch protection,
+  mandatory code review, and protection against tampering with the source code
+  history. Use existing implementations if desired.
+    -   **Record Source Provenance:** Generate and store SLSA source provenance
+    attestations for all source code, capturing information about the author,
+    committer, and the specific commit hash.
+    -   **Issue Source VSA:** Generate a Verifiable Source Attestation (VSA) for the
```
zachariahcox, line 48: I belive this should be 
```suggestion
    -   **Issue Source VSA:** Generate a Verification Summary Attestation (VSA) for the
```

right? https://slsa.dev/spec/v1.2-rc1/verification_summary

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,205 @@
...
+themselves.
+
+## Timeline
+
+There’s no specific deadline, but a reasonable goal would be to have example
+implementations published by September 30th, 2025.
+
+## Stages
+
+For the purposes of the end-to-end story we’ve broken the SDLC into 5 stages
```
zachariahcox, line 29: Mostly as an observation, this stage model combines verify/publication phase for "source" while it's separate for "build"

* source
  * verify
  * publish
* build & test
  * verify 
  * publish
* use
  * verify
  * profit!

Where the "verify" parts are repeated before starting each new phase.

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,205 @@
...
+which we’d expect an end-to-end implementation to cover: Source, Build & Test,
+Verification, Publication, and Use.
+
+### Source
+
+The source stage covers applying SLSA protections to source code leveraging the
```
marcelamelara, line 35: Perhaps a nitpick, but I think we should be clear that SLSA isn't a security mechanism per se.

```suggestion
The source stage covers applying SLSA requirements to source code management leveraging the
```

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,205 @@
...
+    -   **Issue Source VSA:** Generate a Verifiable Source Attestation (VSA) for the
+    source code, which cryptographically guarantees the integrity and provenance
+    of the code.
+
+<p style="text-align: right">
+Potential tools: <a href="https://github.com/slsa-framework/slsa-source-poc">slsa-source-poc</a>, <a href="https://gittuf.dev">gittuf</a></p>
+
+### Build & Test
+
+The build & test stage covers applying SLSA protections to the way software is
```
marcelamelara, line 57: As above
```suggestion
The build & test stage covers applying SLSA requirements to the way software is
```

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,205 @@
...
+[Dependency](/spec/draft/#dependency-track) tracks.
+
+Implementations should:
+
+-   **Generate Build Provenance:** Generate SLSA build Level 3-compliant build
+  provenance, which includes information about the build environment, the
+  specific build steps, and the inputs and outputs of the build process.
+-   **Generate Build Environment attestations**: (Optional) Generate SLSA BuildEnv
+  L2-compliant BuildEnv attestations, which includes integrity information about
+  the build environment at the start of the build.
```
marcelamelara, line 70: I'm ok with leaving this step as optional, but I'm a bit confused as to why we're limiting the BuildEnv track to L2, not L3, given that it is possible to implement L3 today. I suggest adding a line to that effect.
```suggestion
  L2-compliant BuildEnv attestations, which includes integrity information about
  the build environment at the start of the build. If trusted hardware is available,
  additionally generate L3-compliant BuildEnv attestations.
```

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,205 @@
...
+
+Implementations should:
+
+-   **Allow software producers to define expectations used during verification
+  (i.e. a policy):** Provide a mechanism for software producers to define a
+  security policy that specifies the requirements for a successful verification.
+-   **Verification:** Implement a verification process that checks the following:
+    -   **Check build provenance:**
+        -   **Comes from the expected builder:** Verify that the build was performed
+      by a trusted and authorized builder.
```
marcelamelara, line 106: I'm not sure if this is meant to include checking the BuildEnv attestation. If not, I would add another sub-bullet:

```suggestion
        -   **Comes from the expected builder:** Verify that the build was performed
      by a trusted and authorized builder.
      -    **Build was run in the expected build environment (optional):** Verify that
      the build ran in a good known build environment as advertised by the trusted
      builder.
```

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,205 @@
...
+            -   **(optionally) has a desired source level:** Verify that each source
+        dependency meets the minimum SLSA source level defined in the security
+        policy.
+    -   **Check SBOM:** Check that an SBOM from the expected tooling exists.
+    -   **Check vulnerability attestation and triage:** Verify that all identified
+    vulnerabilities have been triaged and that the remediation status is
+    acceptable according to the security policy.
+-   **Issue VSA including:**
+    -   **SLSA build level:** The VSA should include the SLSA build level that the
+    artifact has achieved.
```
marcelamelara, line 122: As above, the VSA should also optionally include the BuildEnv level at which the artifact was built.

marcelamelara (changes requested):
Thanks @TomHennen . I'm generally supportive of this RFE and hope get good submissions. I would like to see the BuildEnv track covered a little more explicitly given that there do already exist (partial) implementations of it, so it's not inconceivable that these could be part of an e2e implementation.

marcelamelara:
> As discussed earlier in the Google doc, I understand where this is coming from, but as someone who has contributed to both SLSA and in-toto, I'd like to make sure both continue to play well, and that we do not accidentally diverge the two.
> 
> Personally, I think this work should really largely be done upstream with in-toto policies as @marcelamelara and I [presented](https://www.youtube.com/watch?v=lX_sAD68k2Q&list=PLVl2hFL_zAh9KQkOh4b2gcLDf35LwkA9S&index=26) recently. This is largely for the same reasons that the Attestation framework exists, and that we would like all relevant attestations such as the Build Provenance and any VSA to live there.
> 
> Now, one might argue that SLSA doesn't require in-toto, but it's hard to see how to enforce SLSA without in-toto. Already people are producing signed attestations (Build Provenance, Source VSA coming soon), and the reason this RFP exists is so that we would like to do something meaningful with them.
> 
> Having said that, and going back to the RFP: agree with many of the requirements (such as using attestations and producing VSAs as a result). Personally, I think the community should collaborate on in-toto policies which will allow for multiple implementations that are allowed to diverge while at the same time adhering to a common standard. I strongly believe this would be the best path forward for all involved, and can be done within the proposed timeline. WDYT?

I wanted to respond since I have a different take. I don't think this RFE precludes in-toto in anyway, nor do I think that in-toto Policies (the spec and Policy definitions) fully address what I think this RFE is trying to accomplish, which is to encourage more implementations and adoptions of SLSA end-to-end. Because in-toto Layouts/Policies are specifically designed to express end-to-end supply chain expectations, they could play a key role in e2e SLSA implementations. But much how we're not being prescriptive in in-toto Policies about the policy language/engine used, SLSA shouldn't be either IMO.

docs/_posts/2025-07-21-slsa-e2e.md
```diff
@@ -0,0 +1,205 @@
...
+    attestations for all source code, capturing information about the author,
+    committer, and the specific commit hash.
+    -   **Issue Source VSA:** Generate a Verifiable Source Attestation (VSA) for the
+    source code, which cryptographically guarantees the integrity and provenance
+    of the code.
+
+<p style="text-align: right">
+Potential tools: <a href="https://github.com/slsa-framework/slsa-source-poc">slsa-source-poc</a>, <a href="https://gittuf.dev">gittuf</a></p>
+
+### Build & Test
```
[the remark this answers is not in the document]
marcelamelara: +1 to this!

marcelamelara (approved):
Thanks for the changes! LGTM

Issue #1462, opened 2025-07-14
Source Track: nitpicks on Source VSA


MarkLodato:
In [Source Track RC1: Source verification summary attestation](https://slsa.dev/spec/v1.2-rc1/source-requirements#source-verification-summary-attestation):

* Requirement 1: The `subject.uri` field needs more explanation. It currently sounds the same as `resourceUri`, but from seeing the example, I now understand that it is a URI of a (say) of web UI, which might be different than the resource URI. I recommend explaining this and giving an example.

* Requirement 2: For non-cryptographic digest, I recommend saying that this SHOULD be a full URI including the version identifier. For example, in Subversion’s case, this should not be “2019” but instead “svn+https://svn.myproject.org/svn/MyProject/trunk@2019”. Otherwise the revision is so general as to be useless.

* Requirement 3: "MAY NOT" is not a valid RFC 2119 keyword. Did you mean “MAY be non-exhaustive”? Also missing a period at end of sentence.

* Requirements 3: `source_refs` should be `sourceRefs` for consistency with other fields.

* Requirements 6: "COULD" is not a valid RFC 2119 keyword. Did you mean “MAY”, or perhaps "For example, this could..."?

PR #1468, opened 2025-07-18
content: draft: fix various source VSA issues.

TomHennen:
fixes https://github.com/slsa-framework/slsa/issues/1462

* Following RFC 2119
* Clarfiied subject.uri
* source_refs -> sourceRefs

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -473,10 +473,10 @@ The source track issues Source VSAs using the [Verification Summary Attestations
 1.  `subject.uri` SHOULD be set to a URI where a human can find details about
     the revision. The URI MUST NOT be used during verification, but SHOULD be used
     to direct a human investigating verification failures.
-    - For example: `https://github.com/slsa-framework/slsa/commit/6ff3cd75c8c9e0fcedc62bd6a79cf006f185cedb`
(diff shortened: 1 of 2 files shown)


MarkLodato (approved):

zachariahcox (approved):
looks great. thanks!

docs/spec/draft/source-requirements.md
```diff
@@ -470,16 +470,20 @@ the changes that contributed to that revision over its history.
 
 The source track issues Source VSAs using the [Verification Summary Attestations](./verification_summary.md) format as follows:
 
-1.  `subject.uri` SHOULD be set to a human readable URI of the revision.
+1.  `subject.uri` SHOULD be set to a URI where a human can find details about
+    the revision. The URI MUST NOT be used during verification, but SHOULD be used
```
lehors, line 474: Isn't "MUST NOT" a bit too strong here?

MarkLodato: I don't have a strong opinion here, but one alternative could be to avoid MUST vs SHOULD and instead say something like, "This field is not intended for policy decisions. Instead, it is intended only for [...]." or something like that.

lehors: > I don't have an opinion here, but one alternative could be to avoid MUST vs SHOULD and instead say something like, "This field is not intended for policy decisions. Instead, it is intended only for [...]." or something like that.

That sounds good to me.

lehors (approved):
LGTM. Thank you!

Issue #1459, opened 2025-07-14
Suggested reorganization of Source Track


MarkLodato:
Hi all! Nice work on the upcoming Source Track! I'm excited to see this happening.

The following is based on [RC1](https://slsa.dev/spec/v1.2-rc1/source-requirements).

At a high level, I basically agree with all of the requirements, but I suggest reorganizing them (and generalizing a few) to make the track easier to understand and to deliver value more incrementally.

#### Suggested reoganization

<table>
<tr><th> Security Requirement <th> L1 <th> L2 <th> L3 <th> L4
<tr><td>

Repositories and revisions are uniquely identifiable, and revisions are immutable

*[unchanged]*

<td> ✓ <td> ✓ <td> ✓ <td> ✓

<tr><td>

Version history is trustworthy, continuous, authenticated, immutable, and retained

*[combination of History, **Source Provenance (moved from L3 and generalized)**, Continuity, Identity Management (strengthened), and Safe Expunging Process (rephrased as retention)]*

<td> <td> ✓ <td> ✓ <td> ✓

<tr><td>

Changes are submitted to a defined "published" branch or tag that enforces continuity and change management processes

*[combination of Protected Branches, Protected Tags, and Enforced Change Management Process **(all moved from L2)**]*

<td> <td> <td> ✓ <td> ✓

<tr><td>

All changes undergo two-party review

*[unchanged]*

<td> <td> <td> <td> ✓

<tr><th> Interoperability Requirement <th> L1 <th> L2 <th> L3 <th> L4

<tr><td>

Security state is communicated (via VSA, API, URI, etc.)

*[generalization of Source VSA requirement]*

<td> ✓ <td> ✓ <td> ✓ <td> ✓
</table>

#### Description of changes

1.  Group requirements into higher level abstractions to aid readability and ensure cohesion. It is much easier to understand if each level requires only one or two things with a clear name. The current RC1 has a seeming grab-bag of requirements that takes a fair amount of thinking to understand why it adds up to a whole.

2.  Split L2 into two levels. In RC1, L2 is too big.

    -   L2 is now about "trustworthy history" or "auditability". When a particular artifact meets Source L2, you can investigate it and find out what changes were made when, why, and by whom. All of the previous requirements (listed above) are sub-requirements that add up to this overall property.

        -   Notably, it is possible for something to meet L2 but not L3. This might be valuable if engineers are writing and deploying experimental code. The bar for experimental might be lower than production, but you still want to audit it.

[...]

PR #1483, opened 2025-09-15
content: reorganize levels, update requirement text

zachariahcox:
Related to points 2 and 3 of https://github.com/slsa-framework/slsa/issues/1459

This PR reorganizes the levels to achieve slightly different goals. 
There is still work to do here 😅 

Major content changes:
1. Level 2 now focuses on providing evidence that a revision has reliable history. Dropped the generic term "protect" in favor of specific requirements for access and branch technical controls.
1. Level 3 now focuses on the _continuous enforcement_ aspect of technical controls. The continuity concept replaces the previous "org intent" / "consumable branches" concepts.
2. Protected tags and branches were consolidated into just "named references." The requirement to prevent tags from moving was moved from the SCS to the Org.

Minor content changes:
1. Lots of minor text edits to requirements text to align with the major changes.
3. Additional examples were added to clear up the difference between "revision ancestry" and branch "pointer history." 
7. Removed the term "change management process" in favor of "technical controls" in all cases.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -561,8 +561,7 @@ Downstream users are expected to be familiar with the method used by the issuer.
 Example implementations:
 
 -   Issue a new VSA for each merged Pull Request and add the destination branch to `sourceRefs`.
--   Issue a new VSA each time a 'consumable branch' is updated to point to a new revision.
--   Issue a new VSA each time a 'consumable tag' is created to point to a new revision.
+-   Issue a new VSA each time a Named Reference is updated to point to a new revision.
 
 #### Example
 


docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,51 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
+over time, enabling attribution of those changes to the actors that made them.
+Source Provenance provides strong, tamper-resistant evidence of the process that
+was followed to produce a Source Revision.
 
 </dl>
 </section>
 <section id="source-l3">
 
-### Level 3: Signed and Auditable Provenance
+### Level 3: Communication of Organization Intent
```
zachariahcox, line 135: This section is still rough. 
I don't dislike the direction though -- "public declaration of intent to empower verifiers" is probably a good level 3

docs/spec/draft/source-requirements.md
```diff
@@ -429,20 +420,35 @@ associated with the revision identifier delivered to consumers and are a
...
-At Source Level 3, Source Provenance MUST be created contemporaneously with the
-branch being updated to use that revision such that they provide a credible,
-auditable, record of changes.
+Source Provenance MUST be created contemporaneously with the branch being
+updated such that they provide a credible, auditable, record of changes.
 
-If a consumer is authorized to access a revision, they MUST be able to fetch the
-corresponding source provenance documents for that revision.
+If a consumer is authorized to access a revision, they MUST be able to access the
+corresponding Source Provenance documents for that revision.
```
zachariahcox, line 428: consolidated this into the source provenance part -- it seems mostly to be a provenance generation control

TomHennen:
This is meant to fix https://github.com/slsa-framework/slsa/issues/1459, right?

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,51 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
-The SCS generates credible, tamper-resistant, and contemporaneous evidence of how a specific revision was created.
-It is provided to authorized users of the source repository in a documented format.
+The Organization clarifies which Named References in a repo are consumable and
+defines which technical controls are guaranteed in those contexts.
 
 <dt>Intended for<dd>
 
-Organizations that want strong guarantees and auditability of their change management processes.
+Organizations with many Branches and Tags or who otherwise need to specify
+the expected way to consume and verify their Source Revisions.
```
TomHennen, line 146:
```suggestion
Organizations with many Branches and Tags or who otherwise need to document
which controls were in place as Source Revisions were created.
```
One of the major uses of this level, IMO, is for orgs that having compliance requirements that include things that dictate how source is created.  Having those controls enforced by the SCS _and encoded in provenance and VSAs_ would make audits much easier.

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,51 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
 
 <dt>Benefits<dd>
 
-Provides information to policy enforcement tools to reduce the risk of tampering
-within the SCS's storage systems.
+A verifier can use this published data to ensure that a given revision was
+acquired in the correct way and to verify that the Source Provenance (or VSA)
+contains evidence of all expected claims and properties.
```
TomHennen, line 152:
```suggestion
A verifier can use this published data to ensure that a given revision was
created in the correct way by examining the Source Provenance or VSA
for all expected claims and properties.
```

docs/spec/draft/source-requirements.md
```diff
@@ -193,27 +199,14 @@ attestations.
...
-The organization MUST specify which branches and tags are covered by the process
-and are intended for use in its own applications or services or those of
-downstream consumers of the software.
-
-> For example, if an organization has branches 'main' and 'experimental' and it
-intends for 'main' to be protected then it MUST indicate to the SCS that 'main'
-should be protected. From that point forward revisions on 'main' will be
-eligible for Source Level 2+ while revisions made solely on 'experimental' will
-not.
+<tr id="configure-controls"><td>Enable technical controls to enforce intent <a href="#configure-controls">🔗</a><td>
```
TomHennen, line 202: I'm not sure about this title it's talking about intent (L3 stuff) but then also does identity management (should be L2).

docs/spec/draft/source-requirements.md
```diff
@@ -193,27 +199,14 @@ attestations.
...
 [Identity Management capability](#identity-management) to configure the actors
-and roles that are allowed to perform sensitive actions on protected branches
-and tags.
+and roles that are allowed to perform sensitive actions.
 
-> For example, an organization may configure the SCS to assign users to a `maintainers` role and only allow users in `maintainers` to make updates to `main`.
+> For example, an organization may configure the SCS to assign users to a
+`maintainers` role and only allow users in `maintainers` to make updates to `main`.
 
 The organization MUST specify what technical controls consumers can expect to be
```
TomHennen, line 211: Here and below should be moved to L3 as it's not a history related control.

docs/spec/draft/source-requirements.md
```diff
@@ -255,13 +247,25 @@ The content may still exist in other copies of the repository, either in backups
...
 
-SCSs SHOULD have technical mechanisms in place which require an Administrator plus, at least, one additional 'trusted person' to trigger any expunging (removals) made under this process.
+SCSs SHOULD have technical mechanisms in place which require an Administrator plus at least one additional 'trusted person' to trigger any expunging (removals) made under this process.
 
 The application of the safe expunging process and the resulting logs MAY be private to both prevent calling attention to potentially sensitive data (e.g. PII) or to comply with local laws
 and regulations which may require the change to be kept private to the extent possible.
 Organizations SHOULD prefer to make logs public if possible.
 
 <td><td>✓<td>✓<td>✓
+<tr id="intent"><td>Declare which revisions are consumable <a href="#intent">🔗</a><td>
```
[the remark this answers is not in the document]
TomHennen: I bet you can take the "The organization MUST specify what technical controls consumers can expect" bit from above and merge in here too (or have another row below it).

docs/spec/draft/source-requirements.md
```diff
@@ -309,60 +313,46 @@ If the SCS DOES NOT generate a VSA for a revision, the revision has Source Level
...
 The SCS MUST prevent tampering with these records on protected branches.
 
 > For example, in systems like GitHub or GitLab, this can be accomplished by
 enabling branch protection rules that prevent force pushes and branch deletions.
 
+If revisions have their own ancestries in your VCS, the SCS MUST ensure that all
+new revisions added to a branch are descendants of the previous revisions.
+
+> For example, if `main` currently points to revision `a`, it may only be
+moved to a new revision, `b`, if `b` has `a` somewhere in its revision ancestry.
```
TomHennen, line 337: Is this strictly needed?  Source Revisions already need to include their parent revision https://slsa.dev/spec/draft/source-requirements#:~:text=author%2C%20timestamp%2C%20and-,parent%20revision,-(s).%20Note%3A%20Path.

docs/spec/draft/source-requirements.md
```diff
@@ -309,60 +313,46 @@ If the SCS DOES NOT generate a VSA for a revision, the revision has Source Level
...
+`refs/heads/releases/*`, but not `refs/heads/experimental/*` using branch
+protection rules (e.g. [GitHub](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets),
 [GitLab](https://docs.gitlab.com/ee/user/project/repository/branches/protected.html))
-which require additional checks to 'pass' (e.g. unit tests, linters) or the
-application and verification of [gittuf](https://github.com/gittuf/gittuf)
+or via the application and verification of [gittuf](https://github.com/gittuf/gittuf)
 policies.
 
+
 <td><td>✓<td>✓<td>✓
```
TomHennen, line 356:
```suggestion
<td><td><td>✓<td>✓
```

TomHennen: This is the way orgs signal which branches are consumable, no?  That's how I'd been thinking about it.

If it's just about communicating which branches meet L2 requirements then the VSA alone can do that and we don't need anything special at L2.

docs/spec/draft/source-requirements.md
```diff
@@ -429,20 +420,35 @@ associated with the revision identifier delivered to consumers and are a
...
+   enforced.
+-   Allow organizations to distribute additional attestations related to their
+   technical controls to consumers authorized to access the corresponding Source
+   Revision.
+-   Prevent organization-specified properties from beginning with any value
+   other than `ORG_SOURCE_` unless the SCS endorses the veracity of the
+   corresponding claims.
+
+
+<td><td>✓<td>✓<td>✓
```
TomHennen, line 449:
```suggestion
<td><td><td>✓<td>✓
```

TomHennen (changes requested):
Thanks for this Zach it's looking promising!

docs/spec/draft/source-requirements.md
zachariahcox: I like your phrasing better. I did have the word "acquire" in there, because I think the verifier is supposed to check the uri? maybe we can skip that concept. 
maybe it also doesn't matter if the consumer gets the revision from the canonical location or not, as long as the provenance was signed correctly or the VSA exists?

TomHennen: Yeah, we have a whole section on how to verify the VSA.  Maybe we just change 'examining' to 'verifying'?

docs/spec/draft/source-requirements.md
zachariahcox: As I think about it, I think this makes sense. 
The main change for L3 is on the org side, "declaration of intent."
What tooling changes would be needed just for L3? Probably none, if we already have to generate source provenance for those controls at L2, and provenance requires knowledge of branches or tags I think.

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,51 @@ Migrating to the appropriate tools is an important first step on the road to ope
 </section>
 <section id="source-l2">
 
-### Level 2: Controls
+### Level 2: History & Provenance
```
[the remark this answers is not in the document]
TomHennen: The main thing is that in the discussion we've been having those additional controls are introduced at level 3.  Are you proposing changing that?

docs/spec/draft/source-requirements.md
TomHennen: > Good question. As a starting point, wouldn't these have to be L2 in order to let the Org configure them?

Some SCSs make this the default. No configuration necessary.

> What tooling changes would be needed just for L3? Probably none, if we already have to generate source provenance for those controls at L2, and provenance requires knowledge of branches or tags I think.

Some SCSs don't have the ability to let orgs configure arbitrary controls.  If we move that to L2 then those systems can only achieve L1.

docs/spec/draft/source-requirements.md
```diff
@@ -309,96 +340,86 @@ If the SCS DOES NOT generate a VSA for a revision, the revision has Source Level
...
 
-The SCS MUST prevent protected tags from being moved or deleted.
+> For example, the main branch currently points to revision `a` and a new
+technical control `t` is configured.
+The next revision on the `main` branch, `b` will be the first revision that can
+claim to have been protected by `t` and `b` is the first revision in the
+"continuity" of `t`.
+If in the future, `t` is disabled, any revisions added to `main` while `t` is
+disabled cannot claim to have been protected by `t` and the "continuity" of `t`
+will be reset.
```
zachariahcox, line 419: A draft of a continuity example, cc @arewm 

maybe defining what `t` is would make it more clear? IE, "no force pushes". If you let that one lapse for a single force push, you have lost a LOT of context on the branch

docs/spec/draft/source-requirements.md
```diff
@@ -429,20 +433,65 @@ associated with the revision identifier delivered to consumers and are a
...
+
+The SCS MUST provide a mechanism for organizations to indicate which branches
+should be protected by technical controls.
+
+> For example, the organization may instruct the SCS to protect `main` and
+`refs/heads/releases/*`, but not `refs/heads/experimental/*` using branch
+protection rules (e.g. [GitHub](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets),
+[GitLab](https://docs.gitlab.com/ee/user/project/repository/branches/protected.html))
+or via the application and verification of [gittuf](https://github.com/gittuf/gittuf)
+policies.
```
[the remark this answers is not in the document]
zachariahcox: they need to be _configured_ to be unmovable. I don't think there are any SCSs that would have them actually _be_ unmovable. I think we can get away by calling it a control required to achieve the history requirement.

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,57 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
+over time, enabling attribution of those changes to the actors that made them.
+Source Provenance provides strong, tamper-resistant evidence of the process that
+was followed to produce a Source Revision.
 
 </dl>
 </section>
 <section id="source-l3">
 
-### Level 3: Signed and Auditable Provenance
+### Level 3: Communication and Enforcement of Intent
```
zachariahcox, line 136: If you want to claim something, you must have documented evidence of mechanically-enforced technical controls for it. You can't just promise to unit test.

We can maybe make that even more clear:

"Organizations mut take a hard look at all things they want to be true about their software. Do they want code review? SAST? Unit testing? 
Which of those things do they commit to actually doing? This represents their intention. 
At this Level, organizations must deploy technical controls that hold them accountable to their commitments. There must be documented evidence of all technical controls that govern the repo, and together, these must combine to ensure that each new revision is a faithful representation of the org's intent for the repo."

docs/spec/draft/source-requirements.md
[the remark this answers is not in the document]
TomHennen: > I can't quite imagine what tools would get you into that situation, but I fully agree with the sentiment.

Custom built ones (I assure you they exist :) )

docs/spec/draft/source-requirements.md
```diff
@@ -182,86 +188,111 @@ Many examples in this document use the [git version control system](https://git-
...
 
 <td><td>✓<td>✓<td>✓
-
 <tr id="safe-expunging-process"><td>Safe Expunging Process <a href="#safe-expunging-process">🔗</a><td>
 
-SCSs MAY allow the organization to expunge (remove) content from a repository and its change history without leaving a public record of the removed content,
-but the organization MUST only allow these changes in order to meet legal or privacy compliance requirements.
-Content changed under this process includes changing files, history, references, or any other metadata stored by the SCS.
+SCSs MAY allow the organization to expunge (remove) content from a repository
+and its change history without leaving a public record of the removed content,
```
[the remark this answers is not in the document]
TomHennen: Seems fine to me but we should probably not use a git specific term like force pushing (should already be covered by history protection?).

Alternatively, leave it alone, if we soften this further it's an easy change that doesn't require a major change.

docs/spec/draft/source-requirements.md
```diff
@@ -182,86 +188,111 @@ Many examples in this document use the [git version control system](https://git-
...
-enforce this requirement and store corresponding [test result attestations] for
-all affected revisions.  They may then embed the `ORG_SOURCE_UNIT_TESTED`
-property in the [Source VSA](#source-verification-summary-attestation). Consumers
-would then expect that future revisions on `main` have been united tested and
-determine if that expectation has been met by looking for the
-`ORG_SOURCE_UNIT_TESTED` property in the VSAs and, if desired, consult the
-[test result attestations] as well.
+If the SCS supports "tags" (or other Named References that do not support
+continuity enforcement mechanisms or change management processes), the SCS MUST
+be configured to prevent tags from being moved or deleted.
```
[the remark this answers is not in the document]
TomHennen: I think this could work here but I think the section could use some tweaking.

One thing that might be causing some trouble is lack of clarify around what the level applies to. As written this is somewhat ambiguous.  As long as the org has protected a single branch does all their source qualify as L2?  I think we understand that the answer is 'no', only revisions created on that branch would qualify as L2 (and there would be a VSA & prov linking the revision to that branch).  But I don't think we communicate this anywhere (should we?).

Contrast this with the current language for tags, "the SCS MUST
be configured to prevent tags from being moved or deleted" if this is using our current definition of tags (something that shouldn't move) this is a bit of a tautology.  If we're using the looser definition of tags (e.g. maybe it could move) then it just conflicts with our other definition.

What we should probably try to communicate is "if you want your users to be able to consume a revision from a named reference verified at L2+ then that named reference must either be a branch with its history protected or a tag that is prevented from being moved or deleted".

docs/spec/draft/source-requirements.md
TomHennen: IDK, maybe I'm daft but I think this is covered by 

"The SCS MUST record the sequence of all Source Revisions ever referenced by a
branch."

If they force push 'legal' commits then that record of the sequence of Source Revisions is gone?

If it is necessary, maybe it can be said more plainly?

docs/spec/draft/source-requirements.md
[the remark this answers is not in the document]
TomHennen: The SLSA Source Tool implementation actually can prevent them from being moved even by admins unless they change their policy, which resets the clock at which they've had the control enabled, making the bypass visible.

_That_ SCS includes GitHub but goes beyond what it provides by adding additional controls on top.

Maybe I'm being pedantic or missing something though.

I really do like the idea of considering the tag protection as a part of preserving history though.

I'm coming around to combining them though. I think this would actually part of the declaring intent that we discussed for L3 at the last meeting [link](https://github.com/slsa-framework/slsa/issues/1459#:~:text=L3%20would%20be%20about%20organizational%20intent%20(%22is%20this%20branch%20meant%20for%20consumption%3F%22).

The existing protecte branches/tags control served to both preserve history and declare intent.  History is now preserved elsewhere, but we could combine them here to say that the SCS must enable orgs to declare which branches and tags are meant for consumption and what controls they have in place (e.g. IMMUTABLE_REFERENCE, TESTED, HISTORY)?

TomHennen (changes requested):
Sorry for the long turnaround on these last few comments, I find I really need some time to focus to think about them.

docs/spec/draft/source-requirements.md
zachariahcox: There are two parts for the history reqirement:
1. reflog: the SCS must keep track of the sequence of revisions ever-pointed-to by a ref.
2. lineage: each new revision on a ref must be a descendent of the previous revision in the reflog (or null).

If your reflog currently has the ordered list [null,a,b,c], you may not update it to be [null,a,b,c,b], because b is not a descendent of c.
If you want to "go back" to the tree id captured by b, you must introduce a new "revert" commit d which descends from c. 

This is the core concept of the "no force pushes" rule.

docs/spec/draft/source-requirements.md
```diff
@@ -182,86 +190,114 @@ Many examples in this document use the [git version control system](https://git-
...
 
-> For example, if an organization has branches 'main' and 'experimental' and it
-intends for 'main' to be protected then it MUST indicate to the SCS that 'main'
-should be protected. From that point forward revisions on 'main' will be
-eligible for Source Level 2+ while revisions made solely on 'experimental' will
-not.
+The organization MUST configure technical controls provided by the SCS to
+enforce the organization's intention for the Source Repository.
+At a minimum, the SCS MUST be configured to produce a reliable
+[Change History](#history) on at least one branch.
```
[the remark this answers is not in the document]
TomHennen: Yes I think that works.  Still need to update the language though I think.  I generally like @MarkLodato's framing around L2 being about History and Access Control.  L3 seems much better positioned to get into 'intent'.

docs/spec/draft/source-requirements.md
TomHennen: Perhaps this can be resolved by adding a requirement that the history of a branch must be linear?  That seems less wonky/git specific.

docs/spec/draft/source-requirements.md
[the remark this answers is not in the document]
zachariahcox: So, if we cut the intent line you'd be ok? That's fine I think.

docs/spec/draft/source-requirements.md
```diff
@@ -309,96 +345,59 @@ If the SCS DOES NOT generate a VSA for a revision, the revision has Source Level
...
-   [Source VSA](#source-verification-summary-attestation) when the corresponding controls are
-   enforced.
--   Allow organizations to distribute additional attestations related to their
-   technical controls to consumers authorized to access the corresponding source
-   revision.
--   Prevent organization-specified properties from beginning with any value
-   other than `ORG_SOURCE_` unless the SCS endorses the veracity of the
-   corresponding claims.
+If revisions have their own ancestries in your VCS, the SCS MUST ensure that all
+new revisions added to a branch are descendants of the previous revisions.
```
zachariahcox, line 366:
```suggestion
If revisions have their own ancestries in the VCS, the SCS MUST ensure that all
new revisions added to a branch are descendants of the previous revision of the branch.
```

docs/spec/draft/source-requirements.md
TomHennen: Hmm, going back to the original point, I'm thinking the language "The SCS MUST record the sequence of all Source Revisions ever referenced by a
branch." provides enough justification for blocking force push?  Doesn't force push get rid of that sequence?

My main motivation here is to try to keep the language as simple as possible.

"If revisions have their own ancestries in your VCS, the SCS MUST ensure that all..." would also seem to fall into a category of being covered well enough?

If not, what additional problem (beyond keeping the history of all the revisions) is this solving?

MarkLodato (changes requested):
Huge improvement! Thanks for the hard work here.

In order to get this to land sooner, you might want to consider splitting this PR in two:

- Pure editorial changes --- grammar, typos, rewording, including the updated table at the top. These are uncontroversial and can just get submitted since there is little disagreement here and it's an obvious improvement.
- Content changes. This is where all of the discussion is. Splitting this to a separate PR might make it easier to discuss these changes in isolation.

Note: I didn't review any of the content changes yet in this PR since there is still a lot of back-and-forth. I'll review once those in-flight edits are made.

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,53 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
 <dt>Summary<dd>
 
-The SCS generates credible, tamper-resistant, and contemporaneous evidence of how a specific revision was created.
-It is provided to authorized users of the source repository in a documented format.
+The SCS is configured to enforce the Organization's intentions for the Source
+Repository using any necessary technical controls.
+
+The Organization clarifies which revisions and Named References in a repo are
+intended for consumption and defines which technical controls are guaranteed in
+those contexts.
```
zachariahcox, line 143: I kinda like this conversational tone better?

```suggestion
The organization clarifies which revisions and Named References in a repo are
intended for consumption and defines which technical controls are guaranteed in
those contexts.
This represents their "intention" for each branch of the repo.

At this Level, organizations must deploy technical controls that hold themselves
accountable to their commitments.
There must be documented evidence of all technical controls that govern the
branch, and how together they combine to ensure that each new revision is a
faithful representation of the organization's intent.
```

zachariahcox: > nit: Is there a case where revisions are allowed here?

I think the core idea is that revisions are the thing which are or aren't compliant -- refs just point to those things. 
A case I keep in mind for this is a CI/CD system on a fast branch where you might have multiple builds in flight for revisions along the same branch. It should be okay to build and deploy a revision as long as it has the right provenance regardless of whether it's on the tip of the branch or even reachable from the branch.

An extreme case of this could be a CD process which is in flight while simultaneously a "safe expungement" operation is happening which will ultimately orphan the revision.
Depending on why the expungement was happening, that commit may still be fine to consume, just the producer needs to change the history for some reason (maybe they're rotating a signing key and need to resign all the commits for example).

docs/spec/draft/source-requirements.md
```diff
@@ -281,6 +275,31 @@ to be kept private to the extent possible. Organizations SHOULD prefer to make
...
+those controls.
+
+> For example, if an organization implements a technical control via a custom
+tool (such as required GitHub Actions workflow), it must indicate the name of
+this tool, what it accomplishes, and how to find its evidence in the provenance
+attestation.
+
+> For another example, if the organization intends to claim that all consumable
+Source Revisions on the `main` branch were unit tested prior to acceptance, this
+MUST be explicitly configured in the SCS, not left to a purely social contract.
```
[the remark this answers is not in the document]
zachariahcox: > you can't implement in a technical control it won't be encoded in the provenance or VSA.

I've never seen anyone try to do that one formally, but I have seen claims like "all members of this team are X, and only members of that team contributed to the code". That kind of contribution policy can be verified and attached to the provenance. 

It does seem reasonable that if you want to claim in a VSA, you have to ("must") provide mechanical evidence for it somehow.

docs/spec/draft/source-requirements.md
```diff
@@ -281,6 +275,31 @@ to be kept private to the extent possible. Organizations SHOULD prefer to make
 logs public if possible.
 
 <td><td>✓<td>✓<td>✓
+<tr id="intent"><td>Declare which revisions are consumable and their technical controls<a href="#intent">🔗</a><td>
```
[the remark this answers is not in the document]
zachariahcox: We could avoid it here in particular by focusing on the activity:


```suggestion
<tr id="intent"><td>Declare where to find consumable revisions and their technical controls<a href="#intent">🔗</a><td>
```

> Consumers can think "anytime I use this named reference I should expect the corresponding source revisions to have evidence of those claims"

Agree, but it's also possible to have multiple consumable revisions reachable from a branch. There's no requirement that _only_ the tip is legal.
EG, if a release branch moves (logically moving from version 1.0 to 1.1), both commits are consumable and probably had the same technical controls.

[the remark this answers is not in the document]
TomHennen: Perhaps we can resolve this by providing more explanation either here or elsewhere.  I'll need to think about it a bit more.

On the terminology point, I've been perhaps a bit too liberal (or not thinking about git internals) when I use the langauge "consume a revision **from** B". What I mean is "consume a revision that **was pointed to by** B". I guess I'm using the branch name as a name for the ["cohesive line of development"](https://slsa.dev/spec/draft/source-requirements#:~:text=cohesive%20line%20of%20development) and not just for whatever is currently being pointed to.

There should be provenance or VSA for the revision that will indicate which named references pointed to it so consumers will have a way of determining where the thing came from.

With that my assumptions are as follows:

1. Organizations will set expectations within the context of a branch or tag.  "Things on branch A [meaning things that A has pointed to] will meet properties X, Y, Z".  This is where the expectations originate.
2. These expectations _transfer_ to the pointed to revisions.  Organizations are unlikely to explicitly set expectations for individual revisions.  E.g. they won't say "You should expect revision M to have properties X, Y" because that wouldn't be especially valuable, you can just say what properties it _does_ have (via the VSA).
3. The SCS will only issue a 'passing' VSA for a revision tied to a named reference if the revision met the expectations of that named reference.
4. Users should validate the VSA before consuming the revision.

In practice this means that when consumers set up their consumption of revisions from a source repo they'll somehow establish their baseline expectations (based on '1'), their tooling will then fetch revisions (somehow), and then check that their baseline expectations are met by the revision.

Now perhaps what you're getting at is "why do branches and tags have to enter into this at all, can't an organization say 'any revision that has properties X, Y, Z should be considered consumable, regardless of what named reference it came from'?".  Is that right?

If so, I think it might have continuity problems, but maybe it could work?

In either case, I'm thinking that it would help clarify the spec if we explained a bit more what we mean by expectations, and where they get set.  That might be something we need to discuss prior to the requirements themselves?  Maybe with the other definitions?

sumonkhan71cf-beep (approved):

docs/spec/draft/source-requirements.md
zachariahcox: > Will consumers need to update their policy once t' comes into play?
Probably yes? There are some rules where the configuration could be monotonically more strict, but I suspect most of the change would be complex like "we used to use workflow x@123 for this, but now we use y@456"

zachariahcox: > any individual commit could be protected by t in theory, but that is not sufficient for ensuring organizational intent is met from a given revision/point in time. Instead, organizations state their intent based on the protections and then acknowledge how long they have been.

Oo, I like this much better. Sort of like "continuity reflects an organization's ongoing committment to a control."
Or "it is the number of revisons in a row (on the reflog only?) where a control was enforced."

While I agree it's important to quantify how committed an org is to a control, I'm not sure how to use that data, it's pretty gameable. Would I write a policy like, i only care about controls that have been in place for 100 commits?

docs/spec/draft/source-requirements.md
```diff
@@ -332,92 +351,55 @@ the SCS-issued [source provenance](#source-provenance) when issuing
...
-<td><td>✓<td>✓<td>✓
 <tr id="history"><td>History <a href="#history">🔗</a><td>
 
-Revisions are created by applying specific code changes (a "diff" in git) on
-top of earlier revisions of a branch. This sequence of changes, the revisions
-they produced, and how they were introduced into a branch constitute the history
-of that branch.
+If a Named Reference is updated, the SCS MUST record the actor making the
+change, the new Source Revision ID and the timestamp of the change.
+The SCS MUST prevent tampering with these records.
```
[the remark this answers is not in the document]
zachariahcox: > What does it mean for an SCS to keep track of the sequence of revisions ever pointed to by a ref? 

this would be the server's reflog (GitHub draws some of that data as the branches view).

> If you create a new branch off of a current one, would the entire commit history be included in the reflog or would it start fresh from that commit?

The revision history would be the same, but I guess the branch history would be new? We've said elsewhere there would be no continutiy or provenance attestations for old commits if you import a repo from another server (ie, transferring from Azure Repos to GitHub), more or less because you don't have to trust what the other server did, you just attest to what you've seen.

docs/spec/draft/source-requirements.md
MarkLodato: Oh, I now see the other thread where this is being discussed in more detail. It sounds like we're not so far off on intent, but rather about how we explain it. I agree with Tom's suggestion that it needed more explanation in the doc.

docs/spec/draft/source-requirements.md
```diff
@@ -193,47 +203,32 @@ attestations.
...
-and tags.
+The SCS MUST be configured to produce a reliable [Change History](#history)
+for all consumable Source Revisions. If your SCS provides this capability by
+design, no additional controls are needed.
 
-> For example, an organization may configure the SCS to assign users to a `maintainers` role and only allow users in `maintainers` to make updates to `main`.
+If the SCS supports "tags", the SCS MUST be configured to prevent them from
+being moved or deleted.
+support change management), the SCS MUST be configured to prevent them from
+being moved or deleted.
```
zachariahcox, line 223:
```suggestion
```

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,53 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
+over time, enabling attribution of those changes to the actors that made them.
+Source Provenance provides strong, tamper-resistant evidence of the process that
+was followed to produce a Source Revision.
 
 </dl>
 </section>
 <section id="source-l3">
 
-### Level 3: Signed and Auditable Provenance
+### Level 3: Communication and Enforcement of Intent
```
zachariahcox, line 133:
```suggestion
### Level 3: Continuous technical controls
```

At level three you must state the controls you commit to enforcing in some contexts.

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,53 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
 
 <dl class="as-table">
 <dt>Summary<dd>
 
-The SCS generates credible, tamper-resistant, and contemporaneous evidence of how a specific revision was created.
-It is provided to authorized users of the source repository in a documented format.
+The SCS is configured to enforce the Organization's intentions for the Source
```
zachariahcox, line 138:
```suggestion
The SCS is configured to enforce the Organization's intentions for specific Named References within the Source
```

TomHennen:
Generally agreed with one tweak

> With that change, we can cut a lot of the complexity of intent and just lean on the VSA **AND PROVENENACE** existence to provide evidence of org intent.

docs/spec/draft/source-requirements.md
```diff
@@ -448,17 +425,53 @@ associated with the revision identifier delivered to consumers and are a
...
+-   Allow organizations to provide
+   [organization-specified properties](#additional-properties) to be included in the
+   [Source VSA](#source-verification-summary-attestation) when the corresponding controls are
+   enforced.
+-   Allow organizations to distribute additional attestations related to their
+   technical controls to consumers authorized to access the corresponding Source
+   Revision.
+-   Prevent organization-specified properties from beginning with any value
+   other than `ORG_SOURCE_` unless the SCS endorses the veracity of the
+   corresponding claims.
```
TomHennen, line 451: Suggestion: these are all about the _additional_ technical controls in "Protected Named References". Can we just move this set of requirements there?

TomHennen:
This is looking really good.  Just a couple of comments. I think we're about ready to merge.

docs/spec/draft/source-requirements.md
```diff
@@ -327,97 +339,61 @@ If the SCS DOES NOT generate a VSA for a revision, the revision has Source Level
...
 <tr id="history"><td>History <a href="#history">🔗</a><td>
 
-Revisions are created by applying specific code changes (a "diff" in git) on
-top of earlier revisions of a branch. This sequence of changes, the revisions
-they produced, and how they were introduced into a branch constitute the history
-of that branch.
-
-The SCS MUST record the sequence of changes, the revisions they created,
-the actors that introduced them and the context they were introduced into.
+There are two key aspects to change history:
```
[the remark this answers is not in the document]
zachariahcox: 🔢 counting.

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,53 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
-### Level 2: Controls
+### Level 2: History & Provenance
 
 <dl class="as-table">
 <dt>Summary<dd>
 
-Clarifies which branches and tags in a repo are consumable and guarantees that
-all changes to protected branches and tags are recorded and subject to the
-organization's technical controls.
+At least one branch’s history is continuous, immutable, and retained, and the
```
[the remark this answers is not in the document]
TomHennen: Sounds great.

docs/spec/draft/source-requirements.md
```diff
@@ -79,10 +79,10 @@ level.
 
 | Track/Level | Requirements | Focus
 | ----------- | ------------ | -----
-| [Source L1](#source-l1)  | Use a version control system | First steps towards operational maturity
-| [Source L2](#source-l2)  | History and controls for protected branches & tags | Preserve history and ensure the process has been followed
-| [Source L3](#source-l3)  | Signed provenance | Tampering by the source control system
-| [Source L4](#source-l4)  | Code review | Tampering by project contributors
+| [Source L1](#source-l1)  | Use a version control system. | Generation of discrete Source Revisions for precise consumption.
+| [Source L2](#source-l2)  | Preserve Change History and generate Source Provenance. | Reliable history through enforced controls and evidence.
+| [Source L3](#source-l3)  | Enforce organizational intent with technical controls. | Consumer knowledge of consumable Source Revisions and their guaranteed technical controls.
```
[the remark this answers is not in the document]
TomHennen: "Enforce organizational technical controls" SGTM!

docs/spec/draft/source-requirements.md
TomHennen: SGTM!

sumonkhan71cf-beep:
Google account service email pricediana31@gmail.com , this Gmail also connected my Facebook business related support fage Rising Bangla Suman khan , I have problem it is set my business Gmail and recovery problem twist verification
Screenshot_2025-10-03-19-19-31-03_a23b203fd3aafc6dcb84e438dda678b6
Screenshot_2025-10-03-18-35-57-80_f7aa348215f5d566f9e4ca860f474209
Screenshot_2025-10-03-18-36-27-14_f7aa348215f5d566f9e4ca860f474209
Screenshot_2025-10-03-19-15-53-24_f7aa348215f5d566f9e4ca860f474209

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,50 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
-It is provided to authorized users of the source repository in a documented format.
+The SCS is configured to enforce the Organization's intentions for specific Named References within the Source
+Repository using any necessary technical controls.
+
 
 <dt>Intended for<dd>
 
-Organizations that want strong guarantees and auditability of their change management processes.
+Organizations with many Named References or who otherwise need to document
+which controls were enforced.
```
zachariahcox, line 145:
```suggestion
Organizations that need to show evidence of additional technical controls.
```

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,48 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
 
-The SCS generates credible, tamper-resistant, and contemporaneous evidence of how a specific revision was created.
-It is provided to authorized users of the source repository in a documented format.
+The SCS is configured to enforce the Organization's intentions for specific Named References within the Source
+Repository using any necessary technical controls.
 
 <dt>Intended for<dd>
 
-Organizations that want strong guarantees and auditability of their change management processes.
+Organizations that need to show evidence of additional technical controls.
```
zachariahcox, line 143: @TomHennen I think it's just this now 😂

zachariahcox: or maybe "evidence of their commitment to technical controls?"

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,48 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
 
 <dt>Benefits<dd>
 
-Provides information to policy enforcement tools to reduce the risk of tampering
-within the SCS's storage systems.
+A verifier can use this published data to ensure that a given Source Revision
+was created in the correct way by verifying the Source Provenance or VSA
+for all expected claims and properties.
```
zachariahcox, line 149:
```suggestion
Provides verifiers strong evidence of all technical controls enforced when updating a Named Reference. Allows verification that the Source Revision is compliant with the expected policy.
```

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,48 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
-### Level 3: Signed and Auditable Provenance
+### Level 3: Continuous technical controls
 
 <dl class="as-table">
 <dt>Summary<dd>
 
-The SCS generates credible, tamper-resistant, and contemporaneous evidence of how a specific revision was created.
-It is provided to authorized users of the source repository in a documented format.
+The SCS is configured to enforce the Organization's intentions for specific Named References within the Source
+Repository using any necessary technical controls.
```
zachariahcox, line 139:
```suggestion
The SCS is configured to enforce the Organization's intentions for specific Named References within the Source
Repository using any necessary technical controls.
The organization declares its intention to enforce those controls continuously.
```

docs/spec/draft/source-requirements.md
arewm: Something along the lines of "evidence of their commitment to technical controls" is more accurate as it isn't a requirement that there are any additional technical controls.

docs/spec/draft/source-requirements.md
```diff
@@ -105,45 +105,48 @@ Migrating to the appropriate tools is an important first step on the road to ope
...
+over time, enabling attribution of those changes to the actors that made them.
+Source Provenance provides strong, tamper-resistant evidence of the process that
+was followed to produce a Source Revision.
 
 </dl>
 </section>
 <section id="source-l3">
 
-### Level 3: Signed and Auditable Provenance
+### Level 3: Continuous technical controls
```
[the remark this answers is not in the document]
arewm: A policy repo isn't _necessarily_ a requirement of the track, but it is an implementation detail of the source tool? I guess one benefit of this type of public visibility is that it allows consumers to reason about continuity of controls as well as the use of safe expunging process.

zachariahcox:
@MarkLodato @TomHennen @arewm @adityasaky I have updated the _description_ of this PR to summarize what I think ended up happening here.

Please feel free to directly edit as you see fit 👍 Thank you!

PR #1499, opened 2025-10-21
editorial: draft: link to tag def

### LINKED ISSUE #1490: clarity: source track: history requirement: what do we mean by tags?
https://github.com/slsa-framework/slsa/pull/1483#discussion_r2426270833

We should clarify here that we mean our narrow definition of tags, namely: a named reference that the SCS uses as if it were immutable. 

We can address this with a hyperlink back to our local definition. 

> If the SCS supports Tags, the SCS MUST be configured to prevent them from being moved or deleted.

TomHennen:
Adding a link to the tag definition to make it clear what we mean by tag in this context.  E.g. we mean the SLSA definition.

fixes #1490

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -28,7 +28,7 @@ The following terms apply to Version Control Systems:
 | Change | A modification to the state of the Source Repository, such as creation of a new Source Revision based on a previous Source Revision, or creation, deletion, or modification of a Named Reference.
 | Change History | A record of the history of Source Revisions that preceded a specific revision.
 | Branch | A Named Reference that moves to track the Change History of a cohesive line of development within a Source Repository. E.g. `main`, `develop`, `feature-x`
-| Tag | A Named Reference that is intended to be immutable. Once created, it is not moved to point to a different revision. E.g. `v1.2.3`, `release-20250722`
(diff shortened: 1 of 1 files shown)


docs/spec/draft/source-requirements.md
```diff
@@ -28,7 +28,7 @@ The following terms apply to Version Control Systems:
 | Change | A modification to the state of the Source Repository, such as creation of a new Source Revision based on a previous Source Revision, or creation, deletion, or modification of a Named Reference.
 | Change History | A record of the history of Source Revisions that preceded a specific revision.
 | Branch | A Named Reference that moves to track the Change History of a cohesive line of development within a Source Repository. E.g. `main`, `develop`, `feature-x`
-| Tag | A Named Reference that is intended to be immutable. Once created, it is not moved to point to a different revision. E.g. `v1.2.3`, `release-20250722`
+| <span id="tag">Tag</span> | A Named Reference that is intended to be immutable. Once created, it is not moved to point to a different revision. E.g. `v1.2.3`, `release-20250722`
```
arewm: Should we consistently create spans around all of these?

TomHennen: My feeling was that unless we're going to use it (or add visible anchors to use them) it's not that helpful.

PR #1504, opened 2025-10-27
content: draft: Remove unnecessary terms

### LINKED ISSUE #1481: Consider reducing the number of defined terms
In the review of #1471 it was discussed that perhaps we have more terms defined than we really need.

Let's review and see if we should remove any.

TomHennen:
These terms don't add much value to the spec. Removing them might make things easier to read & understand.

fixes #1481


arewm (approved):
Sounds reasonable to me. If there is confusion around these later, we can always add some form of them back in.

zachariahcox (approved):
these are good terms, but maybe not needed yet.

PR #1505, opened 2025-10-27
content: draft: Remove provenance requirement

TomHennen:
Creating this PR to get an idea of what it would mean to remove the provenance (and VSA) requirement from the spec.

It seems like the majority of what we're looking for in provenance can be subsumed by other existing requriements we already track with provenance becoming a (convenient) implementation detail.

If we do this we might also want to move the provenance and VSA parts of this page to their own page to keep the level requirements clear.

If we accept this change it would mean that 'provenance' becomes a nice feature of any particular SLSA Source Track implementation but won't be required.

### DIFF
--- a/docs/spec/draft/source-requirements.md
+++ b/docs/spec/draft/source-requirements.md
@@ -379,7 +379,7 @@ technical control.
 
 The SCS MUST record which technical controls were in place during the creation
 of each revision and it MUST ensure each control was in place from a specific
-start revision. If there is a lapse in continuity for a specific control,\
+start revision. If there is a lapse in continuity for a specific control,
 continuity of that control MUST be re-established from a new revision.
 
 Exceptions to the continuity requirement are allowed via the [safe expunging process](#safe-expunging-process).


docs/spec/draft/source-requirements.md
```diff
@@ -270,16 +265,12 @@ if possible.
 <td><td>✓<td>✓<td>✓
 <tr id="technical-controls"><td>Continuous technical controls <a href="#technical-controls">🔗</a><td>
 
-The organization MUST provide evidence of continuous enforcement via technical
-controls for any claims made in the Source Provenance attestations or VSAs (see
-[control continuity](#continuity)).
```
arewm: Are we being too broad with this deletions? I feel like it would still be beneficial for organizations to provide evidence even if we don't specify provenance or VSAs.

docs/spec/draft/source-requirements.md
```diff
@@ -318,22 +309,18 @@ non-plain-text changes (e.g. render images, verify and display provenance for
...
-
-At Source Levels 1 and 2 the SCS MAY issue these attestations based on its
-understanding of the underlying system (e.g. based on design docs, security
-reviews, etc.), but at Level 2+ the SCS MUST use the SCS-issued
-[source provenance](#source-provenance) when issuing the VSAs.
+> For example, an SCS may generate a
+[source verification summary attestation](#source-verification-summary-attestation)
+(Source VSA) to indicate the SLSA Source Level of any revision at Level 1 or
+above.  If the SCS DOES NOT generate a VSA for a revision, the revision is
+assumed to have Source Level 0.
```
arewm: Is it clear enough that this only holds in the example? The lack of a VSA doesn't mean Source Level 0 _unless_ the the VSA is the mode of communication.

docs/spec/draft/source-requirements.md
```diff
@@ -348,6 +335,29 @@ There are three key aspects to change history:
 To answer these questions, the SCS MUST record all changes to Named References,
 including when they occurred, who made them, and the new Source Revision ID.
 
+The SCS MUST ensure this information is protected from tampering and
+falsification from both users and platform administrators.
+
+> For example:
+>
+> 1.  The SCS may record detailed metadata for revisions in signed
+Source Provenance, protecting it from tampering from other components of the
```
arewm: Should these occurrences still be capitalized if it isn't something that we are introducing in "the spec?"

docs/spec/draft/source-requirements.md
```diff
@@ -402,42 +412,6 @@ The SCS MUST document how actors are identified for the purposes of attribution.
...
-<tr id="source-provenance"><td>Source Provenance <a href="#source-provenance">🔗</a><td>
-
-[Source Provenance](#source-provenance-attestations) are attestations that
-contain information about how a specific revision was created and how it came to
-exist on a protected branch or how a tag came to point at it. They are
-associated with the revision identifier delivered to consumers and are a
-statement of fact from the perspective of the SCS. The SCS MUST document the
-format and intent of all Source Provenance attestations it produces.
-
-Source Provenance MUST be created contemporaneously with the branch being
```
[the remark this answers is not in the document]
arewm: @TomHennen , were you suggesting to move the source provenance content elsewhere? It could be part of the repository even if it isn't a core part of the spec itself.

arewm (commented):
This change generally seems to be aligned with Mark's viewpoint expressed in #1010. If we merge this, could we also try to clarify the guidance on new tracks? As it relates to the source track, who are the entities that we need to trust (primarily the SCS) and how would consumers verify/validate their trust in those entities?

adityasaky:
I'd like to understand more of the motivation for this change.

> It seems like the majority of what we're looking for in provenance can be subsumed by other existing requriements we already track with provenance becoming a (convenient) implementation detail.

I think this is a signal we need to do a better job in the source track of explaining what source attestations add over the other requirements. :smile: 

The requirements themselves tell me, a consumer, what controls to enable to get certain security properties. This, in and of itself, does not differentiate SLSA from numerous best practices guides which tell me to enable branch protection rules, disable force pushes to `main`, require code review, etc (e.g., the OpenSSF itself has https://best.openssf.org/SCM-BestPractices/). The attestations enable me to **keep the SCS implementing the requirements honest**. With SLSA, I can verify that I'm safe for every revision that matters rather than rely on enabling the right controls. So, while I understand the emphasis on the source track's requirements, I worry that we'll be losing a core value proposition of SLSA as a whole.

Looking at the alternatives proposed to the (optional) source provenance in this PR, I'd like to understand how they improve some aspect (e.g., adoption) of SLSA over requiring provenance. For instance, logging metadata to a transparency log sounds like the SCS is essentially implementing support for attestations which it then records in a transparency log. Adding an API that returns information again sounds like something the SCS could wrap in an attestation at the time a revision is created. OTOH, without attestations, we lose verifiability, non-repudiation, and the powerful interoperability across SLSA tracks we get using attestations as a common, consistent interface for each track.

docs/spec/draft/source-requirements.md
```diff
@@ -367,10 +377,10 @@ history of a Branch.
 'Control continuity' reflects an organization's ongoing commitment to a
 technical control.
 
-For each technical control claimed in a VSA, continuity MUST be established and
-tracked from a specific start revision.
-If there is a lapse in continuity for a specific control, continuity of that
-control MUST be re-established from a new revision.
+The SCS MUST record which technical controls were in place during the creation
+of each revision and it MUST ensure each control was in place from a specific
+start revision. If there is a lapse in continuity for a specific control,\
```
patzielinski:
```suggestion
start revision. If there is a lapse in continuity for a specific control,
```

docs/spec/draft/source-requirements.md
```diff
@@ -270,16 +265,12 @@ if possible.
...
-
 The organization MUST document the meaning of their enforced technical controls.
 
 > For example, if an organization implements a technical control via a custom
-tool (such as required GitHub Actions workflow), it must indicate the name of
-this tool, what it accomplishes, and how to find its evidence in the provenance
-attestation.
+tool (such as required GitHub Actions workflow) and tracks its status in
+provenance attestations, it must indicate the name of this tool, what it
+accomplishes, and how to find its evidence in the provenance attestation.
```
patzielinski: Is this bit still relevant with these edits?

PR #1506, opened 2025-10-30
editorial: draft: Note the purpose of the provenance build model diagram

### LINKED ISSUE #1290: Contradictory SLSA Model diagrams on slsa.dev
Brought this question to the SLSA meeting on 3/10/2025, and the community agreed that these diagrams need clarification. 

On the Provenance page of SLSA.dev [SLSA • Provenance](https://slsa.dev/spec/v1.0/provenance), the diagram that shows the "Model" shows that external & internal parameters go into the template that go into the build process. 

But in the Terminology page of SLSA.dev [SLSA • Terminology](https://slsa.dev/spec/v1.0/terminology), the External Parameters is going into Interface which then goes straight to the Control Plane? From a threat model perspective, I had thought the Control Plane is a trusted area (and External Parameters are untrusted) and that the Control Plane is supposed to be outside of user control.

Having both of these diagrams creates confusion. Is there a way that we can align on one, or somehow make the intent of each diagram more clear?

TomHennen:
We have two diagrams that are very similar and both depict the build model.

This one, which is specific to the provenance format, and the one on the https://slsa.dev/spec/draft/terminology#build-model page which is less specific.  We probably want both versions.  So, instead of updating the diagrams I've added an explanatory note with a link to prevent confusion.

fixes #1290

### DIFF
--- a/docs/spec/draft/build-provenance.md
+++ b/docs/spec/draft/build-provenance.md
@@ -44,6 +44,10 @@ software artifacts through execution of the `buildDefinition`.
 
 ![Build Model](images/provenance-model.svg)
 
+> NOTE: This diagram depicts how the SLSA Provenance format is structured.  It
+> elaborates on the [SLSA Build Model](terminology#build-model) but only applies
+> to the SLSA Provenance format specifically and not to builds in general.
+
 The model is as follows:
 
 -   Each build runs as an independent process on a multi-tenant build platform.


lehors (approved):
I note that we are inconsistent in how we mark such notes, some read 'NOTE:' while others read 'Note:'... But this can be addressed in a different PR eventually.

PR #1510, opened 2025-11-04
content: Create SLSA 1.2 RC2

TomHennen:
I think we've addressed everything necessary to cut RC2.

Replaced 1.2 RC1 in the nav because we shouldn't need both RC1 and RC2. 

For me the easiest way to review this change is by using a tool like 'meld'.

E.g. `meld docs/spec/draft docs/spec/v1.2-rc2` to compare it against whats in draft or `meld docs/spec/v1.2-rc1 docs/spec/v1.2-rc2` to compare it with the prior RC.

### DIFF
--- a/docs/spec/v1.2-rc2/whats-new.md
+++ b/docs/spec/v1.2-rc2/whats-new.md
@@ -4,7 +4,7 @@ description: The changes brought by SLSA 1.2 RC2.
 ---
 
 This document describes the major changes brought by SLSA 1.2 RC2 relative to
-the prior release, [v1.1] and the prior release candidate [v1.2 RC1].
+the prior release [v1.1] and the prior release candidate [v1.2 RC1].
 
 ## Summary of changes since v1.2 RC1
 


TomHennen:
FYI we need 3 approvals to get this merged.

puerco (approved):
Non-binding lgtm from me. And a small suggestion below :point_down:

lehors:
I also think the whatsnew section ought to explain what has changed between RC1 and RC2 since that's what most people will want to see when this gets published.

lehors:
> > > Replaced 1.2 RC1 in the nav because we shouldn't need both RC1 and RC2.
> > 
> > 
> > I wonder about that. Someone might want to have them side by side, no?
> 
> Added something that might do the trick. This website maintenance is a real pain in the neck.

I see it. I think that works.

> > I also think the whatsnew section ought to explain what has changed between RC1 and RC2 since that's what most people will want to see when this gets published.
> 
> Done

Thanks.

lehors (approved):
One small nit but LGTM otherwise. Thanks for all the effort!

PR #1602, opened 2026-04-12
impl: Generate onepage automatically

lehors:
Maintaining the single page view of the spec has been an ongoing challenge because it relied on maintaining the list of files to be included. Invariably when a file is added we forget to update `onepage.md`. This PR makes `onepage.md` entirely automatic and version independent. Its content is drawn from the navigation menus (`/www/_data/na/*yaml`), reducing the number of files we need to maintain.

Incidentally, I found out that the current draft specification is missing dependency-track which is now included automatically. I rest my case. ;-)

This also fixes how the links within the spec are changed to relative links. In particular, as the spec grows we have more and more anchors with the same name in different files which creates collisions. To prevent this relative links are now prefixed with the filename to ensure uniqueness.

### DIFF
--- a/www/_includes/section.liquid
+++ b/www/_includes/section.liquid
@@ -29,7 +29,6 @@ sure it is.
    {% capture pattern %}<section id="{% endcapture %}
    {% capture anchor %}<section id="{{include.file}}-{% endcapture %}
    {% capture fixedcontent %}{{fixedcontent | replace: pattern, anchor}}{% endcapture %}
-   {% capture fixedcontent %}{{fixedcontent | replace: "</h1>", "</h2>"}}{% endcapture %}
 {% else %}
    {% assign fixedcontent = include.content %}
 {% endunless %}
--- a/www/_includes/section.liquid
+++ b/www/_includes/section.liquid
@@ -26,6 +26,10 @@ sure it is.
    {% capture anchor %}<h2 id="{{include.file}}-{% endcapture %}
    {% capture fixedcontent %}{{fixedcontent | replace: pattern, anchor}}{% endcapture %}
    {% capture fixedcontent %}{{fixedcontent | replace: "</h1>", "</h2>"}}{% endcapture %}
+   {% capture pattern %}<section id="{% endcapture %}
+   {% capture anchor %}<section id="{{include.file}}-{% endcapture %}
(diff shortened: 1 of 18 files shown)


www/_includes/section.liquid
```diff
@@ -1,50 +1,55 @@
...
    {% capture fixedcontent %}{{fixedcontent | replace: "</h2>", "</h3>"}}{% endcapture %}
-   {% capture fixedcontent %}{{fixedcontent | replace: "<h1", "<h2"}}{% endcapture %}
+   {% capture pattern %}<h1 id="{% endcapture %}
+   {% capture anchor %}<h2 id="{{include.file}}-{% endcapture %}
+   {% capture fixedcontent %}{{fixedcontent | replace: pattern, anchor}}{% endcapture %}
+   {% capture fixedcontent %}{{fixedcontent | replace: "</h1>", "</h2>"}}{% endcapture %}
+   {% capture pattern %}<section id="{% endcapture %}
+   {% capture anchor %}<section id="{{include.file}}-{% endcapture %}
+   {% capture fixedcontent %}{{fixedcontent | replace: pattern, anchor}}{% endcapture %}
    {% capture fixedcontent %}{{fixedcontent | replace: "</h1>", "</h2>"}}{% endcapture %}
```
mlieberman85: This looks like you have `</h1> </h2>` twice here. Is that right?

mlieberman85 (approved):
LGTM, no one currently links to the one pager right? If they do wouldn't this break their links?

www/_includes/section.liquid
lehors: Good catch! This is harmless because the second instance won't match anything but it still shouldn't be there. Thanks for the careful review!

lehors:
> LGTM, no one currently links to the one pager right? If they do wouldn't this break their links?

Yes, this will undoubtedly break some links unfortunately but I'm not sure how many people point to those pages. I added a bunch of redirects so that they at least get redirected to the new pages but when it comes to the fragment identifiers there is just no good solution.

Either we leave things as they are and we have a whole bunch of broken links within the pages or we change the code and break external links that point to one of the anchors that are renamed. Given that we don't know if any such links even exist, I think we're better off fixing the links we know are broken than preserving links that may not event exist.

PR #1607, opened 2026-04-20
fix: v1.2-rc2 navigation menu

lehors:
Silly me!...


marcelamelara (approved):
Thanks for following up!

PR #1611, opened 2026-05-15
blog: Add post clarifying what SLSA can and cannot do after Mini Shai-Hulud

arewm:
Reports following the May 2026 Mini Shai-Hulud npm supply chain attack incorrectly described the compromised TanStack packages as carrying valid SLSA Build Level 3 attestations. This post corrects that characterization and maps where SLSA's guarantees actually begin and end.

The post covers why npm's built-in provenance achieves L2 not L3, what each level concretely provides, which gaps are addressable with policy on top of SLSA (Source Track attestations, expected-builder enforcement, VSAs), and what requires different controls entirely. The observability boundary limitation is called out as a documented class of constraint with references to prior community work including "Dirty Dancing" (KjellCon 2025) and "From Mild to Wild" (KubeCon EU 2026).

Primary sources: TanStack's own postmortem and GHSA-g7cv-rxg3-hmpx.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1615, opened 2026-06-02
fix: Use git archive to avoid clobbering working tree in combine-versions.sh

arewm:
## Summary

The `combine-versions.sh` script used `git checkout` to switch to each release branch in order to copy spec files, then switched back to the original branch. This modified the entire working tree during the loop — including `www/` content like blog posts and layout files. The restore step (`git checkout $CURRENT_BRANCH`) ran without `set -e`, so a silent failure in Netlify's shallow-clone environment could leave the working tree on the wrong branch. Netlify would then detect "no change" vs. the previous deploy and skip publishing.

This is the likely explanation for the production deploy failing with "no change detected" after #1611 was merged, though we don't have access to the Netlify build logs to confirm definitively.

Replaces the checkout+copy pattern with `git archive | tar`, which reads spec files directly from each branch's tree object without touching the working tree.

## Test plan

- [x] Ran `combine-versions.sh --nofetch` locally using upstream release refs
- [x] `git status` shows zero modifications to tracked files in `www/` after the script runs
- [x] `www/spec/` populated with 12 version subdirectories (`draft/`, `v0.1/` through `v1.2-rc2/`)
- [ ] End-to-end Netlify deploy (will be confirmed by deploy preview on this PR)

🤖 Generated with the help of [Claude Code](https://claude.ai/code)

### DIFF
--- a/tools/combine-versions.sh
+++ b/tools/combine-versions.sh
@@ -41,12 +41,6 @@ for arg in "$@"; do
   esac
 done
 
-CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
-if [ "$CURRENT_BRANCH" = "HEAD" ]; then
-    # If detached, store commit SHA
-    CURRENT_BRANCH=$(git rev-parse HEAD)
-fi
-
 REPO_ROOT=$(git rev-parse --show-toplevel)
 
 # clean spec dir
@@ -71,17 +65,11 @@ fi
 # Then, populate every remote release branch
 for FULLNAME_BRANCH in $(git branch -r --list 'origin/releases/*'); do
     VERSION="$(basename $FULLNAME_BRANCH)"
-    RELEASE_BRANCH=releases/$VERSION
     echo Populating spec/$VERSION
-    git -C "$REPO_ROOT" reset --hard
-    git -C "$REPO_ROOT" checkout $RELEASE_BRANCH
     mkdir spec/$VERSION
-    cp -r "$REPO_ROOT/spec"/* spec/$VERSION/
+    git -C "$REPO_ROOT" archive "$FULLNAME_BRANCH" -- spec/ | tar -x --strip-components=1 -C spec/$VERSION
 done
 
-# back to the original branch
-git reset --hard
-git checkout $CURRENT_BRANCH
 echo '
 The spec folder is now ready:'
 ls -F spec


lehors:
I was just looking at it @arewm and although at first glance I can't say whether the proposed change would be an improvement over the current method the premise is bogus: the production deploy did NOT fail, the publication of the SPEC was merely stopped because there was no change to the spec (which is what we want). At the same time your blog post was posted all right, so...

arewm:
My blog post was published because we forced a rebuild in Netlify. https://app.netlify.com/projects/slsa/deploys/6a1ec938ca68b600089f80b8 was the original push build.
https://app.netlify.com/projects/slsa/deploys/6a1edc5b0f25bb00d2b4c0fe was the rerun.

I just have public access to Netlify though, so you might see more data than me.

lehors:
Do you know that your change will fix the publication of the blog posts when the spec hasn't changed?

lehors:
I can't say that I understand what happened with the failed deploy of #1611. The [Deploy log](https://app.netlify.com/projects/slsa/deploys/6a1ec938ca68b600089f80b8) tells us that the `ignore` command specified in `netlify.toml` returned 0 instead of 1 as expected:

```
2:14:52 PM: Custom ignore command detected. Proceeding with the specified command: 'bash -c 'git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF . ../spec/ ../tools/ && exit 0 || exit 1''
2:14:52 PM: User-specified ignore command returned exit code 0. Returning early from build.

```
 Unfortunately it doesn't give any more insights and in particular doesn't show what `$CACHED_COMMIT_REF` and `$COMMIT_REF` were set to when the command was run.

But in such a case the build command is NOT run so I don't believe that `combine-versions.sh` was run and is the cause of the problem. Therefore I don't believe changing this script as this PR proposes will prevent any such failure in the future.

This being said, I'm happy to merge this PR anyway because, independently of the failure discussed above, I always found the branch switching to be quite disruptive when I run the script locally and I see value in avoiding doing so. Just beware that this probably have no bearing on future deploy failure. I wish Netlify offered a way to have more insights on the running of the ignore command.

lehors (approved):
With the caveat discussed above, I think this is an improvement.

arewm:
Thanks for that -- and for further investigation into the logs. We need someone to create another blog before a spec bump to see if it happens again. 😄

PR #1616, opened 2026-06-03
fix: Simplify netlify ignore command

lehors:
Investigating the deployment failure of #1611 has put the focus on the ignore command specified in `netlify.toml`.

The current command is unnecessarily complicated, calling bash and explicitly setting the exit code when this will all happen when calling the git command directly.

So this change trims down the ignore command to what's really necessary, in line with [the Netlify documentation](https://docs.netlify.com/build/configure-builds/ignore-builds/).


arewm (approved):
This seems reasonable and we can always revert it if it causes problems. Thanks for continuing to iterate on improving this!

PR #1627, opened 2026-06-29
content: Add dependency track

mlieberman85:
I took some of S2C2F requirements that made sense and were attestable along with some other thoughts.

Normally I would keep the gem update separate but we were on an EOL version.

I used Claude to help structure, do some sanity checking, and to bounce ideas off of.


spec/dependency-provenance.md
```diff
@@ -0,0 +1,497 @@
...
+
+A verifier MAY trust the platform's `integrity.result`,
+`policyEvaluations[].result`, and `upstreamProvenance.verification.result`
+fields based on a trust relationship with the platform (per [Trust
+platforms, verify artifacts](principles.md#trust-platforms-verify-artifacts)),
+or MAY re-perform any of those checks independently.
+
+## Examples
+
+### Minimal L1
```
renewitt, line 384: Could you include an example of an L1 attestation for Git / VCS ingestion? 

That's probably the next most common dependency import path after package managers. Maybe I'm overthinking it IRT to the spec itself, but this proposed schema seems like it could be a bit awkward to use with version control systems because there would either be a lot of duplication of values across fields, or some immediate corner cases.

For example, the subject digest:
Would this just be the commit sha? Would this commit SHA then expect to be repeated in other places like version, or would version in that case be an appropriate tag, even though this is not immutable and could be multiple? If you downloaded a tarball of a git repo the SHA would be the digest of the tarball, but this wouldn't apply to any other git use case. I suppose this could be a dependency platform design choice but it could make the spec confusing for consumers.

spec/dependency-provenance.md
```diff
@@ -0,0 +1,497 @@
...
+> [parsing rules](#parsing-rules) for more information.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in
+[RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
+
+## Purpose
+
+Describe how a single third-party dependency entered an ingestor's
```
renewitt, line 37: Is a "single third-party dependency" considered to be result of fetching a dependency from a prescribed source? I think the "single dependency" language is a bit unclear when it comes to anything outside package managers (though my concern may also apply there). Git is somewhat of the wild west, but also a very typical dependency import path for large projects.

Some repos are notorious for packaging pinned third party sub-dependencies in the parent repo via either vendoring or git submodule eg. https://github.com/google/ruy.  These are often invisible to scanners and SBOM generation pipelines since they were implicitly imported. Anything that could make these more visible would be fantastic.

For clarity, is the single third party dependency that's expected to have the attestation just the top level one?

spec/dependency-provenance.md
```diff
@@ -0,0 +1,497 @@
...
+  "subject": [{
+    "name": "lodash@4.17.21",
+    "digest": { "sha256": "abc123..." }
+  }],
+  "predicateType": "https://slsa.dev/dependency/v1",
+  "predicate": {
+    "ingestor": { "id": "https://example.com/" },
+    "ingestionPlatform": { "id": "https://artifactory.example.com/" },
+    "upstream": {
+      "registry": "https://registry.npmjs.org/",
```
renewitt, line 402: In the git case, a repo path is not necessarily sufficent to uniquely identify a repository. Might be worth adding an optional repo_id or other unique identifier for those who care about this. Low volume org/repo names can be recycled on places like GitHub, which could lead to future confusion about the source of code.

https://github.blog/security/supply-chain-security/how-to-stay-safe-from-repo-jacking/#whats-the-risk-of-repo-jacking

spec/dependency-provenance.md
mlieberman85: This is a good point. I will bring it up in the next SLSA meeting.

We've mostly been considering things downloaded via package managers or just binaries downloaded through something like http.

There's always going to be a challenge with something like vendoring.

spec/dependency-provenance.md
mlieberman85: This is a good point worth discussion in an upcoming SLSA meeting. If you're available I think your thoughts would be valuable.

They happen at 12PM Eastern time on Mondays. See: https://openssf.org/calendar/ for details on the Zoom link and stuff.

The thing we generally recommend with the build track is just using the git commit hash in cases like this. As long as it's clear what the hash refers to via the subject via like a url it should be good. But maybe we should be a bit more specific.

spec/dependency-provenance.md
mlieberman85: Yes, this another good point.

spec/dependency-provenance.md
lumjjb: We had a good discussion around this during our call today - there is a clear need to address when dependencies are included in source (whether it be through vendor-ed code or git submodules). We need to brainstorm on how we can be able to enable these dependency track evaluations on source based dependencies as well (distinguishing between the decision to use a new library or a new version, to evaluating the provenance and properties of the actual bytes ingested). 

The source track needs to either surface that information or include evaluation of those dependencies within its evaluation of its level. I.e. Perhaps a source attestation should have a source Level for provenance, but also expose a dependency level certification for all its dependencies.

renewitt: Great to hear! 

Unfortunately I'm in APAC (Australia), so the US east timezone is very challenging for me to work with, but please let me know if there's a better format to discuss or give feedback or thoughts in other than a PR. I'd love to be involved. 

The other potential ingest types I can see needing to be supported to make this robust for large project use cases would be archive packages from misc download locations i.e not git or package manager releases and potentially binaries for a similar type of case.

lumjjb: I've thought a bit more about this and for this comment, i'm naming a concept "Dependency Fog" - the impeding factors that make it so that dependencies are no clearly articulated and indexable so that they can be evaluated..

Dependency fog not only shows up by the issues with source (binary inclusions, vendored source, etc.) but also within the build space (i.e. having a non-hermetic build can pull in dependencies that we cant validate). 

One approach here would be to provide some indication of evaluation of dependency fog.. This is a hard problem, but I think there can be multiple levels here (from scans providing some heuristics of known fog, vs certainty from using a declarative build system - e.g. bazel). 



Concrete suggestions - please let me know the best way to contribute here.
- **We can effectively reduce the introduced source dependencies into the binary/artifact dependency problem**: The problem today in build attestation as source is treated opaque. Perhaps have a dependency attestation for a source repo that will provide any indications of dependency fog and provide a way to surface those dependencies if they exists (list of binaries, submodules, etc.) that need to be considered in evaluation. - every source input to a build will need this attestation to be evaluated. 
- From a policy perspective, I think we need to provide a way to reason about source and builds (materials complete) as well so that we can certify those sources for some properties of dependency fog. 
- We probably can add in a list of cases that we've considered.

| Case | Ingestion event | Dependency indexable?  | Digest |
|---|---|---|---|
| Vendored source tree (`go mod vendor`, `cargo vendor`, manual copy) | the vendoring commit | yes, at vendoring time | `dirHash`/`gitTree` over the subtree |
| Submodule pin/bump | the gitlink update | yes — URL + commit are in the tree | `gitCommit` |
| Checked-in 3p binary (vendor `.so`, prebuilt toolchain) | the commit adding it | usually, from the vendor's release page | blob digest |
| Checked-in minified JS bundle | the commit adding it | often, from the CDN/registry version | blob digest |

spec/threats.md
```diff
@@ -896,10 +905,21 @@ or source threat to *B* is also a dependency threat to *A*. Furthermore, if
...
-threats.
+SLSA addresses dependency threats through two complementary mechanisms.
+[Applying SLSA recursively] to every dependency would address most of
+these threats if each upstream producer published SLSA Build and Source
+attestations. Most upstream OSS does not yet do so, so the
+[Dependency Track](dependency-track.md), in Working Draft, adds controls
+for cases where upstream guarantees are absent. The track defines a
+per-ingestion attestation, Dependency Ingestion Provenance, and grades
+the platform's claims about each ingested dependency across three levels:
+L1 Inventoried (identity and content), L2 Controlled (controlled ingest
```
lumjjb: I agree on identity and content here, i think inventoried may be confused with knowing all dependencies. I suggest that we rename this - maybe Integrity or Provenance?

tombedfordgit: + 1 for something related to provenance

spec/dependency-track.md
```diff
@@ -1,160 +1,1139 @@
...
 # Basics
 
 ## Objective
 
-Enable a software producer to measure, control, and reduce risk introduced from third party dependencies.
+Enable an organization that ingests third-party software dependencies to
+measure, control, and reduce the supply chain risk those dependencies
+introduce. Define a per-ingestion attestation, Dependency Ingestion
+Provenance, that downstream parties can verify.
+
```
lumjjb, line 16: what is a dependency
```suggestion
## What is a dependency

Specifically, we scope dependencies to blobs that can be ingested into a builder. A dependency as a subject must be able to have a digest and optionally a URI. A NPM package or container image with a hash is considered a dependency. A PURL with no verifiable hash is NOT a dependency.
```

www/current-activities.md
```diff
@@ -29,11 +29,42 @@ is in draft.
...
+[slsa-framework/slsa-proposals](https://github.com/slsa-framework/slsa-proposals).
 
-Join the discussions and ongoing efforts [on the SLSA Dependency Track Slack channel](https://openssf.slack.com/archives/C03THTH3RSM).
+The current Working Draft is published at
+[/spec/draft/dependency-track](/spec/draft/dependency-track); the predicate
+schema is at
+[/spec/draft/dependency-provenance](/spec/draft/dependency-provenance).
+
+Join the discussions and ongoing efforts
+[on the SLSA Dependency Track Slack channel](https://openssf.slack.com/archives/C03THTH3RSM).
```
lumjjb, line 70: this channel is archived, should be reopened or create a new channel

spec/dependency-track.md
```diff
@@ -1,160 +1,1139 @@
 ---
 title: Dependency Track
-description: This page describes the SLSA Dependency track, which enables a software producer to measure, control, and reduce risk introduced from third party dependencies.
+description: SLSA Dependency Track requirements for ingesting third-party build dependencies. Imports the attestable subset of OpenSSF S2C2F. Defines the Dependency Ingestion Provenance predicate.
```
lumjjb: I think the definition of the Dependency track shouldn't be based on another spec
```suggestion
description: SLSA Dependency Track requirements for ingesting third-party build dependencies.
```

spec/dependency-track.md
```diff
@@ -1,160 +1,1139 @@
...
+consumers of upstream OSS, and pure consumers (e.g., enterprises that
+consume OSS internally but do not ship downstream software). Medium and
+large organizations benefit most.
 
 ## Scope
 
-The scope of the SLSA Dependency Track is to properly manage and securely consume build dependencies into the developer workflow to mitigate against supply chain threats targeting the open source ecosystem.
+The scope is third-party build dependencies: software artifacts fetched or
+otherwise made available to the build environment during the build of a
+released artifact. Runtime dependencies, base images, and developer
```
lumjjb: Hmm, why are we specifying these out of scope? How are they semantically different from other dependencies?

lumjjb:
```suggestion
released artifact (dependencies provided through hermetically, through source, or downloaded during the build). Runtime dependencies, base images, and developer
```

spec/dependency-track.md
```diff
@@ -1,160 +1,1139 @@
...
+software dependencies. This includes software producers in their role as
+consumers of upstream OSS, and pure consumers (e.g., enterprises that
+consume OSS internally but do not ship downstream software). Medium and
+large organizations benefit most.
 
 ## Scope
 
-The scope of the SLSA Dependency Track is to properly manage and securely consume build dependencies into the developer workflow to mitigate against supply chain threats targeting the open source ecosystem.
+The scope is third-party build dependencies: software artifacts fetched or
+otherwise made available to the build environment during the build of a
```
lumjjb, line 182: I had an interesting conversation with @AdamZWu , as part of Build levels, dependency completeness is interpreted to include dependencies within source as well... I'm not sure how feasible it is for a build system to do though - i.e. binaries hidden in tarballs vs binaries being built.

spec/dependency-track.md
```diff
@@ -1,160 +1,1139 @@
...
+released artifact. Runtime dependencies, base images, and developer
+tooling are out of scope for this revision and listed under
+[Future Considerations](#future-considerations).
+
+The track imports the attestable subset of the OpenSSF
```
lumjjb, line 187: what do you mean by importing the attestable subset?

spec/threats.md
```diff
@@ -861,10 +866,14 @@ and [Defender's Perspective: Dependency Confusion and Typosquatting Attacks](/bl
...
-investigation or ad-hoc analysis, and can complement source-based typosquatting
-solutions.
+*Mitigation:* The SLSA Dependency Track does not prescribe specific
+typosquat defenses. A verifier requiring Provenance whose
+[recorded admission policies](dependency-track.md#dep-policy-attestation)
+include evaluation against a malicious-packages feed or organizational
+allow-list can refuse Provenance lacking those entries. The integrity
+guarantee from the Dep Track is that whatever defenses the platform
+applied are recorded in the Provenance; the choice of defenses is
+operator policy.
```
lumjjb: suggest reverting this, since it seems like its not being addressed still, we should keep it as not being addressed

```suggestion
*Mitigation:* This threat is not currently addressed by SLSA. That said, the
requirement to make the source available can be a mild deterrent, can aid
investigation or ad-hoc analysis, and can complement source-based typosquatting
solutions.
```

spec/threats.md
```diff
@@ -896,10 +905,21 @@ or source threat to *B* is also a dependency threat to *A*. Furthermore, if
 library *B* uses build tool *C*, then a source or build threat to *C* is also a
 dependency threat to both *A* and *B*.
 
-This version of SLSA does not explicitly address dependency threats, but we
-expect that a future version will. In the meantime, you can [apply SLSA
-recursively] to your dependencies in order to reduce the risk of dependency
-threats.
+SLSA addresses dependency threats through two complementary mechanisms.
+[Applying SLSA recursively] to every dependency would address most of
```
lumjjb, line 905: Broken link?

spec/threats.md
lumjjb:
```suggestion
L1 Inventoried (Integrity & Provenance), L2 Controlled (controlled ingest
```

lumjjb (approved):
Made a couple suggestions, but otherwise i think this is a good start. I do think that we need a more succinct landing page for those that are interested in participating in the dependency track (focused on target audiences , i.e. a 1 min read for folks to understand if this is something they should dig deeper and read the entire draft and term definitions)

PR #1644, opened 2026-08-05
nonspec: remove Verizon's logo

joshuagl:
There are no Verizon employees actively involved with the project.


arewm (approved):

lehors (approved):
I'm fine with this but it raises the question of what the criteria for a company to be listed is. There are other companies currently listed that haven't been active in many years if ever... I think that's something the Steering Committee should look into. /cc @slsa-framework/slsa-steering-committee

joshuagl:
> I'm fine with this but it raises the question of what the criteria for a company to be listed is. There are other companies currently listed that haven't been active in many years if ever... I think that's something the Steering Committee should look into. /cc @slsa-framework/slsa-steering-committee

This is a great question. I have tried to tidy up after myself here (and when I changed roles from VMware --> Verizon), but it would be useful to define these criteria more generally.

We have [contributor defined](https://github.com/slsa-framework/governance/blob/main/5._Governance.md#1-roles) in the governance docs, so the main questions here are 1) what are the criteria for a contributor's company to be listed on the landing page; and 2) what are the rules for when that logo should be removed.

PR #1648, opened 2026-08-30
content: Rename upstreamProvenance to upstreamAttestations

mlieberman85:
## Summary

Follow-on to #1627. Renames and pluralizes the upstream-attestation field
in the Dependency Ingestion Provenance predicate so it can carry verdicts
on any upstream-published attestation, not just SLSA Build Provenance.

- `upstreamProvenance` (object) → `upstreamAttestations` (array of entries)
- Each entry identifies a `predicateType` and carries its own verification
  verdict plus availability marker.
- At L3, at least one entry MUST be present describing the platform's
  upstream-attestation stance for the dep.
- Kept the `#dep-verify-provenance` HTML anchor stable so existing external
  links are not broken.

## Why

The previous singular `upstreamProvenance` field was misleadingly named — the
spec text said it could carry a VSA or any other upstream attestation, but
the name implied Build Provenance only. The singular shape also meant the
platform could only reference one upstream attestation per dep, so an
upstream that publishes both a Build Provenance and a VSA could not be
represented cleanly.

## Files changed

- `spec/dependency-provenance.md` — field, section, matrix, examples, and
  verification-step updates.
- `spec/dependency-track.md` — mode-1 reference, L3 summary and requirement,
  benefits, and S2C2F mapping row.
- `spec/threats.md` — threat-block cross-reference wording.

## Test plan

- [x] `npm run lint` clean
- [x] `./tools/lint.sh` clean
- [ ] Netlify preview renders the section correctly

### DIFF
--- a/spec/dependency-provenance.md
+++ b/spec/dependency-provenance.md
@@ -11,9 +11,10 @@ It records the upstream identity of the consumed dependency, the
 platform that admitted it, and (where applicable) the
 identity-verification verdicts (integrity, publisher signature) and any
 admission-policy evaluations the platform applied. At higher levels of
-the [Dependency Track](dependency-track.md) it also records the
-upstream-provenance verdict and the platform's signing-infrastructure
-and ingestion isolation attestations.
+the [Dependency Track](dependency-track.md) it also records verdicts on
+any upstream-published attestations (SLSA Build Provenance, VSA, or other
+upstream predicate types) and the platform's signing-infrastructure and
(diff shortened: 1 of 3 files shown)


jjrauch32-lab (approved):
Approve

arewm (approved):
I think that this change makes sense. I do wonder if we should lean more into the fact that there are more attestations than just SLSA attestations.

